PR #138, opened 2013-03-25
Umbrella balance

theanax:
This weakens the spell a lot, makes it completly reliant on spell power, and removes trance so it can be used and combat and is not, otherwise, only a mule spell.


jrotunda:
Shall I reopen this as a new issue?  Umbrella is still disabled on 101 and 102 due to this broken code.

PR #168, opened 2013-05-27
Jig does not make players drop token

broodhill:
Jig is to powerful and players can to easily steal tokens by casting it
and then cancelling it.

This makes jig no longer causing a player drop a token they are holding
in their hands.
And to stop people from holding onto tokens forever they will drop the
token when they reach 1 vigor.


Meridian59:
Closing very old requests.

PR #176, opened 2013-06-04
Radius Enchantment Class

M59Gar:
Brought back the 'Jala Overhaul' as a new generalized and simplified class, using Umbrella as the prototype. All Jala songs still work under the old system, so this has minimal immediate impact on gameplay.

Content:
* Players have an internal list of Radius Enchantments similar to how Personal Enchantments work.
* These Radius Enchantments will show up in the Area Enchantment icon bar while active.
* With the new functionality, Umbrella finally has the ability to protect players inside it from players outside of it - but once an attacker is inside the barrier, it no longer helps. Even better, non-combatants can't interfere anymore.
* In addition, I created a much clearer 'conditional' function, TargetIsValid, which lets us easily control who is affected by the spells. For instance, Umbrella works on Guildmates and Enemies only.

Going forward, new Radius Enchantment spells will need their own hooks. There are not yet checks for any other functions besides what Umbrella does (modifyDefensePower, modifyResistance).


Meridian59:
Closing very old requests.

PR #179, opened 2013-06-10
Boss zones restricted to mules

broodhill:
Players are abusing mules to make boss zones much easier than they were
designed to be, one example:

Player X wants to do chest runs in trashers, he places his mules in each
"trap room" so the doors will never close, he can then easily run
through and keep doing just the chest without killing all the trashers
on the way.

Boss zones is supposed to be hard and not easy to do solo as it is now.

If a player needs a mule to carry items they can just place them outside
the side, so for trashers they can just place them within the wall
outside the door and just run in there and handover the loot they
collected.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -11524,6 +11524,7 @@ messages:
       }
       
       if IsClass(oRoom,&GuildHall)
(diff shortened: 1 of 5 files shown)


Meridian59:
I would want a different approach than just keeping people out of a room.  I don't doubt that there was abuse but there should be a more comprehensive solution.

PR #189, opened 2013-06-14
Mark of Dishonor can be casted on monsters

todasj:
A level four spell that cannot be improved by casting it on monsters is
a bit to hard, it can now be casted on monsters.
If casted it on a monster who aggro a player the monster will stop its
aggro towards that player.

### DIFF
--- a/kod/object/passive/spell/dishonor.kod
+++ b/kod/object/passive/spell/dishonor.kod
@@ -101,6 +101,7 @@ messages:
       }
 
       if NOT IsClass(target,&Player)
+         AND NOT IsClass(target,&Monster)
       {
          if NOT bItemCast
          {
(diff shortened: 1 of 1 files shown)


Koveras:
+1

jrotunda:
This should be merged, along with any adjustment that removes the "must cast on an outlaw!" need to work, which is a severely outdated leftover of a different M59.

I understand the need to be able to work spells on monsters, this is particularly troublesome with "EVIL TWIN" and the lack of alternative spells to work Riija.  Riija 5 has 3 spells - 1 requires you to be an outlaw or cast on an outlaw - limiting and annoying since it was added in 1997.

Still I have to agree with "theanax" here that any time you force users to have to cast spells on others in order to work them, you're limiting what a player can do in the game, and they'll seek out a cheesy alternative, in this case a "mule" made for the sole pupose of having spells cast on it.  

The player is left interacting with no one in this instance - with Broodhill's change it will eradicate the need for that type of mule, and force those players out into the world instead to cast the spell on mobs (like how Riija Dement, Vertigo work, and Qor's Enfeeble and Blind)

jrotunda:
Looks like this is out of date. This is a good change.

It should be tested to make sure there are no issues with certain monsters. 

The spell has also been changed since this PR to fix a bug so this will likely require more commits to update it. 

I found reference to Mark of Dishonor in a design To Do based on player feedback from NDS-Era, that it would have this temporary effect to make the aggro vanish.

This kind of spell
Update should be tested with some display
examples. 

The spell shouldn't work on most Boss/Supermonster Class creatures. Similar to
being unable to blind/hold the Lich or Dark Angel. 


For testing:

Player visual: the mobs resist it have a message or it just does not work? 

Does the mob aggro again as soon as you move? Is the hatred level effected? 

Does this cause bugs if cast right before a mob that activates a gate somewhere dies?

davidklaw:
@Meridian59 I think that this can be closed in favor of #393

PR #194, opened 2013-06-17
Reblance: Wedding Ring

marksturm:
The wedding ring is more a roleplayitem and was rarely used by the players, cause nobody knew about the defense boon before the game went open source. The item is simple overpowered on casters or hybrid-toons, with high agility, robe and cess-faction. Even a Purefighter will allways miss the caster/hybrid-toon over the half of the time.

Otherwise we had to create a perma "war-ring" which gives the same boon on offense.

### DIFF
--- a/kod/object/item/passitem/ring/ringwedd.kod
+++ b/kod/object/item/passitem/ring/ringwedd.kod
@@ -157,21 +157,6 @@ messages:
       return FALSE;
    }
 
-   % Things we handle to be a defense modifier:
-   ModifyDefensePower(who = $,what = $,defense_power = 0)
-   "Increase or decrease the defense of the defender."
-   {
-      % Add a small amount of defense.  Only added as a defmod if the user is
-      %  One of the original 2 users, and both user slots are filled.
-      return defense_power + viDefense_bonus;
-   }
-
-   ModifyDefenseDamage(who = $,what = $,damage = $,atype = 0,aspell = 0)
-   "Increase or decrease damage done to defender."
-   {
-      return damage;
-   }
-
 end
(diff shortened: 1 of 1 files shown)


Koveras:
+1 to Morbus' suggestions

ApolloDiamonds:
same: +1 to Morbus' suggestions

davidklaw:
@Meridian59 This PR removes the defensive buff given to the wearer of a wedding ring, if the wearer is either of the two users who first picked it up (presumably to be "wed"), effectively making it a pure cosmetic item. There's some good discussion in the comments about defensive rings, but that might be a topic completely separate from wedding rings. It's also a really friggin' old PR. Would you like to see the wedding ring neutralized or is this a candidate for closing?

kod/object/item/passitem/ring/ringwedd.kod
```diff
@@ -157,21 +157,6 @@ messages:
       return FALSE;
    }
 
-   % Things we handle to be a defense modifier:
-   ModifyDefensePower(who = $,what = $,defense_power = 0)
-   "Increase or decrease the defense of the defender."
-   {
-      % Add a small amount of defense.  Only added as a defmod if the user is
-      %  One of the original 2 users, and both user slots are filled.
-      return defense_power + viDefense_bonus;
```
davidklaw, line 166: Removing this function orphans `viDefense_bonus` under `classvars`. I recommend removing that as well.

akirmse:
I'm OK taking this if there are no objections.

davidklaw:
🚫 No objections found.

PR #200, opened 2013-06-24
Who List Colors fix

M59Gar:
In color.c, created GetPlayerWhoNameColor (based on GetPlayerNameColor)

This function ignores shadow form and frees us from dependency on the
actual displayed colors in names over players' heads.

Anonymity turned out not to cause any problems with the who list colors.
It just replaces text.

Morph had an obscure check in user.kod that temporarily eliminated all
colored name flags (outlaw/PK/dm/creator). This was the problem with the
who list colors, and it isn't necessary at all, as the removal of the
USER_YES flag hides displayed names.  In testing, Morph still works
exactly as it did before - morphed players are indistinguishable from
monsters visually.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/council/twocon.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/council/twocon.kod
@@ -47,17 +47,16 @@ properties:
 
 messages:
 
-   UserEntered( who = $ )
+   UserEntered(who=$)
    {
-      local iFlags;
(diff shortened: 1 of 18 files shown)


jrotunda:
This is still an issue on production.

After casting morph, any colored name reverts to white. The name goes back to the correct color after the client is refreshed. 

This also affects DMs in a permanent way by turning their name from Green to Blue.

PR #204, opened 2013-06-25
Hunt fix - only works on Murderers

M59Gar:
This spell, once given a new prompt upon first casting, was immediately
used by established PKs to track and harass victims, completely
preventing them from playing the game. This should really only work on
Murderer targets.

### DIFF
--- a/kod/object/passive/spell/multicst/hunt.kod
+++ b/kod/object/passive/spell/multicst/hunt.kod
@@ -43,6 +43,8 @@ resources:
 
    hunt_lost_trackers = "You seem to have managed to lose your trackers!"
    hunt_gain_trackers = "You sense that someone knows where you are..."
+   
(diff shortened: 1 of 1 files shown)


jrotunda:
I'd like to get this FIX pushed. Huge mess of comments. This spell gets abused, last summer a WHITE player, Iceman used it to run a new player off the server - the player begged for us to get him to stop, and seeing the constant messages was disturbing to them. I don't blame the player at all and it shows a need for Hunt to not work this way.

Hunt was added so Sword of the Hunt Hunters could hunt Necromancers - who were primarily Red Qor characters.

The scenario got disabled, reworked into the Soldier Shield Faction scenario as a method to get and maintain character boosts and be a part of an active in-game storyline.

The remaining spells were reworked. Some maybe not in the best way - Hunt being an example of it.

**- A character should not be able to stay white to cast this spell.
- The spell should only be able to be cast on outlaws and murderers.**

PR #206, opened 2013-06-27
Who list changed to let players fight/talk/use main client

M59Khan:
Here is an example video of this change:
http://www.youtube.com/watch?v=pFFCOJCRUuE&feature=youtu.be

The who list window currently functions as a modal dialog box.  This
means that when it is activated it does not allow any input to the main
meridian window.

This change makes the who list window a modeless dialog box, allowing
the player to keep it up while they use the main screen.

One of the main reasons players use multiple clients is to watch the who
list.  This change pushes m59 a little more back towards the one client
per player model.  Everyone watches the who list for pkers and enemies
and since it is not possible to use a client with a who list up, players
run 2 clients.  Accounts are free so there's no reason not to use 2
clients to see the who list.

This change fixes that: now you can watch the who list and play the game
at the same time.

A couple quick notes:
1) since the who list is now modaless hitting enter goes to the main screen (so it won't close the window, you will have to click to close the window).  This is intentional because otherwise you could not type to anyone with the who list window up.

2) The german windows seem to have their own properties in the client.rc file so I left them alone.  <---- this is fixed 

3) If someone changes color (murderer/outlaw/innocent status) you will need to relog back in just as it is now before the who list color takes effect.

4) having the who list window in front of the main client results in some choppy movement because meridian is juggling the intput between the main window and the who list window.

5) finally, theres no limit currently to the number of who list windows you can open.  There's no advantage to opening more so I don't know why someone would do it, but I haven't figured out how to check if an instance of the who list window is already open. <---- this is fixed 

I think even with these caveats it is still worth it.  If you keep the window off to the side you will have no problems, and I don't know of any reason why someone would want to move around with the who list blocking their view.  It is better than the current implementation where nothing can be done with the who list window up.


Meridian59:
Seems like this was reverted on 103?  Please resubmit if you have something that has been tested and is known to work.

PR #211, opened 2013-07-05
Changed truce reagents

todasj:
Change from a yrxl sap to a elderberry. There is really no reason for
truce to require such expensive reagent as yrxl sap when it is only
effective against monsters (PvE).

### DIFF
--- a/kod/object/passive/spell/roomench/truce.kod
+++ b/kod/object/passive/spell/roomench/truce.kod
@@ -21,7 +21,7 @@ resources:
    truce_desc_rsc = \
       "The spirit of Shal'ille calls upon her ally Faren, forming a truce with creatures of nature for "
       "several minutes.  "
-      "Requires emerald and yrxl sap to cast."
+      "Requires emerald and elderberry to cast."
 
    truce_unnecessary = "Combat with monsters is already impossible here."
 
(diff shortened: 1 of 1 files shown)


Koveras:
+1

XYLUM:
I say revert truce back to it's original state.

jrotunda:
Disagree with this PR, burn it, it shouldn't need to be done to begin with and another commit needs to be made to fix the actual problem: truce itself.

I think the reagents should be unchanged and truce should be reverted back to its full chaotic power and work for players and mobs. 

This also nerfed players who are trying to save others from a singular PK. It was a poorly thought out and borderline malignant change. 

The "fun part" of truce - players would log on during PvP - someone would discord the truce and they would die to a well timed arrow or slap. This happened more often than people think. It was a whole meta and it was a huge gamble. Whoever took this out had a vendetta and it angered everyone playing.

This change was an early player submission that is contained but not small in effect, throwing in a huge PvP balance change that had been part of the game for two generations of combat, 3DO, NDS and 2010-12.


It should never have been changed.

davidklaw:
@Meridian59 A candidate to close, perhaps? I'm getting the sense that the current state isn't desirable with commits like [this OpenMeridian commit](https://github.com/OpenMeridian/Meridian59/commit/24620fb0ce97f024e0e265c16a616e7084765219) and this PRs comments as evidence. If we close this, I'm happy to put in a new PR that addresses the functionality and not the reagents.

PR #213, opened 2013-07-05
Boss bonus XP

theanax:
This is the ground work for something bigger but I thought until I could work the math out for the rest of it maybe the admins could have some fun.

So the very basic idea here is to allow for monsters to give more then the standard 1-2 XP (Gainchance) that they currently do.

For someone at 20 hp killing a giant rat is the same as killing the ghost XP wise. For a person who is 149 hp killing a Mollusk create is the same as killing the luppog king.

So right now high level players only farm mollusks because they're the easiest end game monster to kill/exploit killing. In the future setting the harder monsters and boss monsters to have a slightly high number will balance the difficulty and we should see some variance in building.

Currently there aren't really any low end or mid-tier bosses to speak of but this can be corrected. The Orc Pitt Boss is the closest thing we have to that, but additional bosses can be added.

As I stated, I'm working on the math with a few other people and will push the addition of the specific pibossbonus numbers at a later time.

---

So here is how it works right now. Currently an admin must set pibossbonus on a monster. If a player is within normal range to gain XP from said monster and gets flagged for the normal 2 xp by both getting hit by and landing the killing blow of the monster they will see the message that they have killed a mighty foe. They are granted the normal 2 XP and the bonus.

The reason I put the bonus only at the point where players get both the normal XP is to avoid players cheesing stronger monsters and bosses by using things like firewalls.

---

I hope to add additional ways to gain XP like this in the future in attempt to move away from the bait/firewall abuse we see today. Making the game more interactive and removing the need/use of mules.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -5787,7 +5787,7 @@ messages:
 
    GetBonusXP()
    {
-       return piBossBonus;
+      return piBossBonus;
    }  
 
 end
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -283,6 +283,8 @@ properties:
    piHit_points = 1
    piMax_hit_points = 1
 
+   piBossBonus = 0
+
    % Who we are dealing with.
    poCustomer = $
 
@@ -5783,6 +5785,10 @@ messages:
       return;
    }
 
+   GetBonusXP()
+   {
+       return piBossBonus;
+   }  
 
 end
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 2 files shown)


davidklaw:
This is interesting. If I understand correctly, this allows dungeon masters (server admins or whatever we refer to them as) to give monster types a bonus given only to the player who a) takes a hit and b) lands the killing blow on a monster (or player) that is has greater hp than them.

And that this change is being proposed because it will encourage high level players to go after more challenging monsters, for the bonus, rather than farming monsters just beyond their current experience level.

Is this something you're open to @Meridian59? If so, is there anything you'd like to see addressed here?

akirmse:
Yeah, this might be occasionally useful for events, I suppose.  I don't think it would get used much as-is it requires an admin to set the boss bonus manually.

jrotunda:
Neat idea but looks abandoned and the current version is not very useful.

From a DM standpoint, it's just as easy to award the player who killed a special "event mob" XP through User > piGainChance in real-time.

When doing events with special mobs, a DM can easily give "bonus XP" to anyone in the room logged on or off contributed to the fight. 

It would be a newbie DM move to just award one person XP here anyway with this code since there are already other creative things a DM can do - like using special engaged returnable items in the monster drop so only the finder gets it, to design out any flaws in those kind of events.

If anything, separate DM commands could be made to make a "event xp" process like that easier, but this PR doesn't address that and it looks like @theanax is inactive now.

PR #223, opened 2013-08-16
Island room_no_combat defaults

M59Gar:
Every update an admin manually has to reset these rooms to enable
combat. This eliminates the non-combat defaults. This is an issue
because the island safe zones are literally adjacent to the three major
building spots and cause all sorts of problems with combat.


davidklaw:
It's been 9 long years, but after asking around, I found non-combat to be the desired state for these rooms within the core project. I move to close this PR and leave it as a reference for any fork wanting to do something similar.

PR #238, opened 2013-09-13

M59Gar:
Popular requests to re-enable webs and spores in doors have been
intense, to say the least, predicated on the basis that they do not deal
damage and can't be used to kill newbies in a cheesy fashion. They are
also very important for crowd control, and best used in doors
specifically. It is, in fact, not that people zone into them and get stuck,
but rather that they don't enter the room at all because of the trap wall.
That strategic element of blocking entrances doesn't translate when used
away from the door.

This pull excludes webs and active & passive spores from deletion by
the room entrance code.

### DIFF
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -3910,8 +3910,10 @@ messages:
       for i in plActive
       {
          each_obj = Send(self,@HolderExtractObject,#data=i);
-         if IsClass(each_obj,&ActiveWallElement)
-            or IsClass(each_obj,&Brambles)
+         if (IsClass(each_obj,&ActiveWallElement)
+            or IsClass(each_obj,&Brambles))
+            AND NOT IsClass(each_obj,&ActiveSporeCloud)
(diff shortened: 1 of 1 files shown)


Koveras:
didn't read most of the comments, but +1 to the pullrequest...

jrotunda:
This is a further refinement of a previous pull, #108 , which Meridian59 quotes as a "really good idea" and could be considered a "hotfix". 

With this change, non-lethal walls can again be used in a room zone location - a healthy compromise since there will not be a lethal wall stacked on to the hapless user.

It should be noted that since the initial pull (#108), this style of pker has vanished from server 101/102, and walls get used by players mainly to hold ghosts and PvE.  

Still, the terror of walking into a carefully led web or spore burst was and can still be a fun part of the game without having fire on added to it, and it works both ways for PKs and Hunters alike.

XYLUM:
I'll +1 this because it is definitely better then the current silly state of wall spells, I would still prefer a full revert though.

PR #253, opened 2013-10-13
Update smallbox.kod

DoomPig:
a set of heartstones for a nicer chest is something im sure we'd all like..since there are theives and its annoying...right?


Koveras:
+1
I wouldn't even mind seeing unlimited numbered items here.

davidklaw:
@Meridian59 Is this a desirable change? If so, this change wouldn't make sense without modifying the item description, and even then, quadrupling its space while maintaining the same visual is a little odd.

> This **small wooden chest** is just large enough to **store a few things**.

If not desirable, I move to close this PR.

PR #255, opened 2013-10-14
Guild Hall wooden chest

DoomPig:
giving more reason to steal guild halls since they can have more fat sacks of l00t


davidklaw:
Perhaps this ought to be closed based on the precedent set by closing #253 - there doesn't appear to be a desire to  increase chest sizes in this core version of the game at this time.

akirmse:
Eh, let's give this a try. Maybe it will spur some more guild conflict.

PR #263, opened 2014-01-13
Invigorate after HP.

fsuphil101:
This is a change that went into 103, after an HP is basically fills vigor and sets HP's to their (new) max. It's based on the idea that building is a massive grind and it's a little pep-in-your-step to be rewarded with full vigor and max hp.

Mayhem requested I submit it to official servers.


Meridian59:
Seems reasonable.

PR #267, opened 2014-02-25
Update splsacid.kod

ghost:
Lowest splash spell range of all (including ill wounds).
Changed the range from 3 to 4. It gives a little better chance for a "pure" Qor to hit a target.

### DIFF
--- a/kod/object/passive/spell/atakspel/splsacid.kod
+++ b/kod/object/passive/spell/atakspel/splsacid.kod
@@ -51,7 +51,7 @@ properties:
 
    piDamageMin = 16
    piDamageMax = 25
-   piRange = 3
+   piRange = 4
    
    % No bonus from Mana Focus
    piManaFocusBonus = 0


davidklaw:
Here are some other spells to compare to. Is this a situation where increasing its range should also come with a decrease in max damage or lowering its minimum damage to match explosive frost?

Spell | Min Damage | Max Damage | Range
-- | -- | -- | --
blast of fire | 15 | 23 | 4
explosive frost | 13 | 25 | 4
shocking fury | 12 | 20 | 5
**splash of acid** | **16** | **25** | **3**
vampiric drain | 12 | 18 | 4

Aesica:
Yeah, it looks like its tradeoff was meant to be "the best damage at the cost of range" just like shocking fury is "the best range at the cost of damage."  With a range increase to 4, it will straight-up outperform all of the others and that's no good.

davidklaw:
Good point @Aesica. I imagine there are many factors that need to be considered before a change like this is made (or not and I'm overcomplicating it), so I'll leave that to the folks that understand the current meta.

PR #273, opened 2014-06-11
Movement update from 103 plus piTimeAttackedPlayer fix.

Delerium1:
This pull request combines pulls #45, #69, #71, #92, #113 and #135 from the 103 repo to implement the 250ms movement update rate along with the speedhack and teleport protection that accompanies it. Client minor version incremented to 17. Also included is a fix for the continued issues from the time offset bug: piTimeAttackedPlayer is added to UpdateTimeValues() in player.kod, followed by running (as you probably already know) send c player UpdateTimeValues to set these to current game time. I think this variable has been added to the game since the last time offset fix, but was never added to that function.

Compiles and runs fine locally, no errors, all works as expected.

### DIFF
--- a/blakcomp/function.c
+++ b/blakcomp/function.c
@@ -59,6 +59,7 @@ function_type Functions[] = {
 {"LoadRoom",            CREATEROOMDATA,  AEXPRESSION,   ANONE},
 {"GetClass",            GETCLASS,        AEXPRESSION,   ANONE},
 {"GetTime",             GETTIME,         ANONE},
+{"GetTickCount",	GETTICKCOUNT,	 ANONE},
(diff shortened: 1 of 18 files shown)


jrotunda:
Definitely would like to get this onto 101 and 102 as soon as possible.  The timer bug has affected thousands of characters and is a time consuming manual reset for every player, and an inconvenience for the players to fix themselves (through PvP combat), or downright impossible for players under 30 HP.

skittles1:
I submitted the property/time bug fix separately (Delerium is a separate Github account I used before I figured out how to make an original repo branch to work in). This pull can be closed, we've since upgraded the movement/speedhack detection again so this is out of date.

PR #278, opened 2014-08-17
keep active window tab (inventory/spells/skills/stats) following system save or logout.

siwithaneyeprime:
The Meridian 59 client has *always* reset to the same client side group (spell list) after a system save. This was very annoying during PVP and in general (e.g. to use equipment you would have to use the mouse to reselect the inventory window)! This update persists the most recently selected client group through systems saves and also client logins. This client side operation has annoyed me for a long time and I hope you will agree that this is an important and useful fix! Implementation wise I have mirrored how the existing GetPlayerInfo method is shared between the meridian client and main library. This update includes a small client version revision and some helpful debugging information.


Meridian59:
I like the idea; please resubmit when you have it working.

PR #279, opened 2014-08-17
Character selection window - new/available slows will now always appear after existing charters! :)

siwithaneyeprime:
Previously, if you only had a single character made on your account, new account slots would appear first. This update orders character alphabetically with new and available slots always appearing last in the list.


Meridian59:
I like the idea but I made a few changes to the implementation (so I'm closing this):

- LBS_SORT also needs to be removed from the dialog in German.
- The ghost of my computer science teacher would haunt me forever if I used your bubble sort, so I changed it to insertion sort.
- I used std::swap.  First use of the STL in the client, I believe!

PR #282, opened 2014-11-11
Severel fixes up on 103 since 1 year - List in discription

Nagath:
Might & Aim calculation fixed
Armor of Gort damage cap bug fix
Rescue Potion - No karma Restriction
Ammo fix (Nerudit Arrows and Silver Arrows no work correctly)
Fixed Gauntlets breaking behavior
Server-side fix for the black color of playernames in the wholist


Nagath:
I adjustet it from % to %s like it is in 103 Codebase https://github.com/Daenks/Meridian59_103/blob/master/kod/object/item/passitem/attmod/gauntlet.kod line 36 (Dont know why I mistyped and forgot the s)

Nagath:
Opend smaller Pulls for the fixes to separete them a little bit and make it easier to add them because less files are changed per Pull

-> I let this open till I was able to work most of it out ^^

XYLUM:
I oppose changing calculations that will effect characters that were built for years under the existing calculations. Changes of that nature can render many existing characters useless gimps.

Nagath:
No calculations aren changed. Every old char will be absolutly the same. Same hitchance, same damage.
The problem now is, Aim 1 or Aim 50 are the same damage because the calculation doesnt work like it should so a 50 aim toon will do around 2 damage more then before while a 1 aim toon will be the same like before.
Also right now if you cast Armor of Gort at the moment with Armor on you, its more Powerfull the it has if you cast it in Kraanan Robe, this will also be fixed.

No worries: I dont want and have no interesst in touching basic mechanisms like defense or offense, they are absolutly fine like they are ^^

XYLUM:
My only concern is that after figuring this out through trial and error, we have built low aim toons that are beastly, I just don't want to see them nullified. The aim was more about "to hit" chance, or so I have always understood.
Date: Fri, 6 Mar 2015 00:49:31 -0800

Nagath:
I agree XYLUM. The last thing I want to see is that all the old toons people worked hard on are killed. Mine too because I've used low agi/aim in most cases too :D
But I realy want to see the mechanisms we have to work propably like they were intended.

Koveras:
I’m more concerned about balance. This potentially adds 2 (?) damage to potential new bow toons (or to old ones with lots of aim).

Is that balanced? Do “we” want this?

Those are the questions we need to be asking.

Nagath:
I admit this is a very good question.
I think it always should already have been like this but the mechanic only didnt work like it was planned.
Right now I'm a little bit unsure... I dont think this will make a big difference because of Armor of Gort.
By the way Melee is buffed with this too.
From my point of view it would make sense like this. High Might=Buff for Melee, High Aim=Buff for Bow but also would simply not effect older toons with low Might/Aim=same damage and hitchance then before.
Would be nice to hear more opinions about this ^^ Maybe we should discuss this a bit more

davidklaw:
More than 7 years later and 3 of the stated proposals remain unsubmitted. Seems like closing this is reasonable.

**Unaddressed by PRs:**
1. Might & Aim calculation fixed
2. Armor of Gort damage cap bug fix
3. Rescue Potion - No karma Restriction

**Either merged or closed:**

- #238
- #313
- #311
- #321

akirmse:
I think there's too much going on in this PR.  Individual PRs that change one thing at a time would be better.

PR #285, opened 2014-11-13
Add custom item color in lists for magic weapons/items

Nagath:
Quality of Life improvement:
coded by skittles1

Any weapon with an enchantment (cursed, glow, holder, Colhorr weapon, enchant weapon etc.) will show up in a separate color in all lists. Currently this defaults to orange and can be set by the client with other list colors (Options -> Colors -> List boxes -> Magic items).

All weapons and any 'defense modifier' (armor, shields, robes, etc.) with an enchantment will show up in either orange, the default color, or whatever the user chooses.


clientd3d/color.h
```diff
@@ -61,6 +62,6 @@ void ColorsRestoreDefaults(void);
 M59EXPORT HBRUSH DialogCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type);
 HBRUSH MainCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type);
 
-WORD GetItemListColor(HWND hwnd, int type);
+WORD GetItemListColor(HWND hwnd, int type, int flag);
```
skittles1: Should be flags, not flag.

kod/object/item/passitem/spelitem.kod
```diff
@@ -76,6 +75,8 @@ properties:
 
    vrName = Spellitem_name_rsc
 
+   viObject_flags = APPLY_YES | GETTABLE_YES
```
skittles1: This line should be in wand.kod instead of spelitem.kod, otherwise it will cause potions and scrolls to have the targeting reticule also.

Nagath:
both adjusted

jrotunda:
This would be a good addition for visual information and intuitive game play, currently there is no way to tell if an item is enchanted without clicking on the description, and would make for a nice visual if also combined with the game's colored lighting system.

Koveras:
+1. this sounds like a great idea!

davidklaw:
Seems like a positive change for the reasons described above, however, it's long out of date and the conflict with `clientd3d/resource.h` isn't one I know how to address to get a compile and provide some test results.

davidklaw:
@Meridian59 Abandoned and a candidate for closure.

PR #287, opened 2014-11-14
Autoloot for Numbered Items

Nagath:
Quality of Life Improvement by M59Gar:

When treasure is generated, a UserGet is called for any dropped number
items, saving the player the hassle of constantly looting. It's a normal
Get attempt, so if the user can't hold the treasure or is too far away,
it will fail.

UserGet does notify the user upon failure due to range, capacity limits, and so on.
Autolooting gives a summary message telling the player whether they tried
to loot all, some, or none of the items from a kill.

This will pick up all numbered items like shillings, reagents, and arrows.

Users can say 'autoloot off' or 'autoloot on' to change their autolooting behavior.
Autoloot is default on.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -195,6 +195,10 @@ resources:
 
    monster_default_attack = "attack"
 
+   autolooted_all = "You loot the corpse clean."
+   autolooted_some = "You loot some of the items from your fallen kill."
+   autolooted_none = "You find nothing easily lootable on your slain enemy."
+
 classvars:
 
    vrKocName = monster_koc_name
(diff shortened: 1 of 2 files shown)


jrotunda:
I think this is good, since it still lets the player choose and get tricked by cursed items, and is more of a quality of life tool for a player just trying to quickly gather or sell stacked items, and I like that it can be turned on or off - that's key.   Ideally, I imagine this as part of the UI with a symbol/button the user could press so users would automatically know.

Koveras:
I think I'd much prefer this if you could set individual items to be looted, since I wouldn't turn it on if it looted ALL numbered items, but definitely if I could set it to automatically loot money and that one reagent I need...
However, shillings could always be autolooted.

davidklaw:
@Meridian59 Is this something you see making its way into this project? If so, is anything missing for you? If not, maybe it's time to close this PR.

PR #290, opened 2014-11-20
Fixed damage type typos in ALL monsters

Nagath:
originaly by M59Gar

Because of a minor typing error the Monsters in Meridian all do only generic damage. They should do other damage types mostly like you can see in the code so this will fix this so Monster will do the damage type they should.

Also changed Avar Shamans to use brittle instead of shatter


Nagath:
I fixed the spell damage and reverted brittle to shatter

PR #293, opened 2015-01-01
DirectX9 upgrade from 103 part 1 of 2

treymd:
Sha'Krunes Dx9 Upgrade (1 of 2) 
Part 2 (Tweaks and fixes coming tomorrow)


cyberjunk:
Thanks for setting this up Keen, had the same idea but was lacking time :-(
Please go ahead and add the second parts, I'm just adding some comments:

Here's the link to the original first pullrequest with the 1:1 porting (without additional fixes):
https://github.com/Daenks/Meridian59_103/pull/765

I also tried to list all DX8.1 -> 9.0 API changes there. I'm more than interested in Zaphods oppinion especially on "7) D3DPRASTERCAPS_ZBIAS".

Because porting this to DX9 really gave me an headache first...

This is the second pullrequest on top of first with fixes/tweaks for Dx9 issues:
https://github.com/Daenks/Meridian59_103/pull/787

But be aware that the following commit from it may be not good:
https://github.com/cyberjunk/Meridian59/commit/e651a2c9fc4bf349962295aaabd26ce32224de99

Because we had reports of very few GPU not capable of using D3DPOOL_MANAGED

PR #295, opened 2015-01-02
DirectX upgrades from the 103 team (complete)

treymd:


treymd:
Shak's comment from the original pull:

Thanks for setting this up Keen, had the same idea but was lacking time :-(
Please go ahead and add the second parts, I'm just adding some comments:

Here's the link to the original first pullrequest with the plain 1:1 porting (without additional fixes, tried to stay as close to the dx8 code as possible for the first step):
Daenks#765

I also tried to list all DX8.1 -> 9.0 API changes there. I'm more than interested in Zaphods oppinion especially on "7) D3DPRASTERCAPS_ZBIAS".

Because porting this to DX9 really gave me an headache first...

This is the second pullrequest on top of first with fixes/tweaks for Dx9 issues:
Daenks#787

But be aware that the following commit from it may be not good:
cyberjunk@e651a2c

Because we had reports of very few GPU not capable of using D3DPOOL_MANAGED

Meridian59:
Thanks.. the documentation on this is super useful.  Z and depth are obviously not the same so I'll have to read up on how people dealt with the removal.  I thought I saw some weird artifacts on the tops of vertical rises on 103; I'll have to check more carefully.

Did you notice any performance changes?  The 103 client seems to have a significantly lower frame rate on my machine than the open source client, but I don't know if that's a result of DX9 or other new features that have been added.

Meridian59:
Also, it's odd that Direct3D itself is checked into the repository.  I'm assuming you don't do that with the Windows SDK (?)

cyberjunk:
>> Z and depth are obviously not the same so I'll have to read up on how people dealt with the removal.

No, unfortunately it's not the same... the ZBIAS porting seemed to be the only real "challenge" when bringing this to DX9. I also linked two documents in the comments (7.) in https://github.com/Daenks/Meridian59_103/pull/765 . One from an Intel guy showing different mechanisms, which was pretty useful (https://software.intel.com/en-us/articles/alternatives-to-using-z-bias-to-fix-z-fighting-issues). I think we've found an acceptable approximation of the old ZBIAS using mentioned new values. But it's still an approximation which only works within a given bound of camera-distances (as far as I understand due to always applied linear z-modification which beyond some "expected ranges" will aways turn into too big/small values and causing "misordering"). These tweaks are part of the second pullrequest (second commit here). There was few clear cases of z-fighting which requires solving (I removed manipulations for others like skybox):
1) Objects which have overlays/underlays, which are rendered on same plane (=at same depth)
2) Player/object names with their second shadow/black text with a slightly offset in background
3) The rendered room wireframe, which should always stay slightly behind the solid/really rendered room (I'm not sure if you're aware of this "hack" in the d3drenderer to close gaps caused by the BSP builder). This wireframe can still give a feeling of very sharp or "black" edges on 103, without the second pullrequest it also causes visible lines on top of solid walls.

>> Did you notice any performance changes?
No, I did not notice a big difference, but I did not pay too much attention to it. Was this with enabled christmas mode and weather effects on 103? 103 currently has enabled snow particles and with quite increased particle counts, including height lookups for each particle from ROO BSP to prevent snow under ceilings..

>> Also, it's odd that Direct3D itself is checked into the repository. I'm assuming you don't do that with >> the Windows SDK (?)

No, we don't plan to add the Win SDK.., but we were tired of explaining how to add dx8.1 includes/libs to the repository, it always was one of the additional steps to get it compile and everyone asked about... We also assumed this has no legal issues or technical disadvantages and the headers/libs seemed acceptable small... Do you see any disadvantages with this?

cyberjunk:
Since I already mentioned the wireframe with the sidenote to a BSP build failure, let me elaborate a bit further.... d3drender.c is doing an "hack" / additional rendering of the room as wireframe, see this screenshot: https://googledrive.com/host/0B-6rsj6uHlolV2ZGMjNQLUZiLXc/wireframe_pass.PNG

As mentioned this wireframe also cause some rendering issues due to z-fighting without further z-hacking we tried to find... search d3drender.c for "wireframe" to find lots of comments about that...

If it is disabled, there are some gaps, see this image (not really the best example, but sufficient):
https://googledrive.com/host/0B-6rsj6uHlolV2ZGMjNQLUZiLXc/wireframe_off.PNG

I tried to investigate this further, made a small checker for bsp leaf vertex coordinates, and it turned out that there is quite a few vertexes which are off by 1 fineness - and these are also involved in the triangles around the edges. However I can't explain it any further yet or give a solution for this.

treymd:
Other 3rd party binary libraries are in both codebases, wavmix (waveplay)
for example.  I think this is a good thing because it simplifies the build
instructions for newcomers.  Is directsound in another lib file?  I think
at some point we should be using that for sound.

Meridian59:
Looks like the DX9 SDK is included with Visual Studio 2013, so once you upgrade, you can remove it from your repo.

Meridian59:
OK, I've made the changes locally.  As you mentioned, the only non-trivial thing is the z bias.  Using depth instead does seem like the right thing to do; the Intel article explains why.  The only artifact I'm seeing so far is that objects are sometimes showing through geometry when both are far away.  I'm having trouble finding that specific fix in all of the formatting changes... do you know offhand which it was?

Meridian59:
One other thing, I wound up using IDirect3DDevice9_UpdateSurface instead of StretchRect.  Did you try that?

cyberjunk:
>> The only artifact I'm seeing so far is that objects are sometimes showing through geometry when both are far away.

Try decreasing the depth modifier, e.g. turn 0.001 into 0.0001 so the modification stays small enough to NOT cause this misordering. It's a bit tricky to balance these all with other aspects (like names and objects still staying in front of walls)

>> IDirect3DDevice9_UpdateSurface instead of StretchRect
You mean in "D3DRenderFramebufferTextureCreate"? I would be surprised if UpdateSurface works for this... according to http://msdn.microsoft.com/de-de/library/windows/desktop/bb205857%28v=vs.85%29.aspx UpdateSurface can only be used for Sysmemory->GPU memory. But this place rather seemed GPU mem -> GPU mem copy to me, so the correct one should be StretchRect. Make sure you also test Invisiblity effects, these require the mentioned RenderToTexture parts.

cyberjunk:
About the depth, we have in D3DRenderInit

>>IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_SLOPESCALEDEPTHBIAS, F2DW(1.0f));
IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_DEPTHBIAS, F2DW(0.0f));

We use -0.00001f as the mulitplier in the Z_BIAS constants transform.

>> IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_DEPTHBIAS, F2DW((float)ZBIAS_WORLD * -0.00001f));

We also modified the "D3DRenderNamesDraw3D" to bring the names as much in front as possible with this small modifier:

>>if (offset == 0)
pChunk->zBias = 255;
else
pChunk->zBias = 254;

possible few other small changes...

cyberjunk:
You probably also want to adopt at least these changes included in second parts, affecting the lost device handling and d3d restart/cleanup:

1) Release gpViewElements[i] textures in D3DRenderShutDown (memleak fix)
Originally included here: https://github.com/cyberjunk/Meridian59/commit/62e8da9a04a8e9d1a691493f5577d64fe4ee7561

2) Release vertex declarations in D3DRenderShutDown (introduces with dx9 port):
Originally included here: https://github.com/cyberjunk/Meridian59/commit/2e9f7398e33e54b9c590bf73a350a8e79a69ceae

3) Correctly handle lost d3d device. The old IDirect3DDevice9_Reset call was quite misplaced and not working.
See here: https://github.com/cyberjunk/Meridian59/commit/2520249a0ac974549fa8fb526da59bd171ad77dc

Meridian59:
I think this is complete now.  Let me know if this code doesn't work for you on 103.

PR #296, opened 2015-01-02
Added icon and message to second wind

Nagath:
Quality of life improvement: Was always annoying not to know its already ended or still effects you so you cant eat ;)

Originally coded by skittles1

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -706,6 +706,8 @@ resources:
 
    player_logged_on_wav_rsc = player_login_or_out.wav
    player_logged_off_wav_rsc = player_logout2.wav
+   
+   second_wind_finished = "You feel more relaxed as your second wind dissipates."
 
 classvars:
 
@@ -1001,6 +1003,8 @@ messages:
       {
          DeleteTimer(ptSecondWind);
          ptSecondWind = $;
+         Send(self,@ShowRemoveEnchantment,#what=Send(SYS,@FindSkillByNum,#num=SKID_SECOND_WIND),
+         	#type=ENCHANTMENT_PLAYER);
       }
 
       Send(self,@CancelRescue);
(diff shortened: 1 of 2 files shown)


skittles1:
There was a hotfix to this one that resolved an issue with the icon disappearing on system save:

    if ptSecondWind <> $
    {
       Send(self,@ShowAddEnchantment,#what=Send(SYS,@FindSkillByNum,
             #num=SKID_SECOND_WIND),#type=ENCHANTMENT_PLAYER);
    }

This needs to be added to SendEnchantmentIcons() in player.kod.

Nagath:
Hotfix added, thanks Delerium ^^

Koveras:
+1

Meridian59:
The changes to secwind.kod look fine.  But why does Second Wind need to be special-cased so many times in Player?

skittles1:
The Second Wind effect isn't an actual enchantment, but we use the enchantment messages to show the icon to the player if the ptSecondWind timer is active. So anywhere enchantments are resent (system save) and when the skill is triggered or ends, we have to send the info (Crystalize Mana on 103 works this way too). Ideally this skill would have been originally coded as an enchantment instead of using a separate timer.

davidklaw:
Tested and discovered an issue with the enchantment icon.

FWIW, I tested by modifying my vigor as through administrative commands. My character was an admin, but in "mortal" mode. I assumed that this wouldn't affect the outcome here.

Running `master`, I found the enchantment to trigger but not display the enchantment icon. I don't think it was necessary to test, but I also executed a system save and the effects of second wind appeared to still be present.

Running this PR, I found the enchantment to trigger, the personal enchantment icon to display for the duration of the effect, and for the effect to remain through a system save.

However, logging out and back in resulted in a second personal enchantment icon. I did a system save and the redundant icon went away. This is captured in the videos below.

Here's what I see in `master`:

https://user-images.githubusercontent.com/467443/175828231-e3d29824-3e63-4308-a1bf-9da144fd5076.mp4

Here's what I see in this PR. I left out the period between casting and dissipation so as not to bore you.

https://user-images.githubusercontent.com/467443/175828233-0d5df9d3-fada-483e-929d-9bd56895230e.mp4

https://user-images.githubusercontent.com/467443/175828234-b1aa1274-1b49-4703-b51d-7951e6c80dc7.mp4

Aside from the obvious bug with the enchantment icon, I don't know exactly what I'm looking for because I'm testing w/o a test plan. If this needs to be tested in a different manner, please provide those details.

skittles1:
This should work correctly if the added code in UserLogonHook is removed, since SendEnchantmentIcons should be called anyway (hence the two sets of icons when you log in).

davidklaw:
@Nagath Where you at! 2015 called and we need you back!

davidklaw:
@Meridian59 Another PR that can probably be closed in favor of the more recent #389 , if you believe that one has gone in a more desirable direction.

PR #297, opened 2015-01-04
Stats reroll

Battlecat1337:
1. Due to many school updates and other updates in general, stats preferences do shift
2. People need to have multiple chars with just small stats differences to use all the possible builds which leads to chartrains and toon spaming
3. New or unexperiences playes still need 6-12 Month to max a char and often realise that they have chosen the wrong stats...  frustrating them.

To solve all these issues my idea is to provide a reroll process for stats:
- should be expensive (~250k sh) and should only allow to switch 10-20 stat points
- should be only possible once a year
- yes you can learn a new school with 20 more int but it should be fine with the other rules mentioned above
- some fresh air, players will have fun to "optimise" their older chars


Meridian59:
2000 files?  Bad upload.

PR #300, opened 2015-01-19
Small XP gain changes

Nagath:
coded by theanax

Added an additional point to gain on kill and on kill with taking damage that should effectively up the HP gain rate a bit.

Changed the newbie bonus XP to scale with whatever the angel setting is rather then stopping at 25.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -7663,7 +7663,7 @@ messages:
          }
 
          % Give newbies a little bonus.
-         if piBase_Max_health = PKILL_ENABLE_HP
+         if piBase_Max_health = < Send(Send(SYS,@GetSettings),@GetPKillEnableHP)
          {
             gain = gain + 1;
          }
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -7631,14 +7631,14 @@ messages:
             if killing_blow AND Send(self,@CheckPlayerFlag,#flag=PFLAG_TOOK_DAMAGE)
             {
                % Player took damage and landed killing blow
-               gain = 2;
+               gain = 3;
                roll = TRUE;
             }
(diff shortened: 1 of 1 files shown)


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -7663,7 +7663,7 @@ messages:
          }
 
          % Give newbies a little bonus.
-         if piBase_Max_health < 26 and random(20,bound(piBase_Max_health,20,25)) = 20
+         if piBase_Max_health = PKILL_ENABLE_HP
```
skittles1: This only gives the bonus if the player's max heal is equal to PKILL_ENABLE_HP. We have this on 103:

     if piBase_Max_health < Send(Send(SYS,@GetSettings),@GetPKillEnableHP)

So you would want to use:

    if piBase_Max_health < PKILL_ENABLE_HP

Nagath:
I adjustet it the same way. Its proven it works well like this on 103 ^^

Koveras:
+1, this would especially help in group scenarios since like this the bonus for getting hit is "only" 50%, rather than 100%.

davidklaw:
Seems harmless, and in some small way offers new players a slightly more encouraging onboarding experience.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -7663,7 +7663,7 @@ messages:
          }
 
          % Give newbies a little bonus.
-         if piBase_Max_health < 26 and random(20,bound(piBase_Max_health,20,25)) = 20
+         if piBase_Max_health = < Send(Send(SYS,@GetSettings),@GetPKillEnableHP)
```
davidklaw, line 7666: @Meridian59 We missed this `= <` and introduced a compilation error. I'm putting together a new PR.

akirmse:
Whoops, I should have seen this.  The pre-submit action will find such
things in the future.  I wasn't able to find a way to run actions against
PRs that predate them.

PR #301, opened 2015-01-19
Doubled Vault size

Nagath:
Doubled the size of Vaults -> less need of Storage Mules


Arkblade:
Would it really matter if vaults were just limited to number of items instead? Say 150 unique items? That way you can store as many stackables as you want but will only have space for 150 items in your vault total?

Koveras:
Not really, if people want more vault space they simply get more accounts, But it would most definitely make a hell of a lot more sense this way, great idea!

Nagath:
Broodhills Pull is better ;)

PR #302, opened 2015-01-19
Increase target outline from 1 pixel to 3 pixels

Nagath:
by skittles1

So players can see their target better.
https://github.com/Daenks/Meridian59_103/pull/766

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -7088,14 +7088,14 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 				}
 			}
 
-			pChunk->xyz[0].x -= 32.0f / (float)pDib->shrink;
-			pChunk->xyz[0].z += 32.0f / (float)pDib->shrink;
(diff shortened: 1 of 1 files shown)


davidklaw:
Preview image cross-posted from https://github.com/OpenMeridian/Meridian59/pull/766

image

More than 7 years old and merge conflicts. Might be a good candidate for closing. In my experience, the target outline can be difficult to see from a distance but is this the direction to go with it? I don't know.

davidklaw:
I didn't dig into resolving the merge conflict, but here's the difference:

**BEFORE:**

https://user-images.githubusercontent.com/467443/175795299-69b29c55-ba57-4f9b-98fd-a7abd45926de.mp4

**AFTER:**

https://user-images.githubusercontent.com/467443/175795306-ae909a13-206f-451e-9d93-995784406c63.mp4

Neither is particularly attractive 😐 but hey, that's just my opinion, man.

akirmse:
I'm sure there's a way to do this in hardware that would look a lot better.  Maybe draw each piece with a scale factor of 1.03 and a constant color, or something like that?

PR #303, opened 2015-01-19
Aqueducts of Merb now only spawn Shadowbeasts

Nagath:
Right now its hard to find a place to build HPs effectivly on your own as +Karma Char after 130HPs. Hunting Shadow Beasts isn't effective because of many Orcs between them so you need way longer then if you use Bait/Illufirewall at the beach.
With this we have another High level building Area as compensation. Still takes longer then beach but way more effective/usefull then before.


davidklaw:
I move to close based on the age of this PR.

PR #207, opened 2013-06-29
New Spell Introspection and Changes requested

theanax:

PR #304, opened 2015-01-19
Added Tougher and Weight Carried bar

Nagath:
Added Tougher and Weight Carried bar to the Userinterface


jrotunda:
Great 2-for-1.  Just out of curiosity, do you have a link for the image/graphic to the Tougher/XP and Encumbrance/Weight bar, and where did the art come from? Are these the same that are on 103?  I didn't see a weight bar making the rounds yet, but would make for a good quality of life improvement and help players understand how inventory works.  There is a note in the old updates notes somewhere that this was something originally planned but never implemented.

Nagath:
Im not a good artist so I thought about a good solution. I didnt want to steal the art they worked out for 103 so I decided to take an existing one (the special effect art thats used for praying debuffs/rose effect) for the tougher bar. Weight Carried bar has/needs no Icon because its in the stats window while tougher bar is under vigor

Daenks:
The artwork submitted to the 103 repo is open source just like our code, no need to replace it, and artwork you submit to this repository will also be open source, which means 103 can use it, unless you include a specific license barring re-use.

skittles1:
Just to add onto what Daenks said, I drew the icon used for the tougher bar and cleaned up the other three (hp, mana, vigor). Please use them :)

Nagath:
Will this work? I remember there were some issues with .bgf files because of crusher/zlib...
Are they still compatible to this repo?

Nagath:
I added the Icon Delerium created for Tougher Bar and changed it in the Code

treymd:
I believe both repos now exclusively use zlib compression rather than the proprietary crusher, someone can correct me if I am mistaken.  Any 103 art should work on 101 and vice versa.

Nagath:
I tested it before I uploaded the .bgf file. it works perfectly fine ^^

Koveras:
+1

XYLUM:
+1

davidklaw:
I believe that this https://github.com/Meridian59/Meridian59/pull/207#issuecomment-1172976741 is evidence that changes like these might not be desirable. However, if they are, I propose we consider displaying the "carry weight" amongst HP, Mana, and Vigor (how are these referred to?) because it gives it constant visibility, whereas the Stats or Inventory tab do not. I don't think a value makes sense because there's no visible weight value given to items.

This is an example of what I mean, using a placeholder icon of a handled weight:
image

**Sidenote:** I've personally always felt like the surprise of a tougher was an enjoyable moment. I'd hate to see that give way to a data-driven grind.

@Meridian59 Are you open to considering changes like these or is this something to close?

akirmse:
This is a slippery slope.  Add this bar, and people would want numbers on the bar, then numbers on each item, etc.

davidklaw:
Fair enough. Crush it!

PR #306, opened 2015-01-23
Removed Karma requirement from VampWeaponAttribute

Nagath:
I did never understand why hold/blind/purge weapons have no karma requirements but vampiric weapons have.
Without it they would be great building weapons!

### DIFF
--- a/kod/object/passive/itematt/weapatt/wavamper.kod
+++ b/kod/object/passive/itematt/weapatt/wavamper.kod
@@ -58,7 +58,7 @@ messages:
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
       
    ModifyDamage(damage = 0, target = $, wielder = $, lData= $)
-   "Wielder has a 1% chance of blinding target with any hit."
+   "Wielder has a 1% chance of stealing life from target with any hit."
    {
       local oSpell;
       if random(1,100) <= piEffect_percent
--- a/kod/object/passive/itematt/weapatt/wavamper.kod
+++ b/kod/object/passive/itematt/weapatt/wavamper.kod
@@ -79,11 +79,6 @@ messages:
 
    ItemReqUse(oItem=$,oPlayer=$)
   {
-      if send(oPlayer,@GetKarma) > -30
-      {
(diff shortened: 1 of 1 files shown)


davidklaw:
Seems like the right move, making vampiric weapons consistent by removing karma requirements. I don't know where the updated description comes into play, but it appears to be inaccurate or not all that user-friendly (i.e. "Wielder gets a damage bonus, equal to (proficiency/4) - 2.") for several of these attributes. This description modification is fine, but assuming that description is user-facing, they could all maybe use some love in a separate PR.

davidklaw:
`ItemReqUse` might benefit from a little bit of cleanup but this seems like a good, simple change to bring vampiric weapon attributes in line with other weapon attributes based on karma-driven attributes, which don't care about karma. Unless this is intentional because of gameplay dynamics/balance, this seems like a candidate for merging. cc @Meridian59

akirmse:
I have no idea why the requirements are as they are---there's no comment there.  Does anyone have any opinions?

Dirhai:
This requirement is silly

davidklaw:
I'll also point out something being brought up in Discord -- these attributes would continue to require the wielder to have at least 50% proficiency in the enchanted weapon's respective Skill.

jrotunda:
Vampiric Drain is Level 3 and when applied via a weapon, its spell power and effect on drain/intake is minimal. 

Blind, hold and purge (weapon perhaps better balanced than the actual spells) enchantments are more powerful in PvE or PvP and require only the skill of the weapon being wielded - it seems strangely punitive to make a meager vampiric weapon have a karma requirement in comparison.

This looks like a design conflict that never got addressed. Vampiric Weapons were added to Meridian 59 a long time ago - before Zjiria, Q and PC enchantments, which were made by PC. 

It may be beneficial to review the requirements, effects or drops rate on those separately - but this PR at least makes the weaker Vampiric Drain effect weapons more useful and consistent. 

It would be a positive building aid and add some gameplay diversity for any player as @Nagath mentioned.

Blood Inheritance (L6 Qor) would also have more importance as it can create these for other weapon types that aren't currently included in any Magic Enchantment drop lists, like Nerudite Swords.

davidklaw:
@Meridian59 Coming back around to suggest this get reviewed and merged!

PR #311, opened 2015-01-27
Fixed Gauntlets breaking behavior

Nagath:
Gauntlets didn't update their own graphic when they broke due to strange
DefendingHit code overwriting item.kod's gear-breaking code. This fixes
that issue, so gauntlets will appear broken in one's inventory
appropriately.


Koveras:
+1

PR #312, opened 2015-01-27

Nagath:
Popular requests to re-enable webs and spores in doors have been
intense, to say the least, predicated on the basis that they do not deal
damage and can't be used to kill newbies in a cheesy fashion. They are
also very important for crowd control, and best used in doors
specifically.


Koveras:
+1

jrotunda:
Dupe of #238

XYLUM:
While I think this is a great move in the right direction, I still support the full on restoration of wall spells to their original state. I am not alone on this.

I suggested making walls not effect chars less the 70 HP, to eliminate the noob killing while restoring the full use of wall spells for trapping, guarding and zone control.

Meridian59:
duplicate

PR #313, opened 2015-01-27
Ammo fix Nerudite/Silver Arrows

Nagath:
Nerudite Arrows haven't been shattering because BaseDamage wasn't
passing who and target. Silver Arrows also weren't doing double damage
to undead.

### DIFF
--- a/kod/object/item/passitem/weapon.kod
+++ b/kod/object/item/passitem/weapon.kod
@@ -308,7 +308,7 @@ messages:
       local iDamage, i, oWeapAtt;
 
       % First, get base damage.
-      iDamage = send(self,@GetBaseDamage);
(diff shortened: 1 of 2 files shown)


Koveras:
+1

Meridian59:
It looks like this might cause some runtime error messages in GoldSword because it doesn't have these parameters, and it should.  Could you please add those to goldswrd.kod:GetBaseDamage?  Then I'll merge this.

davidklaw:
@Meridian59 It looks like the requested changes were made. Might this be ready for a merge?

akirmse:
Yes, I think this is only other instance of GetBaseDamage, so it should be safe.

PR #314, opened 2015-01-27
Add new Item - Light Disciple Robe

Nagath:
Added new Item Light Disciple Robe and added them to Avar Shaman loot

### DIFF
--- a/kod/object/item/passitem/defmod/armor/robe/lightdisc.kod
+++ b/kod/object/item/passitem/defmod/armor/robe/lightdisc.kod
@@ -1,0 +1,439 @@
+% Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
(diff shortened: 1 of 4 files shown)


Arkblade:
What purpose do these serve? ...and I mean other than being sleeveless. =)

Arkblade:
Personally ... I think these robes should weigh half as much as Disciple Robes, offer 5 less spellpower to the dedicated school but offer NO spellpower to any other school. My opinion.

Nagath:
I made them in this version really lighter ^^
30 less weight, +3SP per level for the dedicated school (to a max of 15), +2SP to all others.
My Intension here was to let them weight less but also let them give less Power so they have their own purpose in Inventory Management (More other stuff but less power for one school)

Arkblade:
Cool beans

Nagath:
I added them as loot to Avar Shamans, I think would be more interessting to not let Orc wizards alone drop all kinds of Robes ^^

PR #315, opened 2015-01-28
Add Spellpower Bonus to God Gifts

Nagath:
I remeber the rumors in the good old times about GGs giving Power Bonuses to one School and I think this is a very good Idea, especially because they are all already dedicated to a School like you can see on the Names inside the .kod files.

### DIFF
--- a/kod/object/item/passitem/necklace/frcharm.kod
+++ b/kod/object/item/passitem/necklace/frcharm.kod
@@ -43,6 +43,7 @@ classvars:
    viMax_Disp_hits = 5
 
    viGround_group = 2
+   viSpell_modifier = 10
 
 properties:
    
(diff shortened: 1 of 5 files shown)


Koveras:
Yay for making GGs more relevant, but please add a way to acquire them...

Nagath:
Currently there are plenty of Pulls open, I parted some already because some adjust the same files and this should make it easier to merge them.
Right now I'll wait till I know what Pulls Zaphod is interessted in before I open a new one that changes the same files currently open Pulls already do ^^

davidklaw:
@Nagath Can you please elaborate on the change you're making to aid in review and testing? Were there buffs prior? What were they? What is expected from your changes? FWIW, I think this is a nice change and want to help test it.

davidklaw:
This PR adds a spell modifier (value: 10) to each of the game's "God's Gifts." It's assumed that this was the original intent of these items as they're all associated with a single school in the codebase (i.e. A diamond pendant's class name is `ShalilleCharm`).

Is this giving them too much power? I don't know. It looks like a Faren player wearing a robe, holding a torch, and now wearing a brass medallion gets an extra healthy dose of spell modification. Should _all_ gifts get the school bonus, or just the 20% that have the special message? Any desire to upgrade these gifts @Meridian59 ?

davidklaw:
@Meridian59 Anything more I can add to this to help make a decision? Seems like @Nagath might not return.

akirmse:
I think we should close it.  Kind of an interesting concept, but unbalancing and might lead to yet more buffs.

PR #317, opened 2015-02-08
Remove wrap.lib

broodhill:
Since we are no longer using wrap I do not know why it is still in here and archive.lib is not. Generates errors when trying to compile the source if wrap.lib is in.


akirmse:
Sorry, looks like I changed this locally but forgot to check it in.

PR #318, opened 2015-03-01
Vault change

todasj:
The vault patch from 103 (https://github.com/Daenks/Meridian59_103/pull/175), but reduced the number of items to 150 instead
of their 600 which might be a little overkill. 150 might be to much too
but I leave that to the developers to decide.

The vaults allow you to carry 150 items (stacks of reagents/food count
as one item).
People are building and gathering loot but have to store it on multiple
mules, this would reduce the number of mules people are using.

### DIFF
--- a/kod/object/passive/storage.kod
+++ b/kod/object/passive/storage.kod
@@ -13,6 +13,8 @@ Storage is PassiveObject
 constants:
 
    include blakston.khd
+   
+   VAULT_ITEMS_MAX = 150
 
 resources:
    msg_not_enough_items_on_deposit = "%s%s does not have that many on deposit."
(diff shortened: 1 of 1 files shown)


jrotunda:
A much requested upgrade and have seen good feedback on it since implemented.

Also fits with the goal of reducing the need for 'mule' characters.

Nagath:
+1

XYLUM:
+1

davidklaw:
@todasj Would you mind explaining this to me: is the current functionality based on item weight (with a max of 3000 units)? This change does away with weight units in favor of items and stacks with uncapped counts?

akirmse:
The server 101 save game is pretty huge, and it's almost all junk ancient characters have stored in their inventories and vaults.  I'd be concerned about increasing the number of such "dead" items that hang around in the saved game forever. That increases load/save times, fills up the server's disk, and causes other problems.

davidklaw:
So, a change like this is dependent on factors related to the server. Given that each servers situation is different, keeping this as-is seems preferable -- close it?

Would we be open to considering ways in which vaults might get "uncluttered" in the interest of optimizing the save game?

jrotunda:
Some thoughts:

Do players really need more space? What is being stored? Doing a cursory check after an admin made the vaults on 101 too
High, it seemed to be "I have space so let me throw all kinds of garbage here that I don't need", example : someone with 50 Ivy circlets and every armor instead of an emergency set or three. 

There's not much unique gear that one needs or can even obtain for a given character even in a lengthy player character lifespan. 

Vaults should be for valuables. Everything else is meant to be for activity - you're not active you don't need it actively stored, which is where guild hall and personal housing - currently rentable rooms - come into play.

Who wants this exactly - and why? What is the end result - easier/faster building by having tons of stuff in one place, or making it easier for someone to play individually without others? 

Is it just for players who have perhaps an excessive amount of rare gear? 

Rare items or stacks of reagents "for a rainy day"? Should reagents and food expire? Arrows go missing? 

Maybe you have to log on once a year or bye bye MSH?

This should be closed - the storage situation and design changes if any necessary can be looked at in more detail.

davidklaw:
This might be best to close and kick off as a new Issue regarding unused, stored items. Hell, it might be a conversation outside of GitHub since it doesn't pertain to the larger open source community, yet. How might 101 reduce its save game file, specifically when it comes to inventories and vaults.

I'd love to get my hands on the data.

PR #320, opened 2015-03-09
Fix 3 aggro limit - check if player attacked monster.

skittles1:
The 3 aggro limit is caused by not having monsters check if they're being attacked before cancelling hatred due to the amount of monsters already chasing the attacker.

This change adds a check for TS_FIGHT_ME_VICTIM before hatred is reduced, so that if a player attacks a monster, it will aggro no matter what (i.e. no more monsters sitting around letting themselves be killed).

Fixed up the spacing around the area I edited a little.

### DIFF
--- a/kod/object/passive/brain.kod
+++ b/kod/object/passive/brain.kod
@@ -905,6 +905,7 @@ messages:
          % If they already have a lot of people chasing them then run away,
          % unless they attacked us.
          if NOT (event = TS_FIGHT_ME_VICTIM)
+            OR Send(Send(SYS,@GetSettings),@AlwaysCheckMonsterChasers)
          {
             iHatred = iHatred - (Send(what,@GetMonsterChasers)
                                        * piChaser_factor)/100;
--- a/kod/object/passive/brain.kod
+++ b/kod/object/passive/brain.kod
@@ -891,21 +891,31 @@ messages:
       {
         iHatred = iHatred + 20;
       }
-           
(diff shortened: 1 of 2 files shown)


jrotunda:
Thanks for working on this!  

Something a lot of players (and self, even building) always wanted to see and would be a great way to enhance the PvE and way the AI works, and definitely make the game a little scarier.  This is how I recall the monsters functioning prior to the Revelations update, but I could be wrong.  

Have you noticed any adverse changes in game play with this in effect, and should it perhaps be couple with a change to how 'bait' works?

skittles1:
I'm not exactly sure when this AI system was written, but I'd guess from the code style maybe 1998. It definitely does make a difference in gameplay and enhances PvE as any mob you attack will fight back against you (unless it already has a target and has more hatred for that target).

No adverse effects noted, although the AI/monster movement is very different on 103 now (mobs move a grid of 16 finerow/col instead of 1 row/col, and they have a vision range). Casting EQ, Final Rites or Bait is not advised if you're surrounded by mobs as they will all (if they pass hatred check) attack; our Bait has a smaller range (16 row/cols) but EQ and Final Rites have the same range.

Meridian59:
This sounds great in theory but I'd like to have a way to turn it on and off in case something weird happens.  Can you add a new Setting to enable this (on by default is fine)?

skittles1:
Added setting, if enabled it will check aggro level even if the player attacks the monster.

PR #321, opened 2015-03-10
Server-side fix for the black color of playernames in the wholist

Nagath:
by cyberjunk
unset DRAWFX_INVISIBLE (=OF_INVISIBLE) bits in playerlist


Meridian59:
1) Add a comment in the code as to why this is there

2) Can you check on the  unrelated change around line 4334?  Looks like a character encoding issue.

davidklaw:
If this is desirable we can pick up where previously left off. If not, I move to close this PR.

PR #323, opened 2015-03-11
Migrate RoomEditor to VS2013/OWLNext (old)

cyberjunk:
Outdated, see: https://github.com/Meridian59/Meridian59/pull/324


Daenks:
Does this contain the heightmap changes we made a while ago?

cyberjunk:
---

PR #324, opened 2015-03-12
Migrate RoomEditor to VS2013/OWLNext

cyberjunk:
**Introduction**

---

This my cleaned up pullrequest of the room editor upgrade from the 103 branch.

It will migrate the code to compability with the following two compilers:
1) Visual Studio 2013 (by roomedit.sln - using OWLNext - builds roomedit.exe)
2) Borland C++ 5.02 (by makefile - using OWL 5.x - builds windeu32.exe)

The old BCC 4.52 support has been dropped.
BCC 5.02: https://googledrive.com/host/0B-6rsj6uHlolV2ZGMjNQLUZiLXc/bcc502.zip

---

**Commits**

---

**PART I:** 14fc309
Adds the OWLNExt sources 6.34.3 (there was issues with 6.40).
Downloaded here: http://sourceforge.net/p/owlnext/code/HEAD/tree/tags/V634_03/
Unrequired parts have been ripped away, also the '#pragma' to force link against a specific named lib was removed from owllink.h. Note: You must build OWLNext and roomedit with OWL5_COMPAT set.

---

**PART II:** 6b79b0a
Add VS2013 / OWLNext and BCC 5.02 compatibility

--- 

**PART III:**  3a0afe0
Fixes for issues/crashes, compiler warnings + minor improvements

---

**PART IV:**  211bee2
Additional high resolution grid/heightmap in ROO file. A change we have in the 103 room editor since months... This commit just adds the writer for the ROO file grid. Screenshots of highres/heightmap:
http://wiki.openmeridian.org/index.php/High_Resolution_Grid,_LoS_and_monster_movement_upgrade
*Note: This is just ignored in blakserv unless further implemented*

--- 

**PART V:**  a92f195 + 4395dd2
a92f195 : Merges all remaining changed lines between the 103 pullrequest and this here. This is mostly space characters, new lines - except for more nullptr safe access on the StatusBar gadgets.

4395dd2 : Updated BCC 5.02 (windeu32.exe) and VS2013 (roomedit.exe) binaries

---

**PART VI:** Additional fixes:

3677188: Fixing char[] overflow for deep paths

4506feb: Increases zoom in to edit tiny linedefs

0fe85c4: Fixes an issue in the BSP builder. In 'BSPBuildTree()' (bspmake.cpp) it starts the algorithm with a rectangle (=boundingbox of the room). Any area part of this start-polygon/rectangle, but not part of any sector, is still incorrectly added to sector 0. Also causing function like 'BSPFindLeafByPoint()' incorrectly returning this sector for outter areas. See the following images for such an outter area incorrectly added (this is fams, sector 0 is the bar around paddock)
OLD CLIENT: https://docs.google.com/file/d/0B-6rsj6uHlolVEx1dzVwX2ZuSGM
OGRE3D: https://drive.google.com/file/d/0B-6rsj6uHlolbUY5Nmx0enJRUlU

829a121: Fixes following issue: Sometimes when you unset a side in the roomeditor, save the ROO and load it again, the unset side will appear again (also creating roomeditor warnings and errors on check features). This is due to the roomeditor using different/dedicated linedef/sidedef definitions, which were not saved unset.

1cfbcb0 / 64d9baf: Issue in writing roomeditor own walls (save.cpp/load.cpp). EditorWall coordinates are SHORT in roomeditor, but ROO fileformat provides 4 bytes for them. Roomeditor read 4 bytes from a memory pointer proving coordinates as SHORT(2 bytes) (read also adjacent memory). Then this 16bit value and 16 bit garbage were written for each coordinate to the file. This fix will write "0" values into the unused 16 bit after each coordinate and handle them accordingly as unused payload.

1ea5d6a: Fixes an error introduced by the OWLNext port. The calls to SetTransferBuffer() in window.h must use the OWL5_COMPAT signature (void* argument). This fixes crashes when trying to edit multiple elements and the additional dialog is supposed to come up after clicking "OK" / Save.

932ca67: Fixes elements not correctly updated with new values after using the multi-edit dialogs


cyberjunk:
It's likely I will discover and fix more bugs in the next few days. I won't add updated .exe files, guess you're going to build it yourself anyways...

Nagath:
PART IV will be in fact the end of safe spot in future maps and if someone does this with the current ones in Meridian itself.
I'm sure people will be very unhappy about this!
To be honest it will also be the end of building beyond 130HPs because sadly theres no other way right now without investing incredible amounts of time. I don't like it how I have to build now to max HPs but its absolutely no fun to farm shadowbeasts for 6+ hours only to get a little hitpoint...

Meridian59:
Haven't gotten a chance to look at this yet; probably will be next week.

cyberjunk:
>> PART IV will be in fact the end of safe spot in future maps and if someone does this with the current ones in Meridian itself.

Please read the comment with that part, it clearly states:
>> *Note: This is just ignored in blakserv unless further implemented*

No safe-spot is touched by this pullrequest in any way.
This part is simply to keep the editors of 103 and 101 in sync..

Meridian59:
I'm ready to start importing these but I'd like to do it one commit at a time, starting with Part I (OWL library, no roomedit changes).  I will plan on building with VS 2013 and not using Borland at all.  I'll likely try to construct a VS makefile so that it can be built from the command line with a single command, just like the rest of the system.

Can you make individual pull requests so I don't have to take it all at once?

cyberjunk:
>> Can you make individual pull requests so I don't have to take it all at once?

I'm a bit busy this weekend.... If you don't want to merge the whole thing at once, have you tried "cherry picking" the first commit (and possibly the later ones after that in order) and apply it on your master? I assume this should work... this branch the pullrequest was created from is rather up to date with your master. Otherwise I can have a look at that next week....

>>  I will plan on building with VS 2013 and not using Borland at all. I'll likely try to construct a VS makefile so that it can be built from the command line with a single command, just like the rest of the system.

I was considering to do the same with the makefile, but decided to keep BCC 5.02 support a bit longer. If you drop BC and OWL 5 support fully, you can basically clean up some parts of the pullrequest (e.g. all the preprocessor OWLVersion > checks switchting code accordingly to OWL/compiler)

Meridian59:
Cherrypicking would be painful on my end because git is a
not-quite-identical mirror of our SVN repository where the official version
of the code is stored.  i.e. I don't actually do development with the git
version of the code.  If you don't mind sending these one at a time when
you can, it will be easier.

Meridian59:
A few questions as I start looking at Part II:

- You renamed most of the include guards in the roomedit includes of OWL headers, but I think they can be entirely removed.  Yes?

- Can you say why you're loading zlib dynamically instead of linking it in?  Having it dynamically linked means the path needs to be correct or it will crash.

- I'd prefer to remove all of the OwlVersionchecks.. what are the implications?  i.e. what version is required if  OWLVersion > OWLVERBC502?

- Did you try just #defining HUGE, FAR, far, etc. to nothing?  Much easier than changing all occurrences (and I'm surprised that the OWL headers don't already do this).

- What were the problems with using OWLNext 6.40 ?

cyberjunk:
>> You renamed most of the include guards in the roomedit includes of OWL headers, but I think they can be entirely removed. Yes?

I think turning something like *#ifndef __OWL_CONTROL_H* into *#ifndef OWL_CONTROL_H* is part of the migration. I don't exactly remember, but even in the oldest OWL in BCC 4.5 up to the newest OWLNext in the headers the guardians are written without the "__" prefix.

Here's my two original commits, when working on this on 103, I noted something that this is required for VS:
https://github.com/cyberjunk/meridian59-orig/commit/c5632e6f6e51c4154e31478545f8f0a13432a41b
https://github.com/cyberjunk/meridian59-orig/commit/542f13258123c9c8ad3e62f6f378cdc4dabfff32

I can't tell you if you can remove them at all....

>> Can you say why you're loading zlib dynamically instead of linking it in? Having it dynamically linked means the path needs to be correct or it will crash.

This is because I wanted to support both, BCC 5.02 and VS 2013 they would/currently need different variants of the zlib.LIB (the OMF thing). We may want to update zlib one day, and maintaing both of the LIB variants could get ugly, so I decided for the dynamic linking here. Dropping BCC 5.02 completely I would rather definitely link at compile-time again.

>> I'd prefer to remove all of the OwlVersionchecks.. what are the implications? i.e. what version is required if OWLVersion > OWLVERBC502?

If you drop BCC 5.02 support, you could remove these. Guess all you have to keep is the code for the case of (OWLVersion > OWLVERBC502), what would be the OWLNext/VS code.

>> Did you try just #defining HUGE, FAR, far, etc. to nothing? Much easier than changing all occurrences (and I'm surprised that the OWL headers don't already do this).

No I did not try. Your suggestion might be easier, but is it cleaner for further work on the editor?

>> What were the problems with using OWLNext 6.40 ?

Phew, it's weeks ago, let me try remember: It's about the OWL backward compatibility. OWLNext changed their event handling/response table stuff, but the roomeditor needs the old style (look for block of DEFINE_RESPONSE_TABLE1, for example in lineedit.cpp). OWLNext has a flag OWL5_COMPAT to support these old fashinioned response table definitions. This backward flag seemed to work better with OWL 6.34. 6.40 still has this backward compatibility, but if your response handler signatures are not correct, then you will get an very weird error within an OWLNext header file (caused by the generics in there), but you won't get a clue which of your respone handlers is wrong. 6.34 performs much better here (I assume in many cases of OWL5_COMPAT). If you try to use 6.40, I recommend messing up like one of the ButtonDown/Up/.. signatures slightly (remove const or something), and see the error you are faced with.

cyberjunk:
>> What were the problems with using OWLNext 6.40 ?

>> Phew, it's weeks ago, let me try remember: It's about the OWL backward compatibility. OWLNext changed their event handling/response table stuff, but the roomeditor needs the old style

One more note on this, you may find this information useful:
http://sourceforge.net/p/owlnext/wiki/Upgrading_from_OWL/

See changes mentioning OWL5_COMPAT and especially the part:
>> *Changes in the message dispatch framework*....

>> In particular, OWLNext 6.40 introduced a major overhaul of the message dispatch machinery, and many event handler signatures were updated in the process....

I think unless you do the additional migration work to get rid of OWL5_COMPAT mode completely, version 6.34 is the better choice...

Meridian59:
I took most of the changes except for Part IV.

Many thanks!  I never thought the day would come when this would build with MSVC.

PR #327, opened 2015-03-22
Visual Studio 2013 solution & projects

CyberBill:
Added VS2013 solution and project files, made changes to allow keybind
project (m59bind.exe) and updater project (club.exe) to build.  I also
fixed a number of compiler warnings that would show up with warning
level 3.  Solution builds in Debug and Release.

Both the client and server are included in the solution, either one can be launched and debugged.

Build times are reduced pretty substantially.  Using nmake, both the release and debug versions take me 2m40s to build.  This solution builds in 1m32s in Release and 2m15s in Debug - mostly due to it using the Debug version of BlakComp to build the KOD files.


Meridian59:
I really don't want to have project and solution files---it's impossible to keep them in sync with the makefiles, and I need the makefiles for Linux.  Can you send a pull request with just the code changes?

CyberBill:
I agree that it's a pain having to keep them in sync - but I also want to express how much it's still absolutely worth it, by giving you a short list of two reasons:
1) More developers will contribute to the project if they can use Visual Studio - particularly for debugging
2) The VS workflow allows a MUCH faster iteration cycle, so developers can get more stuff done faster

If it will help sway you, I'd also be happy to make a wrapper VS project that builds the entire makefile from within VS for verification (two of the projects just wrap makefiles already), or a wrapper makefile that builds the VS solution for verification, so that verifying that the build works for both is no more work than verifying an existing pull request.

Meridian59:
I wouldn't mind a makefile target that generates VS project files from the
makefiles.  That way I wouldn't have to maintain the project files as the
makefiles change.

CyberBill:
I looked around a bit and there doesn't seem to be any way to generate them automatically from make files. :(  The change I just committed adds a "makefile Build All" project that is normally not built, and you can specifically build it in Release, Debug, or Final, and all it does is run "nmake all FINAL=1" for example.

If your workflow is going to be always using the makefiles, I'm happy to maintain the whole VS side of things, and now I can easily verify that the makefiles are still working.

PR #329, opened 2015-03-24
Fixing club and keybind, fixing warnings

CyberBill:
This fixes club & m59bind executables (so long as you have a Pro version
of Visual Studio), and also fixes a number of compiler warnings that
happen with warning level 2.


davidklaw:
@Meridian59 Is this something you're interested in? If so, @CyberBill are you up for resolving the conflicts?

akirmse:
I think this is probably obsolete.  Everything builds without warnings these days.

PR #332, opened 2015-09-10
Fix broken .wav headers.

skittles1:
Some .wav files would not open in Windows Media Player due to what
appears to be a broken file header. Replaced with a header from a
working .wav file. Sounds still play in client, and can be opened with
WMP.


davidklaw:
I'm not finding an issue with these particular files opening in WMP 13 under Windows 11. I think that can be closed and then re-opened if @skittles1 confirms this to still be an issue.

skittles1:
> I'm not finding an issue with these particular files opening in WMP 13 under Windows 11. I think that can be closed and then re-opened if @skittles1 confirms this to still be an issue.

I can't reproduce it now either (WMP or VLC, Win10). It could have been specific to Win7 and whatever version of WMP I was using in 2015.

PR #333, opened 2015-09-13
Grammar and spelling fixes in the welcome letters for factions.

HoldOffHunger:
Moved the princess message text around to maintain what appeared to be a 92 char max on e-mail line length.


Meridian59:
Thanks for the fixes!  BTW both worshipped and worshiped are correct.

PR #335, opened 2015-10-20
Add admin cmd to clear unused accounts and compact IDs.

skittles1:
Ported from https://github.com/OpenMeridian/Meridian59/pull/1213.

We recently added this on 103 as 14% of our accounts were never logged on. We though that this might be much more useful if run on 101/102 given how many accounts must be present after a decade.

The admin command "delete unused" will remove any unused accounts (i.e. any that have never been logged on before) and compact the account numbers to reclaim previously used ones.

Kicks off all users before carrying out the delete/compact to avoid any potential issues with sessions.

This took about 40 seconds to run with ~3500 accounts (deleting ~550).

### DIFF
--- a/blakserv/account.c
+++ b/blakserv/account.c
@@ -541,3 +541,46 @@ void DeleteAccountAndAssociatedUsersByID(int account_id)
 	      account_id);
    }
 }
+
+// Iterate through accounts and remove unused ones.
+void DeleteAccountsIfUnused()
+{
+   account_node *a, *temp;
+
+   a = accounts;
+   while (a != NULL)
+   {
+      temp = a;
+      a = a->next;
+      DeleteAccountIfUnused(temp);
+   }
+}
+
+// Deletes an account and associated users if account has never been logged in.
+void DeleteAccountIfUnused(account_node *a)
+{
+   if (a->last_login_time == 0)
+      DeleteAccountAndAssociatedUsersByID(a->account_id);
+}
+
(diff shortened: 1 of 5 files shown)


Meridian59:
This looks OK, but I'm curious how it came up.  Were the extra accounts
causing any problems?

skittles1:
Someone used our webpage account creation to create ~50 accounts to annoy us, which also alerted us to the 14% of accounts having never been logged in. We thought this addition would be useful in the future if that happens on a more serious scale, but we've run it already on 103 to remove the unused accounts (cleans up a couple thousand objects/list nodes).

Meridian59:
I'm sure we have a zillion dead accounts on 101, and FWIW it hasn't caused
any problems.  I'll think about taking this change though.

davidklaw:
I see this branch has conflicts now. Any interest in reviving this?

davidklaw:
@Meridian59 Bumping this. Would this be desirable as core functionality, not necessarily for 101/102?

akirmse:
I've been a little hesitant because this change actually renumbers accounts.  That would make logging less reliable.  We don't use logs much but I could see how someone might want to.

The alternative is to not renumber accounts, but then the server's statistics about the number of accounts would be wrong.

davidklaw:
Ah, yes. I'm thinking close on account of it being abandoned and until a "safer" alternative might come along.

akirmse:
Yeah, agreed.  Not a bad idea though.

PR #337, opened 2015-12-27
First pass at getting blakserv to run on linux

ckirmse:
- make separate osd_windows.[ch] and osd_linux.[ch]
- use epoll on linux
- removed files.[ch] and put them into osd_ files
  - file/directory functions restructured to make easy to get
    to work on linux
- more cleanup would certainly be nice, but things basically
  seem to work on windows still and now on linux
- blakod references to roo filenames have case sensitivity
  problems; that still needs to be fixed
- I've wanted this to exist for about 15 years. Wow!


sicancode:
WOW awesome!! good job Chris!! :))

PR #339, opened 2016-02-08
Translate into German

Rodrigez01:


Daenks:
@Rodrigez01 You've targeted the wrong repository with this pull request I think. This attempts to merge the entire fork into the original repo.

PR #340, opened 2016-03-29
Fix sizeof statement.

skittles1:
Fixed a sizeof statement that was using the wrong value.


Meridian59:
What about the same issue around line 261 (AP_REQ_GAME handler)?

skittles1:
Thanks, missed that one.

PR #341, opened 2016-05-30
Play a sound when you receive a tell

GarMeridian3:
This plays a pre-existing wav (that is not used in the game as far as I
know) whenever the user receives a tell. This is a much needed feature
for those who are AFK, looking at other windows, and so on.

Definitely open to using other sounds. This one is the most appropriate I
could find within existing resources.


davidklaw:
@Meridian59 Do you care for this type of feature? If open to it, is it safe to assume something like this also needs to include an option under Preferences?

akirmse:
I've never heard anyone ask for it.  It would definitely have to come with a client setting to turn it off.  Actually, this should probably be done completely client-side.

PR #342, opened 2016-06-01
Fix StringSubstitute errors.

skittles1:
Some of the string terminations in StringSubstitute were being placed 1 char ahead of the correct position, leading to incorrect string lengths.

I left out the formatting changes/cleanup so the functional changes could be seen, if you like I can add the other changes from https://github.com/OpenMeridian105/Meridian59/commit/527efa8f19471236158cd7783c7a543c9af6ca46


Meridian59:
Looks good.  An alternative is to increase the size of the memcpy by 1, then there doesn't need to be an explicit set to 0 at all.

PR #344, opened 2017-03-04
HP after 160

Maikeljr:
So far I get toughers at 160, but no 161 or even more training points. So I'm guessing training points won't work due to not being at "max" but 160 is my unaltered max? Thought we were allowing stam bonuses to work for max hp.


akirmse:
3000 commits?  I'm guessing this is the wrong repo.

PR #349, opened 2018-08-09
Hold, Blind, and Dazzle visual effects

Meridian59101102103:
This preliminary commit includes animated projectile graphics for Hold,
Blind, and Dazzle. It also includes an animated overlay for players to
visually display the binding rings of purple energy from being Held.

Players who are Held, Blind, or Dazzled will also glow purple, green, or
white respectively, letting other players see that they are afflicted -
and when the affliction wears off.

Note: I am still working on figuring out the differences between codebases. This already works on 103, and I am pretty sure this will work on 101/102, but I can't be 100% certain right now until I finish setting up my test environment. I needed to make this pull request now to make sure my Github is set up correctly. I will be testing and adding tweaks over the next few days until it does. (this is Gar from 103)

We have more to come if these are approved, including projectiles actually bouncing off of a visual magical barrier when the bounce spell Deflect happens.

The effects can be viewed in action here: https://imgur.com/a/cpLbGCK


OfficialFON:
This. Is. So. Cool.

Meridian59101102103:
I've run into a problem (which was actually my hurdle before) - this pull now loads properly with no errors, but I can't get a local client of this repository's build to see the resources in its own localclient/resource folder. It loads black, all resources missing.

Do I need set something somewhere for the localclient to recognize its own resource folder?

Meridian59101102103:
Ok, I am up and running and can confirm the committed projectile changes work properly. The hold rings even display, albeit a little lower and with different layer order; problem - they do not disappear when hold ends. Working on that now.

Unfortunately, I had to put my graphics in the client's /resource folder manually. I put them in the Meridian59/resource/ folder but they did not copy up apparently. Where do I put them to have them included in a build? Researching now.

Meridian59101102103:
Squashed commits after I was done

Meridian59:
Thanks for this.. we're right in the middle of dealing with a timer overflow right before the Steam launch, so this may take awhile to look at.  Supporting Steam is also going to make production updates harder.

The new "premadebgf" directory looks unnecessary and I'm sure we want that removed.  The right way to do this is to build the bgfs by editing resource/graphics/makefile, and adding the new textures in the resource/graphics directory.

tsShergill:
@Meridian59 I know the 105 code is now vastly different from the official code. But I know @skittles1 has helped addressed many issues/bug fixes. He may be able to help out.

OfficialFON:
Ahhhh I had seen the adv. hall globe was popping the 2010 posts to the top just recently - what bad (or good given it didn't happen day-of) timing for that to crop up.  Hope the fix doesn't give anyone some sleepless nights!  Look forward to popping in on "launch" day!

skittles1:
@tsShergill looks like the server/client time fix has been merged already, but @Meridian59 don't forget to run the command `send c player UpdateTimeValues` to reset certain references to time kept by the Player class (same thing that was done last time, see #299).

Meridian59:
Also, could you please not use the trademarked name "Meridian" in your
username?  Especially with the Steam release, we want to be very clear
which are the official servers, and which are the user servers that are
running with our permission.

akirmse:
Obsoleted by a newer PR.

PR #354, opened 2018-09-20
Made broadcast mana penalty a setting that can be changed at runtime 

treymd:
While the cost to broadcast can be useful to prevent flame wars and spamming, it also reduces player ability to ask questions and socialize.  (Puts a damper on trivia contests too!)  This value can be changed to tweak the penalty in real time by an admin

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -3227,7 +3227,7 @@ messages:
 
       if NOT Send(Send(SYS,@GetParliament),@BetaPotionsEnabled)
       {
-         iCost = Bound(piMax_mana/2+1,11,$);
+         iCost = piMana * Send(Send(SYS,@GetSettings),@GetBroadcastManaCost) / 100;
          if piMana < iCost
          {
             Send(self,@MsgSendUser,#message_rsc=player_cant_broadcast);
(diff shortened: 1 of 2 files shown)


kod/util/settings.kod
```diff
@@ -343,5 +349,9 @@ messages:
       return pbAlwaysCheckMonsterChasers;
    }
 
+   GetBroadcastManaCost()
```
Meridian59: Since this is a percentage, a better name is GetBroadcastManaPercent.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -3227,7 +3227,7 @@ messages:
 
       if NOT Send(Send(SYS,@GetParliament),@BetaPotionsEnabled)
       {
-         iCost = Bound(piMax_mana/2+1,11,$);
+         iCost = piMana * Send(Send(SYS,@GetSettings),@GetBroadcastManaCost) / 100;
```
Meridian59: You removed a call to Bound here that previously set the minimum mana required to broadcast to 11.  Without that, suppose piMana is 1.  Then as long as the setting is < 100, the mana cost is 0. Oops...

My suggestion is to put the minimum bound on, and possibly add a special case for when the setting is 0 that bypasses the bound.  Thus an admin could make broadcasts free if desired.

treymd: done (including using piMax_Mana and not piMana to base the cost on, which was not intended).  I also set the default penalty percent to 8% instead of 0 to start with.

kod/util/settings.kod
treymd: done

treymd:
Meh, I'll need to reopen this since I amended the commit (I know, bad!)

PR #355, opened 2018-09-20
Made broadcast mana penalty a setting that can be changed at runtime

treymd:
(Resolved issues with previous  pull request)

While the cost to broadcast can be useful to prevent flame wars and spamming, it also reduces player ability to ask questions and socialize. (Puts a damper on trivia contests too!) This value can be changed to tweak the penalty in real time by an admin


akirmse:
Better, but please make the default 50%, which matches the current behavior.

treymd:
Done.

PR #358, opened 2018-10-10
Hold, Blind, and Dazzle visual effects

Gar103:
This pull contains, fully implemented, and toggleable with two switches
in Settings:

Under pbUseCCProjectiles:
* Purple rotating ring projectile for Hold
* Green alternating ring projectile for Blind
* White alternating star projectile for Dazzle

Under pbUseCCGlow:
* Animated purple rings around a Held player
* Both players and monsters will glow white if Dazzled, green if
Blinded, and purple if Held, in that order of increasing priority. This
way, modern players will be able to see combat's most ubiquitous spells
being cast, and can also tell who is disabled.

Note that monsters will not show the purple hold rings because monsters
are variable sizes. Players are a standard size.

If Hold or Blind is cast by a weapon proc, the projectile is fired from
the wielder.

Monsters casting these spells will also launch projectiles correctly.


OfficialFON:
Looks like I gotta find some players to blind on 103!  Love the glow effect but I want to see the pretty rings!

davidklaw:
While testing, after being hit with blind, this is how my vision returned. It returned to normal after some movement, but something is clearly amiss.

image

I also noticed that even though I had copied the new resources files to my localclient folder, these new effects aren't rendering.

https://user-images.githubusercontent.com/467443/177198937-8bc1990d-7793-409e-b9b7-ebb2a71c947d.mp4

I can try again if I get some instructions for testing.

And while I can't see the combined effect, my initial impression is that the glow is a bit much, but I'll reserve judgement for when these other resources are rendering.

akirmse:
Did you copy both the new .bgf files and changed .rsc files to your resource directory?

Also, try running the client compiled in debug mode.  The debug window will give more information about missing files, etc.

davidklaw:
That was helpful, thanks.

This seems like a neat idea in theory. Personally, I think this sets Hold too far off visually from anything else in Qor or otherwise. Likewise, the green glow of Blind simply feels out of place _to me_.

I did notice a couple of issues while testing:

* If you're far enough away, the visual just simply does not trigger. My target was at the Tos fountain while I had my back to the doors to the Duke's (see video below)
* In some cases (related to distance, I think) the Hold projecticle doesn't move from the caster (see video below)
* Hold begins immediately but the projectile travels, so it makes for poor player experience. I was using 99% spell but I assume that in some cases a target's hold might dissipate before the visual projectile even reaches them (see video below)

**Blind example**

https://user-images.githubusercontent.com/467443/177205692-591fdbfc-1ff6-4033-92de-ca2e972f9e64.mp4

**Hold example**

https://user-images.githubusercontent.com/467443/177205693-1034bc33-1918-4307-8ba6-e0405870b440.mp4

**Dazzle example**

https://user-images.githubusercontent.com/467443/177205695-7755d943-9845-4529-b049-6d186d601713.mp4

**Hold example(s)**

https://user-images.githubusercontent.com/467443/177205710-eba2461c-7356-4779-b63d-4de3c8115a63.mp4

https://user-images.githubusercontent.com/467443/177205718-42407473-188b-48b2-855f-6cbf344c2c87.mp4

**Hold long distance example**

https://user-images.githubusercontent.com/467443/177205728-e944c13c-6c9d-4690-afd4-e7786adb9a17.mp4

**Hold longer distance example, no visual**

https://user-images.githubusercontent.com/467443/177205739-b8cb07d0-03f3-416d-9c6c-7683c59d6658.mp4

**Hold example, broken projectile**

https://user-images.githubusercontent.com/467443/177205749-0fcdb255-b0d8-4597-ba3e-03a448fee83b.mp4

davidklaw:
@Meridian59 There's a few things about this one that just don't quite cut the mustard. The Dazzle effect is neat, but the Hold and Blind effects seem a bit too "unique" and the Hold effect doesn't sync up with the actual player experience (it's a projectile and hold is immediate). Cool ideas here, but maybe close this one?

akirmse:
Sounds like a pretty cool idea.  I'd be up for taking another look in the future if some of the problems you found are fixed.

jrotunda:
In case this comes up again in anyone's search: These visuals look good, the Hold one in particular, and it would be cool to have these tweaked a bit.

- The dynamic lighting can be toned down (if not removed) to 1-5 out of 255
- The projectile should perhaps be nixed on Hold given its short duration, but keep the animation of it around the held player. 
- Lightning bolt projectile anim speed seems like it would fit best for the Dazzle user experience, which lasts longer.
- Blind should be dark blue 
- Design: Blind seems strange with a glow: perhaps the player characters eyes get blanked out while in effect? (admin_icon_blank)
- For testing: Does this also work when the Kara'hol's Curse hold turns on, and for Gaze of the Basilisk Held Battlers? Does it work for all mobs the same?

PR #359, opened 2018-10-28
Sortable Vaults and Chests

marcelkauf:
Users can sort items in their vaults and chests.

### InsertListElem
The first three commits have been cherry-picked from the 105/112 code base with only minor formatting changes. They add an `InsertListElem` function to kod that inserts an element into a list at a specified index. They only consist of technical changes.

#### Functionality
This dialog can be opened via the new `vault` command (or `shift+v` with the new key bindings). A screenshot is attached below (German version only, sorry for that). It has four arrow buttons to move the selected item to the top, bottom, previous, or next position. They are enabled depending on the position of the currently selected item. The _previous_ and _next_ buttons keep the item selected, so that it is always visible in the item list. The _top_ and _bottom_ buttons do not keep the item selected, so that it leaves the visible window of the item list if the list is too long. This has been done because the user probably wants to move more items to the top or bottom and scrolling back to the old position would be annoying.
If the chest or vault is empty, a message is displayed to the user (either a tell from the vaultman or a system message).
vault
#### Implementation
The implementation is similar to that from the _buy_ dialog. With the vault command, objects marked with `OF_CONTAINER` or `OF_VAULT` are searched for. If there is more than one, the default dialog for chosing an object is displayed. The client then sends a `BP_REQ_VAULT` message containing the vaultman or chest object to the server. The server sends a `BP_VAULT_LIST` answer containing the vaultman or chest object and the item list. The user can sort these items in the dialog (`vault.c`). They are sent back to the server in a `BP_REQ_VAULT_ITEMS` message in the new order along with a number list for the number items. The server checks all items against the items stored in the corresponding vault (`holder.kod`). All items must be present in both lists (with the same quantity for number items). If there is a difference, the new order is not accepted. If both lists contain the same items, the item list is replaced by the new list (`monster.kod` and `safebox.kod` for vault, `storebox.kod` for all kind of chests).

### Cluster Items when Put into a Vault or Chest
If a user puts a (non-number) item into a vault or chest, the item is inserted before the first item of the same class. With this change, there is no need to re-sort a vault if a similar item has already been stored. If there is no such item, the item is added to the top just like before. I wanted to keep it simple and therefore refrained from adding more logic like clustering weapons etc. Technically, `safebox.kod` and `storebox.kod` override a new message `HolderDoAddNode` that is called when inserting a non-number item or a number item if it is not already present.


davidklaw:
@Meridian59 I'd say that this can be closed on account of #394 being closed.

CenWulf-Keltic:
Now that everyone loves the ability to sort their personal inventory without dropping on the ground, could we maybe re-open this and also see if we can do the same with chests and vaults?

PR #360, opened 2018-10-28
Consistent Buy, Offer, Deposit, and Withdraw Dialogs

marcelkauf:
The main intent was to fix the size of the withdraw dialog, which has never been made bigger like the other dialogs. All dialogs have also been made as consistent as possible in terms of wording and alignments. German versions below (I have no idea how to change the language to English in Meri).

Buy
buy

Offer
offer

Deposit
deposit

Withdraw
withdraw


davidklaw:
Just so we're clear where we're coming from, this is a lineup of what I see today:

image

davidklaw:
What about something like this? Not necessarily final, but wondering if this sparks any new conversation around this proposed change.

image

akirmse:
I'd be OK taking this as-is, since it also fixes some German typos.  I was trying to see if it's possible to run the workflow first to make sure it compiles.

davidklaw:
I didn't translate all of it yet, but a few translations and some rusty high school German didn't yield anything controversial. Can you translate it well enough to be good with it? 😆

Other than that, I was curious how to enable the German strings, so that I could test how these characters are rendering in game?

akirmse:
You have to put Windows in German mode, which involves installing some kind of language pack.  Not ideal, obviously.  We never wound up translating the server-side strings, so maybe at some point we should take this out, not sure.

Issue #363, opened 2019-01-24
Hardware Renderer doesn't display override depths correctly


Gar103:
Right now I believe Override Depths are only used in the Riija Temple, but I'm trying to use them to create dynamic 'swimming' and 'flying', and I found out that the hardware renderer only displays override depth heights at height 0. As in, a player / item / monster will be drawn at height 0 no matter what you set override depth to, even if that means the player / item / monster is *under the floor.* This means Riija Temple bridge has been broken visually for years and nobody ever noticed because they weren't on one character looking at another crossing it.

I'm guessing that always drawing the object at height 0 means the error's actually pretty simple, maybe a 'return 0' somewhere or something, but I don't know where hardware renderer works with the override depth functions. Anyone have any ideas?

jrotunda:
Addressed by author in PR#365 and can be closed when reviewed.

PR #364, opened 2019-02-16
Fix Override Depth issue in Hardware Renderer

Gar103:
An error in the calculations for Override Depths caused objects on
overriden sectors to be given negative heights and thus drawn at height
0 in hardware renderer mode. This meant the Riija Bridge displayed
incorrect visuals from a 3rd party point of view (i.e. if you watched
another player crossing) since the inception of hardware renderer mode.

With these equation changes, objects will be drawn at the correct
height. Furthermore, case handling was added for OF_HANGING objects like
Kriipa and OF_BOUNCING objects like dragonflies in flight so that they
will also work correctly on overriden sectors.


Gar103:
ref https://github.com/Meridian59/Meridian59/issues/363

clientd3d/d3drender.c
```diff
@@ -3316,6 +3316,7 @@ void D3DRenderNamesDraw3D(d3d_render_cache_system *pCacheSystem, d3d_render_pool
 	for (list = room->contents; list != NULL; list = list->next)
 	{
       float glyph_scale = 255;
+      bool bUsingAlternateDepth;
```
Meridian59: Add a comment: what is alternate depth?

Also, move declaration of variable to its first use.

clientd3d/d3drender.c
```diff
@@ -7483,6 +7540,7 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
 
 					// Set object depth based on "depth" sector flags
 					depthf = sector_depths[SectorDepth(sector_flags)];
+               bUsingAlternateDepth = FALSE;
```
Meridian59: "false"

clientd3d/d3drender.c
```diff
@@ -7492,27 +7550,51 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
 							if (ROOM_OVERRIDE_DEPTH1 & GetRoomFlags())
 							{
 								depthf = GetOverrideRoomDepth(SF_DEPTH1);
+                        bUsingAlternateDepth = TRUE;
```
Meridian59: Use "true" for booleans, not TRUE. Also, since this is set in every branch of the switch, it can be moved up to line 7546 and only set once.

clientd3d/d3drender.c
```diff
@@ -7492,27 +7550,51 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
...
+                        pRNode->motion.z), (float)pRNode->motion.y);
+                  }
+                  else
+                  {
+   					   MatrixTranslate(&mat, (float)pRNode->motion.x, depthf, (float)pRNode->motion.y);
+                  }
+					}
+               else
+               {
+   					MatrixTranslate(&mat, (float)pRNode->motion.x, (float)max(bottom,
```
Meridian59: As before, use a yoffset variable so that MatrixTranslate can be called only once.

Gar103:
Test instructions:  take 2 accounts to the Riija Temple in hardware renderer mode. Walk one character across (or drop an item or make a mob) and, to the other character, they aren't visible at the height of the bridge. Instead, if you look down, you'll see them on the water (height 0) because of the calculation error.  Kriipa (OF_HANGING) and dragonflies in flight (OF_BOUNCING) don't work right either.

Once the fix is in, everything will look as it is supposed to. The monster/player/item will be at the height of the invisible riija bridge.

Meridian59:
OK, thanks.  Let me know when you've addressed the other comments.

clientd3d/moveobj.c
```diff
@@ -232,21 +233,60 @@ Bool ObjectsMove(int dt)
...
+		   depth = sector_depths[SectorDepth(sector_flags)];
+         if (ROOM_OVERRIDE_MASK & GetRoomFlags()) // if depth flags are normal (no overrides)
+         {
+            switch (SectorDepth(sector_flags))
+            {
+               case SF_DEPTH1:
+                  if (ROOM_OVERRIDE_DEPTH1 & GetRoomFlags())
+                  {
+                     depth = GetOverrideRoomDepth(SF_DEPTH1);
+                     bUsingAlternateDepth = TRUE;
```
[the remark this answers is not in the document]
Gar103: The way it's coded, the If is a room flag, but the switch is based on the specific sector, so much of the time (i.e. when the object's not on the special sector) it's actually going with none of these cases - so I can't move bUsingAlternateDepth = true to 246

Gar103:
I've made the requested changes and squashed it into the original commit

clientd3d/d3drender.c
```diff
@@ -3371,18 +3372,21 @@ void D3DRenderNamesDraw3D(d3d_render_cache_system *pCacheSystem, d3d_render_pool
 				if (ROOM_OVERRIDE_DEPTH1 & GetRoomFlags())
 				{
 					depth = GetOverrideRoomDepth(SF_DEPTH1);
+               bUsingAlternateDepth = false;
```
Meridian59: Are you sure that "false" is correct here, and in the other two cases below?  It is never set to true now.

Jatrott:
are these issues still being reviewed? I'd love to help get some of them merged into master.

Meridian59:
Yes, I still review these.  Looks like there are still comments to be addressed on this one in particular.

Jatrott:
Better question more broadly (sorry its in comments, no DMs on GitHub) how can I help more broadly here?

akirmse:
Send small, contained changes.  Explain what they do and why they're
necessary.  Let me know how to test them.

clientd3d/d3drender.c
Gar103: Just saw this, sorry for the delay. Fixed.

Gar103:
Sorry for the delay, just saw this. I've fixed the boolean issue.

davidklaw:
Me again! It looks like the requested changes have been followed up on and this is mostly a ping to revive this PR, but I'll do my best to test tomorrow and provide some validated test results.

davidklaw:
I tested this at the Temple of Riija only. And boy was it a crash course in admin commands.

Before, in `master`:

Players appear to fall only to reappear on the other side.

https://user-images.githubusercontent.com/467443/175203244-6ae3efd0-96f5-4147-b6da-f7acdad0a96c.mp4

Kriipa's and Dragonflies fall to their immediate death.

https://user-images.githubusercontent.com/467443/175203259-c634a231-0236-4845-a5e8-40353775393a.mp4

https://user-images.githubusercontent.com/467443/175203414-87f13f89-cfdd-43f9-9d33-a0e2e9c53fec.mp4

After this change:

Players appear to walk along the invisible bridge.

https://user-images.githubusercontent.com/467443/175203630-e54fad24-91f2-498f-96b0-9b540fcf8eb4.mp4

Kriipa's and Dragonflies die immediately, without falling. While the result here technically makes sense, it looks odd.

https://user-images.githubusercontent.com/467443/175203753-db7e2089-e172-4b1a-aa5d-f17b98f2a4b4.mp4

It was brought to my attention that these changes ought to be tested in other rooms like `koc_apothecary` and anywhere with pits. I wasn't able to get to that in my tests.

davidklaw:
@Meridian59 Anything else I can do or add to this to help nudge it along?

akirmse:
Do you want to try testing in some of the other rooms first?  I like the Riija bridge improvement but I'd like to be more sure that this doesn't break anything.

davidklaw:
I'm not as familiar with the rooms as I was 20 years ago, so if someone can recommend some rooms where override depth comes into play I'd appreciate it. Otherwise, I'll see what I can figure out on my own.

clientd3d/d3drender.c
```diff
@@ -3394,10 +3399,35 @@ void D3DRenderNamesDraw3D(d3d_render_cache_system *pCacheSystem, d3d_render_pool
 
 		MatrixRotateY(&rot, (float)angleHeading * 360.0f / 4096.0f * PI / 180.0f);
 		MatrixTranspose(&rot, &rot);
-		MatrixTranslate(&mat, (float)pRNode->motion.x, (float)max(bottom,
-			pRNode->motion.z) - depth +
-			(((float)pDib->height / (float)pDib->shrink * 16.0f) - (float)pDib->yoffset * 4.0f) +
-			((float)pRNode->boundingHeightAdjust * 4.0f), (float)pRNode->motion.y);
+
```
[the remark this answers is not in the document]
davidklaw: I built on this PR with these thoughts. I'm not seeing a reason why this wouldn't make sense. https://github.com/Meridian59/Meridian59/commit/7987f89b460958ca121d33127f4953ace04eb674

davidklaw (commented):
I'm unclear on what exactly to test besides the Riija bridge to ensure I'm hitting on exactly what this PR might affect. So, I walk around the jungle triggering dragonflies and kriipa and moved around on sloped surfaces in the Ko'catan "city" rooms. I had already tested the bridge itself.

Aside from the bridge fix, I found no difference in how I see other players or bouncing/hanging monsters move before and after. Movement is a bit wonky, but it was before. But I'm also pretty that the type of movement recorded here is unaffected by this change because it's all within sectors at a depth of 0.

Examples:

**Original:**
https://user-images.githubusercontent.com/467443/181597167-2fe95ae0-c650-49b8-8834-0a42f4740ce2.mp4


**PR:**
https://user-images.githubusercontent.com/467443/181597044-9c2310c6-2586-48e2-8c86-b76928ffa5e1.mp4

davidklaw:
I've spent a decent chunk of today looking into this and have maybe reached my limit 😆. Here's where I'm at with it: I understand this change to introduce new logic to the rendered depth of room objects (most notably players and monsters). It looks for an overridden depth value, the only instance of which appears to be the Riija Template bridge height of 384 vs. the room sector height of 0. Per this PR, if this override value is detected, the object being rendered is offset from the original height value accordingly.

OF_HANGING and OF_BOUNCING types were added to this logic, so that creatures like kriipa and dragonfly would render at the overridden depth. The result is that rather than falling to the bottom of the ravine and dying, they die on the invisible space of the bridge (they do not cross). The handling of `OF_` monster types may not be needed. At least not for this lone instance. But it does make sense for other forks that utilize depth overrides.

After testing in several other places, like `kc1`, `kocpoth`, and the Ko'catan square, I didn't see any rendering issues. However, none of these areas should be affected because they don't attempt to override the depth.

I think there's opportunity here to refactor, which I propose [here](https://github.com/davidklaw/Meridian59/commits/update-override-depth). I'm not 100% confident in my understanding of this one, therefore I'm not 100% confident in my tests.

🤷‍♂️

jrotunda:
This is Ready for Review.

It doesn't appear to cause bugs with floating or flying mobs elsewhere - I was unsure if the override depth was tied to these monster classes and suggested the other tests on all pits and cliffs like a taskmaster and unintentionally caused a delay here. 

The issue is contained to this room.

This PR fixes this old visual bug, and will now make the Indiana Jones & the Last Crusade style, leap of faith Inviso-Bridge look freaking cool again when people walk across it with the right sword (and hilarious when they don't and suddenly DROP like a bad penny in front of their friends, family, pets, all now scrambling to steal that non-shrouded cow mask left on the bridge and coming up with a alibi.

A Riija Sword, this bridge and a risky, disarming duel between mages can once more be a thing (it wasn't a thing - but it should be)

It's kind of annoying to know that players were seeing people drop down and walk on water for two decades - if it's your character and screen you have no idea.

This was introduced in 2003 when the graphics engine was completely overhauled by MisterY. 

A lot of the quirky visual bugs in the game occurred from and only with using the updated rendering. 


This wound up taking away one of the bigger visual moments - many players who log on still use the old graphics especially for PVP for speed, simplicity and stability - the old engine and graphics are still the most stable. 

Graphic errors like this can get missed for so long people get used to them and think they're normal. 

@GarMeridian3 Thank you for submitting this added Bridge fix + the Wedding Ring so quickly when I first mentioned it ~

davidklaw:
A recent change has caused a conflict here:

The original PR:
```
depth = sector_depths[SectorDepth(sector_flags)];
bool bUsingAlternateDepth = false; // is kod overriding sector depth?
```

The incoming upstream changes when combined with these changes:
```
float depth = sector_depths[SectorDepth(sector_flags)];
bool bUsingAlternateDepth = false; // is kod overriding sector depth?
```

The change being the explicit `float` type. If @Gar103 isn't around, I can submit a new PR with the conflict resolved but am curious if the same level of testing is needed?

jrotunda:
"OF_HANGING and OF_BOUNCING types were added to this logic, so that creatures like kriipa and dragonfly would render at the overridden depth. The result is that rather than falling to the bottom of the ravine and dying, they die on the invisible space of the bridge (they do not cross). The handling of OF_ monster types may not be needed. At least not for this lone instance. But it does make sense for other forks that utilize depth overrides."

Comparing again to current:
Important effect that you noted - there are currently no other instances that utilize depth overrides where a monster would be around, and agree with your sentiment that OF_HANGING and OF_BOUNCING handling isn't needed, and would look better visually - the current way, they die and fall to the bottom as one would expect, having them float to the top trades one visual aberration for another.

Gar103:
I've got a 101 dev environment up and running again (mostly). What's the status here? Do I need to do something?

Gar103:
I don't understand why it's showing a merge conflict against nothing

Gar103:
Nah, this is the only area in the game currently using the override depth tech. It's for making players / items / monsters appear higher than they should be.

jrotunda:
Is this one ready for review? I'm unsure what the status is on this.

Gar103:
Yep, it's done

PR #372, opened 2020-07-12
Fullscreen

Meridian59:


davidklaw:
@nullpsi @sicoderealm Any chance you'll be coming back to this work?

Dirhai:
Anyone already tested this draft, we have any interface visible progress so far?

davidklaw:
I tested it earlier, but only briefly. The room view gets stretched when the aspect ratio changes. I think it would be cool to wireframe the desired outcome of this.

sicoderealm:
> @sicoderealm Any chance you'll be coming back to this work?

I would like to yes. There is still an unbounded amount of time here - the plan was to have a minimum size for the inventory / stats area (as this section of the client looks very strange when too small or large).

I have an idea to help get this task completed. Let me report back over the next few weeks

jrotunda:
I saw the recent comments and wanted to pile on the support and encouragement here. 

The visual window stretching and the comment about a defined boundary for the other window elements - this is exactly how I would expect this feature to look with Meridian.

Please update if we can help contribute or test to your efforts in any way @sicoderealm - this seems close.

sicoderealm:
I've made some progress here finally - apologies for the delay!
I hope to update my PR and files in the next week or so. Please stand by

davidklaw:
If you're up for it and aren't already here somewhere, come join us in the Meridian 59 Discord.

akirmse:
Implemented in a different PR.

PR #374, opened 2020-10-08
Durability text for chalice, rings, spell wands, and special wands

Aesica:
Adds general durability text to the chalice of rain, rings (except wedding/signet), spell wands, and special wands that's similar to the durability text displayed on weapons and armor.  Also displays a message to the owner of a chalice when they drop it in a place where it can be refilled, but only when it actually gets refilled.


kod/object/item/passitem/chalice.kod
```diff
@@ -29,6 +29,14 @@ resources:
    chalice_empty = "You sip the last bit of water from the Chalice of Rain."
    chalice_combine = "You pour the water from one Chalice into the other."
    chalice_cant_use_pkill = "Only innocents may partake of the chalice."
+   chalice_refilled = "The Chalice of Rain glows faintly for a moment, renewed by the spirit of Shal'ille that seems to permeate the area."
+
+   chalice_condition_exc = " filled to the brim with pure, sparkling water."
+   chalice_condition_exc_mended = " filled to the brim with pure, sparkling water."
+   chalice_condition_good = " is still plenty full."
```
davidklaw, line 36: The difference between "plenty full" and "a few sips left" seems too much. Given that there are a handful of sips before it's empty, maybe it shouldn't start out "filled to the brim".

Some ideas to take or leave:
1. holds pure, potable sparkling water (too wordy but suggests consumable)
2. holds several more sips
3. has a few sips left
4. is almost empty

davidklaw:
I had left some inline comments for the Chalice and am following up here with some other issues/comments. The bit that that messages a user when their chalice is dropped feels out of place here and would be better off in it's own PR, but that's not my call 😄 

The Wand of Healing's durability message does not display.
image

The Wand of Healing's broken state also makes this final durability message redundant.
image

I'm wondering if maybe the Wand of Healing wasn't meant to be included in this. But it raises an interesting question about where to include messaging about durability -- in the description of the item or a separate line?

Similarly, the description of some rings are redundant when paired with these new messages
image

I wasn't able to fully test this yet, but wanted to get in the stuff I'm seeing most imediately.

kod/object/item/passitem/specwand.kod
```diff
@@ -36,6 +43,14 @@ classvars:
    viObject_flags = APPLY_YES | GETTABLE_YES
    viItem_type = ITEMTYPE_WAND
 
+   vbShow_condition = TRUE
+   vrCondition_exc =   wand_condition_exc
```
davidklaw, line 47: These 6 classvars need to be updated to fix the issue I called out in my comments. They're referencing the incorrect resources.

davidklaw:
@Aesica I re-tested and tracked the issue with the durability text to the classvar mappings in `specwand.kod`.

I think we've already discussed a need to extract the chalice changes and other than that, my only beef would be with the messaging. However, I don't know if it's worth going down that rabbit hole in this PR.

I'll return and test again once its been updated/resubmitted.

jrotunda:
Needs a little more work from @Aesica per comments by @davidkla, and this PR contains and relates to too many classes fo keep track of necessary work or changes and has already gotten very confusing.

* For clarity and sanity, I suggest this be closed and resubmitted as smaller, individual PRs per CLASS.

Note: This spawned from an old request of mine for added Chalice condition text resources.

The Wands, Rings and the Jewel of Froz may all benefit from descriptors but should be looked at individually in this case - some items may need to be used faster or broken more easily for balance or another design reason, not knowing when they break can "help" with balance. 

Based on other gear and intuitive gameplay, a user may expect that the Chalice a large cup the player can fill, give some indication it was empty or full, similar to water skins visually emptying. 

Since the cup can be refilled but only a few players may know online, if any, let alone where and how to refill it - I wanted it to tell you if it had been satisfied by your efforts. 

Flavor text descriptions retain RPG immersion for items like this in lieu of say, a catch-all "uses left" Override.

The Chalice is harder to get with casting dispel illusion added as a requirement (you used to have to only kill the Orcs - which still can take a long time) - a solid design change, it leads in really well for your first solo journey opening the Cave since the Chalice can also, which new players may discover naturally, will rescue them and take them through most of the caves, and another sip to the Island Inn, getting them there in seconds.

*BUT* it's no longer an item a newbie can casually get and is more of an achievement in the building path. 

Since they're hard to get, InitUse varies from 3-5, and not everyone knows how to refill them (even those that do still mess up) - the Chalice portion of this seemed like a small QOL rebalance so you can better hold on to them after obtaining one.

PR #378, opened 2021-11-16
Fix Riija Bridge overzealous item dropping

GarAdminTools:
Riija Temple bridge is coded to sometimes drop an item when the player falls. However, it checks if an item is valid to drop with ReqLeaveOwner, which is only ever false for cursed items. Here I've added a check for DropOnDeath, which will exclude very important items like room keys and bonded equipment.


akirmse:
This looks pretty good, but since the logic is duplicated, how about you move it to a separate function (ShouldDropItem or similar)?

GarAdminTools:
Sure, done

PR #379, opened 2021-11-16
Implement Rain and Snow in Software & Hardware Renderer

GarAdminTools:
This adds the capability for Rain and Snow to the client & server.
Admins can send a room:
StartRain
EndRain
StartSnow
EndSnow

This doesn't include any sort of automatic weather, sounds, or texture changes.
If so desired, ice textures, snow textures, snow walking sounds, and rain sounds can be added, as well as natural weather cycles.

But for now, this is admin only.

TESTING: Enter the game, send the room you're in StartSnow or StartRain. Try it out in software and hardware renderer. End the related weather with EndSnow or EndRain.

Particle systems improvements

- Added a new function in client3d.c, GetPointCeiling, which returns the height of the sector for the given x and y values.
- Increased the MAX_ITEMS value in drawbsp.c from 1200L to 2400L, was reaching that limit.
- New variables for particle emitters - bRandomize (used for x, y, z randomizing) now split into bRandomizeXY (map position) and bRandomizeZ (height); bGroundDestroy for particles that can be removed once they hit the floor and bWeatherEffect for weather effects such as snow.
- Weather effects now have randomized z velocities between 0.5 and 1.5 of the emitter's velocity.
- Snow and rain particles are destroyed when they hit the ground.
- Weather effects have a default height that should be relatively near the ceiling for a lot of screens, but low enough that there isn't much delay in drawing them on entry (snow anyway). Half of weather effect particles are given a starting point of the ceiling itself. Since ceiling heights are all over the place, drawing them all from the ceiling (ideal scenario) would cause issues on screens with very large heights.


akirmse:
This could work, but it's very large and there are some non-trivial changes I'd like to see.  How about you split this into two PRs, one for the client and one for the kod?

If you want to get a head start on the client PR, we can't have a function with 20+ parameters (D3DParticleEmitterInit).  Maybe pass in a structure?

GarAdminTools:
Quick question so I understand what I'm trying to fix, what's the issue with functions with 20+ parameters?

GarAdminTools:
Removed the kod side from this PR

akirmse:
The general consensus is that functions with more than 4 or 5 parameters are fragile and hard to maintain.  I recommend the classic books Code Complete, Refactoring, or there are probably some online articles that cover why writing maintainable code is important.  Not that we knew any of this when we originally wrote Meridian :-)

davidklaw:
I'd like to help move this one along, but it seems it can't be tested in its current state. If @GarAdminTools can help, that'd be great. If not and there's still interest in this, I can attempt to revise it with a new set of PRs.

Dirhai:
For me as a player, Rain and Snow would be a huge positive implementation, this would add so much in terms of ambience and immersion, would be really nice if someone make this work properly.

jrotunda:
Needs additional work from @GarAdminTools or steps in feedback from @akirmse to get the code to where it needs to be for merging. 

If author is unable to follow-up at this time perhaps someone else can rectify the last requests and pick up where this was left off. 


Aside: Gar submit this PR at my behest and it's active on other OS servers / cannot tell if they changed their code for it beyond this as their branches dev and commit histories have gotten too confusing and obfuscated for me to follow.

Meridian has went a long time without water or snow but they would be really fun immersive additions for regular ganeplay and event settings.

GarAdminTools:
@davidklaw @jrotunda My bad, I didn't see these pings at the time. I'll get a meridian dev environment going and see what I can do here.

davidklaw:
@GarAdminTools 👀

jrotunda:
Based on comments from Q, this effect never did look right in early testing and is why it was decided to not be added to the game. It's important that this look and feel professional; and hope we can get some visuals in here that can be approved for the next PR as it seems this one hasn't been updated in some time.

Note: There is code, perhaps best in its own PR, for updated textures for the outdoor weather, with pools of water forming in low elevations, and immersive audio for both outside and indoors, that could/should be added along with this if it makes it in for the full effect -  silent rain would be strange.

davidklaw:
Time to close and and tidy up the PR list?

akirmse:
Yeah this seems abandoned.

Issue #224, opened 2013-08-18
Missing soundfile flight.wav in light.kod


cyberjunk:
There is a missing soundfile...
light.kod references a non existant flight.wav

At least I wasn't able to find it in my resources (?)

https://github.com/Meridian59/Meridian59/blob/master/kod/object/passive/spell/roomench/light.kod

See line 33:
light_sound = flight.wav

Meridian59:
I don't have it either.  Does anyone have a copy?

davidklaw:
I was able to install the Insurrection and Dark Auspices discs, but neither had this file present in their resources.

PR #387, opened 2022-06-26
Update light to use generic Faren sfx

davidklaw:
This updates Faren's Light sfx by updating its reference from `flight.wav`, a sfx that quite possibly doesn't exist, to`faren.wav`. It is in response to #224.

After installing and looking through the resources for the only two discs I could installed (Insurrection and Dark Auspices) I was unable to find `flight.wav` or equivalent sound under a different name. I created a potential new sfx that I'll share below, but ultimately chose to do what spells like Bramble and Lightning Wall do: use `faren.wav`. The generic Faren sfx might be one of the better choices considering how frequently a spell like Light might be cast (i.e. less noise, relatively speaking).

Here is the result:

https://user-images.githubusercontent.com/467443/175820684-450abfbc-93ce-4276-83cc-9aff6ed662f3.mp4

And here is the alternative I created:

https://user-images.githubusercontent.com/467443/175820748-fd50e7cb-4357-4a01-8f98-b1f93e50ea24.mp4

Note: this alternate version uses sfx from [freesound](https://freesound.org/), [licensed under the Attribution Noncommercial 4.0 License.](https://creativecommons.org/licenses/by-nc/4.0/)

### DIFF
--- a/kod/object/item/passitem/weapon.kod
+++ b/kod/object/item/passitem/weapon.kod
@@ -308,7 +308,7 @@ messages:
       local iDamage, i, oWeapAtt;
 
       % First, get base damage.
-      iDamage = send(self,@GetBaseDamage);
+      iDamage = send(self,@GetBaseDamage,#who=poOwner,#target=target);
 
       % Then check weapon attributes
       %  Weapon Attributes in general should only + or - damage - no multipliers!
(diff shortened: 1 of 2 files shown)


davidklaw:
Closed due to a snafu with my fork's branches needing change. New PR is in effect as #388

PR #389, opened 2022-06-27
Update second wind enchantment

davidklaw:
This PR builds on the work of @skittles1 and @Nagath in #296 and closes #296 if merged.

As it stands today, Second Wind does not trigger a visual indicator and #296 adds one with a few minor bugs this PR initially addressed. Per the suggestion of @Meridian59, I've pivoted this PR to refactor second wind to mimic personal enchantment spells. No changes to the core second wind functionality should be changed by this PR.

# Known Issues
- The original handling of second wind updates the player's rest threshold from 80 to >80 once second wind dissipates. I don't understand why and it seems like it could be a bug, because I don't see anything that says second wind should permanently affect a player's default rest threshold.

# Testing
## Overview
* Testing requires your player to to be in several states (i.e. vigor below threshold, vigor above, with and without second wind, etc). When testing, I used an admin player with the default stats (25) in mortal mode (`dm mortal`) with all spells and skills (`dm get spells` and `dm get skills`). I cast blink to quickly drain my vigor and trigger second wind.
* If you're comfortable in the code, consider overriding `iWaitTime` in `GetDuration()` to shorten the SW enchantment. Otherwise you'll be waiting 5-10 minutes for it to disippate.

### Players without second wind

- As a playerwithout second wind, no exertion event (anything that negatively impacts vigor) should trigger an increase in vigor (e.g. second wind).

### Players with second wind, triggering

- As a player with second wind, any vigor change that results in vigor > 25 should not trigger second wind.
- As a player with second wind, any vigor change that results in vigor < 25 should trigger second wind, and result in a significant vigor boost (40 + stamina/2 + skill %), the appearance of a personal enchantment icon, an audible sound effect, and a system message.
- As a player with second wind, any vigor change that results in vigor < 25 should modify your rest threshold to 10.

### Players with second wind, enchanted

- As a player enchanted by second wind, consuming food or drink should not raise vigor.
- As a player enchanted by second wind, any activity that reduces vigor should continue to do so at it's normal rate.
- As a player enchanted by second wind, any reduction in vigor should not activate second wind enchantment again.
- As a player enchanted by second wind, my enchantment should continue after logging off and logging back in.
- As a player enchanted by second wind, my enchantment should continue after a system save.

### Players with second wind, dissipated

- As a player whose second wind enchantment has dissipated, my vigor should remain unchanged.
- As a player whose second wind enchantment has dissipated, my rest threshold should update to it's original value: (80 + (skill % + 1)/5)
- As a player whose second wind enchantment has dissipated, my enchantment icon should go away and a system message should alert me to the skill enchantment being removed.
- As a player whose second wind enchantment has dissipated, second wind should trigger again as soon as my vigor falls below 25.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -87,9 +87,6 @@ constants:
    PANTS_MUL  = 0x0100
    
    BASE_NEED = 105
-
-   % If vigor gets this low, see if second wind skill kicks in.
-   SECONDWIND_THRESHOLD = 25
    
    % How steep is the curve for learning new levels of spells?
    %  A higher number means a greater slope (less points at low levels)
@@ -708,8 +705,6 @@ resources:
    player_logged_on_wav_rsc = player_login_or_out.wav
    player_logged_off_wav_rsc = player_logout2.wav
    
-   second_wind_finished = "You feel more relaxed as your second wind dissipates."
-
 classvars:
 
    viHand_space = 2
@@ -877,6 +872,7 @@ properties:
    
    piGender = GENDER_MALE
 
+   % Player vigor thresholds
    piVigor_rest_threshold = 80
 
    % The user's string description, typed when created (and editable in game).
@@ -937,9 +933,6 @@ properties:
 
    piLastDeathTime = 0
 
-   % This keeps the duration of our Second Wind skill's "downtime"
-   ptSecondWind = $
-
    % Measures how long we have to wait before we're rescued.
    ptRescue = $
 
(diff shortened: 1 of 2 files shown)


akirmse:
This doesn't feel right---why is there second-wind-related code in player.kod?  It should all be in secwind.kod.  Maybe that needs to be re-implemented?

davidklaw:
Fair point and I agree. That's a bit beyond me at the moment, but I'll dive in and welcome any and all suggestions or participation from others.

akirmse:
It has been decades since I looked at enchantments. If this can be done as a personal enchantment, I think the code will make more sense. Some others with more recent experience might be able to suggest how to do this.

We never had a style guide, and these days there isn't enough development to need one, but a few general things to watch out for in this change:

- Don't comment out code; delete it if it's dead.
- Don't leave debug messages in (the logs are already full of junk)
- This adds a new string resource, so we'd want to put it in a development branch like you've suggested.

Generally looks pretty good though, lots of good comments in the code.

davidklaw:
> Don't comment out code; delete it if it's dead.

Always.

davidklaw:
This PR is finally ready for review 👍 

I got hung up on how I was checking for an existing second wind enchantment. When the timer expires, my calls to adjust vigor threshold were relying on `IsEnchanted()`. At that point the enchantment was still present, so no vigor threshold change took effect. I followed what I was seeing in other personal enchantments and used a `Post()` call instead of `Send()` to trigger the adjustment. I'm most curious to know if I implemented the removal of the enchantment well/correctly.

jrotunda:
Ready for review.

PR #286, opened 2014-11-14
Create Weapon Enhancement

Nagath:
QoL

Now while in trance you can say the name of the weapon to create it (No Bows only the same as usual)

PR #390, opened 2022-07-05
Add focus and weapon requests to "create weapon" spell

davidklaw:
This PR is my spin on #286 and closes #286 if merged.

In that PR, functionality was added to 🪄 create weapon to make it a focus spell, allowing 5 seconds for the player to `say` a weapon name to generate their weapon of choice. The level at which certain weapons can be created remains the same. The list of possible weapons is below.

The current codebase and this PR included a bug, which this PR attempts to fix among other things. If a player's inventory is full, the created weapon is deleted _then_ enchanted, which created an unnecessary timer. This PR makes it so that the weapon is only enchanted if it stands to be added to the player's inventory.

# Testing

* As a player with create weapon and inventory space, when I cast create weapon and do nothing more, then I expect a random weapon to be generated and added to my inventory.
* As a player with create weapon and inventory space, when I cast create weapon and say one of the weapon names below within 5 seconds, then I expect that weapon to be created and added to my inventory.
* As a player with create weapon and little to no available inventory space, when I cast create weapon, then I the created weapon to instantly vanish, as described by a system message.

## Administration
* Monitor the server timers and notice that when a weapon is created, a new timer is added, but when a created weapon vanishes, the timer increments and decrements.

# Known Issues

* Immortals can not take advantage of the trance option
* Seems wasteful to generate a weapon only to delete it - is there a way around this?

# Notes

Weapons that can be created:

- Mace
- Short Sword
- Hammer
- Axe
- Long Sword
- Scimitar
- Mystic Sword


akirmse:
To be honest I'm not sure whether we want this trance change.  Having a list of hard-coded weapon names as strings is super fragile, for instance.

How about we first do a change that fixes the timer bug, which we definitely want, then we could consider the rest.

davidklaw:
I agree. I'll revisit this PR in the morning.

davidklaw:
Adding this https://github.com/Meridian59/Meridian59/pull/391#issuecomment-1174677666 from @skittles1

davidklaw:
I'm going to close this on account of #391 being merged. If there's a desire, we can revive it.

PR #391, opened 2022-07-05
Fixes timer issue with create weapon

davidklaw:
This PR fixes an issue with "create weapon" in which new timers were being created for weapons that had previously been deleted (on account of not enough inventory space). This was resulting in bad timers. 

**Related PRs:**
* #390 
* #286

# Testing
- As a player with create weapon and inventory space, when I cast create weapon, then I expect a random weapon to be generated and added to my inventory.
- As a player with create weapon and little to no available inventory space, when I cast create weapon, then I expect the created weapon to instantly vanish, along with a system message saying as much.

# Administration
- Monitor the server timers and notice that when a weapon is created, a new timer is added, but when a created weapon vanishes, the timer count does not increase.

### DIFF
--- a/kod/object/passive/spell/creaweap.kod
+++ b/kod/object/passive/spell/creaweap.kod
@@ -118,6 +118,10 @@ messages:
 	      Send(who,@NewHold,#what=oWeapon);
 	      Send(who,@MsgSendUser,#message_rsc=createweapon_cast_rsc,
 	           #parm1=send(oWeapon,@GetCapIndef),#parm2=send(oWeapon,@GetName));
+
+         % Apply attribute only if the weapon has been added to the player's inventory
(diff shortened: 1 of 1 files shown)


davidklaw:
If we're happy with a change like this for now, then merging this should close #390 and #286.

skittles1:
The way we handled this in the forks was to leave the attribute assignment in place and change the Delete call to instead drop the item on the ground, but this works too. A downside of our method is you can get piles of weapons on the ground when players are working the spell, but a (situational) benefit is they can quickly cast super strength or drop something else to pick up the weapon they just made. If you wanted to do that you could adapt [this](https://github.com/OpenMeridian105/Meridian59/blob/develop/kod/object/passive/spell/creaweap.kod#L119) and also grab the modified `createweapon_inv_full_rsc` from that file for dropped rather than vanished weapon.

PR #393, opened 2022-07-09
Update mark of dishonor to allow for monster targets

davidklaw:
This PR attempts to finish what #189 started and if merged, will close #189.

This PR updates mark of dishonor to allow it to be cast upon monsters. If a monster is currently aggro on a player, successfully casting mark of dishonor on the monster will cause it to stop aggro and enter a wait state (the existing  `EnterStateWait()`). The spell's effect will eventually dissipate. While the ability to cast this spell on players is dependent on their karma, a monster's karma does not factor in here -- mark of dishonor can be cast upon any monster.

Per the comments in #189, this is a desirable change because this is challenging spell to level up when it can only be cast upon players. It seems players resort to mules for leveling it up. Other feedback I've seen questions its usefulness in general. As it stands, this PR only focuses on continuing the changes in #189.

# Additional Thoughts
* After working through the math of the vigor threshold changes for a while, I started to wonder if this spell might be simplified to do only what it's description suggests -- take away vigor. Instead of adjusting both the rest threshold and adding exertion, what if mark of dishonor reduced the target's (players only) vigor by a % based on their karma and the caster's ability? Is there actually a strong reason to adjust their rest threshold as well?

# Known Issues

* Occasionally while testing this and #389, I've noticed that my vigor stops draining, then continues after some time. The two changes are completely separate and I'm wondering if a vigor reduction is a consistent drain or if there's some randomness to it.

# Testing

### Cast on player (good)

* As a player with mark of dishonor, when I cast mark of dishonor on a good player, then I expect to see the message: "You may only cast mark of dishonor upon the dishonorable."

### Cast on player (evil)

* As a player with mark of dishonor, when I cast mark of dishonor on a evil player, then I expect their vigor threshold and vigor amount to be negatively affected based on their karma level.
* As a player with mark of dishonor, when I cast mark of dishonor on an evil player, then I expect to see the following message: "The eyes of Shal'ille turn towards %.", where % is the target player's name.
* As a evil player, when I am enchanted with mark of dishonor, then I expect my vigor threshold to lower.
* As a evil player, when I am enchanted with mark of dishonor, then I expect to the mark of dishonor enchantment icon and see the message: "Shal'ille singles you out for punishment.  Your life energies ebb."
* As a evil player, when I am enchanted with mark of dishonor, then I expect the enchantment for last for anywhere from 2-10 minutes.
* As a evil player, when I am enchanted with mark of dishonor, then I expect food to still increase my vigor.
* As a evil player, when I am enchanted with mark of dishonor, then I expect my vigor threshold to be lower than normal.

### Cast on monster

* As a player with mark of dishonor, when I cast mark of dishonor on a monster with aggro, I expect that monster to stop aggro.
* As a player with mark of dishonor, when I cast mark of dishonor on a monster already marked, I expect to see the message: "% is already marked.", where % is the target monster's name.
* As a player with mark of dishonor, when I cast mark of dishonor on a monster with aggro, then I expect to see the following message: "The eyes of Shal'ille turn towards %.", where % is the target monster's name.
* As a player engaging a monster enchanted by mark of dishonor, when I aggro the monster, then I expect the monster to attack.
* As a player engaging a monster enchanted by mark of dishonor, I expect the mark to last for 2-10 minutes, during which time I can not cast mark of dishonor upon the monster again.


davidklaw:
Unless I hear otherwise, I'm going to expand this PR to disallow casting on karmically good players.

kod/object/passive/spell/dishonor.kod
```diff
@@ -132,39 +133,56 @@ messages:
...
+               #parm1=Send(oTarget,@GetDef),#parm2=Send(oTarget,@GetName));
 
-      % Make sure hps aren't over maximum.  Post this so enchantment is finished.
-      post(oTarget,@NewHealth);
+         % Make sure hps aren't over maximum. Post this so enchantment is finished.
+         Post(oTarget,@NewHealth);
+      }
+      
+      % If being cast on a monster
+      if isClass(who,&Monster)
```
davidklaw, line 177: I added this check instead of `else` to be safe.

akirmse:
Looks OK I think, but it adds a resource, so we'd need a development branch.

PR #394, opened 2022-07-10
Add sort to storage dialog

davidklaw:
This PR addresses the merge conflict in #359 and if merged or closed, should close #359.

I don't know if this functionality is desirable or not. The original PR even suggests that this is something players "probably" want. But here it is, only with the merge conflicts resolved.

Note: I'll be AFK for several weeks and this still needs to be more thoroughly reviewed and tested _if desirable_. My vote is to close these these two PRs until it's apparent that this is an issue or need for players.


jrotunda:
This allows players to arrange their storage inventory listing, it's a small change that I've seen requested - and personally wanted back when I was a player. 

Organizing inventory space in general is already a hardship, this PR is a nice, smooth and almost imperceptible optimization of how the Personal Storage Vault should already work - you're requesting access to look at your stuff you already paid a fee to store.

- This removes (annoyance) needing to make space to hold the items you're removing to sort
- No more getting ripped off: constantly paying an additional
fee every time you replace the item or stack to
Move it where you like - this is a big deal for a lot of players!
- time: this saves players large amounts of irreplaceable time

akirmse:
This is a pretty massive change--it actually alters the Blakod language just so that chests can be sorted.  I'm not sure that's the best way.

davidklaw:
@akirmse I only knew how to resolve the merge conflicts of #359. I wouldn't know a better way, so if this solution isn't it I think it's safe to close these PRs. If you or someone else can suggest an alternative, I think that'd be good for whoever might come across these in the future.

akirmse:
A similar approach in smaller pieces might be OK, but I think this is too monolithic.

PR #397, opened 2022-07-23
Make rented room duration maximum editable by admins

### LINKED ISSUE #382: Rented Rooms expire too quickly and there is no system setting to manage the maximum paid stay.
In the current Rented Room Inns (Marion, Barloque) , the cap for rentable rooms is 3 weeks - a very short amount of time. 

This was designed in a period when Meridian had a subscription fee, when constant play and management was required; it's expected people who take the time to design their rooms now be able to take a moderate break from the game and retain their rooms.

Rooms are open to all players but open up more as an endgame feature and rewarding to own a fully decorated one with difficult to obtain resources (Heartstones).  

In terms of Player Housing, you can maintain a guild and guildhall for as long as you offer money to Frular the Guildmaster. Granted, rooms can't be broken into when you're gone for a long time, but you should still be able to maintain them for at least a month, if not 2-3, and this should be able to be adjusted based on player activity.

There is no boolean setting or otherwise integer variant available in the rented room code and instead it's hard-coded at 21 days. This should be available for an administrator of any given server to be able to update on the fly.

davidklaw:
This PR would close #382, which asks that the maximum length at which rooms can be rented be moved from a constant to a property that can be managed by an administrator with a deeper understanding of the state of the server.

To test this, I logged in as an administrator, found the `RentableRoomMaintenance` instance and updated `piRentableDaysMax` from `240` to `365`. I then executed a system save and confirmed that the property value was still `365`.

cc @qorkod

### DIFF
--- a/kod/util/rntmaint.kod
+++ b/kod/util/rntmaint.kod
@@ -14,10 +14,6 @@ constants:
    
    include blakston.khd
 
-   % This is the max number of days that a player can rent a room at a time.
-   % Equal to about 20 offline days.
-   DAYS_MAX = 240
-
 resources:
 
    RoomMaintenance_room_already = "You already have a room here."
@@ -31,8 +27,9 @@ resources:
 properties:
 
    plHoldingDuringRecreate = $
-
    plRoomsRented = $
+   % Max number of days a room can be rented for at one time. ~20 offline days.
+   piRentableDaysMax = 240
 
 messages: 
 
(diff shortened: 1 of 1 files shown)


akirmse:
Do we have a named constant for the 24 factor between real vs. game time?  That would certainly make reading stuff like this easier in the future.

davidklaw:
These are the time constants from `blakston.khd`.

```
%%% Time constants of the world

% MS_PER_DAY = 24 * 3600 * 1000
MS_PER_DAY = 86400000

% Seconds per...
DAY = 86400
HOUR = 3600
MINUTE = 60
```

davidklaw:
And I'll just be honest and say that I don't completely understand what you're asking for, but would like to improve upon this. Are we looking for a named constant to use in place of the `240` or to refer to in the comments, to help understand the `240` to 20 IRL days equivalency?

akirmse:
I thought it would be good to have a named constant for the factor with which time passes faster in Meridian compared to the real world (Is it 12?  24?).  So then you could say:

```
piRentableDaysMax = 20 * MERIDIAN_TIME_SPEEDUP
```

and then it would be obvious that you're talking about real time, not in-game time.  I think such a constant would have a few hundred uses throughout the codebase.

(BTW note how in this case it's actually clearer to say "12 * 20" rather than "240" because it expresses intent better.  The blakcomp compiler will simplify the expression so that there's no runtime performance penalty.)

davidklaw:
Got it, thank you 🙇

Issue #26, opened 2013-01-21
Mail and news columns should be sortable


Meridian59:
Clicking on the column headers in the mail and news dialogs should cause the messages to be sorted appropriately.

PR #398, opened 2022-07-24
Unlock and add newsgroup sorting functionality

davidklaw:
This PR targets part #26 by updating the news globe dialog to have sortable columns.

My solution was to implement a list comparison callback on column click. The sort used was unstable, hence the addition of the article number for stability. This is not an area I have much experience in, so whether or not this is performant in the face of large data sets is unknown to me (and I expect there to be large datasets).

## Known Issues
* My test data is _very_ minimal; only about a dozen list items

### DIFF
--- a/module/mailnews/newsread.c
+++ b/module/mailnews/newsread.c
@@ -11,26 +11,27 @@
 
 #include "client.h"
 #include "mailnews.h"
-
-#ifdef _WIN32
-#include <string.h>
-#define strcasecmp _stricmp
-#else // assuming POSIX or BSD compliant system
-#include <strings.h>
-#endif
+#include "string.h"
 
 HWND hReadNewsDialog = NULL;
 
 /* Stuff for making raw times into human-readable times */
-static int months[] =
-    {IDS_JANUARY, IDS_FEBRUARY, IDS_MARCH, IDS_APRIL, IDS_MAY, IDS_JUNE, IDS_JULY,
-     IDS_AUGUST, IDS_SEPTEMBER, IDS_OCTOBER, IDS_NOVEMBER, IDS_DECEMBER};
-static int weekdays[] =
-    {IDS_SUNDAY, IDS_MONDAY, IDS_TUESDAY, IDS_WEDNESDAY, IDS_THURSDAY, IDS_FRIDAY, IDS_SATURDAY};
+static int months[] = {IDS_JANUARY,   IDS_FEBRUARY, IDS_MARCH,    IDS_APRIL,
+                       IDS_MAY,       IDS_JUNE,     IDS_JULY,     IDS_AUGUST,
+                       IDS_SEPTEMBER, IDS_OCTOBER,  IDS_NOVEMBER, IDS_DECEMBER};
+static int weekdays[] = {IDS_SUNDAY,   IDS_MONDAY, IDS_TUESDAY, IDS_WEDNESDAY,
+                         IDS_THURSDAY, IDS_FRIDAY, IDS_SATURDAY};
 
 static list_type new_articles = NULL; /* Build up new articles arriving from server */
 
 static RECT dlg_rect; // Screen position of dialog
+
+// Constants for news dialog columns
+// enum { COL_TITLE = 0, COL_POSTER = 1, COL_TIME = 2};
+
+#define COL_TITLE 0x0000
+#define COL_POSTER 0x0001
+#define COL_TIME 0x0002
 
 static ChildPlacement newsread_controls[] = {
     {IDC_NEWSEDIT, RDI_ALL},
@@ -65,8 +66,8 @@ void UserReadNews(object_node *obj, char *desc, WORD newsgroup, BYTE permissions
    s.permissions = permissions;
    s.desc = desc;
 
-   DialogBoxParam(hInst, MAKEINTRESOURCE(IDD_NEWSREAD), cinfo->hMain,
-                  ReadNewsDialogProc, (LPARAM)&s);
+   DialogBoxParam(hInst, MAKEINTRESOURCE(IDD_NEWSREAD), cinfo->hMain, ReadNewsDialogProc,
+                  (LPARAM) &s);
 }
 /****************************************************************************/
 /*
@@ -95,7 +96,7 @@ void ReceiveArticles(WORD newsgroup, BYTE part, BYTE max_part, list_type article
 
    if (part == max_part)
    {
-      SendMessage(hReadNewsDialog, BK_ARTICLES, 0, (LPARAM)new_articles);
+      SendMessage(hReadNewsDialog, BK_ARTICLES, 0, (LPARAM) new_articles);
       last_part = 0;
    }
 }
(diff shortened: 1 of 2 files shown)


module/mailnews/newsread.c
```diff
@@ -12,28 +12,39 @@
 #include "client.h"
 #include "mailnews.h"
 
+#ifdef _WIN32
+#include <string.h>
+#define strcasecmp _stricmp
+#else // assuming POSIX or BSD compliant system
+#include <strings.h>
+#endif
```
akirmse, line 20: Not necessary; the client is Windows-only.  I bet you can delete this whole block (the client mostly relies on a single huge precompiled header, so individual source files rarely need to include anything extra).

module/mailnews/newsread.c
```diff
@@ -190,48 +201,48 @@ BOOL CALLBACK ReadNewsDialogProc(HWND hDlg, UINT message, UINT wParam, LONG lPar
...
+
          // Add article to list view
          if (!IsNameInIgnoreList(article->poster))
          {
             lvitem.mask = LVIF_TEXT | LVIF_PARAM;
             lvitem.iItem = ListView_GetItemCount(hList);
             lvitem.iSubItem = 0;
             lvitem.pszText = article->title;
-            lvitem.lParam = (LPARAM) article;
+            lvitem.lParam = (LPARAM)article;
```
akirmse, line 224: These little whitespace changes obscure what this change is about (and actually I disagree with deleting the space here).  Assuming this is some kind of auto-linter in your editor, can you turn it off?

module/mailnews/newsread.c
```diff
@@ -383,14 +399,74 @@ Bool DateFromSeconds(long seconds, char *str)
...
+      break;
+
+   case 1:
+      if (bSortAscending)
+         return strcasecmp(article1->poster, article2->poster);
+      else
+         return strcasecmp(article2->poster, article1->poster);
+      break;
+
+   case 2:
```
akirmse, line 463: Equivalent and shorter is:
```
if (condition)
  return whatever1;
return whatever2;
```

No ```break``` needed then.

module/mailnews/newsread.c
```diff
@@ -383,14 +399,74 @@ Bool DateFromSeconds(long seconds, char *str)
...
+   NewsArticle *article1, *article2;
+   BOOL bSortAscending = (lParamSort > 0);
+   int nColumn = abs(lParamSort) - 1;
+
+   article1 = (NewsArticle *) lParam1;
+   article2 = (NewsArticle *) lParam2;
+
+   switch (nColumn)
+   {
+   case 0:
```
akirmse, line 449: Named constants for the column names would be nice self-documentation.

module/mailnews/newsread.c
```diff
@@ -383,14 +399,74 @@ Bool DateFromSeconds(long seconds, char *str)
...
+   ListView_SortItems(pLVInfo->hdr.hwndFrom, CompareListItems, lParamSort);
+}
+
+int CALLBACK CompareListItems(LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort)
+{
+   NewsArticle *article1, *article2;
+   BOOL bSortAscending = (lParamSort > 0);
+   int nColumn = abs(lParamSort) - 1;
+
+   article1 = (NewsArticle *) lParam1;
```
akirmse, line 444: We use the C++ compiler, so you can declare variables at first use, which is better practice.  i.e.

```
NewsArticle *article1 = (NewsArticle *) lParam1;
```

module/mailnews/newsread.c
davidklaw: I'll have to dig around and see. I'm running Visual Source Code's default code style formatting. Probably not a great idea w/o a full understanding of the desired code style, but I'm running into a lot of difficult to read blocks of code and have been using it sparingly to make the code more readable.

davidklaw: I found myself adopting the use of `clang-format` and curious if there'd be interest in adopting it here as well? We could add a root level `.clang-format` and throw in [everything and anything](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) that we'd like to work towards.

akirmse: Hm, maybe, but certainly not in this change.

davidklaw:
As I've gotten deeper into this I've wondered:

Sorting by date is obvious and simple.

Sorting by any of the other fields brings into question how to handle the comparison of duplicate values. On news globes, many posts have the same author or same subject. The same applies to subjects and authors in mail. How do we want the sorting of these properties to be handled? Should date be incorporated into the sorting when any of the other properties are the sort key? Should sorting by author ascending list that authors most recent posts first, but the opposite when author is sorted in descending order?

I thought we might do this by appending the time to the values being compared, so that they act as a secondary sort order when the title/poster/author is equal. I haven't attempted to refine it much yet, but here's my drafted solution:

```
int CALLBACK CompareListItems(LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort)
{
   BOOL bSortAscending = (lParamSort > 0);
   int nColumn = abs(lParamSort) - 1;

   NewsArticle *article1 = (NewsArticle *) lParam1;
   NewsArticle *article2 = (NewsArticle *) lParam2;

   // Append the time to titles for greater accuracy when titles are the same, like post replies
   // that use the same title
   char title1[MAX_SUBJECT + MAXDATE];
   char title2[MAX_SUBJECT + MAXDATE];

   sprintf(title1, article1->title);
   sprintf(title2, article2->title);

   sprintf(title1 + strlen(title1), "%d", article1->time);
   sprintf(title2 + strlen(title2), "%d", article2->time);

   switch (nColumn)
   {
   case COL_TITLE:
      if (bSortAscending)
         return stricmp(title1, title2);
      return stricmp(title2, title1);
   case COL_POSTER:
      if (bSortAscending)
         return stricmp(article1->poster, article2->poster);
      return stricmp(article2->poster, article1->poster);
   case COL_TIME:
      if (bSortAscending)
         return article1->time - article2->time;
      return article2->time - article1->time;
   }

   return 0;
}
```

akirmse:
This is a well-known problem that depends on whether the sort is "stable", i.e. does it preserve the existing order among tied values?  If so, then the user can get secondary sorting on date by first sorting on date, then sorting on name.  I'd suggest figuring out whether this sort is stable by trying it out or looking it up online (I don't remember, it's been too long).  If so, then probably don't need to do anything.  Otherwise, yeah, adding the date seems reasonable.

davidklaw:
🙇‍♂️ I appreciate the education!

As it turns out, [this sort is unstable](https://docs.microsoft.com/en-us/windows/win32/controls/lvm-sortitems). I've implemented a solution that incorporates `article->num` (article index) behind the scenes and adds stability. It's not the prettiest code, but I don't know a more optimal solution. Additionally, I don't have any idea what the performance will be like on a large data set 😬

I've addressed your other comments. I don't have the confidence to say that this can move beyond draft, but if we like the way it's implemented perhaps we limit this PR to newsgroups and then implement something similar for mail in a separate PR. What do you think?

davidklaw:
I've started to implement sorting to mail in a similar fashion. The solution I'm working towards is using the list view items `lParam` to store a `MailHeader` rather than just the message number. I believe I'm nearly there but throwing this out there in case someone has a better suggestion.

davidklaw:
Sorting mail was proving quite a challenge, so I'm going to draw a line at newsgroup sorting.

jrotunda:
The News Sorting component looks ready for Review.

DK has deferred on the Mail sorting for now (which is less of a QOL need anyway)

Meridian59:
Heads up that I'm reverting this in
https://github.com/Meridian59/Meridian59/pull/406

davidklaw:
Well, that's disappointing. I'll assume this is still a desired improvement and see about fixing the bugs. My bad!

PR #400, opened 2022-07-30
Fixes Mausoleum max monster count issue

### LINKED ISSUE #381: Bug: Room: Raza Mausoleum generates mummies over its max monster setting
The MAUSOLEUM (rid newb6) in Raza will continue to spawn mummies no matter what the monster maximum setting is for the room as long as a user is present. Hundreds of mummies will spawn. 

To reproduce, log in to a server with Raza and go to the Mausoleum. M

Set the GEN PERCENT from 1-100 and a Max over 10 (to account for the automatic generated end puzzle monsters and any pit monsters in the entrance area). Wait and observe that mummies will continue to spawn over the maximum. 

Saving a new setting doesn’t stop the behavior.

It’s expected the room not continue to generate more monsters than its maximum amount setting in the object code for the room on a live server.

davidklaw:
This commit fixes #381 by removing a questionable bit of logic increasing the max monster count of the room by 1 each time the door slam timer is executed. The issue was resulting in an ever-increasing number of monsters in the room (in the hundreds).

Although this change now maintains the max monster count (20), the room's need to spawn 5 pit mummies means that max monster count will often be exceeded if close to 20 monsters are present elsewhere in the Mausoleum. For example, if you complete the pit and it resets while there are 18 monsters elsewhere, the pit will reset and there will be 23 monsters. The room should not exceed 25 monsters.

To test this locally I had to `SLAM_TIME = 2000` to `SLAM_TIME = 5000`, to give myself enough time to pull both levers and access the pit. I killed mummies outside the pit, completed the pit challenges several times, and monitored `piMonster_count_max` and `piMonster_count` throughout.

### DIFF
--- a/kod/object/active/holder/room/monsroom/guest6.kod
+++ b/kod/object/active/holder/room/monsroom/guest6.kod
@@ -447,11 +447,6 @@ messages:
    {
       ptDoor = $;
 
-      if GetTime() > 59871006
-      {
-         piMonster_count_max = piMonster_count_max + 1;
-      }
-
       Send(self,@SomethingWaveRoom,#wave_rsc=guest6_door1_rsc,#what=poLever1);
       Send(self,@SetSector,#sector=SECTOR_DOOR,#animation=ANIMATE_CEILING_LIFT,
            #height=84,#speed=128);


jrotunda:
This is an awful bug as it breaks the whole crypt and is huge resource hog if left unchecked.

I have no idea how it's been in the game this long based on how it's being caused and am trying to understand what happened to cause the break down in the code.

Based on dev history I'm unsure if this bug was there in 1996 when Chris Kirmse coded the tutorial village - which would have meant a lot of busy work for support staff on very active servers back then. 

I never noticed this bug happening during NDS either, but perhaps Fattymoo was fixing it all that time during maintenance.

I noticed it after an admin on 101 put the max count up in the room - and that it just kept going up - and one day logged in and saw 2048 mummies were trying to spawn there. 

But editing it down never stuck - after clearing all active mummy objects and a save it seemed fine but would raise up slowly again.

Just for understanding the original problem - what is the removed code for "monster count +1" supposed to have been doing? 

What was the design intent for the max count of the room to be altered? 

Was the problem that it wasn't calling for the max to be lower again when the room resets? 

It took a long time to figure out at what point the max was going up, as just resetting the max manually appeared fine after a save. 

It was unknown that one of the messages assigned to the .kod was causing the problem until looking at the production code in real time in $ while several newbies were doing a multitude of different actions in the crypt.

I noticed the max change on me in real time as the pillar/puzzle reset and door shut was called, but couldn't see why even when looking at the kod - thank you for figuring this out @davidklaw

akirmse:
Yikes, this code was hideous.  Thanks for removing it.

PR #401, opened 2022-07-31
Introduce Meridian day multiplier

davidklaw:
This commit is in response to [this comment](https://github.com/Meridian59/Meridian59/pull/397#issuecomment-1197227981) from @akirmse. It introduces a constant value that can be used when using an offline day count to calculate the number of Meridian days.

Thoughts: A cursory search of the codebase didn't reveal a whole lot of similar use cases -- days are the unit of measure. For example, the Assassin Game (amusingly named `assgame.kod`) is based on seconds, not days. I wonder: is this constant actually useful elsewhere or is your suggested `12 * 20` the better route?

### DIFF
--- a/kod/include/blakston.khd
+++ b/kod/include/blakston.khd
@@ -35,6 +35,10 @@
    HOUR = 3600
    MINUTE = 60
 
+   % 1 offline day is equal to 12 Meridian days
+   % Use when calculating Meridian days using  
+   MERIDIAN_DAY_MULTIPLER = 12
+
    %%% Object type constants
 
    ACTIVE = 1
--- a/kod/util/rntmaint.kod
+++ b/kod/util/rntmaint.kod
@@ -28,8 +28,9 @@ properties:
 
    plHoldingDuringRecreate = $
    plRoomsRented = $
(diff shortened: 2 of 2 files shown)


davidklaw:
I've been told that this may not be accurate. I'm fine if this gets 🔪

jrotunda:
This code isn't crucial if the constants vary across only a handful of smaller in-game features as DK says, but would be helpful.

 - My knowledge of coding is primitive, but I see the benefit of a constant time calc. for additional content features and the games current ones 

Though they're few in number the ones that do use these constants can have their settings more efficiently managed for a better user experience by an admin or DM.

AssassinGame is a minor feature but a good example, frustrating users by being kicked for inactivity caused by unclear timer issues. 

Then again - it all still works fine without it, isn't causing errors if not changed and no one is freaking out.

As much as I want to see the remaining PRs get closed, killed or merged, if it's a thought away from being figured out, it can't hurt to leave open as this is also a specific ask that may not get addressed to be of relevance to our players if closed. 

AK's call!

davidklaw:
I'm going to close this and consider it reference for future conversation rather than a ready-to-go contribution.

PR #403, opened 2022-07-31
Remove lever area from monster check

### LINKED ISSUE #402: Mausoleum puzzle doesn't complete if monster is nearby
In the Raza Mausoleum, when a monster is present "behind" the right lever, the lowering of the column in the final room will not trigger. I was able to confirm by entering the final room, moving a mummy to the coordinates ROW 27, COL 4, and subsequently completing the pit mummies. Because the monster is present the column does not lower.

The column drop logic is looking for the presence of monsters in the area I've outlined in red below. This looks like it might just be a simple value adjustment for `MIN_FINAL_ROW = 27`.

image

davidklaw:
This commit closes #402 by reducing the range of the Raza Mausoleum's monster check by 1 grid row. Currently, the logic reaches into row 27, which is the grid square the right lever sits in (see image below). It should not be included and since it is, a monster present in that square will block the final room from finishing. With this PR, the check extends to row 26.

It was brought to my attention that a mummy had made its way behind the lever and as a result, the column in the final room wouldn't drop. I was able to reproduce this locally by teleporting a mummy to ROW 27, COL 4 and confirmed that when the final room was cleared, the column dropped and the room reset as expected.

The rows and columns currently checked for monsters:
image

### DIFF
--- a/kod/object/active/holder/room/monsroom/guest6.kod
+++ b/kod/object/active/holder/room/monsroom/guest6.kod
@@ -26,7 +26,7 @@ constants:
    MAX_GEN_ROW = 37
    MAX_GEN_COL = 8
 
-   MIN_FINAL_ROW = 27
+   MIN_FINAL_ROW = 28
    MIN_FINAL_COL = 1
    MAX_FINAL_ROW = 40
    MAX_FINAL_COL = 9


jrotunda:
This bug is noticeable when observing new players run through the tutorial naturally. 

This bug occurs frequently on the production servers due to the MONSTER MAX COUNT issue in the guest6/newb6 roo.kod 

The two issues combined make the tutorial crypt into a mess after a while, as the room fills up with mummies and they stuff themselves around the levers as well as making the room get stuck open. 


Over the last ten years whenever there has been a "Newb Rush"™️, the tutorial crypt has required constant maintenance to keep it functioning as intended. 

This fix helps smooth out the experience. 

Theoretically this also fixes the Hazar Mausoleum (?) since they share properties, messages and functions when ran together on live - or does that also require an update to newb6.kod?

For @akirmse - Is a recreate of the specific room on live called for if singular room kod gets updated? Or does this merge with the resource?

jrotunda:
Another unpleasant aspect of this bug is that even if players figure out what's going on - that the mobs near the gate and levers are keeping the treasure pillar from falling - if they kill the mob while outside the door will slam shut and reset the room.

akirmse:
I don't think the room would need to be recreated for this change.  The new value is loaded when the kod file is reloaded.

PR #405, opened 2022-09-30
Includes and corrects audio resources

### LINKED ISSUE #404: Orc Wizard sound effects are absent due to typos
The Orc Wizard (`orcwiza.kod`) include several resources not being put to use.

```
resources:
OrcWizard_sound_hit = orc_atkh.wav
OrcWizard_sound_miss = orc_atkm.wav
OrcWizard_sound_death = wor_dth.wav
OrcWizard_sound_aware = wor_awr.wav
```

```
classvars:
vrSound_hit = orc_sound_hit
vrSound_miss = orc_sound_miss
vrSound_aware = orc_sound_aware
vrSound_death = orc_sound_death
```

The classvars should be updated to reference their respective resources. Currently, hit and miss will play standard orc sound effects, but so will aware and death states. After this change, hit and miss should play orc sound effects while aware and death should play unique Orc Wizard sound effects.

davidklaw:
This commit adds two newer resources, the Orc Wizard "aware" and "death" sound effects, and updates the Orc Wizard kod file to use them. Previously, the `classvars` were referencing the incorrect resources. As a result, it fixes #404.

Note: This PR updates resources and includes two resource files not currently in the repository.


kod/object/active/holder/nomoveon/battler/monster/orcwiza.kod
```diff
@@ -49,8 +49,8 @@ resources:
 
    OrcWizard_sound_hit = orc_atkh.wav
    OrcWizard_sound_miss = orc_atkm.wav
-   OrcWizard_sound_death = wor_dth.wav
    OrcWizard_sound_aware = wor_awr.wav
+   OrcWizard_sound_death = wor_dth.wav
```
davidklaw, line 53: I re-ordered these to match the order of the classvar references 🤷‍♂️

jrotunda:
Note: these resources can be heard on the live production servers by sending the wave resource to the user class, which is how the issue was found to begin with while running an in-game even (I have used this audio resource for other event monsters as it's been unheard by players until recently.) 

This is a bittersweet fix - it had never been heard in the initial launch, but at least the Orc Wizards still get a chance to scare some players still.

The audio itself: It's a great old school lion growl used in a few other pieces of media. You hear this same intense growl during the climax of POLTERGIEST from the "beast". 

Ganeplay wise, it'll amplify the inhuman setting in the Orc Warrens journey and all Orc raids with each of the four Orc variants having their own battle cries. 

The last two unused Orc audio effects in the files correlateto the original Orc (which looked different) before it was redesigned for the 1997 Revelations update.

Thank you for uploading the kod for this PR @davidklaw

PR #416, opened 2022-11-27
Draggable text area

akirmse:


clientd3d/intrface.h
```diff
@@ -12,7 +12,10 @@
 #ifndef _INTRFACE_H
 #define _INTRFACE_H
 
-#define TEXT_AREA_HEIGHT 100
+// Default text area height as a percentage of the height of the client.
+#define TEXT_AREA_HEIGHT 20
+#define TEXT_AREA_HEIGHT_MIN 0.10
```
akirmse, line 17: 10% feels too small when I tried it.  How about 25%?

clientd3d/statgame.c
```diff
@@ -15,6 +15,15 @@
 /* Current state of game--selecting object, picking character, etc. */
 static int game_state = GAME_NONE;
 
+extern int main_viewport_height;
+extern int main_viewport_width;
+
+bool text_area_resize_zone = false;
```
akirmse, line 21: Assuming these variables aren't accessed from outside this file, declare them static so they're not visible.

clientd3d/statgame.c
```diff
@@ -197,16 +206,78 @@ void GameMouseButtonDown(HWND hwnd, BOOL fDoubleClick, int x, int y, UINT keyFla
...
+    ToolbarGetUnionRect(&rcToolbar);
+
+    /* Resize the text area is possible from the left hand edge of the client window,
+    in-between the viewport and textarea. The vertical resize cursor will be displayed
+    at this time. */
+    int offset = 5;
+    if (text_area_resize_inprogress ||
+        (y >= main_viewport_height + rcToolbar.bottom + EDGETREAT_HEIGHT- offset
+        && y <= main_viewport_height + EDGETREAT_HEIGHT + rcToolbar.bottom + offset
+        && x <= EDGETREAT_HEIGHT + offset))
```
akirmse, line 242: Can we have the right edge of this window extend to the right edge of the text area?  That will be a lot more discoverable.

clientd3d/statgame.c
```diff
@@ -197,16 +206,78 @@ void GameMouseButtonDown(HWND hwnd, BOOL fDoubleClick, int x, int y, UINT keyFla
...
+    {
+        RECT rect;
+        GetClientRect(hwnd, &rect);
+        int width = rect.right - rect.left;
+        int height = rect.bottom - rect.top;
+
+        // Calculate new text area size.
+        int delta_x = text_area_resize_start_x - x;
+        int delta_y = text_area_resize_start_y - y;
+        int new_text_area_height = (((float)config.text_area_size / 100.0) * height) + delta_y;
```
akirmse, line 263: This makes the mouse offset from the start of the drag into a velocity, rather than an absolute value.  As a result, if you move the mouse up a little and then stop, the graphics area continues to get smaller even though you're not doing anything.  I think this would work better if you used the current delta_x and delta_y (renamed) to calculate a new absolute size, rather than adding to the current size.  You might have to introduce variables to store the original sizes from the start of the drag to get that to work.

PR #431, opened 2022-12-11
Adds name change service to Caramo

davidklaw:
This PR adds the ability to change names through Caramo's office and was first implemented [here](https://github.com/OpenMeridian/Meridian59/pull/1045/files) by @skittles1 in the OpenMeridian project. This version differs from their implementation in a couple ways detailed below. **It is submitted now as a draft for review and feedback.**

# Overview
Once deployed, this PR makes it possible for any player to change their name in Caramo's office. Player's can request a scroll, fill out their new name, then return to the scroll to Caramo to process the name change at the predetermined price point. This process does not require a game administrator. The default price for this service is 3,000,000.

## Key Points
- Names are validated using the existing name validation code in `System`, which has been refactored for use outside of the initial character creation flow.
- A record of each name change is logged to the Book of Jala in the Justicar's office.
- At this time, players can change their name as often as they'd like and limited only by the size of their bank account.
- Unlike the OpenMeridian implementation, this PR requires players to have the required money in their inventory and not simply in their bank account. There ought to at least be an ounce of danger involved, in my opinion at least.

# Testing
1. Travel to the Office of the Justicar, `go 970`
2. Request a name change scroll by saying `change name` or `name change`
3. Acquire the necessary funds for the service -- 3,000,000 using `dm get money`
4. Edit the scroll with a new username
5. Offer the scroll to Caramo while holding 3,000,000 shillings
6. Verify that the scroll is now gone, you are lighter by 3,000,000 shillings, and your name has been changed

# Questions
- Should murderers and outlaws be able to change their name? They can currently.

### DIFF
--- a/blakserv/ccode.c
+++ b/blakserv/ccode.c
@@ -930,41 +930,43 @@ blak_int C_SetResource(int object_id,local_var_type *local_vars,
 	
 	str_val = RetrieveValue(object_id,local_vars,normal_parm_array[1].type,
 		normal_parm_array[1].value);
-
-    switch (str_val.v.tag)
-    {
-    case TAG_TEMP_STRING:
-        snod = GetTempString();
-        new_len = snod->len_data;
-        new_str = snod->data;
-        break;
-    case TAG_RESOURCE:
-        r = GetResourceByID(str_val.v.data);
-        if (r == NULL)
-        {
-            bprintf("C_SetResource can't set from bad resource %i\n", str_val.v.data);
-            return NIL;
-        }
-        new_len = strlen(r->resource_val);
-        new_str = r->resource_val;
-        break;
-    case TAG_STRING:
-        snod = GetStringByID(str_val.v.data);
-        if (snod == NULL)
-        {
-            bprintf("C_SetResource can't set from bad string %i\n", str_val.v.data);
-            return NIL;
-        }
-        new_len = snod->len_data;
-        new_str = snod->data;
-        break;
-    default:
-        bprintf("C_SetResource can't set from non temp string %i,%i\n", str_val.v.tag,
-                str_val.v.data);
-        return NIL;
-    }
-
-    r = GetResourceByID(drsc_val.v.data);
+	switch (str_val.v.tag)
+	{
+	case TAG_TEMP_STRING :
+		snod = GetTempString();
+		new_len = snod->len_data;
+		new_str = snod->data;
+		break;
+		
+	case TAG_RESOURCE :
+			r = GetResourceByID(str_val.v.data);
+			if (r == NULL)
+			{
+				bprintf("C_SetResource can't set from bad resource %i\n",
+					str_val.v.data);
+				return NIL;
+			}
+			new_len = strlen(r->resource_val);
+			new_str = r->resource_val;
+			break;
+	case TAG_STRING :
+		snod = GetStringByID(str_val.v.data);
+		if (snod == NULL)
+		{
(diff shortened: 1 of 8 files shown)


akirmse (commented):
A few global comments before getting into the code:

- Is this a good idea?  Think about possible avenues for abuse, keeping in mind that there are probably billions of shillings floating around after so many years.

- A server change is probably not necessary.  How do guild names get set?  Maybe via the temp string, which is already supported?

- The client also removes leading and trailing whitespace when creating character names, so that should be enforced here too (not sure whether this is currently possible in Blakod).

davidklaw:
I wouldn't say I have much skin in the game, so I hope others chime in here. It seemed harmless to me, but I can see how much the name is tied to the identify, history, and therefore reputation of that character. Abuse aside, a change like this would do away with something that contributes to making Meridian what it is.

I'm happy with whatever decision is made. I just thought it would be fun and interesting to do the work and PR this for consideration 😄

jrotunda:
This is good work and does feel right for the Clerk. 

As an admin I wanted this because name changes are something that do get asked for QOL wise and provides a fun way to roleplay, it may also be needed at times, and the less reliance on an active DM or tampering with live code for any issue the better.

The chance of this being abused at the moment is equal to the chance of the same players logging on multiple free accounts to vote themselves as justicar - and that very Minor segment of player isn't active. It also requires a significant amount of money (digression: the ballots for voting should also cost a small fee)

The fee could be changed based on economy and active community - if someone is trying to change the names of their 24 friends this can be made to be not viable for abuse at that time.

The saved up money on 101 from
years ago has been mostly liquidated at this point through auctions, and players have to farm regularly now.

Suggest - A limit with a default of 1, that can be set in the Clerk object or whatever Class is handling the change, so it's balanced regardless of how rich a player is as well. We would want a player to be able to do this at least once

- A dynamic Minimum Age requirement to use the item so people have to invest in the time spent with their name, and keep relative import to the significance of ones identity in meridian 

- A tag in their profile saying who they were once known as (suggested by @aesica)

Testing -

I don't see where the String is for the list of "bad names" - is that something an admin can edit? 

How is it checking for dupe names so as not to cause an issue with duplicate vrNames being used, or is that handled by the temp string? 

Back to the point raised above by AKirmse - the value of Your Name in Meridian, and that they are so
Important - is more the factor to consider. Names in Meridian have weight. Do we want this new feature which lessens that? 

Player feedback and more time to think this over.

vi-dominik:
While I like the idea and see the QoL improvement I'd still want to see to have it not abused. My ideas:
- Limit name changes to 1 each month?
- calculate cost depending on the summary of isk being held on the server? (How to calculate to be decided, I got no idea)
- you need to do a complicated quest to unlock the option?
- <your ideas here>

jrotunda:
* Gives some more fun and immersive features to the Justicar/Notary

* The amount of shillings isn't as high as it used to be in the current community, as seen in recent auctions, and most players are still a long ways of getting their first million in savings

* Limit name change to 1x with a boolean code in player.kod; vrNameChanged = 1/0; prevents abuse/confusion along with the record of the name being public, and keeps it from being ridiculous and names meaningless

* Many players won't use this, but for those that do-  it would become important for them, and perhaps even a goal for some

* piNameChangeCost = 3000000 as a setting allows for balance for varying server world economies, probably good to let them spend their money on this as a later game expenditure

davidklaw:
This is not a priority of mine, therefore I'm going to close it. If it serves as the basis for a future proposal then that's awesome, but whoever comes across this should definitely review the discussion above.

Issue #424, opened 2022-12-08
Missing Texture: Raza Mausoleum missing Upper Wall near water pit


jrotunda:
In the Raza Mausoleum, while going up the steps from the small water pit in the SE portion of the room, there is a missing UPPER WALL texture as the top of the staircase touches the next level of floor texture.

The user is able to see through the entire map here.

Note: This doesn't appear missing in the mirror variants of this room, Hazar Mausoleum (guest6) or Resting Place of Marion's Ancestors (Marion Crypt 1)

Room:

newb6.roo
rid_newb6

davidklaw:
![image](https://user-images.githubusercontent.com/467443/208278685-402fffc6-4cc8-4cf5-ad7f-b9f1e071414f.png)

davidklaw:
![image](https://user-images.githubusercontent.com/467443/208278699-943e3f8d-c067-42d4-a0af-e7d9c22f913e.png)

PR #432, opened 2022-12-17
Texture fix for raza Mausoleum

TheDiggie:
https://github.com/Meridian59/Meridian59/issues/424


davidklaw:
Visually, the fix looks like it's working to me.

The issue was with a missing text seen here:
image

This is a screenshot of the change looking of the same surface from a different angle: 
image

Highlighting the change I see, but I know very little about map editing. The updated version is on the left, the original on the right. The updated version has an "Upper texture" set for the "First SideDef" while the original did not.
image

TheDiggie:
sorry for not giving a full explanation but yes, the step was missing its upper texture thats all

Issue #425, opened 2022-12-08
Floating Texture: East Ende


jrotunda:
In the East Ende, there's a noticeably large floating ground texture directly above the door/exit to the Tos Adventurer's Hall.

davidklaw:
Outside Adventurer's Hall:
![image](https://user-images.githubusercontent.com/467443/208278629-fd855237-97e4-4689-916a-2f90b07e76d8.png)

Nearby Tos Bank:
![image](https://user-images.githubusercontent.com/467443/208278615-db266e80-a249-42ea-a1bb-3fc889de0e90.png)

TheDiggie:
![image](https://user-images.githubusercontent.com/8981364/208280991-c61c3546-ea09-46b3-9515-30449f56a295.png)

PR #433, opened 2022-12-17
Texture fix for east ende

TheDiggie:
fix floating textures above adventure hall and bank sign


TheDiggie:
https://github.com/Meridian59/Meridian59/issues/425

davidklaw:
@TheDiggie I'm relatively new to map editing here and since there's no great way to compare, could you clue me into how you resolved this? Was it just adjusting these two x offset values?

image

davidklaw:
Visually, these changes look good to me 👍 

Screenshots using the new room files:
image

image

image

image

TheDiggie:
@davidklaw unfortunately for this one i do not have a good answer. the floating texture isnt on any of the walls so technically shouldn't be there at all. i added the offset because the sign needs it, if u look next to the wall by the sign u can see where it starts to repeat. after i set the offset the floating texture simply isnt there anymore. no idea why :p

davidklaw:
Ah, okay. That makes sense. A good PR would include what you just shared so that we can double-check the updated position of the sign (I was only focused on the issue in the sky). Thanks for knocking these two issues out. There's a bunch more calling your name 😉

Issue #422, opened 2022-12-08
Missing Texture: Ancient Smuggler's Crossing


jrotunda:
There are missing textures on the UPPER WALL of the Ancient Smuggler's Crossing.

1) Entering from Barloque Sewers - look up while walking down the first hallway. 

Three wooden boards on the ceiling are missing textures on one side facing South in Overhead/Map view.

Room: 
sewking.roo
rid_sewking

[needs screenshot]

TheDiggie:
![image](https://user-images.githubusercontent.com/8981364/208283427-34274a19-c387-424b-90d5-549407bf25fb.png)

PR #434, opened 2022-12-18
texture fix for smugglers way

TheDiggie:
added upper the missing upper wall texture to these three walls

image


TheDiggie:
https://github.com/Meridian59/Meridian59/issues/422

PR #440, opened 2023-01-20
Dark Mode Client

Dirhai:
Dark Client:

fullll

So far in the server side we have the darken texture, and text collors modifications. and localclient modification in some text area collors. 

Thanks @Aesica for provide the text RGB that makes sense with the dark client texture.


akirmse:
Unilaterally changing the look of the client at this point probably isn't a good idea.  You could make it into a configuration option if you wanted, though.

akirmse:
Closing due to inactivity.

We've since changed the text colors to address this problem, too.

PR #441, opened 2023-02-10
Update jas-west.roo

Dirhai:
This PR is a polish update to jas-west.roo

It <del>removes an extra pub sign pole, and</del> fixs the floating texture in both pub signs of this room.

Current version:
odd

Update:
fix2

%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

Current version:
olderrorjonas

Update:
jonasfixed


davidklaw:
The floating texture fix is good in my opinion, no problems there.

But I think the extra inn bracket is for a lantern, much like the East end of Jasper. The lighting in your before photo seems to suggest it as well.

image

Dirhai:
Updated, now this PR only removes the floating textures!

PR #446, opened 2023-03-17
Allows guild allies to enter guild hall

davidklaw:
This commit re-enables previously removed logic and now allows allied guild members above the rank of Initiate to enter a guild hall. The code change was made back in 2013 in #163 at a time when I'm sure the community was a bit different, but there's been conversation about bringing it back.

My personal opinion is that the community ought to be able to police themselves on this one.


akirmse:
Might want to give people a heads up that this is coming (eventually), so that they can break any alliances they don't trust.

Issue #29, opened 2013-01-21
Make RecreateAll preserve factions, shrines


Meridian59:
System.RecreateAll should preserve the state of factions and shrines.

jrotunda:
There appear to be some fixes for this @davidklaw  as linked by @Serpentiel-zz

davidklaw:
@Meridian59 What are your thoughts on [this implementation for flags](https://github.com/skittles1/Meridian59/commit/ed19967b2655d8ea938a0bc310ef94f0e314307e)? I confirmed flag factions aren't recreated today and preliminary tests show this works, but is there a more optimal/desired way of handling this?

Meridian59:
That general approach is fine: get a list of state to save, recreate, then
reapply the state.  The particular implementation leaves something to be
desired, e.g. it would fail if there were ever more than one flag per
room.  There's also some duplication of existing code that can be avoided.
But it's the right idea.

davidklaw:
Shrine recreate support added via #449

PR #449, opened 2023-03-20
Adds support for recreating shrines

davidklaw:
This PR adds `RecreateAll` support for shrine allegiance and power level, as requested in #29. To test, I chose two random shrines and dedicated one to Kraanan and the other to Shal'ille, then verified that these dedications and bonuses were accurately recreated on `RecreateAll`. If there's something else I should consider, please let me know!

### Testing

1. Give yourself the necessary offerings

```
dm get misc
```

1. Check shrine bonuses:

```
send o 0 GetShrineBonus school int 1
send o 0 GetShrineBonus school int 2
send o 0 GetShrineBonus school int 3
send o 0 GetShrineBonus school int 4
send o 0 GetShrineBonus school int 5
send o 0 GetShrineBonus school int 6
```

| 1 | 2 | 3 | 4 | 5 | 6 |
|-|-|-|-|-|-|
| 0 | 0 | 0 | 0 | 0 | 0 |

1. Dedicate shrines

Note: I learned late in testing this that you can just teleport directly to objects using the admin panel (or related command). Do that instead of trying to find the right coordinates to move to the shrine like I did 🤦.

```
go rid_e5 (The Forest Shrine)
send object <room#> somethingmoved what object <player#> new_row int 11 new_col int 15 fine_row int 52 fine_col int 47
```

Drop Sweat of a Warrior King, dedicating this shrine to **Kraanan**

```
go rid_orc_cave5
send object <room#> somethingmoved what object <player#> new_row int 67 new_col int 42 fine_row int 32 fine_col int 32
```

Drop Tears of a Newborn, dedicating this shrine to **Shal'ille**

2. Verify state by running shrine bonus commands:

| 1 | 2 | 3 | 4 | 5 | 6 |
|-|-|-|-|-|-|
| 6 | 0 | 6 | 0 | 0 | 0 |

4. Recreate

```
send o 0 RecreateAll
```

6. Confirm state by running shrine bonus commands:

| 1 | 2 | 3 | 4 | 5 | 6 |
|-|-|-|-|-|-|
| 6 | 0 | 6 | 0 | 0 | 0 |

This is based on the work of @M59Gar and @skittles1 in the OpenMeridian project 🙇‍♂️

### DIFF
--- a/kod/object/passive/shrine.kod
+++ b/kod/object/passive/shrine.kod
@@ -223,7 +223,7 @@ messages:
       return piAllegiance;
    }
 
-   SetAllegiance(allegiance = $)
+   SetAllegiance(allegiance=0)
    {
       piAllegiance = allegiance;
       return;
--- a/kod/object/passive/shrine.kod
+++ b/kod/object/passive/shrine.kod
@@ -223,6 +223,12 @@ messages:
       return piAllegiance;
    }
 
+   SetAllegiance(allegiance = $)
+   {
+      piAllegiance = allegiance;
+      return;
+   }
+
    GetPower()
    {
       return piShrine_power;
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -2750,7 +2750,8 @@ messages:
                n = n + 1;
             }
          }
-         else {
+         else 
+         {
             Debug("RecreateShrineAllegiances:: Length mismatch between shrines and shrine allegiances");
          }
       }
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -365,6 +365,7 @@ properties:
 
    plShrines = $
    plShrine_Powers = $
+   plShrine_Allegiances = $
 
    % Bonus spellpower set by admins for events.
    plBonusSpellpower = $
@@ -426,9 +427,8 @@ messages:
       Send(self,@RecreateAllBrains);
       Send(self,@RecreateLibrary);
       Send(self,@RecreateAllRooms);
-      % Creates the Faction games
       Send(self,@RecreateParliament);
-
+      Send(self,@RecreateShrineAllegiances);
       Send(self,@RecreateAssassinGame);
       Send(self,@RecreateQuestEngine);
       Send(self,@RecreateNodeAttack);
@@ -468,7 +468,6 @@ messages:
 
       Send(self,@RefigureAllAbilityTotals);
       Send(self,@RefigureAllAbilityTotals);
-      Send(self,@RecomputeShrineTotals);
       Send(self,@RecreateLearnAdvice);
 
       psTemp = CreateString();
@@ -524,6 +523,7 @@ messages:
       Send(self,@RecreateLibrary);
       Send(self,@RecreateAllRooms);
       Send(self,@RecreateAssassinGame);
+      Send(self,@RecreateShrineAllegiances);
       % The following two lines are handled in Parliament
       % Send(self,@RecreateTerritoryGame);
       % Send(self,@RecreateTokenGame);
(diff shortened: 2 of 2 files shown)


davidklaw:
Committed some very minor cleanup (just comments) and this is ready for review.

davidklaw:
Sorry about the back and forth on draft and review here. At the last second I noticed that `RecreateAll` was doubling up shrine objects, but garbage collection got rid of them. Is that expected?

kod/util/system.kod
```diff
@@ -2339,7 +2338,7 @@ messages:
    "CreateAllRoomsIfNew.\nUNSAFE to call from admin mode, as items"
    "in rooms lose state.  Use RecreateAll instead."
    {
-      local i, bFeast_Hall_Locked, oFeast_Hall;
+      local i, n, bFeast_Hall_Locked, oFeast_Hall;
```
akirmse, line 2341: Maybe call it "shrine" instead of "n"?  n usually means a number.

kod/object/passive/shrine.kod
```diff
@@ -223,6 +223,12 @@ messages:
       return piAllegiance;
    }
 
+   SetAllegiance(allegiance = $)
```
akirmse, line 226: Default value should probably be a number rather than $ ?

kod/util/system.kod
```diff
@@ -2720,6 +2726,28 @@ messages:
       return;
    }
 
+   RecreateShrineAllegiances()
+   "Resets shrine allegiances and power levels"
+   {
+      local i, n;
```
akirmse, line 2732: i -> shrine ?

kod/util/system.kod
```diff
@@ -2720,6 +2726,28 @@ messages:
...
+
+      if plShrine_Allegiances <> $
+      {
+         n = 5;
```
akirmse, line 2736: Don't hard code the number of shrines here.

kod/util/system.kod
```diff
@@ -2720,6 +2726,28 @@ messages:
...
+
+         for i in plShrines
+         {
+            Send(i,@SetAllegiance,#allegiance=(Nth(plShrine_Allegiances,n)));
```
akirmse, line 2740: This is confusing that the list of allegiances is backward compared to the list of shrines.  Can you build up the list so that it's in the same direction?

davidklaw: I've been racking my brain on this for a bit and between my fading knowledge and what I see in the manual, the only solution I could come up with is creating a temporary local list that I then use to map to `plShrine_Allegiances` in reverse order, so as to sync up with the order of `plShrines`. Is there a better way to go about this?

akirmse: You could iterate backward through plShrines in RecreateAllRooms(), rather than here, and then the lists would be in the same order.

davidklaw:
I want to believe that I addressed all of the previous comments. Everything looked good in my local tests. I even learned that you can jump right to objects using the admin panel, which is something I wish I would have known when I was trying to find just the right coordinates to move myself to these shrines.

kod/util/system.kod
```diff
@@ -2720,6 +2733,35 @@ messages:
...
+         {
+            n = 1;
+
+            for shrine in plShrines
+            {
+               Send(shrine,@SetAllegiance,#allegiance=(Nth(plShrine_Allegiances,n)));
+               n = n + 1;
+            }
+         }
+         else {
```
akirmse, line 2753: This is super minor but you should be consistent with brace placement.

Otherwise looks good.

davidklaw: I care about the minor stuff! Thanks for spotting 👍

PR #450, opened 2023-03-23
Adding support for flag states during recreateall

davidklaw:
This commit adds the restoring of territory flag states when recreating the parliament game. It was initially influenced by the work in [this commit](https://github.com/skittles1/Meridian59/commit/ed19967b2655d8ea938a0bc310ef94f0e314307e), but takes a different approach.

The ability to recreate the flag states during a recreate was essentially in place already. However, `SetFaction`, which was being called during the recreate process, was overwriting the flag states with the default ones. The result was that the first saved flag would be reloaded, but the rest would not.

My solution -- detailed more [below](https://github.com/Meridian59/Meridian59/pull/450#issuecomment-1484154619) -- adds a flag to `SetFaction` that makes recalculating flag counts and saving state optional and something reserved for cases where only an individual flag's faction is set. In other cases where flags are being processed in bulk, like during (re)creation, we disable this and handle it after the flags have been processed.

This PR is loosely related to #449 and solves 1/2 of #29.

### DIFF
--- a/kod/object/active/flag.kod
+++ b/kod/object/active/flag.kod
@@ -695,8 +695,7 @@ messages:
       return poClaimer;
    }
 
-   SetFaction(faction=FACTION_NEUTRAL)
-   "for admin use"
+   SetFaction(faction=FACTION_NEUTRAL,claim=TRUE)
    {
       piFaction = faction;
       if poOwner <> $
@@ -709,7 +708,11 @@ messages:
          }
       }
 
+      % If setting the faction as part of a bulk process (i.e. recreate),
+      % pass `claim=FALSE` and be sure to manually recalculate the flags 
+      % and save the flag states, as `FlagClaimed` does.
       if (send(SYS,@GetTerritoryGame) <> $)
+         AND claim
       {
          send(send(SYS,@GetTerritoryGame),@FlagClaimed,#flag=self);
       }
--- a/kod/util/factgame/territry.kod
+++ b/kod/util/factgame/territry.kod
@@ -397,7 +397,6 @@ messages:
    }
 
    ReloadFlagStates(lFlagStates=$)
-   "Used during recreates to save flag states."
    {
       local i, iRID, oRoom, oFlag;
 
--- a/kod/util/factgame/territry.kod
+++ b/kod/util/factgame/territry.kod
@@ -397,6 +397,7 @@ messages:
    }
 
    ReloadFlagStates(lFlagStates=$)
+   "Used during recreates to save flag states."
    {
       local i, iRID, oRoom, oFlag;
 
--- a/kod/util/factgame/territry.kod
+++ b/kod/util/factgame/territry.kod
@@ -396,25 +396,18 @@ messages:
       return;
    }
 
-   ReloadFlagStates(lFlagStates=$)
-   "Used during recreates to save flag states."
+   ReloadSavedFlagStates()
    {
       local i, iRID, oRoom, oFlag;
 
-      if lFlagStates = $
+      if plFlagStates = $
       {
-         lFlagStates = plFlagStates;
+         DEBUG("ReloadSavedFlagStates:: nothing saved, nothing reloaded");
       }
 
-      if lFlagStates = $
-         OR NOT IsList(lFlagStates)
+      if length(plFlagStates) <> length(plFlagRIDs)
       {
-         DEBUG("ReloadFlagStates:: nothing saved, nothing reloaded");
-      }
-
-      if length(lFlagStates) <> length(plFlagRIDs)
-      {
-         DEBUG("ReloadFlagStates:: error in saved flag states - length mismatch");
+         DEBUG("ReloadSavedFlagStates:: error in saved flag states - length mismatch");
       }
 
       i = length(plFlagRIDs);
@@ -425,16 +418,16 @@ messages:
          oRoom = send(SYS,@FindRoomByNum,#num=iRID);
          if oRoom = $
          {
-            DEBUG("ReloadFlagStates:: got nil room for RID=",iRID);
+            DEBUG("ReloadSavedFlagStates:: got nil room for RID=",iRID);
             continue;
          }
          oFlag = send(oRoom,@FindHoldingActive,#class=&Flagpole);
          if oFlag = $
          {
-            DEBUG("ReloadFlagStates:: got nil flag in",send(oRoom,@GetName));
+            DEBUG("ReloadSavedFlagStates:: got nil flag in",send(oRoom,@GetName));
             continue;
          }
-         send(oFlag,@SetFaction,#faction=nth(lFlagStates,(i+1)));
+         send(oFlag,@SetFaction,#faction=nth(plFlagStates,(i+1)));
       }
 
       return;
@@ -481,7 +474,7 @@ messages:
          DEBUG("ReloadSavedFlagStates:: encountered errors during save; not replacing older save.");
       }
 
-      return plFlagStates;
+      return;
    }
 
    FlagClaimed(flag=$)
@@ -886,11 +879,6 @@ messages:
    GetMinUsers()
    {
       return piMinUsers;
-   }
-
-   GetFlagStates()
-   {
-      return plFlagStates;
    }
 
    GetMinAllies()
--- a/kod/util/factgame/territry.kod
+++ b/kod/util/factgame/territry.kod
@@ -175,8 +175,6 @@ messages:
          else
          {
             send(self,@CountFlags);
-            % do this after, so that bonus town recalcs saved
-            send(self,@SaveFlagStates);
          }
 
          send(self,@RecountFlagItems);
@@ -186,7 +184,10 @@ messages:
       if piGameMode = GAME_PART
       {
          send(self,@UpdateFlagsPartialGame);
+         
       }
+
+      send(self,@SaveFlagStates);
 
       return;
    }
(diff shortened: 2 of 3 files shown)


kod/util/factgame/territry.kod
```diff
@@ -396,18 +396,24 @@ messages:
...
-   ReloadSavedFlagStates()
+   ReloadFlagStates(lFlagStates=$)
    {
       local i, iRID, oRoom, oFlag;
 
-      if plFlagStates = $
+      if lFlagStates = $
       {
-         DEBUG("ReloadSavedFlagStates:: nothing saved, nothing reloaded");
+         lFlagStates = plFlagStates;
```
akirmse, line 405: Having the caller always pass in lFlagStates would be better than this strange fallback.

kod/util/factgame/territry.kod
```diff
@@ -396,18 +396,24 @@ messages:
...
+      if lFlagStates = $
+         OR NOT IsList(lFlagStates)
       {
-         DEBUG("ReloadSavedFlagStates:: error in saved flag states - length mismatch");
+         DEBUG("ReloadFlagStates:: nothing saved, nothing reloaded");
+      }
+
+      if length(lFlagStates) <> length(plFlagRIDs)
+      {
+         DEBUG("ReloadFlagStates:: error in saved flag states - length mismatch");
```
akirmse, line 416: Probably just return here, since it's about to crash anyway (falling off the end of a list)?

kod/util/factgame/territry.kod
```diff
@@ -396,18 +396,24 @@ messages:
       return;
    }
 
-   ReloadSavedFlagStates()
```
akirmse, line 399: Did you rename this at all call sites?  Make sure you find them all.

kod/util/factgame/territry.kod
```diff
@@ -474,7 +480,7 @@ messages:
          DEBUG("ReloadSavedFlagStates:: encountered errors during save; not replacing older save.");
       }
 
-      return;
+      return plFlagStates;
```
akirmse, line 483: Is there a need to return anything?

kod/util/system.kod
```diff
@@ -4382,9 +4382,21 @@ messages:
 
    RecreateParliament()
    {
+      local plSavedFlagStates;
```
akirmse, line 4385: Don't start with "p", this isn't a property.

kod/util/system.kod
```diff
@@ -4382,9 +4382,21 @@ messages:
...
+
       if poParliament <> $
       {
+         plSavedFlagStates = Send(self,@ListCopy,#source=Send(Send(self,@GetTerritoryGame),@GetFlagStates));
```
akirmse, line 4389: Why is copying the list necessary here?  Does Recreate modify the list in place?

kod/util/system.kod
```diff
@@ -4382,9 +4382,21 @@ messages:
...
+        
          Send(poParliament,@Recreate);
+         
+         if plSavedFlagStates <> $
+         {
+            Send(Send(self,@GetTerritoryGame),@ReloadFlagStates,#lFlagStates=plSavedFlagStates);
+            Send(Send(self,@GetTerritoryGame),@SaveFlagStates);
+            Send(Send(self,@GetTerritoryGame),@CountFlags);
+            plSavedFlagStates = $;
```
akirmse, line 4398: Unnecessary as this is a local variable.

kod/util/system.kod
```diff
@@ -4382,9 +4382,21 @@ messages:
...
       {
+         plSavedFlagStates = Send(self,@ListCopy,#source=Send(Send(self,@GetTerritoryGame),@GetFlagStates));
+        
          Send(poParliament,@Recreate);
+         
+         if plSavedFlagStates <> $
+         {
+            Send(Send(self,@GetTerritoryGame),@ReloadFlagStates,#lFlagStates=plSavedFlagStates);
+            Send(Send(self,@GetTerritoryGame),@SaveFlagStates);
+            Send(Send(self,@GetTerritoryGame),@CountFlags);
```
akirmse, line 4397: These lines need some comments.  After the flag states are reloaded, why do they need to be immediately saved and counted?

kod/util/system.kod
davidklaw: What I'm seeing (now that I'm looking more closely) is that Recreate overwrites flag states. It clears and creates new flags. Following the breadcrumbs, I see that `SetFaction` leads to `FlagClaimed` and then to `SaveFlagStates`, where the flag states are replaced.

I'm going to need at this with fresh eyes another time because I just went cross-eyed figuring this out.

davidklaw: My understanding is that Recreate creates a new set of flags (`PlaceFlagsFullGame()`) and in the process replaces `plFlagStates` with the starting states, by way of `SetFaction()` to `FlagClaimed()` to `SaveFlagStates()`. I'm not sure what to do with this information yet.

akirmse: (I think) that process builds up a new list and does not modify the old one.  If so, you wouldn't need to make a copy of the old list here.  (Note that ListCopy is used only once or twice in the whole codebase, in very unusual circumstances where a list is chopped up into pieces.)

davidklaw:
I went back to the drawing board and this PR now offers a completely different approach for consideration.

As I dug deeper, I found that setting the faction would recount flags and save flag states. This makes sense when setting the faction of a single flag, but when processing flags in bulk, like during `Recreate` the flag states that could be used for reloading are overwritten.

The new parameter I added to `SetFaction` makes it possible to skip recount and saving state with the caveat that these things must be done manually if processing flags in bulk, as is done when (re)creating the game. In those cases, counting the flags and saving state is executed after the fact.

I realize that all instances of `SetFaction` are passing `claim=FALSE` and therefore maybe recounting and saving state should be removed from `SetFaction`, but I didn't want to take that leap yet without feedback.

To test this, I inspected the faction points and flag counts prior to making any changes to flags:

||||||
| --- |  --- |  --- |  --- |  --- | 
| **Flags:** | Duke: 4 | Princess: 4 | Jonas: 4 | Other: 24 | 
| **Points:** | Duke: 111 | Princess: 111 | Jonas: 111 | Other: 24 | 

I then converted two flags (outside of towns) to opposing factions, resulting in these flags and points:

||||||
| --- |  --- |  --- |  --- |  --- | 
| **Flags:** | Duke: 4 | Princess: 5 | Jonas: 3 | Other: 24
| **Points:** | Duke: 111 | Princess: 138 | Jonas: 83 | Other: 24

I then ran `Recreate` and confirmed the flags and points persisted.

Feedback at this point is very welcome.

kod/object/active/flag.kod
```diff
@@ -709,7 +708,11 @@ messages:
          }
       }
 
+      % If setting the faction as part of a bulk process (i.e. recreate),
```
akirmse, line 711: Make this a function comment (the string that can go right after the function declaration) rather than burying it the code.

kod/object/active/flag.kod
```diff
@@ -709,7 +708,11 @@ messages:
...
+      % pass `claim=FALSE` and be sure to manually recalculate the flags 
+      % and save the flag states, as `FlagClaimed` does.
```
akirmse, line 713: Comments like "as FlagClaimed does" typically "rot", as they're referring to code that can be changed independently.  10 years from now, someone will look for "FlagClaimed", and maybe it will be gone.

kod/util/factgame/territry.kod
```diff
@@ -229,6 +230,10 @@ messages:
          return;
       }
 
+      % todo: Combine these loops because the new SetFaction accepts
```
akirmse, line 233: Customarily "TODO" is all capitalized to make it stand out and easier to search for.

kod/util/factgame/territry.kod
```diff
@@ -292,6 +297,10 @@ messages:
...
```
akirmse, line 300: Here too.

kod/object/active/flag.kod
```diff
@@ -695,8 +695,7 @@ messages:
       return poClaimer;
    }
 
-   SetFaction(faction=FACTION_NEUTRAL)
-   "for admin use"
+   SetFaction(faction=FACTION_NEUTRAL,claim=TRUE)
```
akirmse, line 698: The new parameter seems good, but the name isn't super clear.  I'd say either come up with a name that is self-explanatory, or describe what it does in the function comment.

Also, I'm guessing there is an existing call of SetFaction somewhere in the territory game,*not* in Recreate(), that doesn't set "claim" and thus gets TRUE?  Maybe change that call to explicitly set TRUE so it's obvious that you mean to set TRUE, rather than maybe you forgot to update it?  (Default arguments are dangerous that way.)

kod/util/factgame/territry.kod
```diff
@@ -175,8 +175,6 @@ messages:
          else
          {
             send(self,@CountFlags);
-            % do this after, so that bonus town recalcs saved
-            send(self,@SaveFlagStates);
```
akirmse, line 179: Just double-checking: previously, SaveFlagStates wasn't called for the "partial" game mode, but now it is.  That's OK?

kod/object/active/flag.kod
davidklaw: On second (third?) thought, `SetFaction` is only used in batch processing loops of `PlaceFlagsPartialGame`, `PlaceFlagsFullGame`, `ReloadSavedFlagStates`, `RecalcTownFlags`, and `UpdateFlagsPartialGame`. Unless there's a practical need for DMs to set the faction and update the territory game, perhaps `claim` and the the call to `FlagClaimed` should be removed?

If there is an admin need, maybe `saveGame` is a more appropriate param name?

kod/util/factgame/territry.kod
davidklaw: 🤔 I need to revisit `Recreate` because this doesn't address recreating the partial game (which was why I added this call to save here). 🤦

davidklaw:
I'm going to close this for now. There are things about this, like the territory game's relationship to the parliament game, that I didn't (and still don't) fully understand. I don't want this to collect dust while I figure that out.

PR #452, opened 2023-03-29
Removes handling of animate as a config option

davidklaw:
This PR:
- Removes code that slows down animations when the game is backgrounded
- Speeds up animations when dialogs are open to match non-dialog animations speeds
- Removes code for handling a config option related to animations that is no longer necessary.

Because the config option logic touches on lots of animations, I tested and validated the following:

* Guild hall doors animate as expected
* Blurry, wavering vision when drunk
* Animated sandstorm "dibs"
* POV shakes with earthquake cast
* Zooming the map in and out is smooth
* Firewalls animate
* Movement is slowed while walking through water
* Fireballs and lightning bolts animate towards their target
* Animations when going up or falling down vertically are present and consistent with production
* Animations are "smooth" and complete while app is not in the foreground
* Animations behind an in-game dialog are "smooth" and complete

### DIFF
--- a/clientd3d/animate.c
+++ b/clientd3d/animate.c
@@ -27,7 +27,7 @@
 
 #include "client.h"
 
-#define ANIMATE_INTERVAL 40       // ms between background animation updates
+#define ANIMATE_INTERVAL 8  // ms between background animation updates (8.3333 = 120fps)
 #define FLICKER_LEVEL (LIGHT_LEVELS/2)
 #define FLASH_LEVEL (LIGHT_LEVELS/2)
 #define TIME_FLASH 1000
--- a/clientd3d/animate.c
+++ b/clientd3d/animate.c
@@ -111,13 +111,10 @@ void AnimationTimerProc(HWND hwnd, UINT timer)
    ModuleEvent(EVENT_ANIMATE, dt);
 
    /* Send event to non-module child windows */
-   if (config.animate)
-   {
-      Lagbox_Animate(dt);
-   }
+   Lagbox_Animate(dt);
 
    /* Animate the first-person view elements */
-   if (config.animate && GetGameDataValid())
+   if (GetGameDataValid())
    {
       // Avoid short-circuiting OR
       need_redraw |= ObjectsMove(dt);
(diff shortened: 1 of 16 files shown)


akirmse:
You should also remove the animate field from the Config struct in config.h.  Note that modifying this struct will require that we ship new versions of every module in the next client update, even those that haven't otherwise changed.

clientd3d/animate.c
```diff
@@ -27,7 +27,7 @@
 
 #include "client.h"
 
-#define ANIMATE_INTERVAL 165       // ms between background animation updates
+#define ANIMATE_INTERVAL 40       // ms between background animation updates
```
davidklaw, line 30: 40ms is an animation interval I pulled from `firewand.kod`, but I'm out of my depth here. Is this the right move to eliminate the difference in background animations or should it be a number close to 0?

davidklaw: The results of 40ms vs. 0ms seem obvious, but here's footage for comparison (when the dialog is present) for anyone following along:

`ANIMATE_INTERVAL 40`

https://user-images.githubusercontent.com/467443/228704161-6f1a45a3-7dc9-434a-a2a2-924886d20cba.mp4

`ANIMATE_INTERVAL 0`

https://user-images.githubusercontent.com/467443/228704244-04f19ce8-a036-47ce-8d65-6a2b81ed43b7.mp4

akirmse:
Does setting it to 0 peg the CPU at 100% ?  You should compare that value, too.

davidklaw:
For me, it seems to make very little difference. This recording covers the same troll setup as above with the interval set to 0, both with and without a dialog present.

`Processor: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz, 3600 Mhz, 8 Core(s), 8 Logical Processor(s)`

Screenshot 2023-03-29 220152

davidklaw:
I chose to PR it at 8ms or close to 120fps (8.3333ms). This gave me better results on my machine than 16ms (~60fps). Here it is running at these settings w/ a dialog present.

https://user-images.githubusercontent.com/467443/228996569-a34cbe07-aa5a-42b6-b81f-905c72e3389a.mp4

akirmse:
Can you also try this in software mode?  I just want to be sure we're not shooting ourselves in the foot performance-wise by cranking up the frame rate by such a huge factor.

davidklaw:
Software mode at 8ms:

https://user-images.githubusercontent.com/467443/229001922-68a5c489-613c-4f50-92bd-a16a914f1fce.mp4

PR #455, opened 2023-04-11
Fixes persistent effects of Umbrella

davidklaw:
Umbrella adds a resistance modifier that is not removed when the caster's enchantments are removed, by DM or by purge being two example actions that have an affect. Logic within `EndEnchantment` is designed to only end the enchantment if the time elapsed is greater than the duration of the spell, which I believe is unique to Umbrella. Under these conditions, a player could cast umbrella and purge themselves until they reach a maximum level of resistance. As far as I can tell, nothing short of a DM clearing their resistances would wipe out this effect.

Because `RemoveEnchantmentEffects` is call in these unhandled scenarios, I implemented that to update the state object to have a favorable condition and pass it to `EndEnchantment`.

To test this, I monitored `plResistances` before, during, and after umbrella, as a solo caster and as a caster and another player moving in and out of the umbrella's protection. I also tested to make sure DM commands like `RemoveAllEnchantments` work as expected here.


kod/object/passive/spell/umbrella.kod
```diff
@@ -216,9 +216,18 @@ messages:
...
    "Need to override spell.kod's thing, since we do this ourselves specially."
    {
+      local iElapsed;
+
+      % Update state so that elapsed time matches duration of the effect,
+      % effectively calling an end to the enchantment
+      iElapsed = Nth(state,1);
+      SetNth(state,4,iElapsed);
+
+      Send(self,@EndEnchantment,#who=who,#state=state,#report=report);
```
akirmse, line 229: Should this maybe propagate instead?

davidklaw:
Something seems strange about the history of this PR, so I'm moving back to a draft to investigate.

akirmse:
You got one change per change in master when you rebased.  That's noisy but
OK.  There's an option about how to handle this on a branch (my git
knowledge isn't great but I know it can be flattened into a single merge
commit).

davidklaw:
I knew I had made the propagate change and somehow managed to get my local branch all out of whack with this one. It's resolved now and moving it back to ready for review.

I mentioned above that the description is no longer accurate. Is that something you'd like to see included in this PR or a separate one? This statement is not true.
> ...the spell dissipates if caster breaks concentration by running or combat.

akirmse:
Sure, why don't you fix up the description while you're at it.

davidklaw:
Description updated. The caster can move around, attack, and be attacked without losing umbrella. I don't know if this creates a gameplay balance issue or not, but now the description reflects the current state of the spell.

PR #458, opened 2023-04-13
Feature: added bow condition descriptions

adrienlaws:
**Background**

- The current bow.kod uses the standard weapon condition descriptions.
```
   weapon_condition_exc = " is in flawless condition."
   weapon_condition_exc_mended = " is in excellent condition, but has been repaired before."
   weapon_condition_good = " is slightly tarnished but in generally good condition."
   weapon_condition_med = " is notched and stained with blood but is still an effective weapon."
   weapon_condition_poor = " is well worn and may not last much longer."
   weapon_condition_broken = " has been shattered by a powerful blow."
```

**The Problem**

- These two do not make sense for **bow** descriptions. A player reported it.
```
   weapon_condition_good = " is slightly tarnished but in generally good condition."
   weapon_condition_med = " is notched and stained with blood but is still an effective weapon."
```

**The Change**

-   Added two new descriptions for bow_condition_good and bow_condition_medium.  
-   The other inherited condition states make sense and do not need any changes.

```
   bow_condition_good = " is slightly worn but in generally good condition."
   bow_condition_med = " is starting to crack and fray but is still effective."
```

### DIFF
--- a/clientd3d/srvrstr.c
+++ b/clientd3d/srvrstr.c
@@ -261,16 +261,12 @@ typedef struct {
 } FormatCode;
 
 static FormatCode code_table[] = {
-{ 'r', CODE_COLOR, PALETTERGB(128,   0,   0) }, // Maroon
-{ 'g', CODE_COLOR, PALETTERGB(  0, 100,   0) }, // Green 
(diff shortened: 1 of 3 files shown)


adrienlaws:
Added a broken bow condition description

PR #460, opened 2023-04-14
feature: improved safety on_off messages

adrienlaws:
The Problem / Opportunity

- Safety on off messages have a huge impact on the player experience
- The current on-off messages are not as visible as they could be
- Safety On message is currently in italics vs the Safety Off being bold.  It's easier to see if they are both bold.
- With PR#456 merged (brighter colors) we can use them to greater effect with safety on off messages

The Changes

- changed verbose start to very short, capitalized, colored messages "SAFETY ON" vs "your safety is now ON" 
- aligned the text so SAFETY ON and OFF are always in the same position

safety_on_off

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -193,9 +193,9 @@ resources:
    user_guildmate_logoff = "~IYour guildmate %s has just departed."
 
    user_safety_on = \
-      "### ~B`gSAFETY ON   `nYou can no longer strike innocents."
+      "### Your safety is now ~I~gON~n:  You can no longer strike innocents."
    user_safety_off = \
-      "### ~B`rSAFETY OFF  `nBe careful, you are now able to "
+      "### Your safety is now ~B~rOFF~n:  Be careful, you are now able to "
       "hurt those around you."   
 
    user_cant_go = "You are unable to go anywhere."
@@ -4327,12 +4327,12 @@ messages:
          string = psPlayerDescription;
       }
       
-      if StringContain(psPlayerDescription,"ï¿½")
+      if StringContain(psPlayerDescription,"¶")
       {
          iCount = 0;
-         while StringContain(psPlayerDescription,"ï¿½")
+         while StringContain(psPlayerDescription,"¶")
          {
-            StringSubstitute(psPlayerDescription,"ï¿½","");
+            StringSubstitute(psPlayerDescription,"¶","");
             iCount = iCount + 1;
             if iCount > 10
             {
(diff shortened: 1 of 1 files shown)


kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -4327,12 +4327,12 @@ messages:
          string = psPlayerDescription;
       }
       
-      if StringContain(psPlayerDescription,"�")
+      if StringContain(psPlayerDescription,"�")
```
akirmse, line 4330: These spurious changes are probably due to different file encodings (probably ASCII vs. UTF-8), but in any case I don't think they're safe.

adrienlaws, line 4330: I don't know why these files show as changed.  I can't even see a change in them.  They appear identical in a search.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -193,9 +193,9 @@ resources:
    user_guildmate_logoff = "~IYour guildmate %s has just departed."
 
    user_safety_on = \
-      "### Your safety is now ~I~gON~n:  You can no longer strike innocents."
+      "### ~B`gSAFETY ON   `nYou can no longer strike innocents."
```
akirmse, line 196: Changing italic to bold is a good idea.  I don't know that the language change is an improvement, though.  If I'm a new user and one of the first things I read is "SAFETY ON" in bold green text, I might think I can't be killed.  "Your safety is now ON" certainly isn't super clear either, but it's arguably better.

kod/object/active/holder/nomoveon/battler/player/user.kod
adrienlaws: I will try and fix this.  I'm noticing this happens with user.kod for me in other test branches.

adrienlaws: This is fixed due to removal in commit d2c5198, resolving

adrienlaws: This is also fixed due to removal in commit d2c5198, resolving

kod/object/active/holder/nomoveon/battler/player/user.kod
adrienlaws: Good feedback.  I changed it back to the normal text, but made the "ON" bolded, removed italics because they made the text harder to read, removed the colon because green and purple clashed, and aligned the message to look a lot better (IMO) since the spacing aligns.  Attaching screenshot example.
new_safety_off_on

PR #468, opened 2023-04-16
Disallowing Karahol to be re-cast prior to expiration

paul-bargerstock:
Applying anonymity logic to Karahol; removing the ability for users to "hold" themselves immediately upon re-self cast.

### DIFF
--- a/kod/object/passive/spell/persench/karahol.kod
+++ b/kod/object/passive/spell/persench/karahol.kod
@@ -98,16 +98,10 @@ messages:
 
    CanPayCosts(who = $, lTargets = $)
    {
-      if lTargets = $
-      {
-         lTargets = [who];
-      }
-      
       % check for enchantment already applied
-      if Send(first(lTargets),@IsEnchanted,#what=self)
+      if Send(who,@IsEnchanted,#what=self)
       {
-         Send(first(lTargets),@MsgSendUser,#message_rsc=KaraholsCurse_already_enchanted_rsc,
-              #parm1=Send(self,@GetIndef),#parm2=Send(self,@GetName));
+         Send(who,@MsgSendUser,#message_rsc=KaraholsCurse_already_enchanted_rsc);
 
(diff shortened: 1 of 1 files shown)


paul-bargerstock:
To paraphrase the feedback from Discord, it isn't intuitive to allow self-punishment due to recasting Karahol.

kod/object/passive/spell/persench/karahol.kod
```diff
@@ -96,6 +96,25 @@ messages:
...
 
+   CanPayCosts(who = $, lTargets = $)
+   {
+      if lTargets = $
+      {
+         lTargets = [who];
+      }
+      
+      % check for enchantment already applied
+      if Send(first(lTargets),@IsEnchanted,#what=self)
```
davidklaw, line 107: Pass in `who` instead of `first(lTargets)`

kod/object/passive/spell/persench/karahol.kod
```diff
@@ -96,6 +96,25 @@ messages:
...
+      {
+         Send(first(lTargets),@MsgSendUser,#message_rsc=KaraholsCurse_already_enchanted_rsc,
```
davidklaw, line 109: Pass in `who` instead of `first(lTargets)` and this parm1 and parm2 are unnecessary -- only the following is needed:
```
Send(who,@MsgSendUser,#message_rsc=KaraholsCurse_already_enchanted_rsc);
```

davidklaw (changes requested):
With a few simple revisions I think this one is good for final review. The resource is already present, so all this really needed was for this message to be implemented.

davidklaw:
@paul-bargerstock Think you'll be able to get back to this?

paul-bargerstock:
Made the requested changes.  Tested inability to re-cast while buff is active, message is correctly displayed.

davidklaw:
Nice job @paul-bargerstock !

PR #474, opened 2023-04-19
fix: prevent prism deletion on DestroyDisposable room function call

adrienlaws:
TLDR; prisms sometimes disappear when placed on the floor, this is bad.  This PR fixes that.

Issue
* Prisms are extremely rare items that must be dropped on the floor to use
* They are used during major events such as breaking into a guildhall which may take extended amounts of time
* the DestroyDisposable function will destroy objects in a room including prisms if they are not bonded  (code from room.kod pasted below for reference)
* Given that prisms are one of the only (perhaps only) items that must be dropped on the floor to use, it doesn't make sense they would be at risk of dissapearing due to a room cleanup function
* There is anecdotal evidence of his happening according to chat in the 101 discord where a user said they lost their prism within 3 minutes

The Fix
* return FALSE on DestroyDisposable() for prism.kod

```if NOT pbUser_in_room
      {
         % If no one is here, then try to destroy everything.
         Send(self,@DestroyDisposable);

         return;
      }
      else
      {
         % Someone's in the room.  If there are more than 5 things in the room,
         %  delete the 20% that are the oldest.  Note:  this may be screwed up
         %  by permanent, non-deleting objects like trees and braziers.

         iLen = Length(plPassive);
         if iLen > 5
         {
            delete_index = 4*iLen/5;
            count = 0;
            for i in plPassive
            {
               count = count + 1;
               if count > delete_index
               {
                  Send(First(i),@DestroyDisposable);
               }
            }
         }
      }
```

### DIFF
--- a/kod/object/item/passitem/prism.kod
+++ b/kod/object/item/passitem/prism.kod
@@ -512,7 +512,13 @@ messages:
 
       return;
    }
-   
+
+   %%% Disposability
+
+   DestroyDisposable()
+   {
+      return FALSE;
+   }
 
 end
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%


davidklaw:
> that must be dropped on the floor to use

I like this fix for that simple fact alone. 👍 from me.

akirmse:
With this change, is it possible to delete a prism at all, or will they accumulate forever?

davidklaw:
> With this change, is it possible to delete a prism at all, or will they accumulate forever?

Oops. In that case, would it make sense to base this off a timer? Maybe one that is created when the prism is dropped and remains unused?

akirmse:
Maybe?  Timers are very easy to get wrong (as we can see in the error logs), but perhaps with care it could be done.

adrienlaws:
There are 5 items in the game that currently have DestroyDisposable return False

- Riija Swords
- Totems
- Signet Rings
- Wedding Rings
- faction flags (not sure if in use at the moment)

Wedding Rings and Signet Rings also have an additional property that protects them from the "dm clear inventory command"
` ReqDMDelete()
   "Will this item be deleted by a DM Clear Inventory command?"
   {
      return FALSE;
   }
`

In short, prisms will accumulate forever except through shatter, a room being deleted with one inside of it, or a character being deleted while holding them.

This may or may not be a problem.  The rate of generation seems quite low.  From what I understand only through the box in the basement of CV will there be a small chance one is generated, or through events.  I would argue there are a lot more Riija swords out there than prisms, but I don't know that.

If we must find a way make these "destroyable" or weak over time, perhaps a number of max uses would be the best way. (50 casts or something)

akirmse:
You're right---in a somewhat old saved game I have, there are ~1700 Riija Swords and ~970 prisms.  I guess I'm OK with this.

PR #475, opened 2023-04-23
Umbrella revisions

davidklaw:
This PR follows on the heels of #455 and revises umbrella in the following ways:

* Umbrella is now a trance spell and the caster must maintain focus to keep the umbrella up. Running, attacking, or taking damage will all cause umbrella to dissipate.
* As a trance spell, Umbrella comes at a cost of 1 mana per second. When the caster's mana is completely drained, the umbrella will dissipate.
* Umbrella protects any other player near the caster, even attackers. They are notified when they enter and leave the protection of the spell.
* The higher the caster's spell power, the higher the likelihood that they maintain their trance when hit. At maximum spell power, this offers a 50% chance of holding the trance. A new message has been added for this event.
* It offers a much higher resistance to attack spells than before (`spellpower/5` to `50 + spellpower/2`) but is balanced by the fact that the caster must hold trance.
* Umbrella now uses a periodic enchantment timer, not a basic enchantment timer. I believe it's the more appropriate enchantment timer for this spell and as a result, the PE icon no longer flickers.

**Trance Breaking**
A few of us talked about the current state, which feels too easily disrupted by any type of damage. Rather than swinging too wildly in the opposite direction, I chose to build in odds based on spell power. While the existing defense modifier helps avoid a lot of ranged attacks (bows in particular), one hit would spell the end of the umbrella. I felt that the odds made it more interesting for higher level players and less likely to be abused by mules, but am open to suggestions.

**Flickering PE Icon**
Umbrella is built upon a 1-second timer that checks for players entering or leaving the umbrella each cycle. But because of how `StartEnchantment` was being called, it was removing and redrawing the PE icon. In my testing it was most obvious late in the spell. I re-implemented it using the periodic enchantment timer, which won't call an end to the enchantment until `lastcall=true`.

**Defense and Resistance Modifiers**
Umbrella adds defense and resistance modifiers to the caster and anyone within range. These effects can be seen added to and removed from players entering and leaving the umbrella. The strength of these modifiers has been returned to their former values, which offer a great deal of defense but also put the caster at risk. It was suggested that we start here and see how it's used before making any tweaks there. When testing, it's important to check that the resistance and defense modifiers come and go as expected on the caster and everyone else in range.

### DIFF
--- a/kod/object/passive/spell/umbrella.kod
+++ b/kod/object/passive/spell/umbrella.kod
@@ -60,6 +60,8 @@ classvars:
    viChance_To_Increase = 35
    viCast_Time = 10000
 
+   viManaDrain = 1 % Drain is amount used every UMBRELLA_CHECK_TIME milliseconds
+
 properties:
    
 messages:
@@ -121,15 +123,16 @@ messages:
       iDuration = bound(iDuration,50,240);
       iDuration = iDuration * 1000;
 
-      % Put spell maintenance info in casters enchantment list.
+      % Kick off a periodic timer for handling players that enter and leave the umbrella
       Send(who,@StartEnchantment,#what=self,#time=UMBRELLA_CHECK_TIME,
-           #state=[iElapsed,lEnchanted,iSpellpower,iDuration]);      
+         #state=[iElapsed,lEnchanted,iSpellpower,iDuration],#lastcall=FALSE);      
+      Send(who,@SetTranceFlag);
 
       propagate;
    }
 
    BreakTrance(who = $, state = $, event = 1)
-   "Clears trance, removes enchantment(s), and sends notifications to any previously enchanted players"
+   "Clears trance, removes enchantment and notifies any players affected"
    {
       local oUser,lEnchanted,iSpellpower;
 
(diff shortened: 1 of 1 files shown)


kod/object/passive/spell/umbrella.kod
```diff
@@ -125,6 +127,25 @@ messages:
...
+
+      Send(who,@ClearTranceFlag);
+      
+      iSpellpower = Nth(state,3);
+      lEnchanted = Nth(state,2);
+      
+      for oUser in lEnchanted
+      {
+         Send(oUser,@MsgSendUser,#message_rsc=Umbrella_break_trance);
+         Send(oUser,@RemoveEnchantment,#what=self);
```
davidklaw, line 143: The original spell called `RemoveEnchantment` on the caster and then again on all enchanted, which in my testing included the caster and would throw a debug message. `RemoveEnchantment` has the same effect when you dig into it.

The original `BreakTrance`:
```
BreakTrance( who = $, state = $ )
{
      local oUser,lEnchanted,iSpellpower;

      iSpellpower = Nth(state,3);

      % Clean up Trance.
      Send(who,@ClearTranceFlag);
      Send(who,@RemoveEnchantment,#what=self);

      lEnchanted = Nth(state,2);
      for oUser in lEnchanted
      {
         Send(oUser,@MsgSendUser,#message_rsc=Umbrella_break_trance);
         Send(self,@LeaveUmbrella,#who=oUser,#iSpellpower=iSpellpower);
      }

      return;
}
```

davidklaw:
Umbrella is currently undergoing heavy discussion in Discord. To be continued...

davidklaw:
Closing to propose a rework of the spell.

PR #477, opened 2023-04-26
Dusk rat mask

TheDiggie:


davidklaw:
Description? Screenshots?

TheDiggie:
https://gyazo.com/56dbcaac7d9c28400ed87fff4405d2df

image

akirmse:
The description should explain why a change exists.  What is this for?

TheDiggie:
Well. I don't have a "reason" for it. Its just a mask I made that fits the meridian theme and style to expand on the available mask we have. Same goes for the troll mask. It's a purely cosmic item that players love

akirmse:
OK.  Closing this as I think it's already possible to get better art through an in-game color translation.

For future reference we'd rather have the original art checked in, instead of the post-processed .bgfs.

PR #481, opened 2023-05-02
UpdatePolishQuest - Give 20 get 10

CardiovascularMike:


CardiovascularMike:
I don't know why some not-changed lines are highlighted.  My first PR.  Any comments appreciated - change is super simple.  Changed text to say that you get half of the gems back as polished gems and changed cargo to 20 gems and reward to 10 gems.  There already exists a mechanism where the NPC will tell you 'give me a few more' if you only give, for example, 15 gems.  Still needed to test: what happens if you give too many?  Likely the NPC takes them and gives you 10 still.  I'm not sure what the testing requirements are prior to the PR as well.

akirmse:
Probably the encoding of the files is wrong, which is causing non-ASCII characters to get changed.  You'll need to check into your editor to see what's going on there.

As for the change... why are the old numbers bad, and the new numbers better?  Why 20 and not 5 or a thousand?

CardiovascularMike:
> Probably the encoding of the files is wrong, which is causing non-ASCII characters to get changed. You'll need to check into your editor to see what's going on there.
> 
> As for the change... why are the old numbers bad, and the new numbers better? Why 20 and not 5 or a thousand?

Thanks for the feedback.  I'll look into the text encoding problem.

The old quest was give 3 gems and get 1 polished.  The problem is that 1 polished gem has no utility to most players.  I've been offered the quest when I had the 3 gems in my inventory numerous times and never bothered to complete it because what am I going to do with 1 gem?  As far as the 20 and 10 figures - well these seemed like a reasonable compromise but are essentially arbitrary.  

There are two gameplay reasons to keep the numbers reasonably small:
1. You don't want the required number of uncut gems to be large enough that players will have to drop other reagents or change their inventory around significantly due only to the extra aggravation on what is already a fetch-quest. 
2. The upper bound also has to consider the value of polished gems.  20-40 per quest isn't enough to impact the game's economy but its always worth considering these factors.

Your questions have made me think through the change more carefully and so I just checked a few things:
1.  Weight: we want to make sure the reward can fit in our players inventory always.  This is virtually guaranteed because uncut gems weigh the equivalent of 5 emeralds and polished gems weigh 3 emeralds so even at 2:1 instead of 3:1 we are very safe.
2. In terms of economy:  Uncuts can be purchased 54 shills and sell for 24 shillings.  Polished you can buy for 200 shillings and sell for 80 shillings.  Total sale value of 20 uncut seraphym is 480 shillings and you receive 800 shillings of value you can sell to an NPC.  These are all tiny numbers in the grand scheme of things but its good to make sure the quest isn't negative value for the players and isn't wildly lucrative either.  The profit for the player in terms of NPC sales would be 320 shillings while the savings for a player who is buying polished anyhow (assuming they buy the uncuts to complete the quest) would be 920 shillings.

Lastly - a better change would be to accept any number of shillings up to some reasonable max and just return half as polished seraphym.  I'm not sure exactly how to accomplish this inside the questengine file as quests generally have a cargo and a reward table.  The cargo always seems to be a fixed quantity.  In terms of what the reasonable max is I'd say 200 would be the highest I'd be comfortable with - that would equate to 3,200 shillings in saleable profit and a sizeable 9,200 shillings.

akirmse:
That makes sense.  It's a good idea to put explanations like this in the PR description first.  The idea is that someone reading it in the future will understand why the change was made, rather than just seeing one set of magic numbers change to another set of magic numbers without a reason.

CardiovascularMike:
I'm going to close this PR and then fix the encoding issue and re-submit the PR as a draft for comment.

PR #482, opened 2023-05-07
Add minor items to Raza's shops

Aesica:
This is intended to make the newbie experience in Raza a little less spartan by adding items to shops that should be more useful to them while in town.  Complete overview of everything added/changed:
- Raza Smith:  Added small round shield, chain armor.  Removed long sword, scale armor.  The former is a bit of a trap item they can't use properly yet, and the latter is overkill armor.
- Raza Apothecary:  Added mushrooms, blue mushrooms, herbs, elderberries, emeralds, flasks, and torches.  In particular I wanted to make sure most lowbie spell reagents (especially those for touch spells) were covered, and now they are except for a few qor spells (such as acid touch) because they use entroot berries.  The torches and flasks should also prove handy.
- Raza Bartender:  Added green shirt, orange shirt, gray tight pants, gray baggy pants, orange long skirt.  Not the most fashionable items, just a taste of how clothing works.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzapoth.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzapoth.kod
@@ -39,9 +39,16 @@ messages:
 
    SetForSale()
    {
-      plFor_sale = [ [ Create(&RedMushroom)],
-                     $,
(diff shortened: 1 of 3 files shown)


davidklaw:
Thoughtful changes for Raza folk👍

PR #483, opened 2023-05-07
Raza portal vigor fix

Aesica:
Currently, exiting Raza sets the player's current vigor to 100 regardless of what it's currently at.  Since players generally leave Raza right after hitting 25 HP, and since gaining a hit point sets your vigor to 200, this is sort of an outdated/weird penalty more than anything else.  This change will bump their current vigor up to 200 (assuming it isn't at or around already) when they exit the Raza portal instead of setting it to 100.,

### DIFF
--- a/kod/object/active/portal/newbport.kod
+++ b/kod/object/active/portal/newbport.kod
@@ -92,7 +92,7 @@ messages:
 
          % Nearly everyone will be leaving Raza at around 200 vigor after
          % hitting 25, so top them off before sending them on their way.
-         Send(what, @EatSomething, #nutrition=200);
+         Send(what, @AddExertion, #amount=-2000000);
 
          % teleporting them out of region will clean inventory, etc.
          Post(what,@AddRealWorldObjects);
--- a/kod/object/active/portal/newbport.kod
+++ b/kod/object/active/portal/newbport.kod
@@ -90,7 +90,9 @@ messages:
             Send(what,@AdminGotoSafety);
          }
 
-         Send(what,@AddExertion,#amount=(send(what,@GetVigor) - 80) * 10000);
+         % Nearly everyone will be leaving Raza at around 200 vigor after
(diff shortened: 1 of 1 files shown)


kod/object/active/portal/newbport.kod
```diff
@@ -90,7 +90,9 @@ messages:
             Send(what,@AdminGotoSafety);
          }
 
-         Send(what,@AddExertion,#amount=(send(what,@GetVigor) - 80) * 10000);
+         % Nearly everyone will be leaving Raza at around 200 vigor after
+         % hitting 25, so top them off before sending them on their way.
+         Send(what, @EatSomething, #nutrition=200);
```
davidklaw, line 95: Out of curiosity, why `EatSomething` over `AddExertion`? Less convoluted? And it looks like not passing in `filling` means they get 200 vigor w/o the fullness effect?

davidklaw (commented):
Seems like a nice change for new characters without any drawbacks I can think of.

kod/object/active/portal/newbport.kod
Aesica: Yeah, less convoluted and easier to look at since the nutrition value and vigor are 1:1.  And yeah, no filling value should default it to 0.  There was at least one other thing in the game that used the same approach which is where I got the idea, but I can't remember where I saw it exactly.  If addexertion is preferred, it's easy enough for me to switch it to that.

akirmse: I'm not 100% convinced this is safe, given that it changes the function that is called, and the existing one is known to work.  Plus I'm not sure this is really much of a problem, so the risk outweighs the reward for me.  If you use AddExertion instead it will likely look like a safer change.

akirmse:
I think this is probably OK---any objections?

davidklaw:
No objection here. Tested and verified.

PR #485, opened 2023-05-13
Refresh Umbrella

davidklaw:
This PR completely changes umbrella, making it a self-enchantment that adds defense against undead battlers.

Highlights of this change:
- Refactors `Umbrella` into a `PersonalEnchantment`, and a self-enchantment only at that
- Reduces cast time and duration to match bless
- Reduces the amount of defense power by 1/2 (ability to avoid being hit)

Damage is reduced based on spellpower and a scaling factor, for which sample data is below. This PR uses a factor of 10.

### Sample data

The following table provides some examples of how much damage is absorbed for a range of damage and spellpower values. The PR sets the damage reduction factor to 10, but this table provides sample data for a factor of 5 for comparison. Notice that a value of 5 provides a pretty substantial advantage at higher spellpower and probably shouldn't be used.

Dmg | Sp | Absorbed(Factor=10) | Absorbed (Factor=10) | Absorbed(Factor=5) | Absorbed(Factor=5)
-- | -- | -- | -- | -- | --
5 | 10 | 1 | 20% | 1 | 20%
5 | 42 | 1 | 20% | 2 | 40%
5 | 77 | 1 | 20% | 3 | 60%
5 | 99 | 2 | 40% | 4 | 80%
10 | 10 | 1 | 10% | 1 | 10%
10 | 42 | 2 | 20% | 4 | 40%
10 | 77 | 3 | 30% | 6 | 60%
10 | 99 | 4 | 40% | 8 | 80%
15 | 10 | 1 | 7% | 1 | 7%
15 | 42 | 3 | 20% | 6 | 40%
15 | 77 | 4 | 27% | 9 | 60%
15 | 99 | 6 | 40% | 12 | 80%
20 | 10 | 1 | 5% | 1 | 5%
20 | 42 | 4 | 20% | 8 | 40%
20 | 77 | 6 | 30% | 12 | 60%
20 | 99 | 8 | 40% | 16 | 80%

### Class hierarchy change
This PR refactors the umbrella down the class hierarchy and per Zaphod, needs to be handled with care, "especially if there's any saved state in Umbrella when the user logs out".

#### How I tested
To test this locally, I first ran a fresh save game using the `master` branch (not what's currently on 101/102). I created a character, gave them umbrella, cast it, then logged them off before saving the game. At this point there was only one umbrella class and one object.

Then, I shut the server down and did a clean build with the PR. There were no errors present in the log on startup nor during a save game. The same single umbrella class and object were present -- same class id and same object id.
Note that the player's resistance modifier from the original Umbrella remains and will need to be dealt with through some other means.

Let me know if there's a better way to test this!

# Deployment notes
After deploying this, you must called `ResetReagents` on Umbrella for the reagent cost change to take effect.


kod/object/passive/spell/persench/umbrella.kod
```diff
@@ -0,0 +1,143 @@
...
+
+   vbCanCastOnOthers = FALSE
+
+properties:
+   
+messages:
+
+   ResetReagents()
+   {
+      plReagents = $;
```
akirmse, line 64: Can you do this instead?
```
plReagents = [ [&Emerald,2], [&PolishedSeraphym,1] ];
```

davidklaw: Done and marking as resolved.

PR #490, opened 2023-05-18
feature: add item rarity grade system

adrienlaws:
# What

- Adds item rarity grades to allow players to more easily differentiate rare items from common ones and each other
- An additional full INT is sent with as an `item_rarity_grade` enum value from the server
- On the client, adds a `rarity` member to client objects
- This change also allows for item text to be colored in dialog boxes (look, get, etc).
- Rare item titles have a suffix with their grade designation, for example:  `feathered choker (rare)`
- There are 5 new rarity grades with associated colors in dialog and list boxes:
  - magic
  - rare
  - legendary
  - unrevealed
  - cursed
- Important: a client object is only assigned ONE rarity grade value at a time
- The rarity grade allows for other presentation options in the future (minimap ping for a unique item, etc)

Example listbox with rarity grade suffixes and text coloring
rarity_grades_example

Example "rare"-grade gold sword item description with a rarity grade suffix in the title
image

# Why

- To make it easier to decide what items to pick up
- It can be difficult for players to identify items with magical properties
- Many items on the floor make it tedious for the user to click on each one and read the description
- Rare items should be highlighted

# Usage
The rarity is configured through 4 different checks
- by class assignment for specific items (MSH = legendary, God's Gifts = rare)
- unrevealed?
- cursed?
- desirable item attribute count?
  - 1 = magic
  - 2 = rare
  - 3+ = legendary

See the following logic diagram to see how competing rarity assignments are resolved

## Logic Diagram
image

## What is a "desirable attribute?"
This is a baseline proposal of attributes that would play into the item rarity grade calculation.  If one or more of these attributes are applied to an item, it will show up as magic, rare, or legendary (if not cursed or unidentified)
```
   IA_TRANSCENDANT
   IA_DURABLE
   IA_MISDIRECTION
   IA_BONDED
   WA_FACTION
   WA_GUILD_HALL_DEFENDER
   WA_EXPERT
   WA_PUNISHER
   WA_CEREMONIAL
   WA_BLINDER
   WA_PARALYZER
   WA_BONKER
   WA_VAMPER
   WA_TWISTER
   WA_SPELLCASTER
   WA_PURGER
```
# Edge Cases
## Rings
Although this PR is primarily intended to address items like weapons and armor, not rings or potions, rings are included for unidentified and cursed rarity grades.  

If we did not assign a rarity grade for unidentified rings, then cursed rings would appear as cursed due to the global logic on `GetRarity`.  The current implementation is that unidentified rings show up as unidentified (orange at the moment) and cursed rings will only show as cursed (or red) if they are revealed.

The end result is 
- Is it unidentified?  If so apply that rarity grade
- Is it cursed? If it's unidentified then don't apply the cursed rarity grade.  If it's revealed then do so.
- Is it revealed?  Then apply normal rarity grade

image

## Potions
Potions are by nature common and are not included in rarity grade calculations.
- Potions drop plentifully and since most are UNIDENTIFIED when dropped from monsters, it would overwhelm the inventory with UNIDENTIFIED grade notifications

Potions are not assigned the UNIDENTIFIED rarity grade when unidentified
- There is no easy way to do this as the functions that check whether something is identified don't apply to potions
- Some potions use the default name
  - For rings I checked if the real name is the same as the display name to see if it's unidentified
  - Since some potions use the default name it's difficult to check for an identified state

## Amulets of Shadows
Amulets of Shadows are not unidentified but are cursed
- This means they will appear as cursed and the rarity grade assignment will take away some of the surprise
- This is probably okay since everyone knows except newbies that they are cursed
- If we want to fix this, we could put in a PR or a fix in this PR to make amulets of shadows unidentified

## `WA_SPELLCASTER` and `WA_ATTACKSPELLTYPE`
Both of these item attributes are related to having a weapon cast a spell
- Note: some other attributes have a spell build in like `WA_BLINDER` (blind) and `WA_PARALYZER` (hold)
- `WA_ATTACKSPELLTYPE` is a flag to show the description ex: `The weapon is hot to the touch`
  - It doesn't add the ability, only the visibility
- `WA_SPELLCASTER` adds the actual spell to the weapon

I assigned `HasRarity()` = `TRUE` for both
- this means that a `NORMAL` item that has both of these will appear as `RARE` and skip `MAGIC`
  - +1 count from each = +2 rarity grade assignment
- I think this is okay because these items are much more rare than `WA_BLINDER` and other common item attribute drops.

# How can I disable this? I hate it.
Under Game->Preferences there is a new configuration option for "Show item rarities"
- uncheck the checkbox to immediately remove rarity suffixes and coloring
image
meridian_pnzvcw2MKV

# Attribution / Inspiration
* skittles1's PR here https://github.com/OpenMeridian/Meridian59/pull/502

# Other changes
- Added wrapper to `ExtractObject` to reduce code duplication with `ExtractObjectNoLight`
- Reveal and Artifice modified
  - Now sends SomethingChanged() to the object owner so the rarity presentation is updated immediately in the client
  - If this modification wasn't done, sometimes things would appear with a rarity grade when they should, and vice versa


davidklaw:
Personally, I'm a fan of this feature. I recently started building a character to experience some of this stuff first hand and the loot is excessive in the couple of build spots I've settled into. Excessive to the point where I've just started ignoring it. Even if the loot count was lower, a change like this would be a nice quality of life change. There's nothing fun about clicking through each item for consideration.

My only question is, is "magic" too limiting? Are there other non-magic modifiers that would make a weapon special, like color or engraving, that should be considered here?

@Meridian59 How do you feel about this? Anything else you'd like to see from it?

clientd3d/color.c
```diff
@@ -403,19 +404,25 @@ HBRUSH DialogCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type)
...
-		return COLOR_LISTFGD;
-	case UNSEL_BGD:
-		return COLOR_LISTBGD;
-	case SEL_FGD:
-		return COLOR_LISTSELFGD;
-	case SEL_BGD:
-		return COLOR_LISTSELBGD;
+		if ((flags & OF_ITEM_MAGIC) != 0)
+				return COLOR_ITEM_MAGIC_FG;
+		else
```
davidklaw, line 413: My only comment is that the indentation of these statements seems excessive. I don't know if that's GitHub or the code.

include/proto.h
```diff
@@ -407,6 +408,7 @@ enum {
 #define GetPlayerFlags(flags)   ((flags) & OF_PLAYER_MASK)
 #define GetDrawingEffect(flags) ((flags) & OF_EFFECT_MASK)
 #define GetDrawingEffectIndex(flags) (((flags) & OF_EFFECT_MASK) >> 20)
+#define GetItemFlags(flags) ((flags))
```
akirmse, line 411: This can't be right---it does nothing.  Is it actually used anywhere?

clientd3d/color.c
```diff
@@ -403,19 +404,25 @@ HBRUSH DialogCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type)
 * GetItemListColor:  Get given color id # for given owner-drawn list box.
 *    (Inventory has different colors than popup dialog lists)
 *    Doesn't return color itself so that caller can use id to call GetBrush.
+*    Now colors magic items; any future item colors should be added here.
```
akirmse, line 407: "Now" will mean nothing to a reader in 10 years.  Better to remove that and instead document what "flags" does.

clientd3d/ownerdrw.c
```diff
@@ -345,16 +345,27 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    SetBkMode(lpdis->hDC, OPAQUE);
    obj = (object_node*)lpdis->itemData;
 
-   hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, (selected? SEL_BGD : UNSEL_BGD)));
+   hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, (selected? SEL_BGD : UNSEL_BGD), NULL));
```
akirmse, line 348: NULL is not an int.  Did you mean 0?

clientd3d/ownerdrw.c
```diff
@@ -345,16 +345,27 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
    if ((style & OD_ONLYSEL) && (style & (OD_DRAWOBJ | OD_DRAWICON)))
-      hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, UNSEL_BGD));
+      hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, UNSEL_BGD, NULL));
```
akirmse, line 350: NULL is not an int.  Did you mean 0?

clientd3d/ownerdrw.c
```diff
@@ -345,16 +345,27 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
 
    FillRect(lpdis->hDC, &lpdis->rcItem, hColorBg);
 
    SetBkMode(lpdis->hDC, TRANSPARENT);
-   crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD)));
+   /* Send object flags for objects with icons in lists for coloring magic weapons.
```
akirmse, line 355: Use C++ style comments here.

clientd3d/ownerdrw.c
```diff
@@ -345,16 +345,27 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+   Character select screen causes a client crash because the character name is
```
akirmse, line 356: I don't understand this comment about the character select screen.  That is in char.dll, which isn't even part of this directory.  This if check looks pretty hacky though I don't know why it's here.

clientd3d/ownerdrw.c
```diff
@@ -345,16 +345,27 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+   obj; this statement causes NULL to be sent in that case */
+   if (style & (OD_DRAWOBJ | OD_DRAWICON))
+   {
+      crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD), obj->flags));
+   }
+   else
+   {
+      crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD), NULL));
```
akirmse, line 364: 0 instead of NULL?

Gar103:
Have some experience with this and how this function might evolve. At the very least, I would suggest:

* using one color for unidentified, and other(s) for identified
* adding a map icon for the magic item on the floor, so users don't have to constantly 'look' at piles of loot, they know exactly where it is the moment it appears
* consider adding a text string or special sound when it drops off a mob, not strictly necessary though

Aesica:
I definitely like the idea of different coloring for identified vs unidentified, but I'm not sure about the proposed map icon or string/special sound.  Inspecting each scimitar when a daemon drops 4 at a time was awful, but people should at least look at their loot, identify that one item sticks out, and acquire it that way rather than "DING DING DING ALERT!  A MAGIC ITEM DROPPED!"

Gar103:
I would agree if Meridian's UI was better, but there are a number of issues that make that continually problematic. Items appear and disappear when too many are on the ground (which basically always happens while building), the look list doesn't give any indication of distance or location, and the order of the look list isn't always preserved because players move around - so you might think you checked a certain section of the list already, but a magic item snuck in there. I'm not a fan of adding sounds to stuff like drops myself, but others seem to like that. I'm more interested in a map icon.

Also I think that asking players to hit 'get' or 'look' after every single kill is kind of a crazy ask. If you don't, and you wait, you risk the item silently disappearing forever. It's a cluster of 90's UI issues that make building and looting way more annoying than it needs to be

clientd3d/color.c
```diff
@@ -403,19 +404,27 @@ HBRUSH DialogCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type)
...
-		return COLOR_LISTFGD;
-	case UNSEL_BGD:
-		return COLOR_LISTBGD;
-	case SEL_FGD:
-		return COLOR_LISTSELFGD;
-	case SEL_BGD:
-		return COLOR_LISTSELBGD;
+		if ((flags & OF_ITEM_SPECIAL) != 0)
+				return COLOR_ITEM_SPECIAL_FG;
+		else
```
akirmse, line 415: You don't need "else", since the "then" always returns.

clientd3d/color.h
```diff
@@ -38,6 +38,7 @@ enum {
    COLOR_BAR4,            /* Bar graph color #4 (numbers in graphs) */
    COLOR_INVNUMFGD,       /* Inventory number foreground */
    COLOR_INVNUMBGD,       /* Inventory number background */
+   COLOR_ITEM_SPECIAL_FG, /* Color for special items in lists*/
```
akirmse, line 41: Missing space at end of comment

clientd3d/ownerdrw.c
```diff
@@ -341,16 +341,29 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+      hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, UNSEL_BGD, 0));
 
    FillRect(lpdis->hDC, &lpdis->rcItem, hColorBg);
 
    SetBkMode(lpdis->hDC, TRANSPARENT);
-   crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD)));
+   /* 
+      Color special item text if the appropriate flags are set.  The character selection screen
+      sends extra data (lParam) when calling this function, so an if statement was added to 
+      prevent dereferencing a null pointer by setting the parameter to 0 instead.
```
akirmse, line 354: Try to repeat less code:
```
int flags = 0;
if (style & (OD_DRAWOBJ | OD_DRAWICON))
  flags = obj->flags;
crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD), flags));
```
Also, readers don't need to know when you added the if statement; simplify comment.

clientd3d/ownerdrw.c
```diff
@@ -341,16 +341,29 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+      crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD), obj->flags));
+   }
+   else
+   {
+      crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD), 0));
+   }
+
    if ((style & OD_ONLYSEL) && (style & (OD_DRAWOBJ | OD_DRAWICON)))
-	crColorText = GetColor(GetItemListColor(lpdis->hwndItem, UNSEL_FGD));
+	crColorText = GetColor(GetItemListColor(lpdis->hwndItem, UNSEL_FGD, obj->flags));
```
akirmse, line 366: Change this to "flags" after previous change.

akirmse:
The code looks OK now.  But one general question on the approach.  With this change, there is a single "special/not special" setting on an object.  Are you sure that only one "special" color is all that you're ever going to want?  If you wanted more (rare/unique/whatever, Diablo style), a different approach would be to have the "special" flag you added be a trigger to parse more color information from the object when it's sent from the server, like perhaps an entire int, which would allow arbitrary colors.  Just checking, as this will be hard to change your mind about later.

akirmse:
What you could do is if the "special" bit is set, there would be an
additional int field in the object in the protocol that determines its
color, something like that.  Certainly much more complex, up to you whether
it's worth it.  Depends on the intent of the change.  We don't want a
special secret code with hundreds of colors, but not sure if only one is
enough.

clientd3d/server.c
```diff
@@ -359,7 +365,12 @@ void ExtractObjectNoLight(char **ptr, object_node *item)
    Extract(ptr, &item->icon_res, SIZE_ID);
    Extract(ptr, &item->name_res, SIZE_ID);
    Extract(ptr, &item->flags, 4); // includes drawfx_mask bits
-
+   Extract(ptr, &item->special_type, SIZE_ID);
```
akirmse, line 368: ExtractObjectNoLight is an abomination---it means code duplication, and requires you to make changes twice, as in this case.  We should make a separate change that merges the code in ExtractObject and ExtractObjectNoLight.  You can do it if you'd like, or I can do it, and then you'd have to merge my change into your branch (and probably resolve a merge conflict).

include/proto.h
```diff
@@ -491,6 +492,16 @@ enum {
    SL_FLICKER_OFF      = 2,         // Turn flickering off
 };
 
+/* Special item types */
```
akirmse, line 495: The name "special" isn't doing it for me.  I don't think a reader would understand what a special object is.  This property is really only about how the object is drawn, so perhaps you can come up with a name that is more descriptive of what it does.

clientd3d/color.c
```diff
@@ -403,19 +408,38 @@ HBRUSH DialogCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type)
 * GetItemListColor:  Get given color id # for given owner-drawn list box.
 *    (Inventory has different colors than popup dialog lists)
 *    Doesn't return color itself so that caller can use id to call GetBrush.
+*    Also colors special items.  When the client receives a OF_ITEM_SPECIAL flag
+*    for an object, the client will color the object's text in lists based on the
+*    value of special_type and the corresponding COLOR_ITEM_SPECIAL in color.h
 */
-WORD GetItemListColor(HWND hwnd, int type)
+WORD GetItemListColor(HWND hwnd, int type, int flags, enum special_type special_type_value)
```
akirmse, line 415: You don't "enum" here (I didn't even know you could put it here actually).

clientd3d/ownerdrw.c
```diff
@@ -330,6 +330,7 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    AREA area;
    COLORREF crColorText;
    HBRUSH hColorBg;
+   enum special_type special_type_value = SPECIAL_TYPE_NONE;
```
akirmse, line 333: No "enum" here.

include/proto.h
```diff
@@ -491,6 +492,16 @@ enum {
...
+enum special_type {
```
akirmse, line 496: To avoid having to use "enum" everywhere, typically this is done:

typedef enum { ... } special_type;

include/proto.h
```diff
@@ -491,6 +492,16 @@ enum {
...
+    SPECIAL_TYPE_NONE      = 0,
+    SPECIAL_TYPE_MAGIC     = 1,
+    SPECIAL_TYPE_RARE      = 2,
+    SPECIAL_TYPE_LEGENDARY = 3,
+    SPECIAL_TYPE_UNKNOWN   = 4,
+    SPECIAL_TYPE_CURSED    = 5,
+};
```
akirmse, line 503: Enums are unique values, meaning that for example, you could never have magic and cursed at the same time.  Just making sure that's what you want, because it can't easily be changed.  If they could be combined, you'd make these bitflags instead.

clientd3d/ownerdrw.c
```diff
@@ -341,16 +342,33 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+
+   // Now we need to do the 2 foreground text options
+   // First get the flags if we are checking for an item not the player object like at the character select screen
+
+   int flags = 0;
+   // Color special item text if the appropriate flags are set. Do this by getting the flags and the special_type
+   if (style & (OD_DRAWOBJ | OD_DRAWICON) && obj != NULL)
+   {
+      flags = obj->flags;
+      special_type_value = (enum special_type)obj->special_type; // Retrieve the special type value from the object
```
akirmse, line 365: No "enum"

clientd3d/ownerdrw.c
```diff
@@ -341,16 +342,33 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+   }
+   crColorText = GetColor(GetItemListColor(lpdis->hwndItem, (selected? SEL_FGD : UNSEL_FGD), flags, special_type_value));
```
akirmse, line 367: Some of the lines in this function are super long; please split them across multiple lines.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -2414,6 +2414,16 @@ messages:
 
       AddPacket(4,iFlags);
 
+      % Send special type if special flag is set
```
akirmse, line 2417: Maybe this would be better done by a new function (on Item?  Not sure) that returns the "special type".  So then this would just be

```AddPacket(4, Send(what, @GetSpecialType)); ```
(with a better name perhaps)

Come to think of it, if you're always adding 4 bytes to every message anyway, what is the purpose of ITEM_SPECIAL_YES?  Could the client simply test if the special type is 0?

File: clientd3d/color.c (hunk 15 above)
Comment (adrienlaws) [reply clientd3d/color.c]: changed `enum` to `typedef enum` in `proto.h` as suggested and adjusted related function parameters

File: clientd3d/ownerdrw.c (hunk 16 above)
Comment (adrienlaws) [reply clientd3d/ownerdrw.c]: changed `enum` to `typedef enum` in `proto.h` as suggested and adjusted related function parameters

File: include/proto.h (hunk 17 above)
Comment (adrienlaws) [reply include/proto.h]: Changed as suggested.

File: include/proto.h (hunk 18 above)
Comment (adrienlaws) [reply include/proto.h]: I thought about this and I think it is ok if they are discrete values.  The main issue this PR tries to address is to quickly identify desirable items without reading the item's description.  Unknown (unrevealed) should always come first, then cursed, then a single value of magic, rare, or legendary as applicable.

I could see some small edge cases where perhaps there's a legendary weapon that is cursed which would be interesting to know about.  But I think that's an outlier case and isn't really what this PR's intent is about.

The text colors should help answer the question: "Should I pick up this item?"

File: clientd3d/ownerdrw.c (hunk 19 above)
Comment (adrienlaws) [reply clientd3d/ownerdrw.c]: Updated this reference after changing `proto.h` to include `typedef enum {} special_type` instead of `enum special_type`

File: kod/object/active/holder/nomoveon/battler/player/user.kod (hunk 21 above)
Comment (adrienlaws) [reply kod/object/active/holder/nomoveon/battler/player/user.kod]: Agreed.  I have created a new function in `item.kod` as suggested and will refine it.

I have also eliminated the `ITEM_SPECIAL_YES` flag completely, so there's still that bit available for something else one day.

File: include/proto.h (hunk 14 above)
Comment (adrienlaws) [reply include/proto.h]: Understood.  I removed all references for special in this PR.  I changed the enums to be more specific so there's no danger of misunderstanding what they're for, and how they are different than other "special" constants or properties.
`ITEM_TEXT_COLOR_NONE`, `ITEM_TEXT_COLOR_MAGIC`, `ITEM_TEXT_COLOR_RARE`, etc.

File: clientd3d/ownerdrw.c (hunk 20 above)
Comment (adrienlaws) [reply clientd3d/ownerdrw.c]: Split across lines as requested.

File: kod/object/item.kod [hunk 22]
```diff
@@ -613,6 +613,28 @@ messages:
       return 1;
    }
 
+   GetItemTextColor(what = $)
```
adrienlaws, line 616: I'm still working on this part.

clientd3d/object.h
```diff
@@ -79,6 +79,7 @@ typedef struct {
 		                     this field gives amount of object */
    DWORD     temp_amount;         /* Scratch field used when user is selecting amount of object */
    int       flags;               /* Flags describing various properties of objects */
+   int       text_color_type;     /* Item text color type */
```
akirmse, line 82: This comment is not exactly illuminating.  What is a text color type?

clientd3d/ownerdrw.c
```diff
@@ -341,16 +342,39 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    SetBkMode(lpdis->hDC, OPAQUE);
    obj = (object_node*)lpdis->itemData;
 
-   hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, (selected? SEL_BGD : UNSEL_BGD)));
+   hColorBg = GetBrush(
+                  GetItemListColor(
+                     lpdis->hwndItem,
+                     (selected? SEL_BGD : UNSEL_BGD),
+                     text_color_value));
```
akirmse, line 349: It looks like text_color_value can only be ITEM_TEXT_COLOR_NORMAL at this point?

clientd3d/ownerdrw.c
```diff
@@ -341,16 +342,39 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+
    if ((style & OD_ONLYSEL) && (style & (OD_DRAWOBJ | OD_DRAWICON)))
-      hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, UNSEL_BGD));
+      hColorBg = GetBrush(
+                     GetItemListColor(
+                        lpdis->hwndItem,
+                        UNSEL_BGD,
+                        text_color_value));
```
akirmse, line 356: here too

clientd3d/server.c
```diff
@@ -389,7 +365,7 @@ void ExtractObjectNoLight(char **ptr, object_node *item)
 object_node *ExtractNewObject(char **ptr)
 {
    object_node *item = ObjectGetBlank();
-   ExtractObject(ptr, item);
+   ExtractObject(ptr, item, true);
```
akirmse, line 368: You don't need to put true everywhere.. that's what the default value is for.

module/mailnews/mailnews.c
```diff
@@ -291,7 +291,7 @@ Bool HandleLookNewsgroup(char *ptr, long len)
    Extract(&ptr, &newsgroup, SIZE_NEWSGROUP_ID);
    Extract(&ptr, &permission, 1);
 
-   ExtractObject(&ptr, &obj);
+   ExtractObject(&ptr, &obj, true);
```
akirmse, line 294: No need to change these once the default value is in the header.

clientd3d/server.c
```diff
@@ -316,7 +316,7 @@ void ExtractDLighting(char **ptr, d_lighting *dLighting)
  * ExtractObject: Get an object_node from ptr, and increment
  *   ptr appropriately.  Place data in given object node.
  */
-void ExtractObject(char **ptr, object_node *item)
+void ExtractObject(char **ptr, object_node *item, bool includeLight = true)
```
akirmse, line 319: Yay, thank you!  But, it turns out the default value needs to go in the header file, due to weirdnesses about how default arguments work in C++:

https://stackoverflow.com/questions/14902129/default-parameters-in-h-and-cpp-files

kod/include/blakston.khd
```diff
@@ -40,6 +40,14 @@
    ACTIVE = 1
    PASSIVE = 2
 
+   %%% Item text color enum
```
akirmse, line 43: The reader wants to know what an "item text color" is.

include/proto.h
```diff
@@ -491,6 +491,16 @@ enum {
    SL_FLICKER_OFF      = 2,         // Turn flickering off
 };
 
+/* Item text color types */
+typedef enum {
```
akirmse, line 495: I believe you can actually leave out the "typedef" in C++, give it a try.

clientd3d/server.c
adrienlaws: Refactored `ExtractObjectNoLight` as requested so it's contained within this PR.

clientd3d/server.c
adrienlaws: Changed all references  to remove the `true` parameter after adjusting the default value in `server.h`

Also used `True` instead of `true` since the capitalized version is what is used throughout the rest of the `server.c` file.

module/mailnews/mailnews.c
adrienlaws: Reverted to remove the `true` parameter.

clientd3d/server.c
adrienlaws: This was very interesting to learn about.  I moved the default value for the `includeLight` parameter to the header file declaration.

include/proto.h
adrienlaws: I tried but if I remove the typedef I have to use `enum` every time I declare a variable.

example without `typedef`:
`enum item_text_color text_color_value= ITEM_TEXT_COLOR_NORMAL;`

example with `typedef`:
`item_text_color text_color_value= ITEM_TEXT_COLOR_NORMAL;`

I might be wrong, but if the above is true then I think I would recommend sticking with a `typedef` declaration so the rest of the code is a bit cleaner when using `item_text_color`

kod/include/blakston.khd
adrienlaws: Updated comment based on feedback.

clientd3d/object.h
adrienlaws: updated object member to be `rarity` and updated comment

clientd3d/ownerdrw.c
adrienlaws: Yes it is essentially = 0 at this point.  These two functions are fetching the background color.

For my change, I am only affecting the foreground text color.  The logic of `GetItemListColor` first checks to see if the text_color_value is anything other than 0 or `ITEM_TEXT_COLOR_NORMAL`.  If it is not set, then it checks to see what the background color is.  If if it is set, then it will return the color for the foreground text.

In these two functions (which existed previously) I am passing it the parameter 0 or `ITEM_TEXT_COLOR_NORMAL` so that it will skip the foreground text color section.

Should I try and refactor `GetItemListColor`?  The logic works but I definitely need to pass 0 for the third parameter in some form currently.

Aesica:
A few things real quick:

1. Besides cursed (obvious) and magic (assuming it means items with 1 or more special properties that drops from enemies), how do the various tiers fit into the game?  In particular, rare, legendary, and mysterious.  I'm not against having several tiers, but they should be meaningful in terms of how they fit into the game. 

2. What happens if a spirit helm, which would make sense as being "legendary" has special properties that would also place it into the "magic" category?

3. I'd like to recommend another tier--junk/poor, for items with little to no actual use.  Currently the only truly worthless item is the egg shell, but in a future PR I'd like to add some "junk" items to diversify monster drops, and this kind of rarity would be a great way to tell players they can safely discard, ignore, or sell the items to an npc.

adrienlaws:
Good questions.

The way I was thinking of approaching it (and definitely open to feedback) is actually in 2 PRs:
1. PR to implement the protocol and logic (this one)
Implement the item rarity grade system into the client and server.  This means it works every time, without issue, when the appropriate constants are sent through Blakod to the client.  It makes item list names the correct color and appends the correct suffix to the item name.  Set it to default normal item in all cases but everything is ready for the next PR.

2. PR to Baseline the rarity grade system.  
Decide what is rare, what is magic, etc.  But the beauty in this is if the first PR (this one) is implemented correctly, this is only done in one new Blakod function `GetItemRarityGrade(what = $)` which returns the appropriate rarity.  In this PR we could debate and come to a consensus on what should be normal, magic, rare, legendary, mysterious and cursed.

My thought was
- magic        = 1 item attribute
- rare           = 2 item attributes
- legendary = 3 item attributes
- cursed      = check if cursed with the existing IsCursed() function
- mysterious = unrevealed.  
   I didn't want to use the word unrevealed because that links it to a spell.  But what it really means is, it has item attributes but we can't see them (until it's revealed and then the other checks above would show the appropriate rarity grade.)
 
You could also do things like - check for item and if it is that item, return a rarity grade.  Perhaps items like Magic Shield Helms or Gold Swords or Gods Gifts should be in one of the tiers... but which one?  

I feel like that debate should happen in a future Baseline PR.  But people could easily submit a PR for the `GitItemRarityGrade` function as needed and the change wouldn't be that big.  If the first PR is completed (this one), then the item rarity grade system is available.

Aesica:
> One question I have from your comment - what do you envision a junk/poor tier color being?

Following the commonly-used convention seen in nearly all games with a rarity system, gray would be ideal.  Not saying we have to go this route, but it brings a sense of familiarity that is instantly recognizable by anyone who has played one or more games with multiple rarities, so I'll post it as a guieline:

- Gray: inferior
- White: common
- Green: uncommon
- Blue: rare
- Purple: epic
- Orange: legendary
- Red: whatever's above legendary, but from here it breaks down and different games go their own way.  Cursed is fine for red, however it would completely eliminate the "surprise" factor.
- Yellow/Gold:  Sometimes comes above red, but this might be good for items with attributes that haven't been exposed to the player (unrevealed) yet.

I know in the meridian sense, blue being below green makes more sense (blue names are below green names) but someone coming into meridian might be confused and find it a bit jarring.

Also I really like the number of attributes affecting the rarity, however a few more thoughts on this:

- Some attributes should probably be excluded.  An item with the temporary attribute (made via create weapon) would already be 1 grade above common.  Put a temporary weapon enchantment on it as well and it's now 2 grades above common.
- Certain items should be able to return their own value, forcing a certain rarity to show regardless of special attributes applied.  Ideally it would be something an admin could manipulate as well for special event weapons, armor, and such.  For example, a colored scimitar could be set to be 1 grade above common even without any special attributes.

clientd3d/dialog.c
```diff
@@ -651,7 +652,40 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
...
+       break;
+    case ITEM_RARITY_GRADE_CURSED:
+       sprintf(desc, "%s (cursed)", info.name);
+       info.name = desc;
+       break;
+    default:
+       // Handle unknown rarity gracefully (no prefix)
+	   debug(("Unknown rarity grade\n"));
+       break;
+    }
```
akirmse, line 687: This block is too repetitive:

- Instead, set a new variable to the text depending on type, e.g. "rare", "magic", "cursed", or the default of nullptr.
- If the new variable is not nullptr, then build up the new description, but only in one place.
- Use snprintf instead of sprintf to avoid buffer overflows.

clientd3d/ownerdrw.c
```diff
@@ -68,21 +68,59 @@ void WindowEndUpdate(HWND hwnd)
...
-   /* If item is a number object, add its amount after the item's name */
+
+   // Suffix the name based on the object's rarity
+   switch(obj->rarity) 
+   {
+   case ITEM_RARITY_GRADE_NORMAL:
+      strcpy(suffix, ""); // Empty string
+      break;
+   case ITEM_RARITY_GRADE_MAGIC:
+      strcpy(suffix, " (magic)");
```
akirmse, line 86: This duplicates all of the names from the previous code.  That's a recipe for problems in the future.

Instead, make a single function that takes a rarity and returns a string.

Also, to make future translation possible, move these text strings to the .rc file.

clientd3d/ownerdrw.c
```diff
@@ -68,21 +68,59 @@ void WindowEndUpdate(HWND hwnd)
...
+   {
+      name = nameWithSuffix;
+   }
+
    /* If player using an object, say so */
    if (IsInUse(obj->id))
    {
-      sprintf(desc, "%.*s %s", MAXNAME, name, GetString(hInst, IDS_INUSE));
-      name = desc;
+      sprintf(name, "%.*s %s", MAXNAME, name, GetString(hInst, IDS_INUSE));
```
akirmse, line 123: Change any sprintfs to snprintfs.

clientd3d/server.c
```diff
@@ -351,33 +353,7 @@ void ExtractObject(char **ptr, object_node *item)
...
-
-   if (OF_BOUNCING == (OF_BOUNCING & item->flags))
-   {
-      item->bounceTime = (WORD)(rand() % 1000);
-   }
-   if (OF_PHASING == (OF_PHASING & item->flags))
-   {
-      item->phaseTime = (WORD)(rand() % 1000);
-   }
+   ExtractObject(ptr, item, False); // Disable light extraction
```
akirmse, line 356: bool has the values true/false, not True/False

kod/object/item.kod
```diff
@@ -613,6 +613,30 @@ messages:
       return 1;
    }
 
+   GetItemRarityGrade(what = $)
+   "Returns the appropriate text coloring tier for the item."
+   {
+   % first check if it's revealed, if not send unrevealed color.
```
akirmse, line 619: This is all commented out?

clientd3d/ownerdrw.c
```diff
@@ -341,16 +380,39 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    SetBkMode(lpdis->hDC, OPAQUE);
    obj = (object_node*)lpdis->itemData;
 
-   hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, (selected? SEL_BGD : UNSEL_BGD)));
+   hColorBg = GetBrush(
+                  GetItemListColor(
+                     lpdis->hwndItem,
+                     (selected? SEL_BGD : UNSEL_BGD),
+                     item_rarity_value));
```
akirmse, line 387: It looks like item_rarity_value can only be ITEM_RARITY_GRADE_NORMAL here, so use that constant instead to make it easier for the reader.

clientd3d/ownerdrw.c
```diff
@@ -341,16 +380,39 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+
    if ((style & OD_ONLYSEL) && (style & (OD_DRAWOBJ | OD_DRAWICON)))
-      hColorBg = GetBrush(GetItemListColor(lpdis->hwndItem, UNSEL_BGD));
+      hColorBg = GetBrush(
+                     GetItemListColor(
+                        lpdis->hwndItem,
+                        UNSEL_BGD,
+                        item_rarity_value));
```
akirmse, line 394: here too

clientd3d/dialog.h
```diff
@@ -83,7 +84,7 @@ typedef struct {
 
 M59EXPORT void SetDescParams(HWND hParent, int flags);
 M59EXPORT void DisplayDescription(object_node *obj, BYTE flags, char *description, 
-                                  char* fixed_string, char *url);
+                                  char* fixed_string, char *url, int rarity = ITEM_RARITY_GRADE_NORMAL);
```
akirmse, line 87: We want to avoid proliferating default values; they are kind of a bad practice.  The one for ExtractObject is a convenience because there are so many callers, and it would be a pain to change them all.  But there are only a few callers of this, so no need for a default value.

clientd3d/dialog.c
```diff
@@ -11,7 +11,7 @@
 
 #include "client.h"
 
-#define PAGE_BREAK_CHAR '�'	  /* For multi-page descriptions */
+#define PAGE_BREAK_CHAR '¶'	  /* For multi-page descriptions */
```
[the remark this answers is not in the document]
adrienlaws: Changed my Visual Studio Code encoding to `Windows 1252` from `UTF-8` and that seems to have fix this.  By default it uses `UTF-8`.

clientd3d/dialog.h
adrienlaws: Understood.  Adjusted all callers and removed this default parameter.

clientd3d/ownerdrw.c
adrienlaws: Changed all occurrences as requested.

clientd3d/ownerdrw.c
adrienlaws: Done
- Created new function `GetRaritySuffix(int rarity)` which returns a string as requested.
- Moved text strings to .rc file and referenced appropriately

clientd3d/server.c
adrienlaws: This one really caught me, because I checked beforehand.  The capitalization seemed off.

However, there are so many instances in this file with capitalization that I wasn't sure what to use:
- 78 instances of `return True;`
- 84 instances of `return False;`

It looks like there is an enum in `client.h` for these capitalized booleans
- `enum {False = 0, True = 1};`

I don't mind changing to the lowercase values of `true` and `false`.  Should I do that here?  Or keep the capitalized values so they're the consistent within this file?

clientd3d/ownerdrw.c
adrienlaws: Changed parameter to `ITEM_RARITY_GRADE_NORMAL` constant as requested.

clientd3d/ownerdrw.c
adrienlaws: Changed parameter to `ITEM_RARITY_GRADE_NORMAL` constant here also, as requested.

clientd3d/ownerdrw.c
adrienlaws: Adjusted default parameter as requested in other comments.

clientd3d/ownerdrw.c
adrienlaws: Adjusted default parameter as requested in other comments.

clientd3d/client.rc
```diff
@@ -1785,6 +1785,15 @@ BEGIN
     IDS_SIGNUPMSG           "Your new account [%s] was successfully created!\n\nBefore you can login and play...\n\nCheck your e-mail (%s) to verify your account now."
 END
 
+STRINGTABLE
+BEGIN
+    IDS_RARITY_GRADE_MAGIC       "(magic)"
```
akirmse, line 1790: The parentheses shouldn't be in here.  That makes it impossible to use the names without parentheses elsewhere.

clientd3d/dialog.c
```diff
@@ -663,6 +674,41 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
...
+* GetRaritySuffix:  Provide a rarity suffix based on the given rarity type.
+*/
+std::string GetRaritySuffix(int rarity) 
+{
+	std::string suffix;
+	
+	switch(rarity)
+	{
+        case ITEM_RARITY_GRADE_NORMAL:
+            suffix = "";
```
akirmse, line 686: Just return "" here.

clientd3d/dialog.c
```diff
@@ -663,6 +674,41 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
...
+			break;
+        case ITEM_RARITY_GRADE_MAGIC:
+            return GetString(hInst, IDS_RARITY_GRADE_MAGIC);
```
akirmse, line 689: Don't repeat all these calls to GetString.  Instead set a variable to the string to fetch, then call GetString once after the switch.

clientd3d/dialog.c
```diff
@@ -663,6 +674,41 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
...
+			break;
+        case ITEM_RARITY_GRADE_MYSTERIOUS:
+            return GetString(hInst, IDS_RARITY_GRADE_MYSTERIOUS);
+			break;
+        case ITEM_RARITY_GRADE_CURSED:
+            return GetString(hInst, IDS_RARITY_GRADE_CURSED);
+			break;
+		default:
+			debug(("Unknown rarity grade\n"));
+            suffix = ""; // Handle unknown rarity gracefully (no suffix)
```
akirmse, line 705: return "" here

clientd3d/dialog.c
```diff
@@ -663,6 +674,41 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
...
+            return GetString(hInst, IDS_RARITY_GRADE_LEGENDARY);
+			break;
+        case ITEM_RARITY_GRADE_MYSTERIOUS:
+            return GetString(hInst, IDS_RARITY_GRADE_MYSTERIOUS);
+			break;
+        case ITEM_RARITY_GRADE_CURSED:
+            return GetString(hInst, IDS_RARITY_GRADE_CURSED);
+			break;
+		default:
+			debug(("Unknown rarity grade\n"));
```
akirmse, line 704: Log the value of the unknown rarity, too.

clientd3d/server.c
akirmse: The short answer is there are two different types.  There is "Bool" that has the values True/False, and "bool" that has the values true/false.  Don't mix them.

The longer answer is that when Meridian was written, there wasn't a good C++ compiler, so rather we were unable to use C++'s builtin "bool" type, and instead we defined our own "Bool" type (that enum you found).  Today we would not do this, but instead we'd simply use "bool" everywhere.  New code should use bool.

Windows is even older, and has its own BOOL type that has the values TRUE and FALSE.  This is to be avoided except when dealing with Windows calls that require it.

kod/object/item.kod
```diff
@@ -613,6 +613,52 @@ messages:
...
+         return ITEM_RARITY_GRADE_NORMAL;
+      }
+
+      if Send(self,@IsCursed)
+      {
+         return ITEM_RARITY_GRADE_CURSED;
+      }
+
+      % Static class rarities
+      if IsClass(self,&helm)               % Magic Spirit Helm
```
akirmse, line 634: This is not good object-oriented design.  We are up in "Item" here, which is an extremely high-level and general class.  There should be no mention of lower classes like Helm.

kod/object/item.kod
```diff
@@ -613,6 +613,52 @@ messages:
...
+      {
+         irarity = ITEM_RARITY_GRADE_LEGENDARY;
+      }
+
+      % Want to refactor to check a list of classes to see if 
+      % the item class is in there and assign appropriately
+      if IsClass(self,&KraananCharm)       % Steel Torc
```
akirmse, line 641: Same thing here.

kod/object/item.kod
```diff
@@ -613,6 +613,52 @@ messages:
...
+         OR IsClass(self,&RiijaCharm)      % Feathered Choker
+         OR IsClass(self,&ShalilleCharm)   % Diamond Pendant
+         OR IsClass(self,&MagicBow)
+         OR IsClass(self,&NeruditeBow)
+         OR IsClass(self,&GoldSword)
+      {
+         irarity = ITEM_RARITY_GRADE_RARE;
+      }
+
+      if IsClass(self,&RiijaSword)         % Feathered Choker
```
akirmse, line 653: Same thing here.

It sounds like perhaps you want a function GetRarity that takes no arguments, and returns a class variable, itself defaulted to ITEM_RARITY_GRADE_NORMAL in Item.  This function could check if the item is cursed.  Subclasses would override the class variable if their rarity isn't NORMAL.

clientd3d/dialog.c
adrienlaws: Modified 2 instances as requested to `return ""`.  Normal and Default / Unknown rarity.

clientd3d/dialog.c
adrienlaws: Modified 2 instances as requested to `return ""`.  Normal and Default / Unknown rarity.

clientd3d/client.rc
adrienlaws: Moved the parentheses out of the `client.rc` file as requested.  They are now in the calling `snprintf` functions.

clientd3d/server.c
adrienlaws: Thank you for the explanation!  I adjusted the `bool` parameter for `ExtractObject` to use `true` and `false` (lowercase).

This was changed in 2 places:
-  the `server.h` file for the default `bool` value of `true`
- `server.c`'s wrapper function `ExtractObjectNoLight` with `bool` parameter `false`

clientd3d/dialog.c
adrienlaws: Added rarity value to unknown rarity grade debug message as requested.

clientd3d/dialog.c
adrienlaws: Changed `GetRaritySuffix` to only have one call to `GetString` based on a new `StringID` variable determined by a switch case.

clientd3d/dialog.c
adrienlaws: I don't know if my changes addressed this properly.  Does the current implementation of `GetRaritySuffix` work?  I did not use a nullptr in the implementation I but did address the other 2 points.

kod/object/item.kod
adrienlaws: Thank you, I refactored the rarity logic based on this feedback.  It's starting to make sense to me.  This works really well for item classes.

I would like to, if possible, implement 2 other ways items could have a higher rarity grade than "normal."
- some logic to count up the item attributes and return a rarity of magic (1 attribute), rare (2+ attributes)
- a way for administrators or DMs to assign the rarity grade to something they make

I will try and modify `GetRarity` to do allow for those 2 conditions.

kod/object/item.kod
adrienlaws: Refactored as suggested.

kod/object/item.kod
adrienlaws: Refactored as suggested.

kod/object/item.kod
adrienlaws: Sorry was still working on this, resolving this comment as I now have some code attempts for this function.

Gar103:
I wouldn't worry about a complex tier system. Meridian doesn't really have that. 2 mods on something doesn't even mean better than 1 mod, and almost all magic items are just 1 mod. The only thing we really need item colors for is to pick magic items out of a loot list on the ground. So this definitely doesn't need to go into magic/rare/epic/legendary territory

akirmse:
This change has gotten so huge that it is too hard to review.  It needs to be split up.  One way that might make sense is 3 PRs:

1) client-side support
2) Blakod infrastructure changes
3) Changes to individual Blakod classes to set their rarity

Done in this order, the code will always be shippable.  There may be code that's never activated, but nothing should break.  This is important because there may be a client update somewhere in the middle of getting these PRs checked in.

One thing to look at when refactoring this into multiple changes is that I strongly feel that there should not be an option to turn off the display of rarity colors.  If the system is good, people won't want to turn it off.  "Make it an option" is often a way of giving up on improving a design and asking the players to do extra work.  (e.g. you can't turn the colors off in Diablo I don't think)

PRs that are open for this long (almost a year!) tend to run into merge problems and feature creep.  Something to think about for next time.

adrienlaws:
Makes sense.  I will close this PR and open smaller, refactored ones in the order you mentioned.

Also, that was interesting feedback on the "option to disable."  I will remove it.  Thank you.

PR #492, opened 2023-05-19
Chalice update

Aesica:
This update to chalice.kod does the following:

- Displays durability text in the chalice's description to give the user a general idea of how many remaining uses it has left.
- Sends a message to the user and plays a sound to inform them when a chalice is successfully refilled.
- When picking up a second chalice, allows the user's primary chalice to inherit the max hits of the secondary one if it has a better max hits value.
- Displays a slightly clearer message when the final sip is used to inform the user know they're supposed to lose it.


kod/object/item/passitem/chalice.kod
```diff
@@ -26,9 +26,18 @@ resources:
...
    chalice_use = "You sip a bit of the water out of the Chalice of Rain."
-   chalice_empty = "You sip the last bit of water from the Chalice of Rain."
+   chalice_empty = "The Chalice of Rain crumbles to dust in your hand as you sip the last bit of water from it."
    chalice_combine = "You pour the water from one Chalice into the other."
    chalice_cant_use_pkill = "Only innocents may partake of the chalice."
+   chalice_refilled = "The Chalice of Rain glows faintly for a moment, renewed by the spirit of Shal'ille that permeates the area."
+   chalice_refilled_sound = water_scoop01.wav
+
+   chalice_condition_exc = " is filled to the brim with pure water."
+   chalice_condition_exc_mended = " is filled to the brim with pure water."
```
akirmse, line 36: These two strings are the same?

kod/object/item/passitem/chalice.kod
```diff
@@ -48,7 +57,17 @@ classvars:
 
    viValue_average = 500
 
+   vbShow_condition = TRUE
+   vrCondition_exc =   chalice_condition_exc
+   vrCondition_exc_mended = chalice_condition_exc_mended
+   vrCondition_good =  chalice_condition_good
+   vrCondition_med =   chalice_condition_med
+   vrCondition_poor =  chalice_condition_poor
+   vrCondition_broken = chalice_condition_broken
```
akirmse, line 66: Spacing after equals signs don't match up.

kod/object/item/passitem/chalice.kod
```diff
@@ -48,7 +57,17 @@ classvars:
...
+
 properties:
+
+   poLastOwner = $
```
akirmse, line 70: This extra variable is pretty awkward.  Can't you instead send the message and sound to the user at the end of ReqNewOwner, and remove this variable?

kod/object/item/passitem/chalice.kod
```diff
@@ -151,7 +190,7 @@ messages:
 
          Post(self,@Delete);
       }
-
+
```
akirmse, line 193: Undo whitespace change.

kod/object/item/passitem/chalice.kod
```diff
@@ -76,15 +95,22 @@ messages:
       }
 
       % If we drop it in a room with a high Shal'ille bonus, refill the chalice
+      % And inform the last owner (if any) that the chalice has been refilled
```
akirmse, line 98: Fix grammar (lowercase And?  Two sentences with periods?)

kod/object/item/passitem/chalice.kod
Aesica: Yeah.  Chalices can't be mended so having alternate text for edge cases like dm-created chalices with only 1 or 2 max hits doesn't seem necessary.  I can change it to something different if you want, but players are unlikely to ever see it under normal circumstances.

kod/object/item/passitem/chalice.kod
Aesica: I could, but I'd also probably have to move the refill logic there as well, since I only wanted to notify the user on a successful refill (feels more polished) rather than any time it's dropped somewhere it can be refilled.  If that's fine, I'll go ahead and do that, but yeah I do agree that while I'm not fond of the variable, I wasn't sure of any other way to remember who the message needed to be sent to.

kod/object/item/passitem/chalice.kod
Aesica: Oof, I'll have to keep an eye out for those in the future.  I blame VS~

kod/object/item/passitem/chalice.kod
Aesica: I tend to be a bit lazy with comments and grammar, but consider it done. :)

davidklaw:
Bump! I'm revisiting my audio file and was curious if there's been any progress on the comments that might not be pushed up. I'm happy to review as well when all of that is ready.

Aesica:
Yeah, I've addressed most of them, just haven't fully tested them enough to meet my own satisfaction yet.  That shouldn't affect the sound effect part though, so if you come up with an improved version keep me posted.

Aesica:
Bit of a bump for this, since I've updated it and hopefully addressed all of the points brought up, and I *think* those changes should be reflected here now.  Any further thoughts/feedback/etc?

akirmse:
I think this looks OK.  Does it include the final wav file?

Aesica:
Yeah, just uploaded it a moment ago.

PR #494, opened 2023-05-29
feature: add "show time" client command

adrienlaws:
### Add "show time" client command to see the current M59 hour, day, and year

**What**

- Adds a new client command "show time"
- Adds protocol UC_REQ_TIME,  client command RequestTime(), and server based time response through user.kod

**Atrribution / Inspiration**
- treymd's PR here https://github.com/OpenMeridian/Meridian59/pull/1193

**Why**

- There are time based quests and timers within M59 but no easy way to know the time.  There are only 2 clocks in the entire M59 world as far as I know (BQ water clock + Cor Noth)
- Many things are based on time, for example:
  - Room rents are based on M59 "days"
  - Reroll "suicide" timers say "give it a day"
  - A quest may happen at midnight (ukgoth)
- Some players may not be able to see or may not notice the daytime / nighttime graphics - this complements the visual queues with text

Example
image

### DIFF
--- a/module/merintr/merintr.c
+++ b/module/merintr/merintr.c
@@ -349,7 +349,7 @@ static TypedCommand commands[] = {
 { "safetyon",    CommandSafetyOn, },
 { "safetyoff",   CommandSafetyOff, },
 { "show time",   CommandTime, },
-{ "zeigen zeit", CommandTime, }, // "show time" in German
+{ "zeit anzeigen", CommandTime, }, // "show time" in German
 { NULL,          NULL},    // Must end table this way
 };
 


module/merintr/merintr.c
```diff
@@ -347,6 +348,8 @@ static TypedCommand commands[] = {
 { "tguild",      CommandTellGuild, },
 { "safetyon",    CommandSafetyOn, },
 { "safetyoff",   CommandSafetyOff, },
+{ "show time",   CommandTime, },
+{ "zeit anzeigen", CommandTime, }, // "show time" in German
```
akirmse, line 352: German translations don't go here---notice how there are no others in this list.

module/merintr/merintr.c
```diff
@@ -347,6 +348,8 @@ static TypedCommand commands[] = {
 { "tguild",      CommandTellGuild, },
 { "safetyon",    CommandSafetyOn, },
 { "safetyoff",   CommandSafetyOff, },
+{ "show time",   CommandTime, },
```
akirmse, line 351: Why "show time" and not "time"?  All other commands are one word.

akirmse (commented):
I'm not so sure about this for a few reasons:

- There are already 3 ways of telling time in the game (the 2 clocks, and the sun and the moon, which is currently broken in the hardware renderer).  In-game methods are always preferable to magic commands that no one can discover.  I'd really like to get the sun and moon fixed.

- Adding specific commands for things is generally less desirable than using existing game elements.  Like, maybe saying "time" to an NPC, and they can have a few different interesting responses?  Or some other creative idea or item.

davidklaw:
I like the idea of NPCs being timekeepers. I could see barkeeps and innkeepers giving the time.

I do appreciate the reasoning behind this change. In all my years I've never understood how to read the BQ clock and seem to recall relying on a 3rd-party app to do it.

Issue #463, opened 2023-04-16
Client: In-Game: Objects vanish from peripheral view after manual adjustment


jrotunda:
In-game on the fullscreen client, when manually adjusting the player view screen, objects will not render properly and leave the view when close to any of the peripheral edges.

This is most noticeable if the view window is made small, distracting if it is minimized.

This doesn't occur if the view window is maximized. 

It's expected that no matter what adjustment is made by the user that the view of objects will stay consistent and render within the players full view even when close to the edges of the window.

The issue is more noticeable with a small view.

To reproduce:

- Load in-game to a screen with a sizeable amount of objects (ex: rid_cornoth; rid_tos)

- Put the view screen to maximum 

- Walk along the street and note that objects don't noticeable leave your view, and you can pass by them without issue 

- Minimize the window manually by placing your cursor on the edge of the view screen, put the view window 1/2 the size or smaller (the bug occurs no matter the adjustment)

- Walk along the room and notice that objects vanish from the periphery sooner than they should 

- Without movement: Use mouselook while standing still to move your view screen close to an object (such as a tree) and notice when it vanishes 

(Video to be attached shortly)

davidklaw:
This occurs vertically as well.

https://github.com/Meridian59/Meridian59/assets/467443/98585049-a6ba-4784-94d6-a1e0816f9ab6

PR #495, opened 2023-06-08
Prevent objects vanishing from view after manually resizing the client

sicoderealm:
Minor adjustment to FOV calculations to cater for all client window sizes.

Resolves: #463

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -21,11 +21,11 @@ extern int main_viewport_height;
 // Define field of views with magic numbers for tuning
 inline float FovHorizontal(long width)
 {
-	return width / (float)(MAXX) * (-PI / 3.6f);
+	return width / (float)(main_viewport_width) * (-PI / 3.6f);
 }
 inline float FovVertical(long height)
 {
-	return height / (float)(MAXY) * (PI / 6.0f);
+	return height / (float)(main_viewport_height) * (PI / 6.0f);
 }
 
 #define Z_RANGE					(200000.0f)
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -25,7 +25,7 @@ inline float FovHorizontal(long width)
 }
 inline float FovVertical(long height)
 {
-	return height / (float)(main_viewport_height) * (PI / 6.0f);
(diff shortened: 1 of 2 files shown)


akirmse (commented):
I tried this and the FOV is way too small.  See screenshots.

Before:
before

After:
after

akirmse:
Looking at this some more, I don't think this code is the source of the problem.  The FOV is correct for rendering, but wherever objects are tested for visibility there must be a bug that's still assuming the old fixed width and height values.

clientd3d/d3drender.c
```diff
@@ -19,14 +19,15 @@ extern int main_viewport_width;
...
+inline float FovHorizontal(long width)
 {
-   return (diff / (float)(main_viewport_width) * (-PI / 3.6f));
+	return width / (float)(main_viewport_width) * (-PI / 3.6f);
 }
-inline float FovVertical(long diff)
+inline float FovVertical(long height)
 {
-   return (diff / (float)(main_viewport_height) * (PI / 5.6f));
+	return height / (float)(main_viewport_height) * (PI / 6.0f);
```
akirmse, line 28: Should this 6.0 go back to 5.6?

clientd3d/draw3d.c
```diff
@@ -401,7 +401,7 @@ void UpdateRoom3D(room_type *room, Draw3DParams *params)
    horizon = area.cy/2 + PlayerGetHeightOffset();
    num_visible_objects = 0;
    t1=timeGetTime();
-   DrawBSP(room, params, area.cx, FALSE);
+   DrawBSP(room, params, main_viewport_width, FALSE);
```
akirmse, line 404: This is in the software renderer.  The bug is in the hardware renderer.

davidklaw:
Can anyone explain the change here? Was the code previously only drawing objects within a limited range, but now does so for the entire viewport width? Is viewport height factored in or irrelevant, because I noticed that the vanishing object issues existed at all 4 viewport edges.

davidklaw:
> solves this issues for all but the very smallest of sizes.

I'm testing at 2560x1440 (native) and at 1920x1080 and I only see this issue when the viewport is smaller than I would imagine anyone would ever want.

PR #496, opened 2023-06-08
Scale player overlay equipment correctly with the Software Renderer

sicoderealm:
Logical correction to scaling of player overlay equipment when using small game client windows.

### DIFF
--- a/clientd3d/graphics.c
+++ b/clientd3d/graphics.c
@@ -202,7 +202,7 @@ void GraphicsAreaResize(int xsize, int ysize)
    main_viewport_width = view.cx;
    main_viewport_height = view.cy;
    player_overlay_scaler = (float)(main_viewport_width - CLASSIC_WIDTH) / CLASSIC_WIDTH;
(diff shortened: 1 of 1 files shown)


clientd3d/graphics.c
```diff
@@ -202,7 +202,7 @@ void GraphicsAreaResize(int xsize, int ysize)
    main_viewport_width = view.cx;
    main_viewport_height = view.cy;
    player_overlay_scaler = (float)(main_viewport_width - CLASSIC_WIDTH) / CLASSIC_WIDTH;
-   player_overlay_scaler = (player_overlay_scaler > 0) ? 1.0f + player_overlay_scaler : 1.0f - player_overlay_scaler;
+   player_overlay_scaler = (player_overlay_scaler > 0) ? 1.0f + player_overlay_scaler : 1.0f + player_overlay_scaler;
```
akirmse, line 205: Both branches yield the same value, so it looks like these two lines can be simplified, maybe to
```
player_overlay_scaler = ((float) main_viewport_width) / CLASSIC_WIDTH;
```

PR #498, opened 2023-06-09
Stops Jala casters from teleporting away from their room enchantments

### LINKED ISSUE #497: Indefinite Jala spells when user teleports away from room
An issue was found in which Jala spells would remain in a room indefinitely when the caster left the room via teleport, which is a method used by DM commands _and_ Rescue. The latter of which was determined to the be culprit in these cases. The expectation is that once a caster leaves the room, the Jala trance is broken and the spell ends. What's happening is that the caster is leaving the room (via Rescue) and eventually logging off, leaving behind a room enchantment that can't reference its caster. As a result, the timer craps out, the enchantment remains up, and players entering the room are shown an update dialog. And in some cases the game simply crashes.

To reproduce:

1. Cast a Jala spell wearing a Jala necklace. The necklace is important as it allows the caster to maintain a trance through events that normally would break it, like casting other spells.
2. Rescue away from the room and log off.
3. Enter the room using another character and you should receive a download popup. You should also see the spell remain indefinitely, as the caster is no longer online and their resource cannot be reference.

Mentioned in #380
> During a special case scenario (I'm debugging the exact trigger) - when the active object (user) who created the song is no longer in the room, the enchantment timer turns negative and causes the enchantment to become a permanent enchantment, but in an unnatural way that causes a resource update error - perhaps due to the two music resources conflicting - which leads to a crash when ANY NEW active users enter the room and hits YES in the resource update active dialogue window that pauses their gameplay (B1 Level).
> At its best, the bug allows a Jala spell to stay up indefinitely. At its worst it causes nonstop resource errors, gameplay interruptions and crashing.

davidklaw:
This PR fixes #497 by intercepting users, in trances, who have left a room by teleportation. The two specific cases this has in mind are DM teleports and Rescue, which share teleport functionality. The issue prompting this is that Jala casters, wearing Jala necklaces, could Rescue out of a room and leave behind an indefinite room enchantment. In the same way, a DM could also cause this to happen if they cast a room enchantment under the same conditions and teleported away from the room. As result, users would crash or be ask to update their game.

### DIFF
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -1907,10 +1907,7 @@ messages:
       }
 
       % Break trance if the user is in one
-      if IsClass(what,&User) AND Send(what,@CheckPlayerFlag,#flag=PFLAG_TRANCE)
-      {
-         Send(what,@BreakTrance,#event=EVENT_NEWOWNER);
-      }
+      % Send(what,@BreakTrance,#event=EVENT_NEWOWNER);
 
       % unenchant this one occupant
       for i in plEnchantments
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -1906,6 +1906,12 @@ messages:
          propagate;
       }
 
+      % Break trance if the user is in one
+      if IsClass(what,&User) AND Send(what,@CheckPlayerFlag,#flag=PFLAG_TRANCE)
+      {
+         Send(what,@BreakTrance,#event=EVENT_NEWOWNER);
+      }
+
       % unenchant this one occupant
       for i in plEnchantments
       {


kod/object/active/holder/room.kod
```diff
@@ -1906,6 +1906,12 @@ messages:
          propagate;
       }
 
+      % Break trance if the user is in one
+      if IsClass(what,&User) AND Send(what,@CheckPlayerFlag,#flag=PFLAG_TRANCE)
```
akirmse, line 1910: You don't need to check the class.. the code just about this does it.

davidklaw: I was hoping you'd comment on that. I did so because of the `plActive` loop that follows checking for class and thought whoever wrote that knew something I didn't.

akirmse: That's different---that's checking each *other* active object in the room to see if there are any other users still in the room.

akirmse:
This will fix the bug but I want to double-check that it doesn't introduce a new one.  In Player.NewOwner, there's a similar call: `Send(self,@BreakTrance,#event=EVENT_NEWOWNER);` that does not check the player flag.  Thus, both the player's old and new rooms will receive a BreakTrance message if the user was in a trance and changes rooms.  Is that always safe?

davidklaw:
I believe it is safe. Did you have any particular scenario in mind? I tested this by:

* Casting mirth with a Jala necklace, then teleporting away
* Casting mirth with a Jala necklace, then rescuing away
* Casting enchant weapon (to trance), then teleporting away

I removed the flag check, as it's not necessary. So, now it's just the weirdness of calling BreakTrance in both of these places.

Alternatively, I could add the trance breaking to Rescue and DM teleporting, but it's a bit messier.

akirmse:
Your change looks good, but what's weird is the call to BreakTrance in Player.NewOwner, *after* the room change has happened.  I guess it's safe since it's been that way for many years.  But I wonder if it's needed after this change.  Anyway, accepting this to fix  the bug.

PR #500, opened 2023-06-20
Polish Quest update - give 20 get 10

CardiovascularMike:
The old quest was give 3 gems and get 1 polished. The problem is that 1 polished gem has no utility to most players. I've been offered the quest when I had the 3 gems in my inventory numerous times and never bothered to complete it because what am I going to do with 1 gem? As far as the 20 and 10 figures - well these seemed like a reasonable compromise but are essentially arbitrary.

There are two gameplay reasons to keep the numbers reasonably small:

1. We don't want the required number of uncut gems to be large enough that players will have to drop other reagents or change their inventory around significantly due only to the extra aggravation on what is already a fetch-quest.

2. The upper bound also has to consider the value of polished gems. 20-40 per quest isn't enough to impact the game's economy but its always worth considering these factors.

Some other considerations:

Weight: we want to make sure the reward can fit in our players inventory always. This is virtually guaranteed because uncut gems weigh the equivalent of 5 emeralds and polished gems weigh 3 emeralds so even at 2:1 instead of 3:1 we are very safe.

In terms of economy: Uncuts can be purchased 54 shills and sell for 24 shillings. Polished you can buy for 200 shillings and sell for 80 shillings. Total sale value of 20 uncut seraphym is 480 shillings and you receive 800 shillings of value you can sell to an NPC. These are all tiny numbers in the grand scheme of things but its good to make sure the quest isn't negative value for the players and isn't wildly lucrative either. The profit for the player in terms of NPC sales would be 320 shillings while the savings for a player who is buying polished anyhow (assuming they buy the uncuts to complete the quest) would be 920 shillings.

Testing: I tested this change and was happy to see that if you try to give the NPC 15 gems or some number less than 20 gems he will tell you how many he really wants.  This means the players can naturally discover how many gems to give the NPC through dialogue.

Still to test: What happens if you give him too many gems?  Will try this and update in a comment.


CardiovascularMike:
Ok - the quest does allow you to give more than 20 but you still only get 10 polished gems.  I'm not sure this is a problem other than the potential for some disappointed players.  Is there some dialogue that we could use to hint that a few handfuls of gems is all that we want to offer here?

PR #506, opened 2023-06-23
Correction to visible objects area height and width

sicoderealm:
As `params->width` & `params->height` are now always equal to `main_viewport_width` and `main_viewport_height` the divide by 2 operation is incorrect. It turns out that as `DrawBSP` takes the `width` as a parameter but also uses `horizon` and therefore both values need to be correctly updated and so `area` should use the `main_viewport_width` and `main_viewport_height` values directly.

Previously we resolved issues with horizontal objects disappearing from view by avoiding using the use of `area.cx` when drawing the BSP switching to use `main_viewport_width` instead. This was only half of the fix as we should have updated `area.cx` and `area.cy` directly to always be that of `main_viewport_width` and `main_viewport_height`. This change implements this correction and should resolve outstanding visible object issues as before the refactor.

### DIFF
--- a/clientd3d/draw3d.c
+++ b/clientd3d/draw3d.c
@@ -387,8 +387,8 @@ void UpdateRoom3D(room_type *room, Draw3DParams *params)
 
    // Size of offscreen bitmap.
    area.x = area.y = 0;
-   area.cx = min(params->width / 2, main_viewport_width);
-   area.cy = min(params->height / 2, main_viewport_height);
+   area.cx = main_viewport_width;
+   area.cy = main_viewport_height;
 
    // Force size to be even.
    area.cy = area.cy & ~1;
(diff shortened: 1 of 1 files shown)


davidklaw:
Tested and got visual confirmation that this fixes the most recent sightings of this issue.

I wonder, does `DrawRoom3D` need to be updated similarly? I don't know this part of the code well, but it seems to suggest that when the room is first drawn, the horizon will be incorrectly calculated like this previously was. FWIW, I'm not seeing that in my testing.

```
DrawRoom3D()
/* Size of offscreen bitmap */
area.x = area.y = 0;
area.cx = min(params->width / 2, main_viewport_width);
area.cy = min(params->height / 2, main_viewport_height);
```

```
UpdateRoom3D()
// Size of offscreen bitmap.
area.x = area.y = 0;
area.cx = main_viewport_width;
area.cy = main_viewport_height;
```

akirmse:
It looks like DrawRoom3D is only called in software mode.
https://github.com/Meridian59/Meridian59/blob/4668a09fecddef72322b59d9f9d145805ce5b554/clientd3d/draw.c#L80

davidklaw:
Please hold until I/we have had a chance to test software mode.

davidklaw:
No visible issues in software mode that I can find.

PR #510, opened 2023-06-28
Restore players most recently chosen stat group following a system save or login

sicoderealm:
The game client always reset to the same client side group (spell list) after a system save.

This update persists the most recently selected stat group (inventory, spells, skills or stats) through systems saves and also client logins for the same session without restarting the client.

image
image
image
image

### DIFF
--- a/module/merintr/stats.c
+++ b/module/merintr/stats.c
@@ -375,7 +375,7 @@ void RestoreActiveStatGroup()
 	StatsShowGroup(!inventory_group);
 	ShowInventory(inventory_group);
 
-	// Show the inventory, stats, spell	s or skills group.
+	// Show the inventory, stats, spells or skills group.
 	if (inventory_group)
 	{
 		InvalidateRect(GetHwndInv(), NULL, FALSE);


clientd3d/client.def
```diff
@@ -189,6 +189,8 @@ EXPORTS
 	GetRoomObjectById
 	FindVisibleObjectById
 	GetPlayerInfo
+	SetActiveGroup
+	GetActiveGroup
```
akirmse, line 193: Just "group" is too generic.  Can you rename to "StatGroup" / stat_group everywhere?

module/merintr/mermain.c
```diff
@@ -124,6 +124,38 @@ void InterfaceResizeModule(int xsize, int ysize, AREA *view)
...
+  else
+  {
+    list_type stat_list;
+    if (StatCacheGetEntry(GetGroup(), &stat_list) == True)
+    {
+      DisplayStatGroup(GetGroup(), stat_list);
+    }
+    else
+    {
+      debug(("Setting active group to %d\n", GetGroup()));
```
akirmse, line 153: Probably should remove this debug; will spam the logs without much utility.

clientd3d/game.c
```diff
@@ -15,6 +15,9 @@ player_info player;
 room_type current_room;
 BOOL dataValid = FALSE;
 
+// Default to group 5 (STATS_INVENTORY).
+BYTE active_group = 5;
```
akirmse, line 19: Size isn't really a concern here so I'd just make this an int (change everywhere).

module/merintr/merintr.c
```diff
@@ -1783,3 +1783,17 @@ player_info *GetPlayer(void)
...
 }
+
+extern void SetActiveGroup(BYTE group);
+
+void SetGroup(BYTE group)
+{
+   SetActiveGroup(group);
+}
+
+extern BYTE GetActiveGroup(void);
```
akirmse, line 1794: Needed?

module/merintr/mermain.c
```diff
@@ -124,6 +124,38 @@ void InterfaceResizeModule(int xsize, int ysize, AREA *view)
    InterfaceDrawResize(xsize, ysize, view);
 }
 /****************************************************************************/
+/*
+ * RestoreActiveGroup:  Restore the players activated group.
```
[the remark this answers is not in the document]
akirmse: The terminology is confusing.. "group", "current group", "activated group".  Improve names or documentation.

module/merintr/mermain.c
```diff
@@ -134,13 +166,17 @@ void InterfaceRedrawModule(HDC hdc)
...
   StatsDraw();
+
   if( StatsGetCurrentGroup() == STATS_INVENTORY )
   {
     InvalidateRect( GetHwndInv(), NULL, FALSE );
     ShowInventory(TRUE);
     InventoryRedraw();
   }
+  
+  RestoreActiveGroup();
```
akirmse, line 177: Is the order of this correct with the lines immediately preceding?  If the current group is INVENTORY, we show the inventory, then we immediately restore the group to some other value?

clientd3d/game.c
```diff
@@ -687,12 +684,12 @@ player_info *GetPlayerInfo(void)
    return &player;
 }
 
-void SetActiveGroup(BYTE group)
-{
-    active_group = group;
+void SetActiveStatGroup(int stat_group)
+{    
+    config.active_stat_group = stat_group;
```
akirmse, line 689: Does this actually preserve the current group across restarts of the client?  I'd think you'd need to also call ConfigSave() in order for the new value to make it to disk.

module/merintr/stats.c
```diff
@@ -16,13 +16,13 @@
 #include "client.h"
 #include "merintr.h"
 
-HWND hStats;                        // Window containing stats area
-static list_type stats = NULL;      // List of stats currently displayed
+HWND hStats;                        // Window containing stats area.
+static list_type stats = NULL;      // List of stats currently processed.
```
akirmse, line 20: I don't understand the change to this file.  Was "displayed" incorrect?  What does "processed" mean?

module/merintr/mermain.c
```diff
@@ -144,14 +146,14 @@ void RestoreActiveGroup()
...
+    if (StatCacheGetEntry(active_stat_group, &stat_list) == True)
     {
-      DisplayStatGroup(GetGroup(), stat_list);
+      DisplayStatGroup(active_stat_group, stat_list);
     }
     else
     {
-      debug(("Setting active group to %d\n", GetGroup()));
-      RequestStats(GetGroup());
+      debug(("Setting active stat group to %d\n", active_stat_group));
```
akirmse: We don't want to check these debugs in, right?

module/merintr/statbtn.c
```diff
@@ -420,7 +420,7 @@ void StatButtonCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
 
 	// Record which group is now active (inventory, spells, skills or stats)
 	// Used to restore correct ui state following a system save.
-	SetGroup(group);
+	SetStatGroup(group);
 	debug(("Player picks new active group -> %d\n", group));
```
akirmse: Same with this debug.

module/merintr/mermain.c
```diff
@@ -134,13 +164,9 @@ void InterfaceRedrawModule(HDC hdc)
   StatsDrawBorder();
   StatsMainRedraw();
   StatsDraw();
-  if( StatsGetCurrentGroup() == STATS_INVENTORY )
-  {
-    InvalidateRect( GetHwndInv(), NULL, FALSE );
-    ShowInventory(TRUE);
-    InventoryRedraw();
-  }
+  RestoreActiveStatGroup();
```
akirmse, line 167: I'm still confused by the logic here.  This is in the window redraw function.  We potentially need to fetch stats from the server every time the window is redrawn?  Seems like there could be some cases where the client can generate tons of messages if the client is (e.g.) resized while we're waiting to get the stats back from the server.

Is the idea that RestoreActiveStatGroup() needs to be called after a system save?  Every redraw is overkill for this obviously.  I wonder if there's a function that's called less often, or another mechanism for hooking into restores after system saves.

PR #511, opened 2023-06-30
For each object rendered we increase the z-depth to prevent z-fighting

sicoderealm:
Rendered objects in the hardware renderer suffers from z-fighting, to prevent this we increase the z-depth for each object drawn.

You could easily spot the z-fighting taking place, such as below:

image

With this small update we can see many objects without such visual artifacts:

image

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -6841,6 +6841,9 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 
 	anglePitch = PlayerGetHeightOffset();
 
+	// For each object rendered we increase the z-depth to prevent z-fighting.
+	int z_depth_inc = 0;
+
 	// base objects
 	for (curObject = 0; curObject < nitems; curObject++)
 	{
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -6979,7 +6982,17 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 		pChunk->numPrimitives = pChunk->numVertices - 2;
 		pChunk->xLat0 = xLat0;
 		pChunk->xLat1 = xLat1;
-		pChunk->zBias = ZBIAS_BASE;
```
akirmse, line 6982: A clearer way to write this is

```
pChunk->zBias = ZBIAS_BASE;
// Nodes with a bound height adjust are part of other players' upper bodies.
if (pRNode->boundingHeightAdjust == 0)
{
	// Typical items such as reagents, keys, etc.
	pChunk->zBias = ZBIAS_DEFAULT+(z_depth_inc++);
}
```

Is this always totally safe?  As in, no "typical items" are allowed to have overlays?

PR #517, opened 2023-07-10
Updating FPS counter to be an average and use chrono time

sicoderealm:
FPS is now calculated as an average and includes time spent clamping when the maximum FPS is achieved.
We also switch from using `timeGetTime()` to `std::chrono` to improve the time resolution.

### DIFF
--- a/clientd3d/graphics.c
+++ b/clientd3d/graphics.c
@@ -15,7 +15,6 @@
 #include <vector>
 #include <numeric>
 #include <chrono>
-#include <cwchar>
 
 //	Duplicate of what is in merint\userarea.h.
 #define USERAREA_HEIGHT 64
@@ -301,13 +300,6 @@ int GetMSDrawFrame(void)
  */
 void RedrawForce(void)
 {
-   HDC hdc;
-   int oldMode;
-   char buffer[32];
-   
-   steady_clock_time_point endFrame, startFrame;
-   std::chrono::duration<double> elapsedTime;
-
    if (GameGetState() == GAME_INVALID || /*!need_redraw ||*/ IsIconic(hMain) ||
        view.cx == 0 || view.cy == 0 || current_room.rows == 0 || current_room.cols == 0)
    {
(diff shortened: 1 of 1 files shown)


clientd3d/graphics.c
```diff
@@ -41,6 +42,12 @@ int main_viewport_width;
 int main_viewport_height;
 extern float player_overlay_scaler;
 
+// Variables to store frame times and FPS
+std::vector<double> frameTimes;
+const int windowSize = 60;  // Number of frames to consider in the rolling window
+int average_fps = 0;
```
akirmse, line 48: These variables should all be static (i.e. not visible outside this file).

clientd3d/graphics.c
```diff
@@ -330,13 +337,33 @@ void RedrawForce(void)
 	 Sleep(msSleep);
       }
    }
+
+   // Calcaute the FPS again aftter any adjustments from clamping.
+   fps = 1000 / (timeGetTime() - lastEndFrame);
```
akirmse, line 342: Watch for divide by zero?

clientd3d/graphics.c
```diff
@@ -330,13 +337,33 @@ void RedrawForce(void)
...
+   // Update the last end frame and end the time period.
    lastEndFrame = endFrame;
    timeEndPeriod(1);
 
+   // Add the fps to the rolling window
+   frameTimes.push_back(fps);
+
+   // Remove the oldest fps if the window is full
+   if (frameTimes.size() > windowSize) {
+       frameTimes.erase(frameTimes.begin());
```
akirmse, line 353: Hm, this is going to happen every frame, even if showFPS is false, which is not great.  Use a std::deque instead?  Only do this if showFPS is on?

clientd3d/graphics.c
```diff
@@ -330,13 +337,33 @@ void RedrawForce(void)
...
+   }
+
+   // Calculate the average FPS over the rolling window
+   double sumFPS = 0.0;
+   for (const double frameTime : frameTimes) {
```
akirmse, line 358: Use "auto" as the type.

clientd3d/graphics.c
```diff
@@ -12,6 +12,10 @@
  */
 
 #include "client.h"
+#include <vector>
+#include <numeric>
+#include <chrono>
+#include <cwchar>
```
akirmse, line 18: #include no longer needed?

clientd3d/graphics.c
```diff
@@ -290,10 +302,11 @@ int GetMSDrawFrame(void)
...
 {
    HDC hdc;
-   static DWORD lastEndFrame = 0;
-   DWORD endFrame, startFrame;
-   int totalFrameTime, oldMode;
+   int oldMode;
    char buffer[32];
+   
+   steady_clock_time_point endFrame, startFrame;
+   std::chrono::duration<double> elapsedTime;
```
akirmse, line 309: I think you can declare most of these at the point of first use?  oldMode for sure.

PR #518, opened 2023-07-11
Control z-depth by the position of each item in the 3D world

sicoderealm:
When a relatively large number of items are rendered in the 3d scene some items (for example the urn in Tos by the GY) would render with an incorrect depth. We now control z-depths by the location of items in the 3d world rather than increasing z depths for each item in the scene.

image

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -6854,7 +6854,7 @@ ZPoint calculateBin(const ZPoint& point, int binSize) {
 	return { point.x / binSize, point.y / binSize, point.z / binSize };
 }
 
-// Function to compare points (needed to use Point as a key in std::map)
+// Function to compare points (needed to use ZPoint as a key in std::map)
 bool operator<(const ZPoint& p1, const ZPoint& p2) {
 	return std::tie(p1.x, p1.y, p1.z) < std::tie(p2.x, p2.y, p2.z);
 }
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -6871,7 +6871,7 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 	// This is to mitigate z-fighting by incrementing z-depths for each unique object in the same position.
 	// The key is composed of the x and y coordinates of the object and the value is the current
 	// count of objects found at that location.
-	std::unordered_map<int64, BYTE> depth_adjustment_map;
+	std::unordered_map<int64, int> depth_adjustment_map;
 
 	// base objects
 	for (curObject = 0; curObject < nitems; curObject++)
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -6842,6 +6845,20 @@ d3d_render_packet_new *D3DRenderPacketFindMatch(d3d_render_pool_new *pPool, LPDI
 	return pPacket;
 }
 
+struct ZPoint {
+	int x, y, z;
+};
+
+// Function to calculate the bin for a given point and bin size
+ZPoint calculateBin(const ZPoint& point, int binSize) {
```
akirmse, line 6853: Functions local to a file should be static.

clientd3d/d3drender.c
```diff
@@ -7012,8 +7029,15 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 		// Nodes with a bound height adjust are part of other players' upper bodies.
 		if (pRNode->boundingHeightAdjust == 0)
 		{
+			// For objects in a similar position we increase the z depth
+			// to prevent z fighting at that location.
+			float binSize = 5.0f;
```
akirmse, line 7034: We have FINENESS constants and others that I think you can reuse rather than introducing another one.

clientd3d/d3drender.c
```diff
@@ -7012,8 +7029,15 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
...
+			// For objects in a similar position we increase the z depth
+			// to prevent z fighting at that location.
+			float binSize = 5.0f;
+			const ZPoint position {pRNode->motion.x, pRNode->motion.y, pRNode->motion.z};
+			const ZPoint bin = calculateBin(position, binSize);
+			objection_position_bins[bin].push_back(position);
+
 			// Typical items such as reagents, keys, etc.
-			pChunk->zBias = ZBIAS_DEFAULT + (z_depth_inc++);
+			pChunk->zBias = ZBIAS_DEFAULT + (BYTE)(objection_position_bins[bin].size() & 0xFF);
```
akirmse, line 7040: You want the z offsets to wrap around at 256?  Not sure I understand this.

clientd3d/d3drender.c
```diff
@@ -7012,8 +7016,14 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 		// Nodes with a bound height adjust are part of other players' upper bodies.
 		if (pRNode->boundingHeightAdjust == 0)
 		{
-			// Typical items such as reagents, keys, etc.
-			pChunk->zBias = ZBIAS_DEFAULT + (z_depth_inc++);
+			// Typical items such as reagents, keys, etc. are drawn at the default depth
+			// offset by the number of items already drawn at this location.
+
+			// Combine objects x and y position into a single int64 for the map key.
+			int64_t key = ((int64_t)pRNode->motion.x << 32) | (int32_t)(pRNode->motion.y & 0xFFFFFFFF);
```
akirmse, line 7023: Since there's no longer any binning, this will only affect objects precisely on top of each other.  That will take care of piles of objects dropped simultaneously, which might be enough to solve the problem, just checking that it's what you want.

clientd3d/d3drender.c
```diff
@@ -6866,8 +6867,11 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 
 	anglePitch = PlayerGetHeightOffset();
 
-	// For each object rendered we increase the z-depth to prevent z-fighting.
-	int z_depth_inc = 0;
+	// We track all objects that are in similar positions in the 3D world.
+	// This is to mitigate z-fighting by incrementing z-depths for each unique object in the same position.
+	// The key is composed of the x and y coordinates of the object and the value is the current
+	// count of objects found at that location.
+	std::unordered_map<int64, BYTE> depth_adjustment_map;
```
akirmse, line 6874: Let's make this an int instead of a BYTE.  I guess cast to BYTE later (I didn't realize z bias is only a BYTE... not sure why that is, maybe old video card limitations?).

Also pick int64 or int64_t consistently in this file.  I'd lean toward int64 for readability.

PR #519, opened 2023-07-12
Realtime updates for the latency information tooltip and latency dialog

sicoderealm:
Realtime updates for the latency information tooltip and latency dialog in real time once a second + includes a correction to an invalid server status url. Adding a new configuration option to expose the classic client spinning latency cube and also give latency its own dedicated dialog for future proofing.

cubenew

image

### DIFF
--- a/clientd3d/lagbox.c
+++ b/clientd3d/lagbox.c
@@ -305,7 +305,7 @@ void Lagbox_Command(HWND hWnd, int id, HWND hwndCtrl, UINT uNotify)
 
 	latencyDialogActive = true;
 
-	SetDescParams(hWnd, !DESC_DROP);
+	SetDescParams(hWnd, DESC_NONE);
 	DisplayDescription(pobjLagbox,
 		0/*flags*/,
 		pszLagboxDescription, 


akirmse:
Eww, I'm not a fan of this.  The spinning was super distracting (and didn't add anything), and putting three digits of precision on latency always on the screen seems like way too much information (and is redundant with the cube).  Meridian is kind of old-school in this sense.  We don't even put damage numbers on the screen like modern games and I think we should keep it that way.

Dirhai:
I agree with akirmse its too much.

Dirhai:
I love this new version; it almost feels like nothing has changed. The spinning cube looks really great, and the option to turn it on and off works perfectly (I personally prefer to play with it off). The real-time update latency is a good quality of life improvement; it's annoying to need to close and reopen the latency window every time.

Great work @sicoderealm 


> **Testing Approach:**
> 
> **Default Testing Steps:**
> 1. Fetch the PR
> 2. Run `nmake clean`
> 3. Execute `nmake`
> 
> **Testing Method for this PR:**
> 1. I turned the option on and off several times.
> 2. I constantly checked the terminal for errors.
> 3. I walked around and fought monsters with the new option turned on and off.

include/proto.h
```diff
@@ -364,6 +364,7 @@ enum {
 #define OF_APPLYABLE     0x00001000    // Set if object can be applied to another object
 #define OF_SAFETY        0x00002000    // Set if player has safety on (self only)
 #define OF_PLAYER_MASK   0x0001C000    // Mask to get player flag bits
+#define OF_LATENCY       0x0002C000    // Set if object contains latency information
```
akirmse, line 367: This is not how flags work.  Each flag must be a single bit that does not overlap with any others.  This value overlaps with the player flags, which could cause unforeseen problems if an object sets a flag and the client misinterprets it as a result.  The multi-bit values like OF_PLAYER_MASK are masks used to extract only those flags in a particular group (in this case, player-specific flags).

There are only 32 bits available (possibly only 28, would have to do a detailed server-side analysis), and I don't think we want to "spend" one on the single-purpose latency cube.

It also looks weird that this value is defined in the protocol, but never used on the server.  That's a sign that this isn't the right place for it.

I'll hold off reviewing the rest of this PR as I suspect this will require substantial changes.

clientd3d/dialog.c
```diff
@@ -661,8 +662,18 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
 	TooltipReset();
 	SetDescParams(NULL, DESC_NONE);
 }
-
-/*****************************************************************************/
+/************************************************************************/
+/*
+* SetDialogExtraString:  Update the extra string for the dialog.
```
akirmse, line 667: What is an "extra string"?  Can this be described more clearly?

clientd3d/dialog.c
```diff
@@ -14,6 +14,8 @@
 #define PAGE_BREAK_CHAR '�'	  /* For multi-page descriptions */
 #define MAX_PAGE_DESCRIPTION_TEXT MAXMESSAGE
 
+static DescDialogStruct info;	  /* Info for current Description dialog */
```
akirmse, line 17: Did this need to be moved from the stack to a global variable?  This uses more memory and makes the lifetime of the variable longer than is necessary.

clientd3d/lagbox.c
```diff
@@ -284,30 +297,23 @@ LRESULT Lagbox_OnTooltipCallback(HWND hWnd, TOOLTIPTEXT* pTTT)
...
-	lagbox.animate = &lagbox.normal_animate;
-	lagbox.overlays = &lagbox.normal_overlays;
-	lagbox.flags |= OF_PLAYER | PF_OUTLAW;
 
-	Lagbox_OnTooltipCallback(hWnd, &ttt);
+	latencyDialogActive = true;
 
-	SetDescParams(hWnd, DESC_NOAGE);
-	DisplayDescription(&lagbox,
+	SetDescParams(hWnd, !DESC_DROP);
```
akirmse, line 308: This looks suspicious.  SetDescParams takes an int, but you're passing in false (!DESC_DROP).

PR #525, opened 2023-07-14
Prevent graphical lighting and other artefacts when vision is restored from blind

sicoderealm:
When restoring vision from the blind effect the game world view would often appear to display issues with lighting and other artefacts.

This issue could also cause more severe impact when testing under more extreme conditions - such as creating popup error messages in the client - such messages could appear when running with multiple clients.

Here we make a small switch to obtain simultaneous read rather than write access to a rooms roo file which prevents these issue.

Kudos to m59's Mayhem for helping to identify and resolve this issue :pray: 

image

image

### DIFF
--- a/clientd3d/bspload.c
+++ b/clientd3d/bspload.c
@@ -60,7 +60,7 @@ Bool BSPRooFileLoad(char *fname, room_type *room)
    int node_pos, wall_pos, sidedef_pos, sector_pos, offset_adjust;
    file_node f;
 
-   if (!MappedFileOpenCopy(fname, &f))
+   if (!CliMappedFileOpenRead(fname, &f))
       return False;
 
    // Check magic number and version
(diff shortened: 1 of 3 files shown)


mayhem:
I don't think you meant me...

clientd3d/bspload.c
```diff
@@ -60,7 +60,7 @@ Bool BSPRooFileLoad(char *fname, room_type *room)
    int node_pos, wall_pos, sidedef_pos, sector_pos, offset_adjust;
    file_node f;
 
-   if (!MappedFileOpenCopy(fname, &f))
+   if (!CliMappedFileOpenRead(fname, &f))
```
akirmse, line 63: Weird, I wonder why it was originally done this way.

It looks like this is the only use of MappedFileOpenCopy, so you can delete that function now.

Issue #528, opened 2023-08-02
F10 Hotkey Not Working With Inventory Selected


elladour:
For some reason, you cannot use the F10 hotkey while you have the inventory window selected. You can use any other hotkey, but not F10.

PR #529, opened 2023-08-02
Allow F10 to propagate from inventory and support held hot keys

sicoderealm:
While the inventory was activate the F10 key was being consumed to activate the main game menu - we now allow this key to propagate and also support throttled held hot keys from the inventory as per the main viewport.

Fixes: #528


Dirhai:
I tested this pull request and it worked perfectly, the bug with the F10 key has been resolved.

module/merintr/inventry.c
```diff
@@ -796,7 +796,7 @@ Bool InventoryClick(HWND hwnd, BOOL fDoubleClick, int x, int y, UINT keyFlags)
 Bool InventoryKey(HWND hwnd, UINT key, Bool fDown, int cRepeat, UINT flags)
 {
    ID id;
-   Bool held_down = (flags & 0x4000) ? True : False;  /* Is key being held down? */
+   Bool held_down = (flags & 0x4000) ? true : false;  /* Is key being held down? */
```
akirmse, line 799: This change is incorrect.. the allowable values for Bool are True and False.  You can change it to bool if you want.

module/merintr/inventry.c
```diff
@@ -835,25 +831,24 @@ Bool InventoryKey(HWND hwnd, UINT key, Bool fDown, int cRepeat, UINT flags)
       break;
 
    default:
-      inform = False;   // Modules will be informed in PerformAction call below
+      inform = false;   // Modules will be informed in PerformAction call below
```
akirmse, line 834: Incorrect change (next several instances too).

module/merintr/inventry.c
```diff
@@ -473,6 +464,15 @@ LRESULT CALLBACK InventoryProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lP
 	  StatsDrawBorder();
       InventoryReleaseCapture();
       break;
+   case WM_SYSKEYDOWN:
+      // Prevent activation of the window's menu bar and propagate to parent
+      if (wParam == VK_F10)
+      {
+          if (HANDLE_WM_KEYDOWN_BLAK(hwnd, wParam, lParam, InventoryKey) == True)
+              return 0;
+          break;
```
akirmse, line 473: I don't think this break is needed.

clientd3d/intrface.c
```diff
@@ -27,6 +27,8 @@ static Bool interface_created = False;  /* True iff interface has been created *
 extern font_3d				gFont;
 extern d3d_driver_profile	gD3DDriverProfile;
 
+static DWORD last_text_command;
```
akirmse, line 30: This name makes it sound like it's the last text command that was entered.  last_text_command_time, and add a comment?

clientd3d/intrface.c
```diff
@@ -522,10 +474,17 @@ void PerformAction(int action, void *action_data)
       break;
 
    case A_TEXTCOMMAND:
-      TextInputSetText((char *) action_data, False);
-      ParseGotText((char *) action_data);
-      break;
-
+   {
+       DWORD now = timeGetTime();
```
akirmse, line 478: I'm not convinced this part is a good idea.

- It overrides the keyboard repeat rate from Windows system settings.
- Does this allow super fast macroing?

module/merintr/inventry.c
```diff
@@ -809,10 +806,6 @@ Bool InventoryKey(HWND hwnd, UINT key, Bool fDown, int cRepeat, UINT flags)
    if (action == A_NOACTION)
       return False;
 
-   // Skip held-down keys, except for moving cursor around
```
akirmse, line 812: Presumably we were doing this for a reason?  What has changed?

PR #530, opened 2023-08-07
Added items and updated loot tables

Aesica:
Added items to add variety to monster loot drops:
- Stinky-cap mushroom (poison/diseased food, sellable as reagent)
- corpse mushroom (highly poisonous food, sellable as reagent)
- golden mushroom (sellable as reagent)
- bone dust (sellable as reagent)
- blackened bone dust (sellable as reagent)
- avar feather (sellable as reagent)
- troll blood (sellable as reagent)
- mystery meat (food)
- milk (food)
- ancient sword/sword of pure ice* (weapon)

* this weapon has variable properties, intended to produce a variety of weapon variants as for-flavor monster drops without clogging up the class pool with dozens of new weapons

Changed the following item from a non-stackable misc item to a numbitem/reagent type to use in loot tables, because it currently has zero use in game, and just because it just makes sense:
- egg shell

Updated loot tables for the following monsters to reduce item clutter in rooms, which would lead to eventual blinking/disappearing objects:  (in short, removed or greatly reduced the rates of normal weapons, armors, potions, etc and replaced them with other things, including the new flavor items)
- battered skeleton
- tusked skeleton
- daemon skeleton
- avar warrior
- troll
- narthyl worm

Skeletons (all types) now drop ancient swords (CustomWeapon class, 35% chance, down from 100% when they dropped normal long swords) of varying quality.  They're mostly low/medium quality, but have a very rare (1:1000) chance to drop a high quality variant with a shiny white blade.  This variant is Meridian's first true "rare drop" item.

Updated loot tables for the following monsters, mostly just for flavor:
- baby spider
- spider
- giant rat/snow rat (shared table)

Updated the following loot tables because they were lousy for being bosses:
- queen spider

Updated the following loot tables for being over the top compared to the difficulty level of the monster:
- frogman

Added loot tables to the following monsters, which had none until now:
- ice creature
- cow
- guard cow

Ice creatures have a 15% chance to drop a blade of pure ice (CustomWeapon class) that is medium quality, deals slash + cold damage.  It "melts" (loses durability) over time until it breaks/deletes itself.  This drop chance is separate from the normal loot tables, similar to how skeleton swords drop.

Finally, set the unused cowX image resource to cows and guard cows to give them a corpse, but I'll admit it's pretty low-quality in its current state.  It's just the normal cow image, facing forward.  May be better off as blank.bgf, but a corpse of some type is needed so the trestype class doesn't freak out.

### DIFF
--- a/kod/include/blakston.khd
+++ b/kod/include/blakston.khd
@@ -1577,6 +1577,11 @@
    TID_DARK_ANGEL    = 51
    TID_MUMMY         = 52
    TID_THRASHER      = 53
+   TID_ICE_PERSON    = 54
+   TID_FROGMAN       = 55
+   TID_COW           = 56
+   TID_BABY_SPIDER   = 57
+   TID_SPIDER        = 58
 
    %%% English articles
 
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -4175,7 +4175,6 @@ messages:
          OR IsClass(what,&Book)
          OR IsClass(what,&Arsenic)
          OR IsClass(what,&SpiderEgg)
-         OR IsClass(what,&SpiderEggShell)
          OR IsClass(what,&Key) 
       {
          return TRUE;
--- a/kod/object/active/holder/nomoveon/battler/monster/cow.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/cow.kod
@@ -40,6 +40,10 @@ classvars:
    vrName = cow_name_rsc
    vrIcon = cow_icon_rsc
    vrDesc = cow_desc_rsc
+   vrDead_icon = cow_dead_icon_rsc
+   vrDead_name = cow_dead_name_rsc
+
+   viTreasure_type = TID_COW
 
    viMove_delay_min = 2000
    viMove_delay_max = 10000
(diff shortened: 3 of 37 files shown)


Aesica:
So I'd probably better start from the beginning regarding this PR.  I originally started it as a way to trim down the loot tables of certain monsters that are known for clogging up rooms very quickly with unstackable garbage.  By replacing all the scimitars, scale armors, knight shields, potions, etc dropped by these monsters with stackable items, it takes a lot longer for a room to fill up to the point players need to leave and wait for it to reset.  All of the newly-made items (besides CustomWeapon) are stackable, and I made them because you can only add red/blue/edible/inky mushrooms and gems to so many monsters before all the loot tables start to feel the same.

As for CustomWeapon (ancient sword variants, sword of pure ice variants) in particular, as it might seem like an odd approach at first, I'll discuss it in particular.  I made it how it is because skeletons seem like they should be wielding a wide variety of different swords, long or short, good or bad quality.  Currently they all drop the same long sword forged by the same guy, which has always seemed a bit weird.  So ancient swords come in a number of different qualities, and can be either long swords or short swords.  They are thrusting type in terms of damage and overall stats, but deal nonmagical slashing damage as a way to make them at least seem somewhat different compared to the vanilla long sword and short sword.  This applies to the sword of pure ice dropped by ice creatures as well.

Aesica:
After careful consideration as well as feedback, I think this PR is probably too large and intimidating, so I'm going to close it and break it up into several parts:

- New quasi-reagents (created to be sellable drops to add variety) to loot tables
- New foods and food/reagent hybrids (same as above)
- New weapon (same as above, but its by itself for being fairly complex in comparison)
- Revised loot tables (cut down on excessive nonstackable loot spam--the overall goal of this endeavor, depends on all of the above whoever)

PR #531, opened 2023-08-07
Add blink success message to the user

CardiovascularMike:
This PR adds two text messages sent to the user after succeeding in casting blink.  Currently no message is sent and the user's text window simply states the user is concentrating on the spell and the only indication of success is the animation.  This can lead to confusion in cases where you are casting blink often and happen to look away and miss the animation and sound effect.

There are two rooms where you blink and are teleported to a different room as a secret.  These situations now have their own separate success message.


CardiovascularMike:
One question for the veterans: Should I switch the order of the post and send functions for the special blink rooms?  We are posting the teleport and sending the text message immediately.  I think its OK but I could switch to post then post.  I am not sure it matters at all.

PR #535, opened 2023-08-09
Reflections and Evil Twins don't change karma upon being killed, except at the Riija Temple

Gar103:
By player request, Evil Twins and Reflections no longer ruin your karma except at the Temple of Riija.

Notably, this preserves them actually having karma values for the purposes of spells and abilities. It just doesn't alter a player's karma upon killing one unless they are at the Temple of Riija. The purpose of the Temple exception is 1: because illusions are more real there, and 2: to preserve the player behavior of 'fixing your karma' by purposely killing the reflections of an ally or mule. You can still do that, you'll just have to do it at the Riija Temple.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -4802,6 +4802,12 @@ messages:
       }
 
       if IsClass(what,&Monster)
+         AND NOT (IsClass(what,&Reflection)
+            AND NOT (Send(self,@GetOwner) <> $
+               AND Send(Send(self,@GetOwner),@GetRoomNum) = RID_TEMPLE_RIIJA))
+         AND NOT (IsClass(what,&EvilTwin)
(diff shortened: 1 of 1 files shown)


CardiovascularMike:
I would really prefer to make this PR more simple.  I think we should just set getKarma() to return zero or $ in reflectn.kod and eviltwin.kod

akirmse:
This is a good illustration of why IsClass() is almost always the wrong thing to use.  Now Player, which generally shouldn't know anything about individual spells, has a direct dependency on two individual spells, as well as the ID of a particular room.

Also, the logic is highly suspect.  "what" is a Monster, but then it's checked against Reflection?  That doesn't make much sense.  Does this even work?

Also, this code calls GetOwner 4 times in rapid succession.  That should set off alarm bells.

Closing.

Gar103:
Of course it works. I don't submit code I didn't build or test.

PR #536, opened 2023-08-11
Additional commands and copied key control for rented room management

Aesica:
Currently, room rental is a bit opaque in terms of commands the innkeeper accepts as well as time remaining.  Additionally, control over key copies is pretty much nonexistent, and once stolen, your room is compromised.  This PR enhances these aspects of room management:

- Room Help:  Say this to any innkeeper to get a list of commands related to renting rooms, including prices, maximum rentable days, etc.

- Change Lock:  Say this to an innkeeper you have a room with and they'll destroy any copies of your key that are currently in circulation.

- Rent:  Say this to an innkeeper you have a room with and they'll tell you how much time is left on your room, including a conversion to real life days.

Increased security for copied keys:  Key copies, which are not intended to be dropped on death, shattered, swapped, vaulted, etc can still be dropped accidentally.  This PR changes this behavior so that only the room owner can drop or trade copied keys to other players freely.  Any player who receives a copy of someone else's room key can only offer it back to the room's owner, or give it to an innkeeper to destroy it.  They'll be informed of this if they try to discard it improperly.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/mrinnk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/mrinnk.kod
@@ -33,6 +33,7 @@ resources:
    MarionInnkeeper_cost4 = " shillings to make.  I know you'll like your room."
 
    MarionInnkeeper_days_left = "You like it here, I see.  There are %q days until you should pay me again."
+   MarionInnkeeper_days_left_check = "The room is yours for %q Meridian days.  ~I(%q days in real time.)"
    MarionInnkeeper_too_long = "Oh, I can't be responsible for holding so much money!  Pay me again later when you don't have so many days left on your tab."
 
    MarionInnkeeper_destroy_key_copy = "~kMorrigan tells you, \"Oh, yes. I can get rid of this key for you.  No problem.\""
@@ -41,6 +42,8 @@ resources:
    MrIk_rent_command = "rent"
    MrIk_cost_command = "cost"
    MrIk_copy_key_command = "copy key"
+   MrIk_change_lock_command = "change lock"
+   MrIk_help_command = "help"
 
 classvars:
 
@@ -62,7 +65,7 @@ messages:
    % special handling to make him laugh at Tova's chatter
    SomeoneSaid(what=$,string=$)
    {
-      local oRoomMaint, rSayRsc;
+      local oRoomMaint, rSayRsc, iDaysLeft;
 
       % if it's Tova talking
       if IsClass(what, &MarionBartender)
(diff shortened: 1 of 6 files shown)


kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/mrinnk.kod
```diff
@@ -107,6 +110,39 @@ messages:
             return;
          }
 
+         if (StringContain(string, MrIk_room_command) AND StringContain(string, MrIk_help_command))
```
akirmse, line 113: I haven't read through this whole change yet, but definitely repeating the exact same code in 3 different places (the 3 innkeepers) is a big no-no.

Aesica: I agree and imo the entire room system is pretty gross under the hood, but that's how it was done for the original commands that are already in place.   I could set up a function inside the room maintenance object that each innkeeper would call to check for and handle these new player-issued commands in one place instead, but the originals would still be there in their current messy form.

Aesica: After thinking about this a bit more, the following might be the best approach, which will mean I close this PR and break it up into several smaller chunks:

- Move the existing duplicated code in SomeoneSaid from the 3 innkeepers into monster.kod for a proper subclass/superclass relationship.
- Add a new constant for viAttributes (MOB_LANDLORD) which must be set to enable room rental behavior (similar to how MOB_SELLER, MOB_BANKER, etc works).
- Apply MOB_LANDLORD to existing npcs that rent rooms (morrigan, pritchett, moxal)
- Add the copy key drop protection as PR 1
- Add the new commands as PR 2, maybe even just 1 new command per PR

Thoughts on this?

akirmse:
Somewhat better, but a few questions:

- How many attribute bits do we have remaining if we create MOB_LANDLORD?  Just want to be sure we don't run out.

- If this new attribute is added, what will happen to the existing innkeepers on 101/102?  Will they get the new attribute when the Blakod is updated, or not?  I think the answer is yes but my memory is hazy on this.

- monster.kod is for code that applies to all monsters, not to dump all NPC speech handling.  It's already 6000 lines long, which is absurd.  Ideally there would be an innkeeper class these 3 NPCs could inherit from, but I don't think that's practical given the existing very strange class hierarchy based on towns.  So maybe we add a new class that encapsulates innkeeper behavior (e.g. "InnkeeperBehavior"), and have the 3 innkeepers each contain an instance of this object and pass the appropriate calls to it?

Aesica:
- Currently the highest is 0x200000, so MOB_LANDLORD 0x400000 would make it.  I'm not sure what type of integer these use, however.
- It *should* be added to them just by editing their kod files, since it's a classvar.
- Oh yeah I would've never wanted to lump it into monster.kod had it not been for the current hierarchy structure.  The behavior object approach could work.  The RentableRoomMaintenance class already exists and could have a function added to handle all the dialog branches.  The only tricky part I can think of with this route is getting the right local-flavored replies, based on which innkeeper is talking.  (Morrigan's casual speech vs Pritchett's "I thanketh thee" dialect).  I hate the idea of passing several dozen objects, values, resources, etc through function calls, but everything else I can think of is kind of icky.  (and involve adding code to monster.kod).  While less fun, the local dialects could just be dropped, but I dunno if that's ideal either.

akirmse:
Sounds there are plenty of flags left.

I think you could preserve the per-NPC dialog by passing in the appropriate resources to the behavior object.  Hopefully there aren't too many differences, or you can gather similar ones together into lists, etc.

Aesica:
The flag won't be needed if it's all processed through the room maintenance/behavior object, since that route is just a single Send(...) line in each room-renting innkeeper's SomeoneSaid.  There's currently about 8 resources worth of responses plus the 3 npc-specific room cost values, however 4 of those (used by "room cost") could probably be cut since it's redundant and a lot less informative than the "room help" that was part of this PR.

With that, I think I've got a good direction for this, so unless you had any further thoughts, this can be closed.

PR #538, opened 2023-08-11
Fix for the bugged/inaccurate "has been repaired before" text on low/high/nerudite quality weapons

Aesica:
Newly-created maces and short swords always show "...but has been repaired before" in their descriptions due to the low quality initial hits modifier being negative, and the repaired-before text displaying if piHits_init is below viHits_init_min, which is possible due to the negative durability value applied to low quality weapons.

While less obvious, high and nerudite quality weapons were shy about showing "has been repaired before" right away because they had an extra buffer of bonus hits before they were considered below the minimum possible durability.

This addresses both issues by removing the quality-based adjustment to piHits and piHits_init done in weapon.kod upon object creation and instead sets the intended values directly in each low, high, and nerudite quality weapon's kod file.  This gives each weapon access to the correct viHits_init_min value for determining whether or not it's been repaired before.

### DIFF
--- a/kod/object/item/passitem/weapon.kod
+++ b/kod/object/item/passitem/weapon.kod
@@ -53,11 +53,6 @@ constants:
    WEAPON_RANGE_MID  = 2
    WEAPON_RANGE_HIGH = 3
 
-   % Quality mods to number of hits
-   % Deprecated
-   WEAPON_LOW_QUALITY_HITS  = -50
-   WEAPON_HIGH_QUALITY_HITS = 100
-   WEAPON_NERUDITE_HITS = 200
 
    % Quality mods to hit modifier
    WEAPON_LOW_QUALITY_HITMOD  = 0
--- a/kod/object/item/passitem/weapon.kod
+++ b/kod/object/item/passitem/weapon.kod
@@ -54,6 +54,7 @@ constants:
    WEAPON_RANGE_HIGH = 3
 
    % Quality mods to number of hits
+   % Deprecated
    WEAPON_LOW_QUALITY_HITS  = -50
    WEAPON_HIGH_QUALITY_HITS = 100
    WEAPON_NERUDITE_HITS = 200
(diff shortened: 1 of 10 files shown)


kod/object/item/passitem/weapon.kod
```diff
@@ -54,6 +54,7 @@ constants:
    WEAPON_RANGE_HIGH = 3
 
    % Quality mods to number of hits
+   % Deprecated
```
akirmse, line 57: I don't see that these are referenced any longer.  Remove?

akirmse (commented):
I'm kind of unclear of the scope of this change.  Does viWeaponQuality have any meaning now, and if not, should it be removed, along with its associated constants?

Separately, I looked up uses of WEAPON_QUALITY_LOW and found one class whose hits you didn't adjust (Black Dagger).  How does a reader know if this was intentional or a mistake?  Did you get all of the classes for all of the other qualities?

https://github.com/search?q=repo%3AMeridian59%2FMeridian59%20WEAPON_QUALITY_LOW&type=code

This seems rather invasive for fixing a rarely-seen erroneous description message for maces, and there are a lot of magic number changes that make me wary of the risk.

kod/object/item/passitem/weapon.kod
Aesica: Yeah they shouldn't be, but I left them as a reference for future weapon creation, to give future coders a general idea.  They're probably better off being commented out, so I'll do that next update.

Aesica:
viWeaponQuality is still used to determine a host of other things, such as hit chance, damage, spellpower modifier, disarm bonus, etc, and while I think it's a bit yucky to have all those values lumped into weapon.kod instead of each individual weapon, they're overall harmless except viHits_init_min and viHits_init_max, which cause the aforementioned anomalies in affected weapons' descriptions.

For black dagger, I didn't change it because its hits don't matter and aren't displayed like normal weapons.  It doesn't take normal durability damage and only "breaks" when certain criteria in the assassin game are met.

I know it looks a bit far-reaching, but at some point, somebody changed weapon.kod to just automatically calculate most weapon values (including hits and hits_init) based on a combination of viWeaponQuality and viWeaponType.  However, just as nearly every item declares its own values for weight, bulk, average value, and such, each weapon should probably declare its own durability thresholds.

As for the updating, I got all the high, low, and nerudite quality weapon classes, but there was no need to touch the average quality weapons because they can continue to inherit from weapon.kod like they did before or declare their own and be just fine.  Their hits/hits_init had no modifiers applied so this problem never affected them.

Which magic numbers were you curious about?  If you're wondering about the values I used for viHits_init_min and viHits_init_max on each weapon, those were calculated based on the inherited value from weapon.kod (250 for min, 300 for max) adjusted by the values that were applied for their quality modifier.  So a mace, being low quality (-50), would be 200 and 250.  A mystic sword, being high quality (+100), would be 350 and 400, and nerudite swords (+200) would be 450 and 500.  Special cases, such as the Riija sword (high quality), declared their own hits, but also needed those adjusted since weapon.kod wouldn't be doing it anymore.

Finally I should note that it's not an uncommon occurrence.  Every single newly-created mace or short sword (bought from npc, made by dm, dropped, etc) will show "has been repaired before" unless it lucks out and gets its maximum possible durability (250).

kod/object/item/passitem/weapon.kod
Aesica: On second thought, you're 100% right.  They should just be removed entirely.

PR #539, opened 2023-08-12
Remove 'Unresolved Externals' spam

Gar103:
There's a message that spams rather heavily during nmake that makes it difficult to see if anything went wrong. I'd like to remove that message - I don't believe it's serving a purpose. This commit simply comments out the 'unresolved externals' warning.

### DIFF
--- a/blakcomp/kodbase.c
+++ b/blakcomp/kodbase.c
@@ -547,8 +547,8 @@ int save_kodbase()
    external_list = table_get_all(st.missingvars);
    numexternals = save_externals(kodbase, external_list);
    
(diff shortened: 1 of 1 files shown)


akirmse:
This does serve a purpose.  If the number of unresolved externals is non-zero when building the last file, it means there is a build error (often a typo in a class or method name referencing another file), and the game will not load correctly.  Ideally we'd be able to enforce this on the very last built file, and then this warning could become an error instead.

Also, we should never simply comment out code.  If the code doesn't belong, it should be deleted.

Closing.

PR #541, opened 2023-08-12
Updates and polish to disciple robe coloring

Aesica:
This change allows disciple robes to be created in the color matching their school rather than default blue.  It also switches them back to their school color when unequipped.

### DIFF
--- a/kod/object/item/passitem/defmod/armor/robe/robedisc.kod
+++ b/kod/object/item/passitem/defmod/armor/robe/robedisc.kod
@@ -92,9 +92,46 @@ messages:
       }
       
       send(self,@SetSchool,#school=school);
-      send(self,@SetXLATTranslation,#translation=viXLAT_no_knowledge);
       
       propagate;
+   }
+
+   Constructed()
+   {
+      piItem_flags = Send(SYS, @EncodeTwoColorXLAT,
+         #color1=Send(self, @GetMasterXLATBySchool), #color2=XLAT_TO_SKIN2);
+
+      propagate;
+   }
+
+   GetMasterXLATBySchool()
+   {
+      if piSchool = SS_JALA
+      {
+         return viXLAT_jala_master;
+      }
+      if piSchool = SS_RIIJA
+      {
+         return viXLAT_riija_master;
+      }
+      if piSchool = SS_SHALILLE
+      {
+         return viXLAT_shalille_master;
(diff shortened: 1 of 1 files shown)


kod/object/item/passitem/defmod/armor/robe/robedisc.kod
```diff
@@ -279,6 +288,13 @@ messages:
          Send(poOwner,@RemoveOverlay,#what=self);
       }
 
+      piItem_flags = Send(SYS, @EncodeTwoColorXLAT,
+         #color1=Send(self, @GetMasterXLATBySchool), #color2=XLAT_TO_SKIN2);
+      if IsClass(poOwner, &User) OR IsClass(poOwner, &Room)
```
akirmse, line 293: IsClass() is gross, and in this case I think unnecessary?  Just always call SomethingChanged.

kod/object/item/passitem/defmod/armor/robe/robedisc.kod
```diff
@@ -279,6 +288,13 @@ messages:
          Send(poOwner,@RemoveOverlay,#what=self);
       }
 
+      piItem_flags = Send(SYS, @EncodeTwoColorXLAT,
```
akirmse, line 291: This always sets all item flags, even those unrelated to colors.  Just double-checking that's what you want.

kod/object/item/passitem/defmod/armor/robe/robedisc.kod
Aesica: I was mostly just worried about calling functions that don't exist on certain objects, and for some reason I didn't think SomethingChanged existed on all objects.  After checking, it does so the IsClass checks are pointless.  Removed.

kod/object/item/passitem/defmod/armor/robe/robedisc.kod
Aesica: No, I don't want that and didn't realize there were other flags used by that property besides color.  In light of that, it looks like SetPaletteTranslation should always be used instead of setting the property directly going forward.  All fixed.

Gar103: piItem_flags is only for color as far as I know. Any other properties are typically itematt passive objects

PR #534, opened 2023-08-09
Fellowship groups

Gar103:
Created by player request, as seen in the Feedback & Ideas Discord channel.

A Fellowship is an incredibly simplified association limited to 10 members max. Members are kicked when they log off, they can join and leave at any time, and the Fellowship ceases to exist when no members remain. The Fellowship does not alter any PvP state except preventing members from attacking each other, so that whites / oranges / reds / soldiers can safely play together if desired.

I have not yet added any sort of UI for a non-admin to actually create and manage a Fellowship. There are a great many ways that could actually be done, which is a whole other discussion. This pull request is intended to gauge interest, no sense in making some complex interface unless this is actually a desired addition. In the event we want it, I would suggest it's just a command we give all players, 'create fellowship,' and perhaps create a simple screen like the guild UI that is accessed by typing 'fellowship,' complete with invites, member list, and some options to check.

System.kod change:
Adding a global list of fellowships that can be added to or removed from.

Player.kod change:
Fellowships prevent members from attacking each other.

User.kod changes:
Fellowship members appear as 'ally' green dots.
Users are removed from a fellowship upon logoff.

Corpse pointer change:
Fellowships allow members to loot from each other's kills without waiting for the loot timer, if that option is set for the fellowship.

PR #543, opened 2023-08-16
feature: hear other players' tougher sounds

adrienlaws:
## What

This change makes the "tougher" sound hearable to others in the same room as the player who gets "tougher."

## Why

The improve sound can be heard by other players in the same area.  It can be motivating in a group to hear other party members improving, especially in Thrasher's.  During a recent Thrasher's run people were asking "did you get any toughers?" etc.  This change does not include a text message, so it is pseudo-anonymous.

Toughers are more rare than improves so it will not result in excessive spam.  It may also help party members get a sense for how long is "normal" for a tougher in their group.  There was one player on Discord a few weeks ago who seemed to not have a good sense for how often he should get tougher.  If they were building near others, they would hear the toughers periodically and be able to estimate their own progress better.


Aesica:
In fairness there is a precedent for this, as you can hear other players' skill/spell improve sounds.  Also, it's pretty common in other MMOs today to have some sort of (usually visual) cue that somebody near you gained a level/hp/etc.  It seems like it'd mostly be one of those things to (hopefully) promote positive gameplay, where someone else in the same room can hear it and offer up a "congrats" or something.

If nothing else, we could run an informal poll to see how people generally feel about it, but I'd be surprised if it didn't get an overall positive reception.

davidklaw:
My personal opinion: no thanks. Not globally at least. However, I could see this being an interesting addition amongst guildmates or those in fellowship groups like those suggested in #534, where players have opted into closer relationships.

Gar103:
I would love option 5. Information in Meridian can sometimes matter quite a bit. Here's an example - maybe I want to watch someone as part of the black dagger game. So, I pretend to be imping. Other players being able to hear that I'm not imping is information I don't want to give away. I just tried this earlier this week and my target knew I was faking and really just watching them because they noticed I never gave off an imp sound.

Aesica:
I like 5.  Some may want to share their progress indicators with others, some might not.  Personally I'm indifferent either way, but giving players a choice is usually good.

akirmse:
I'd approve whichever of 1, 2, or 3 players want.  The other options are overkill.

adrienlaws:
From the results of a Discord poll, it seems the majority favor this change (option 3).
Below is a pie chart showing the results by % and by count.

That said, there seem to be a lot of opinions.

One anecdote that stood out to me was that a player uses the ding improvement sounds to count how many improves they have until they hit the max for the room.  They said they would prefer all improvement sounds to be private so that they could keep track and not get confused with others' improvement sounds.

I think we could perhaps make all sounds public, but also add a vague text notification in another PR when the player has hit their improvement limit in the current room.  This would encourage players to move and also keep them from being disenfranchised if they learn they have been wasting reagents and times when improvement was not possible
> You feel you have learned all you can here... for now
> You feel an urge to go explore other places
> Your mind is numb and desires novelty

In any case, I think we can either close this PR and leave things as they are (option 1) or go with all sounds public (merge as it stands with option 3).

```mermaid
pie showData title Hear Tougher and Imp Sounds?
    "No Change" : 7
    "All Private" : 5
    "All Public" : 12
```

davidklaw:
I don't have a strong opinion one way or the other on this one. But, I look at those results as split, since _no change_ and _all private_ are both conservative. If we're going to have a decision be driven by community vote, it seems like it ought to be a strong majority. In an effort to move PRs along, perhaps this should get closed?

akirmse:
Yeah, agreed, there's no clear consensus on this.

PR #544, opened 2023-08-16
QoL Feature:  Removed durability damage from shirts, pants, and skirts

Aesica:
As cosmetic items, it's a bit silly that shirts, pants, and skirts take durability damage and eventually break with use.  This leads to people preferring to just hunt naked, or to not bother with these items at all due to the inconvenience of having to pick up new ones regularly.

Removing the durability aspect encourages people to dress up and look their best wherever they go--something they can already do with another cosmetic item:  masks.

This changes the following:
- Shirts (shirt, royalshirt, tanktop) and pants (pantsa, pantsb, pantsc, pantsd, skirt, and shortskirt) to no longer suffer a reduction in hits when the wearer is damaged.
- No longer displays their condition in descriptions.
- No longer mendable, since that doesn't matter anymore.

### DIFF
--- a/kod/object/item/passitem/defmod/pants.kod
+++ b/kod/object/item/passitem/defmod/pants.kod
@@ -20,16 +20,9 @@ resources:
    pants_desc_rsc = \
    "These pants will cover your legs but won't help much in the way of protection."
 
-   pants_condition_exc = " are in immaculate condition."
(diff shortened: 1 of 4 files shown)


davidklaw:
Yes! ➕1️⃣ from me on this one for reasons stated in the description.

PR #551, opened 2023-08-23
Rentable room update 1: Consolidation of duplicate code found in all …

Aesica:
…rent-enabled innkeepers

Each innkeeper that rents rooms has the same large chunks of copy-pasted logic for parsing commands and handling offered items, and that's sloppy.  This PR consolidates all of that logic into the rented-room maintenance object and now only requires that the innkeeper pass in the appropriate values to the appropriate function calls to the room maintenance object.

Additionally, Moxal's room rental speech lines were near-exact copy pastes of Morrigan's and had a few glitches, so I changed them to fix the problems and reflect his personality better.

Aside from Moxal's speech changes, this PR should consist of entirely behind-the-scenes changes.  The user experience should feel exactly the same.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
@@ -38,7 +38,7 @@ resources:
    BarloqueInnkeeper_days_left = "I thanketh thee.  Thou hast %i days left at my luxurious inn."
    BarloqueInnkeeper_too_long = "It pleaseth me that ye liketh thine room, but thou hast paid enough in rent already!  I canst not accept more from you right now."
 
-   BarloqueInnkeeper_destroy_key_copy = "~kPritchett tells you, \"I will discardeth the key for thou.\""
+   BarloqueInnkeeper_destroy_key_copy = "I will discardeth the key for thou."
    
 classvars:
 
@@ -151,7 +151,7 @@ messages:
    SomeoneSaid(what = $,type = $,string = $)
    {
       Send(Send(SYS, @GetRentableRoomMaintenance), @ParseInnkeeperCommands,
-         #who=self, #what=what, #innRid=RID_BAR_INN, #string=string,
+         #who=self, #what=what, #innRid=RID_BAR_INN, #speech=string,
          #initCost=piInitialRoomCost, #perDayCost=piPerDayCost, #keyCopyCost=piKeyCopyCosts,
          #rscRoomRented=BarloqueInnkeeper_room_rented, #rscRoomCost=BarloqueInnkeeper_cost);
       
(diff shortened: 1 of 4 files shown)


kod/util/rntmaint.kod
```diff
@@ -74,6 +100,150 @@ messages:
       return;
    }
 
+   ParseInnkeeperCommands(who=$, what=$, innRid=$, string=$,
```
akirmse, line 103: Can you come up with a more meaningful name than "string"?

kod/util/rntmaint.kod
```diff
@@ -74,6 +100,150 @@ messages:
...
+            #string=rscRoomCost, 
+            #parm1=initCost, #parm2=perDayCost, #parm3=keyCopyCost);
+
+         return;
+      }    
+
+      return;
+   }
+
+   % Checks items offered (ReqOffer) to an innkeeper to see if they're related to rooms
```
akirmse, line 174: Period at the end of the sentence.

Make this into a doc string rather than a comment?  That way it's readable from admin mode.

kod/util/rntmaint.kod
```diff
@@ -24,6 +27,29 @@ resources:
...
+      "If you want to stay longer, I charge %i per day if you pre-pay.  "
+      "Key copies cost %i shillings to make."
+   RoomMaintenance_innkeeper_say_days_left = \
+      "You like it here, I see.  There are %i days until you should "
+      "pay me again."
+   RoomMaintenance_innkeeper_say_too_long = \
+      "Oh, I can't be responsible for holding so much money!  Pay me again "
+      "later when you don't have so many days left on your tab."
+   RoomMaintenance_innkeeper_say_destroy_key_copy = \
+      "%q tells you, \"Oh, yes. I can get rid of this key for "
```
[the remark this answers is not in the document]
Aesica: Woops, good catch.  I was actually going to implement it as a template (seen in the latest update) so that each innkeeper's response didn't require all that extra formatting (~kMorrigan tells you, \"blah blah\") to make it all less error prone should somebody add rentable rooms to any other npcs later on.  Somehow, I forgot about that, but now it's set up and working.  All the other issues should be addressed, too.

kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
```diff
@@ -30,21 +30,16 @@ resources:
...
+      "The rooms here are the finest.  I needeth %i shillings when thou ~Irent~neth a "
+      "~Iroom~n for for 85 days.  After that, I requireth %i per day up front.  Key copies "
+      "costeth %i shillings if thou wishest to share thine room with a special someone."
 
-   BarloqueInnkeeper_days_left = "I thanketh thee.  Thou hast %q days left at my luxurious inn."
+   BarloqueInnkeeper_days_left = "I thanketh thee.  Thou hast %i days left at my luxurious inn."
    BarloqueInnkeeper_too_long = "It pleaseth me that ye liketh thine room, but thou hast paid enough in rent already!  I canst not accept more from you right now."
 
-   BarloqueInnkeeper_destroy_key_copy = "~kPritchett tells you, \"I will discardeth the key for thou.\""
+   BarloqueInnkeeper_destroy_key_copy = "I will discardeth the key for thou."
```
akirmse, line 41: This should be "for thee" (I know it was wrong originally).

Aesica: Should be fixed now.

PR #552, opened 2023-09-04
Fix node attacks by checking for existing `ptAttack` timers

### LINKED ISSUE #27: Xeo attacks are overly aggressive
When there > 100 players online (or whatever the threshold is set to be), Xeo attacks seem to spawn continually, instead of being periodic.

davidklaw:
This PR adds ~two~ three timer checks to the node attack event.

The first check is made with each new game day that passes. Because the randomness of the timer can be greater than the length of an in-game day, I'm almost certain this logic was adding additional `StartAttackTimers`. We confirmed many of these timers exist on live.

The other timer check occurs when an actual node attack is triggered. In that case, any existing `ptAttack` timer is deleted before being the property is reused for sending player notifications.

Edit: A third check was added for existing servers already dealing with many "bad" StartAttackTimer timers. In that case, disable the event (set pbAttackEnabled=0), wait for these timers to expire, and then re-enable the game.

Ideally, once deployed, this change should stop new and unwanted StartAttackTimers from being created, allow the current set of StartAttackTimers to expire, and settle the node attack event down to its intended cadence.

Fixes #27 

Credit to https://github.com/OpenMeridian/Meridian59/commit/3b1958bdc96c65809af4bc7b43ed944e1aaf5a7c

### DIFF
--- a/kod/util/nodeattk.kod
+++ b/kod/util/nodeattk.kod
@@ -256,6 +256,7 @@ messages:
 
       ptAttack = $;
       if NOT Send(self,@DoNodeAttack)
+         AND pbAttackEnabled
       {
          % Between 1 and 2 hours
          ptAttack = CreateTimer(self,@StartAttackTimer,Random(3600,7200)*1000);
--- a/kod/util/nodeattk.kod
+++ b/kod/util/nodeattk.kod
@@ -223,7 +223,9 @@ messages:
 
       % Check the wait time, see if we're up for attacking the node again!
       piWaitTime = piWaitTime - 1;
-      if piWaitTime <= 0 AND poAttackedNode = $
+      if piWaitTime <= 0 
+         AND poAttackedNode = $
+         AND ptAttack = $
       {
          % Generate a wait between a minute and 4 hours
          ptAttack = CreateTimer(self,@StartAttackTimer,Random(60,14400)*1000);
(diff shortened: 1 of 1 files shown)


davidklaw:
While testing the original PR in my local environment by simulating a live server will more than one StartAttackTimer timers, I noticed that when one would expire, another would take its place. There should only ever be one of these timers. What I found was that each time one of these timers expired and called `StartAttackTimer()`, it would attempt a node attack, fail because conditions aren't meant, and then create yet another timer.

Commit https://github.com/Meridian59/Meridian59/pull/552/commits/e4deaf8099c69e7a4ce13e92f343db93e08a1432 addresses this by skipping new timers if the node attack event is disabled, so server admins can disable the event (set `pbAttackEnabled=0`), let all `StartAttackTimer` timers expire, and then re-enable the event, at which point they should only ever see one of these timers at a time.

I tested this in my local environment by calling `NewDay()` many times, enough to find that a number of `StartAttackTimer` timers were active (`show timers` in console). I let some of those timers expire to confirm that they were replaced by another "bad" timer. I saved the game, deployed these changes, then let the server run for ~5 hours so that these timers could expire. I then re-enabled the event and let the server run for another ~10 hours. As expected, only a single timer remained active. The only issue I came across was that `piWaitTime` went negative (-15) but that may have been the result of something I did to manipulate my environment for testing. In any case, it'll reset the next time a node attack occurs or if Recreate is called, or if updated by an admin.

PR #554, opened 2023-09-06
Update Containers: Sorted with number items on top then alphabetical, allow withdraw specific quantities

CardiovascularMike:
This PR updates storage boxes like guild hall chests, personal chests in rooms, and other chests that are available throughout the game (IE: Thrashers).

The goal is to make managing inventory easier for players. To that end we make two major changes:
1. We allow players to pickup a user-selectable number of items for number items
2. We sort all items in the list box.  The sort is two steps: First number items are put at the top and sorted alphabetically then not-number items are sorted alphabetically against each other and appended to the bottom of the list.

Finally we add two minor changes:
3. Personal storage boxes have increased bulk.  This is to align them with all the other storage changes that have recently been made.  Vaults now are much larger and guild hall chests have all had their bulk hold increased but Personal storage boxes were left behind.  These storage boxes are still much smaller than guild hall chests.
4. When a user selects numerous items to grab we provide a success message if they are able to carry everything.  If the user is unable to hold everything instead they get a message about only the items they are unable to hold.

Attached is a screenshot of the sorted listbox:
GroupSort_ChestItems

### DIFF
--- a/blakserv/sprocket.c
+++ b/blakserv/sprocket.c
@@ -34,6 +34,7 @@ client_def_table_type client_def_table[] =
 	{ BP_REQ_INVENTORY,        { {0, DONE_PARM} } },
 	{ BP_REQ_TURN,             { {4, TAG_OBJECT}, {2, TAG_INT}, {0, DONE_PARM} } },
 	{ BP_REQ_GET,              { {4, TAG_OBJECT}, {0, DONE_PARM} } },
+	{ BP_REQ_GET_FROM_CONT,    { {2, LIST_OBJ_PARM}, {0, DONE_PARM} } },
 	{ BP_REQ_DROP,             { {4, TAG_OBJECT}, {0, DONE_PARM} } },
 	{ BP_REQ_INVENTORY_MOVE,   { {2, TAG_INT}, {2, TAG_INT}, {0, DONE_PARM} } },
 	{ BP_REQ_PUT,              { {4, TAG_OBJECT}, {4, TAG_OBJECT}, {0, DONE_PARM} } },
--- a/blakserv/sprocket.c
+++ b/blakserv/sprocket.c
@@ -34,7 +34,7 @@ client_def_table_type client_def_table[] =
 	{ BP_REQ_INVENTORY,        { {0, DONE_PARM} } },
 	{ BP_REQ_TURN,             { {4, TAG_OBJECT}, {2, TAG_INT}, {0, DONE_PARM} } },
 	{ BP_REQ_GET,              { {4, TAG_OBJECT}, {0, DONE_PARM} } },
-	{ BP_REQ_GET_FROM_CONT,    { {2, LIST_OBJ_PARM}, {0, DONE_PARM} } },
+	{ BP_REQ_GET_FROM_CONTAINER,    { {2, LIST_OBJ_PARM}, {0, DONE_PARM} } },
 	{ BP_REQ_DROP,             { {4, TAG_OBJECT}, {0, DONE_PARM} } },
 	{ BP_REQ_INVENTORY_MOVE,   { {2, TAG_INT}, {2, TAG_INT}, {0, DONE_PARM} } },
 	{ BP_REQ_PUT,              { {4, TAG_OBJECT}, {4, TAG_OBJECT}, {0, DONE_PARM} } },
--- a/clientd3d/gameuser.c
+++ b/clientd3d/gameuser.c
@@ -154,11 +154,11 @@ void GotObjectContents(ID object_id, list_type contents)
    {
       if (IsNumberObj(((object_node *)(l->data))->id))
       {
-         stacklist = list_add_sorted_item(stacklist, (object_node *)(l->data), CompareObjectNameRsc);
+         stacklist = list_add_sorted_item(stacklist, (l->data), CompareObjectNameRsc);
       }
       else
       {
-         nostacklist = list_add_sorted_item(nostacklist, (object_node *)(l->data), CompareObjectNameRsc);
+         nostacklist = list_add_sorted_item(nostacklist, (l->data), CompareObjectNameRsc);
       }
    }
    //Concat the two lists into one
--- a/clientd3d/gameuser.c
+++ b/clientd3d/gameuser.c
@@ -148,14 +148,14 @@ void GotObjectContents(ID object_id, list_type contents)
 	 GameMessagePrintf(GetString(hInst, IDS_EMPTY), LookupNameRsc(r->obj.name_res));
       return;
    }
-   //Separate contents into number items and other items and alpha sort
+   // Separate contents into number items and other items and alpha sort
    for (l = contents; l != NULL; l = l->next)
    {
       sorted_list = list_add_sorted_item(sorted_list, (l->data), CompareObjectNameAndNumber);
    }
 
    sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), sorted_list, LD_MULTIPLESEL | LD_AMOUNTS);   
-   //Request Pickup from container if any items are selected
+   // Request pickup from container if any items are selected
    if (list_length(sel_list) > 0)
       RequestPickupFromContainer(sel_list);
 
--- a/clientd3d/gameuser.c
+++ b/clientd3d/gameuser.c
@@ -136,10 +136,9 @@ void UserPickup(void)
  */
 void GotObjectContents(ID object_id, list_type contents)
 {
-   list_type sel_list, l, stacklist, nostacklist;
+   list_type sel_list, l, sorted_list;
    room_contents_node *r;
-   stacklist = NULL;
-   nostacklist = NULL;
+   sorted_list = NULL;
 
    r = GetRoomObjectById(object_id);
 
@@ -149,29 +148,19 @@ void GotObjectContents(ID object_id, list_type contents)
 	 GameMessagePrintf(GetString(hInst, IDS_EMPTY), LookupNameRsc(r->obj.name_res));
       return;
    }
-   //Separate contents into number items and other items
+   //Separate contents into number items and other items and alpha sort
    for (l = contents; l != NULL; l = l->next)
    {
-      if (IsNumberObj(((object_node *)(l->data))->id))
-      {
-         stacklist = list_add_sorted_item(stacklist, (l->data), CompareObjectNameRsc);
-      }
-      else
-      {
-         nostacklist = list_add_sorted_item(nostacklist, (l->data), CompareObjectNameRsc);
-      }
+      sorted_list = list_add_sorted_item(sorted_list, (l->data), CompareObjectNameAndNumber);
    }
-   //Concat the two lists into one
-   stacklist = list_append(stacklist, nostacklist);
 
-   sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), stacklist, LD_MULTIPLESEL | LD_AMOUNTS);   
+   sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), sorted_list, LD_MULTIPLESEL | LD_AMOUNTS);   
    //Request Pickup from container if any items are selected
    if (list_length(sel_list) > 0)
       RequestPickupFromContainer(sel_list);
 
    ObjectListDestroy(sel_list);
-   ObjectListDestroy(stacklist);
-   ObjectListDestroy(nostacklist);
+   ObjectListDestroy(sorted_list);
 }
 /************************************************************************/
 /*
--- a/clientd3d/gameuser.c
+++ b/clientd3d/gameuser.c
@@ -136,8 +136,10 @@ void UserPickup(void)
  */
 void GotObjectContents(ID object_id, list_type contents)
 {
-   list_type sel_list, l;
+   list_type sel_list, l, stacklist, nostacklist;
    room_contents_node *r;
+   stacklist = NULL;
+   nostacklist = NULL;
 
    r = GetRoomObjectById(object_id);
 
@@ -147,13 +149,29 @@ void GotObjectContents(ID object_id, list_type contents)
 	 GameMessagePrintf(GetString(hInst, IDS_EMPTY), LookupNameRsc(r->obj.name_res));
       return;
    }
+   //Separate contents into number items and other items
+   for (l = contents; l != NULL; l = l->next)
+   {
+      if (IsNumberObj(((object_node *)(l->data))->id))
+      {
+         stacklist = list_add_sorted_item(stacklist, (object_node *)(l->data), CompareObjectNameRsc);
+      }
+      else
+      {
+         nostacklist = list_add_sorted_item(nostacklist, (object_node *)(l->data), CompareObjectNameRsc);
+      }
+   }
+   //Concat the two lists into one
+   stacklist = list_append(stacklist, nostacklist);
 
-   sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), contents, LD_MULTIPLESEL);   
-
-   for (l = sel_list; l != NULL; l = l->next)
-      RequestPickup(((room_contents_node *) (l->data))->obj.id);
+   sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), stacklist, LD_MULTIPLESEL | LD_AMOUNTS);   
+   //Request Pickup from container if any items are selected
+   if (list_length(sel_list) > 0)
+      RequestPickupFromContainer(sel_list);
 
    ObjectListDestroy(sel_list);
+   ObjectListDestroy(stacklist);
+   ObjectListDestroy(nostacklist);
 }
 /************************************************************************/
 /*
--- a/clientd3d/lookdlg.c
+++ b/clientd3d/lookdlg.c
@@ -444,12 +444,16 @@ void LookCommand(HWND hDlg, int ctrl_id, HWND hwndCtl, UINT codeNotify)
 	      if (ListBox_GetSel(info->hwndListBox, i) > 0)
          {
             if (IsNumberObj(obj->id))
+            {
                DWORD amount = (DWORD)ListBox_GetItemData(info->hwndQuanList,i);	
 	            obj->temp_amount = amount;
 	            selection = list_add_item(selection, obj);
+            }
             else
+            {
                obj->temp_amount = 1;
                selection = list_add_item(selection, obj);
+            }
          }
 	      else 
          {
--- a/clientd3d/lookdlg.c
+++ b/clientd3d/lookdlg.c
@@ -439,8 +439,8 @@ void LookCommand(HWND hDlg, int ctrl_id, HWND hwndCtl, UINT codeNotify)
       
       for (i = 0; i < num_entries; i++)
       {
-	      /* If item is selected, add to selection list, else free */
-	      	 obj = (object_node *) ListBox_GetItemData(info->hwndListBox, i);
+	 /* If item is selected, add to selection list, else free */
+	 obj = (object_node *) ListBox_GetItemData(info->hwndListBox, i);
 	 if (ListBox_GetSel(info->hwndListBox, i) > 0)
 	    selection = list_add_item(selection, obj);
 	 else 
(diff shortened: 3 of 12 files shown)


clientd3d/lookdlg.c
CardiovascularMike, line 1: There is no functional change here.  At one point during development of this PR I had different logic in this spot of LookDlg.C.  I was able to return this spot to the original code - although I left the added brackets in place for now.

clientd3d/object.c
CardiovascularMike, line 1: CompareObjectNameRsc takes in objects and uses LookupNameRsc to get the name_res for each object and then forces the return from LookupNameRsc into a string.  Then we can use C's strcmp function to do the actual comparison.

All that is required of CompareObjectNameRsc is to return a negative number if obj1 is earlier in the alphabet than obj2, zero if they are the same object name, and a positive number if obj1 is later in the alphabet than object2.

I'm not sure but I may be able to do away with the casting to object_node if I change the arguments from void in the initial function call?

CardiovascularMike: In case it's not clear, this function is a required argument for list_add_sorted_item which is our insertion sort routine.

clientd3d/gameuser.c
```diff
@@ -147,13 +149,29 @@ void GotObjectContents(ID object_id, list_type contents)
...
+      {
+         stacklist = list_add_sorted_item(stacklist, (object_node *)(l->data), CompareObjectNameRsc);
+      }
+      else
+      {
+         nostacklist = list_add_sorted_item(nostacklist, (object_node *)(l->data), CompareObjectNameRsc);
+      }
+   }
+   //Concat the two lists into one
+   stacklist = list_append(stacklist, nostacklist);
```
akirmse, line 165: Rather than building up two lists and concatenating them, it will be shorter and easier to understand if you just do

```
   for (l = contents; l != NULL; l = l->next)
      sorted_list = list_add_sorted_item(sorted_list, (object_node *) (l->data), CompareObjectNameAndNumber);
```

where the new comparison proc works this way:

- Compute two variables (say, `number1` and `number2` to determine whether the items are number items.
- If number1 != number2, return a value (1 or -1) that puts the number item first.
- Otherwise, look up the two names and compare them, like you do in CompareObjectNameRsc.

Make sure that this new comparison proc has a comment describing its purpose (sort number items before other items, and within each group, sort alphabetically).

clientd3d/lookdlg.c
CardiovascularMike: I've reverted this non-functional change so that there is no change at all even one of formatting.

clientd3d/gameuser.c
```diff
@@ -147,13 +148,19 @@ void GotObjectContents(ID object_id, list_type contents)
 	 GameMessagePrintf(GetString(hInst, IDS_EMPTY), LookupNameRsc(r->obj.name_res));
       return;
    }
+   //Separate contents into number items and other items and alpha sort
+   for (l = contents; l != NULL; l = l->next)
+   {
+      sorted_list = list_add_sorted_item(sorted_list, (l->data), CompareObjectNameAndNumber);
+   }
```
CardiovascularMike, line 155: Here I've updated the method so that the compare function of the insertion sort so that the comparison does the stacking of number items and also the alphabetical sort of number items together and then not number items together.  As suggested by Zaphod.  That comment appears to have disappeared with the edits.

CardiovascularMike: Also we are no longer casting the data to object_node here because list_add_sorted_item passes off a void to the comparison function anyhow and forcing the issue in the comparison function call looked like it would require changing list_add_sorted_item which is used elsewhere for other tasks.

clientd3d/lookdlg.c
```diff
@@ -439,12 +439,16 @@ void LookCommand(HWND hDlg, int ctrl_id, HWND hwndCtl, UINT codeNotify)
       
       for (i = 0; i < num_entries; i++)
       {
-	 /* If item is selected, add to selection list, else free */
-	 obj = (object_node *) ListBox_GetItemData(info->hwndListBox, i);
-	 if (ListBox_GetSel(info->hwndListBox, i) > 0)
-	    selection = list_add_item(selection, obj);
-	 else 
-	    ObjectDestroyAndFree(obj);
+	      /* If item is selected, add to selection list, else free */
```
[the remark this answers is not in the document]
CardiovascularMike: Done.

kod/object/active/holder/storebox/smallbox.kod
```diff
@@ -27,7 +27,7 @@ classvars:
    vrIcon = SmallBox_icon_rsc
    vrDesc = SmallBox_desc_rsc
 
-   viBulk_hold_max = 3000
+   viBulk_hold_max = 5000
```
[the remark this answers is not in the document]
CardiovascularMike: Done.

blakserv/sprocket.c
```diff
@@ -34,6 +34,7 @@ client_def_table_type client_def_table[] =
 	{ BP_REQ_INVENTORY,        { {0, DONE_PARM} } },
 	{ BP_REQ_TURN,             { {4, TAG_OBJECT}, {2, TAG_INT}, {0, DONE_PARM} } },
 	{ BP_REQ_GET,              { {4, TAG_OBJECT}, {0, DONE_PARM} } },
+	{ BP_REQ_GET_FROM_CONT,    { {2, LIST_OBJ_PARM}, {0, DONE_PARM} } },
```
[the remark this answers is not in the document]
CardiovascularMike: Done.

clientd3d/gameuser.c
CardiovascularMike: Good idea and done.

clientd3d/object.c
```diff
@@ -50,6 +50,25 @@ int CompareRoomObjectDistance(void *r1, void *r2)
 {
    return ((room_contents_node *) r1)->distance - ((room_contents_node *) r2)->distance;
 }
+/*************************************************************************/
+// Alphabetical compare of objects based on display name
+//
+int CompareObjectNameRsc(void *obj1, void *obj2)
```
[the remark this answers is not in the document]
CardiovascularMike: Tried declaring object_node but the change doesn't play well with List_Add_Sorted_Item.  Instead have removed declaration in GotObjectContents.

clientd3d/object.c
```diff
@@ -50,6 +50,25 @@ int CompareRoomObjectDistance(void *r1, void *r2)
...
+    // Cast obj1 and obj2 to the correct type (assuming object_node*)
+    object_node *node1 = (object_node *)obj1;
+    object_node *node2 = (object_node *)obj2;
+
+    const char *string1;
+    const char *string2;
+
+    //Get string1 and string2
+    string1 = LookupNameRsc(node1->name_res);
+    string2 = LookupNameRsc(node2->name_res);
```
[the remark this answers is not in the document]
CardiovascularMike: Done.

clientd3d/object.c
```diff
@@ -50,6 +50,30 @@ int CompareRoomObjectDistance(void *r1, void *r2)
...
+      //Node2 before Node1 so that number object are on top
+      return (IsNumberObj(node2->id) - IsNumberObj(node1->id));
+   }
+   else
+   {
+      // Extract strings to compare
+      const char *string1 = LookupNameRsc(node1->name_res);
+      const char *string2 = LookupNameRsc(node2->name_res);
+
+      // Use stricmp to compare the name_res strings
```
akirmse, line 73: This comment simply mimics the code; remove it.

clientd3d/gameuser.c
```diff
@@ -147,13 +148,19 @@ void GotObjectContents(ID object_id, list_type contents)
 	 GameMessagePrintf(GetString(hInst, IDS_EMPTY), LookupNameRsc(r->obj.name_res));
       return;
    }
+   //Separate contents into number items and other items and alpha sort
```
akirmse, line 151: Space after //

clientd3d/gameuser.c
```diff
@@ -147,13 +148,19 @@ void GotObjectContents(ID object_id, list_type contents)
...
+   {
+      sorted_list = list_add_sorted_item(sorted_list, (l->data), CompareObjectNameAndNumber);
+   }
 
-   sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), contents, LD_MULTIPLESEL);   
-
-   for (l = sel_list; l != NULL; l = l->next)
-      RequestPickup(((room_contents_node *) (l->data))->obj.id);
+   sel_list = DisplayLookList(hMain, GetString(hInst, IDS_GET), sorted_list, LD_MULTIPLESEL | LD_AMOUNTS);   
+   //Request Pickup from container if any items are selected
```
akirmse, line 158: Space after //
Pickup should be lowercase

clientd3d/object.c
```diff
@@ -50,6 +50,30 @@ int CompareRoomObjectDistance(void *r1, void *r2)
...
+//
+int CompareObjectNameAndNumber(void *obj1, void *obj2)
+{
+    // Cast obj1 and obj2 to the correct type
+   object_node *node1 = (object_node *)obj1;
+   object_node *node2 = (object_node *)obj2;
+
+   if (IsNumberObj(node1->id) != IsNumberObj(node2->id))
+   {
+      //Node2 before Node1 so that number object are on top
```
akirmse, line 64: Space after //

akirmse: object -> objects

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3685,6 +3695,123 @@ messages:
       return;
    }
 
+   UserGetFromContainer(items_list=$, numbers_list=$)
+   {
```
akirmse, line 3699: The main issue here is that lots of this code was copied from UserGet().  (You can read about the "DRY principle" to learn why this is a problem).  So before I review this in detail, how about you make an attempt to share the common code between the two functions.

akirmse (commented):
One basic question about this whole approach: why should picking up from a container be different from picking up from the ground?  I don't understand why a separate message is needed.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: I did think about this a bit. In terms of gameplay the two actions are fundamentally different.  In terms of coding things are murkier.  I'll start with gameplay and may ask about some coding options on Discord and then respond here or make the changes as needed.

The fundamental difference in terms of gameplay is that players will withdraw a stack of items from the chest once or twice per game session.  This is much like how a player will go to a vendor and buy a stack of reagents to be fully ready for a battle.  Here the user wants to manage their inventory and is planning for an activity.

Conversely picking items off the ground is an action the player will do 40-80+ times during a gameplay session if they are fighting monsters.  Here they are simply accumulating goods.  Each pickup is in the range of 2-5 items in small stacks.  I don't want the user to have to tell the game that yes, they want to pick up all 3 of those rainbow ferns.  In practice, most experienced players will pick up items from the ground with hotkeys.  For me this is 'G' for get menu, then down arrow to the first item I want, then spacebar to select, then arrows for more items and enter to pickup everything highlighted.  It would be a problem if the user had to mouse to the quantity tab for each pickup.

So in terms of player action there is a difference.  That said there is no need to turn on the LD_AMOUNTS flag for the get/look menu outside of GotObjectContents.  In that case the UserGet could have a default condition of number_items = $.  The only problem with that is that the UserGet routine does balloon in length a bit.  Now we will need to check if number_items <> $ in a bunch of spots before doing the oSplit function.

Another issue is one I struggled with greatly (I even had two working options coded).  This is somewhat fundamental.  The issue is: do we loop over the selected items in C and then UserGet each item one at a time in KOD or do we send over the list and then UserGetFromContainer the whole list in a loop in KOD.  In this case I opted for the second option because I wanted a success message when everything is picked up and I didn't want to add the two properties that would be required to do that while looping over the selection in C. (The process involved sending the length of the list as an argument and then incrementing a property for each loop and a second property for each successful pickup.  When the loop property equals the length argument if the success property also equals the length argument you can send the success message).

The result is that right now UserGetFromContainer loops over a list in KOD and UserGet loops over the list in C and picks up each element of the list one at a time.  If we want to combine these functions we are adding a bunch of checks for (Number_Items = $) and we are probably moving UserGetFromContainer away from the list method and towards looping in C.  If instead we want to change UserGet to loop in KOD that is do-able but seemed risky to me.  UserGet is pretty fundamental to the gameplay and any change of that scale would require some amount of additional testing.

That said, if this is the direction we want to go I think I can implement either of these options.

A third path exists as well: We can offload a lot of the safety checks and make these into sub-routines and then UserGet and UserGetFromContainer would both call these subroutines first and if that passes then we move onto the meat of getting the goods.  I think this would pull 20-30 lines out of each method.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3685,6 +3695,123 @@ messages:
...
+      for oitem in items_list
+      {
+
+         iNumber = First(numbers_list);
+         oOldOwner = Send(oitem,@GetOwner);
+
+         % Make sure we're not trying to get something we've already got!
+         if oOldOwner = self
+         {
+            return;
```
CardiovascularMike, line 3711: I may remove this return and the one on line 3722 and instead send a message to the user.  Right now if you trip either of these safety checks the looping over items stops and further action is aborted for safety.  Instead we could send a message like: "You reach for the %ItemName but its nowhere to be found!"  The only way I was able to force these safeties to trip is if I had two characters try to grab the same stuff simultaneously.  In that instance one character gets the stuff and the other gets a message that the item is not accessible.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3685,6 +3706,78 @@ messages:
...
+                     #parm1=Send(oitem,@GetName));
+                  Send(self,@NewHold,#what=oSplit);
+                  iCantGet = 1;
+               }
+            }
+            % iCan_hold > iNumber
+            else
+            {
+               oSplit = Send(oitem,@Split,#number=iNumber);
+               Send(self,@NewHold,#what=oSplit);
```
CardiovascularMike, line 3749: We used to check ReqNewOwner again here (any any time before @NewHold) but ReqNewOwner is already checked in the safety checks.  This redundancy exists in UserGet as well.  We can add back in or also remove from UserGet as appropriate.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: Attempt Made.  Sorry for long reply above.

clientd3d/gameuser.c
CardiovascularMike: Done.

clientd3d/object.c
CardiovascularMike: Done.

clientd3d/object.c
CardiovascularMike: Done.

clientd3d/gameuser.c
CardiovascularMike: Done.

CardiovascularMike:
> 

I did think about this a bit. In terms of gameplay the two actions are fundamentally different.  In terms of coding things are murkier.  I'll start with gameplay and may ask about some coding options on Discord and then respond here or make the changes as needed.

The fundamental difference in terms of gameplay is that players will withdraw a stack of items from the chest once or twice per game session.  This is much like how a player will go to a vendor and buy a stack of reagents to be fully ready for a battle.  Here the user wants to manage their inventory and is planning for an activity.

Conversely picking items off the ground is an action the player will do 40-80+ times during a gameplay session if they are fighting monsters.  Here they are simply accumulating goods.  Each pickup is in the range of 2-5 items in small stacks.  I don't want the user to have to tell the game that yes, they want to pick up all 3 of those rainbow ferns.  In practice, most experienced players will pick up items from the ground with hotkeys.  For me this is 'G' for get menu, then down arrow to the first item I want, then spacebar to select, then arrows for more items and enter to pickup everything highlighted.  It would be a problem if the user had to mouse to the quantity tab for each pickup.

So in terms of player action there is a difference.  That said there is no need to turn on the LD_AMOUNTS flag for the get/look menu outside of GotObjectContents.  In that case the UserGet could have a default condition of number_items = $.  The only problem with that is that the UserGet routine does balloon in length a bit.  Now we will need to check if number_items <> $ in a bunch of spots before doing the oSplit function.

Another issue is one I struggled with greatly (I even had two working options coded).  This is somewhat fundamental.  The issue is: do we loop over the selected items in C and then UserGet each item one at a time in KOD or do we send over the list and then UserGetFromContainer the whole list in a loop in KOD.  In this case I opted for the second option because I wanted a success message when everything is picked up and I didn't want to add the two properties that would be required to do that while looping over the selection in C. (The process involved sending the length of the list as an argument and then incrementing a property for each loop and a second property for each successful pickup.  When the loop property equals the length argument if the success property also equals the length argument you can send the success message).

The result is that right now UserGetFromContainer loops over a list in KOD and UserGet loops over the list in C and picks up each element of the list one at a time.  If we want to combine these functions we are adding a bunch of checks for (Number_Items = $) and we are probably moving UserGetFromContainer away from the list method and towards looping in C.  If instead we want to change UserGet to loop in KOD that is do-able but seemed risky to me.  UserGet is pretty fundamental to the gameplay and any change of that scale would require some amount of additional testing.

That said, if this is the direction we want to go I think I can implement either of these options.

A third path exists as well: We can offload a lot of the safety checks and make these into sub-routines and then UserGet and UserGetFromContainer would both call these subroutines first and if that passes then we move onto the meat of getting the goods.  I think this would pull 20-30 lines out of each method.

CardiovascularMike:
> One basic question about this whole approach: why should picking up from a container be different from picking up from the ground? I don't understand why a separate message is needed.

I did think about this a bit. In terms of gameplay the two actions are fundamentally different.  In terms of coding things are murkier.  I'll start with gameplay and may ask about some coding options on Discord and then respond here or make the changes as needed.

The fundamental difference in terms of gameplay is that players will withdraw a stack of items from the chest once or twice per game session.  This is much like how a player will go to a vendor and buy a stack of reagents to be fully ready for a battle.  Here the user wants to manage their inventory and is planning for an activity.

Conversely picking items off the ground is an action the player will do 40-80+ times during a gameplay session if they are fighting monsters.  Here they are simply accumulating goods.  Each pickup is in the range of 2-5 items in small stacks.  I don't want the user to have to tell the game that yes, they want to pick up all 3 of those rainbow ferns.  In practice, most experienced players will pick up items from the ground with hotkeys.  For me this is 'G' for get menu, then down arrow to the first item I want, then spacebar to select, then arrows for more items and enter to pickup everything highlighted.  It would be a problem if the user had to mouse to the quantity tab for each pickup.

So in terms of player action there is a difference.  That said there is no need to turn on the LD_AMOUNTS flag for the get/look menu outside of GotObjectContents.  In that case the UserGet could have a default condition of number_items = $.  The only problem with that is that the UserGet routine does balloon in length a bit.  Now we will need to check if number_items <> $ in a bunch of spots before doing the oSplit function.

Another issue is one I struggled with greatly (I even had two working options coded).  This is somewhat fundamental.  The issue is: do we loop over the selected items in C and then UserGet each item one at a time in KOD or do we send over the list and then UserGetFromContainer the whole list in a loop in KOD.  In this case I opted for the second option because I wanted a success message when everything is picked up and I didn't want to add the two properties that would be required to do that while looping over the selection in C. (The process involved sending the length of the list as an argument and then incrementing a property for each loop and a second property for each successful pickup.  When the loop property equals the length argument if the success property also equals the length argument you can send the success message).

The result is that right now UserGetFromContainer loops over a list in KOD and UserGet loops over the list in C and picks up each element of the list one at a time.  If we want to combine these functions we are adding a bunch of checks for (Number_Items = $) and we are probably moving UserGetFromContainer away from the list method and towards looping in C.  If instead we want to change UserGet to loop in KOD that is do-able but seemed risky to me.  UserGet is pretty fundamental to the gameplay and any change of that scale would require some amount of additional testing.

That said, if this is the direction we want to go I think I can implement either of these options.

A third path exists as well: We can offload a lot of the safety checks and make these into sub-routines and then UserGet and UserGetFromContainer would both call these subroutines first and if that passes then we move onto the meat of getting the goods.  I think this would pull 20-30 lines out of each method.

CardiovascularMike:
Closing this PR as the update was refactored in a better format as PR#565.

PR #555, opened 2023-09-06
Rentable room update 2:  Adds "change lock" command to rentable room innkeepers

Aesica:
Currently, there's nothing that can be done about a key copy that gets stolen, lost, or just somehow slips away from you.  This adds a command to innkeepers, "change lock" which let you to delete all existing copies of your room key for a particular inn.

As an added note, "change lock" will also allow any "ghost" key copies (copies offered to an innkeeper to be destroyed that don't get disposed because they're still in the master key's plCopies list) to be properly garbage collected.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
@@ -39,6 +39,9 @@ resources:
    BarloqueInnkeeper_too_long = "It pleaseth me that ye liketh thine room, but thou hast paid enough in rent already!  I canst not accept more from you right now."
 
    BarloqueInnkeeper_destroy_key_copy = "I will discardeth the key for thee."
+   BarloqueInnkeeper_change_lock = \
+      "Oh, hast thou misplaced one of thine key copies?  I shalt changeth thine room lock at once, "
+      "denying any unwashed miscreants access therein!"
    
 classvars:
 
@@ -153,7 +156,8 @@ messages:
       Send(Send(SYS, @GetRentableRoomMaintenance), @ParseInnkeeperCommands,
          #who=self, #what=what, #innRid=RID_BAR_INN, #speech=string,
          #initCost=piInitialRoomCost, #perDayCost=piPerDayCost, #keyCopyCost=piKeyCopyCosts,
-         #rscRoomRented=BarloqueInnkeeper_room_rented, #rscRoomCost=BarloqueInnkeeper_cost);
+         #rscRoomRented=BarloqueInnkeeper_room_rented, #rscRoomCost=BarloqueInnkeeper_cost,
+         #rscChangeLock=BarloqueInnkeeper_change_lock);
       
       propagate;
    }
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
@@ -27,20 +27,20 @@ resources:
    BqInn_mace_sale = "Aye, ye might needeth one of them if ye plan on leavingeth our cultured city. I got some ye can buyeth."
    BqInn_unwanted_give = "~k%s%s tells you, \"Whyfore dost you offer me that?~n~k\""
 
-   BarloqueInnkeeper_room_rented = "Here is the key to thine room.  I trust you will findest "
+   BarloqueInnkeeper_room_rented = "Here is the key to thy room.  I trust you will findest "
       "our rooms to be most well-appointed.  Enjoyest thy stay!"
 
    BarloqueInnkeeper_cost = \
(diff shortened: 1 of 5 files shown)


kod/util/rntmaint.kod
```diff
@@ -169,7 +175,26 @@ messages:
             #parm1=initCost, #parm2=perDayCost, #parm3=keyCopyCost);
 
          return;
-      }    
+      }
+      
+      % Command: Change Lock
+      if StringContain(speech, RoomMaintenance_command_change_lock)
+         OR StringContain(speech, "change lock")
```
akirmse, line 182: Isn't this checking for the same thing twice?

kod/util/rntmaint.kod
```diff
@@ -169,7 +175,26 @@ messages:
...
+      {
+         rSayRsc = Send(self, @ChangeLock, #who=what, #iLocation=innRid);
+
+         % On success, rSayRsc will be $, so respond in the innkeeper's local flavor
```
akirmse, line 186: The non-existent documentation for ChangeLock doesn't say what its return value is.  Add a document string to it.

kod/util/rntmaint.kod
```diff
@@ -391,7 +440,32 @@ messages:
 
       % RenewRental returns the new days left.
       return send(oRoom,@RenewRental,#iTimeAdded=iDays);
-   }      
+   }
+
+   FindKeyByPlayer(who=$, iLocation=$)
```
akirmse, line 445: Add documentation

kod/util/rntmaint.kod
Aesica: It is, but the other commands were like that as well so I just followed what I saw to be the pattern.  It seems kind of pointless to include them twice though, so if you want, I'll remove the extra check from this and the other commands as well.

kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
```diff
@@ -39,6 +39,9 @@ resources:
    BarloqueInnkeeper_too_long = "It pleaseth me that ye liketh thine room, but thou hast paid enough in rent already!  I canst not accept more from you right now."
 
    BarloqueInnkeeper_destroy_key_copy = "I will discardeth the key for thee."
+   BarloqueInnkeeper_change_lock = \
+      "Oh, hast thou misplaced one of thine key copies?  I shalt changeth thine room lock at once, "
```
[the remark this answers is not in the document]
Aesica: It's worth pointing out that Pritchett incorrectly uses "thine" when he means "thy" several times, so I figured that was just how he was meant to speak even if incorrect.  In light of this feedback, I've changed it and the other incorrect uses of "thine" to "thy."

kod/util/rntmaint.kod
```diff
@@ -391,7 +445,34 @@ messages:
...
+   "Returns the gold room key matching the specified inn RID held by the "
```
akirmse, line 451: The key it returns is always gold?  If that ever changes color, someone will know to change this comment?  You get the point....

kod/util/rntmaint.kod
akirmse: Yes, please fix this everywhere you see it (after testing to make sure it still works, of course).  It's worse than pointless: if you ever decide to rename "change lock" to something else, you go and change the resource, but "change lock" still works?!?!

kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
```diff
@@ -27,18 +27,21 @@ resources:
    BqInn_mace_sale = "Aye, ye might needeth one of them if ye plan on leavingeth our cultured city. I got some ye can buyeth."
    BqInn_unwanted_give = "~k%s%s tells you, \"Whyfore dost you offer me that?~n~k\""
 
-   BarloqueInnkeeper_room_rented = "Here is the key to thine room.  I trust you will findest "
+   BarloqueInnkeeper_room_rented = "Here is the key to thy room.  I trust you will findest "
```
akirmse, line 30: Mayhem made the point that maybe these "thine"s were wrong intentionally?  If so I'm OK with it.  Either way.

kod/util/rntmaint.kod
Aesica: Done, and gladly so.

kod/util/rntmaint.kod
Aesica: No, that's just the name of the key.  But I see exactly what you mean and have changed it to reference it as its class name (RoomKey) as well as what its purpose is (master/original key).

kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
Aesica: I reverted the fixes (thy back to thine, as well as a thee back to thou from earlier) and left a note so that it hopefully won't cause any confusion to anyone else going forward.

PR #556, opened 2023-09-06
Disarm drop fix

Aesica:
Currently, disarming a player has a chance to force the weapon to drop to the ground instead of just unequipping it.  That's fine in most cases, but it can also do this in situations where it either isn't an expected behavior, or shouldn't happen at all:

- Disarmed player is a DM (big no no)
- Disarmed weapon has soft  white light (transcendence aka bond, supposed to stay with you)
- Disarmed weapon shouldn't leave the player's inventory for other reasons (unused currently by any weapon besides the defunct sword of the hunt, but leaving the door open for future use)

Now, if any of these conditions are met, disarm will unequip the weapon when it triggers, even if the roll to drop the target's weapon to the ground is successful.

I tested with a series of debug statements showing me the results of these checks on a successful disarm.  Target weapons included an ordinary mace (can disarm drop), a soft white light scimitar (cannot disarm drop), a sword of the hunt (cannot disarm drop), and any weapon held by a DM (cannot disarm drop).


kod/object/passive/skill/disarm.kod
```diff
@@ -140,7 +140,17 @@ messages:
          {
             % Currently, have 1/5th your percent plus your modifier chance to make opponent drop weapon.
             iAbil = (iAbil + 1 + iModifier) / DISARM_FACTOR;
-            if (random(1,100) < iAbil) and (not Send(Send(oTarget,@GetOwner),@IsArena))
+
+            % Cannot force a weapon to drop if...
```
akirmse, line 144: This comment is too explicit; it simply repeats the code and will thus become out of date if the code ever changes.  What the reader wants to know is *why* are these checks here.

kod/object/passive/skill/disarm.kod
```diff
@@ -140,7 +140,17 @@ messages:
...
+            % ...the target is a DM (generally, an event target)
+            % ...the target is in an arena
+            % ...the target's weapon doesn't drop on death (soft white light/transcendent)
+            % ...the target's weapon shouldn't leave them for other reasons
+            if (Random(1, 100) < iAbil)
+               AND NOT IsClass(oTarget, &DM)
```
akirmse, line 150: Hm, these kind of checks are fragile.  e.g. if someone ever added a new kind of special room that was like an arena but wasn't IsArena, there would be hundreds of places to edit, which shows that the design isn't right.

So. @ReqLeaveOwner: yes, definitely, this is intended usage.
@IsClass(DM): unpleasant use of IsClass, but I don't think there's currently a better alternative.  It would be better if we had a @CanBeAffectedByMortals or something, but I think we don't.

As for the other two, they're kind of gross, and 4 checks certainly feels like too many.  Maybe there's no other way around it, but can you do some digging and make sure there's not a better existing check?

Aesica: - @IsArena: I didn't find anything more fitting as a substitute for this.  The ROOM_SAFE_DEATH flag is the closest alternative, but that seems even worse.  Plus, after some further digging, IsArena is used similarly by several other spells/skills to restrict behaviors in various ways, so it might be the best option in terms of consistency.  Players have come to accept that arenas change the rules for several spells and abilities already.
- @DropOnDeath, IsClass DM: Would a cleaner approach be to add something like a pbCanDropWeaponOnDisarm property to battler.kod and CanDropWeaponOnDisarm() to battler.kod and itematt.kod?  Normally TRUE for players and itematts, DMs and the Soft-White-Light item attribute would override this with FALSE. (Plus, DMs could set their property to TRUE should they want that functionality to exist for an event or something)

The result would now require this set of checks to all be TRUE for disarm to drop the target's weapon to the ground:  Random 1-100 < iAbil, ReqLeaveOwner, IsArena, CanDropWeaponOnDisarm

Definitely want your thumbs-up or thumbs-down before I proceed with this, though.

PR #557, opened 2023-09-08
Rentable room update 3: Time remaining and other details in key descriptions

Aesica:
Currently, the only way users get notified about how much time is left on their room(s) is through a periodic (and somewhat annoying) mail.  This replaced that with the following changes:

- Room key descriptions updated to indicate the number of meridian days remaining.
- RoomKey objects (master/original keys) in particular also show the number of key copies they have in circulation, if any.
- Additionally, room key description generation in general has been cleaned up a bit.  Building all the extra details in DoBaseDesc was probably not how it should've been done since it interferes with added itemattribute descriptions, so now the extra details are built in AppendDesc.
- When a rented room's remaining days value is below a threshold specified in the RentableRoomMaintenance object, the user is also issued warning tell messages from the landlord NPC every time they log in.
- The periodic mails have been removed because nobody likes spam, however the checkout/expiration mail is still sent.

image

### DIFF
--- a/kod/object/active/holder/room/rentroom.kod
+++ b/kod/object/active/holder/room/rentroom.kod
@@ -88,10 +88,9 @@ resources:
       "could opt to have nothing on the wall."
 
    RentableRoom_expiry_warning = \
-      "Subject: Room check-out reminder notice\n"
-      "This is a courtesy notice to remind you that your room rental will "
-      "expire in %q days.  Please be sure to clear your room of any personal "
-      "effects before then.  I hope you are enjoying your stay."
+      "~k%s tells you, ~I\"Your room is about to expire.  You only have %i "
+      "Meridian days left to pay your rent.\""
+
    RentableRoom_expiry = \
       "Subject: Room check-out\n"
       "Your room rental has expired, and the room has been cleared of any "
@@ -1387,13 +1386,6 @@ messages:
          return;
       }
 
-      % Send warning every 12 Meridian days (2 real days), and when we have 4 hours left
-      if (piDaysLeft <= 2)
-         OR ((piDaysLeft mod 12) = 0)
-      {
-         Send(self,@SendExpiryWarningMail);
-      }
-
       return;
    }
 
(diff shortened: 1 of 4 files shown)


akirmse:
Hmmm.. one email is spam. but now you get a tell every time you log in?  It would be worth asking players whether they want this change.  If applied elsewhere to similar circumstances, I could easily imagine login tells quickly becoming a huge text block that everyone ignores.  At least you can read emails at your leisure.

Aesica:
I should note a couple things about the mails vs the tells:

- The mail was removed in this PR because it's now redundant with the ability to just check the key anytime.  It didn't actively warn them that their room was about to expire.  It was sent out every real life day no matter the remaining duration, and just gave them a report of how much time was left on their room.
- The tells only get sent once the remaining days on a room drop below a certain threshold (can be adjusted by DMs, or disabled completely by setting it to 0), as a warning to get the user's attention and let them know their room's about to expire.  Mail notifications of any time can be easy to miss/disregard due to certain other NPCs mailing excessively (Frisconar, Moxal, and to a lesser extent, Alzahakar)

Aesica:
So I've left a poll about tells vs mails running for awhile, and although it only received 16 replies as of this response, here were the results:

- Tell vs mail when the room is about to expire:  Overwhelmingly in favor of the on-login tell (87.5% vs 12.5%)
- Do people like or dislike npc mails:  Mixed, with a slight bias toward dislike (7 in varying degrees of "fine" vs 9 in varying degrees of "dislike")
- Do people like or dislike npc tells on login:  Overwhelmingly fine with them (14 in varying degrees of "fine" vs 2 in varying degrees of "dislike")

So I think overall, the on-login, tell-based method of expiration warnings will be seen as a fairly positive change to the system, and much harder for people to miss compared to the old mail system.  Additionally, being able to check the key anytime for the days remaining will be a huge boon and should un-clutter people's mailboxes.

akirmse:
OK, that's good enough for me.

PR #558, opened 2023-09-08
Minor tweaks to ring description behavior:  added durability, itematt bugfix, ring of lethargy self-reveal

Aesica:
This PR does a few things related to ring descriptions:

- Adds generral durability condition text to rings' descriptions, if applicable, since currently there's no way to determine how many hits a ring has remaining, short of remembering.
- Fixes a bug where rings with unrevealed item attributes can't have those attributes revealed.
- Changed ring of lethargy to now officially reveals its name/description/color/etc when put on, because it already informally does so the moment a player picks it up.  Until then, it continues to blend in perfectly with beneficial rings.

image

### DIFF
--- a/kod/object/item/passitem/ring.kod
+++ b/kod/object/item/passitem/ring.kod
@@ -22,6 +22,13 @@ resources:
    ring_identified_name_rsc = "generic magical ring"
    ring_identified_description_rsc = "The has some magical effect."
 
+   ring_condition_exc = " is in pristine condition."
+   ring_condition_exc_mended = " is in pristine condition, although it appears to have been weakened somehow."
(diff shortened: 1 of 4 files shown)


kod/object/item/passitem/ring.kod
```diff
@@ -63,14 +78,25 @@ messages:
 
    RevealHiddenAttributes()
    {
+      local i, bDone;
+
+      bDone = FALSE;
+      for i in plItem_Attributes
```
akirmse, line 84: This loop could use a comment.  It's not obvious what's going on here.

Aesica: Done.

PR #560, opened 2023-09-13
Admin-friendly QoL changes to OrnamentalObject

Aesica:
Currently, it's a bit tough and cumbersome for admins to spice up the game with ornamental objects.  This adds a number of features to the OrnamentalObject class to greatly simplify this process:

- DM create ornament:  Adds a dm command to admin.kod and above to place an ornamental object at their current location.
- Adds two signposts to the preset ornament list.
- SetFlickering(value):  Adds a function to allow easier control over an OrnamentalObject's flicker flag.
- SetLookable(value):  Adds a function to set an OrnamentalObject to be viewable with the look command.
- SetCustomDesc(text):  Adds a function to set safe (no dynamic resources) custom descriptions onto OrnamentalObjects.
- ClearCustomDesc():  Adds a function to clear custom descriptions.
- StartDecayTimer(d, h, m, s):  Adds a function to start a decay timer onto an OrnamentalObject, automatically deleting it when the timer expires.  The duration can be entered using any or all of the parameters.  (Just days, just seconds, just hours, minutes and seconds together, etc)
- ClearDecayTimer():  Adds a function to clear the current decay timer, if one is present.
- PlaceAt(what):  Adds a function to easily move the OrnamentalObject to the target object's current room/location.  Generally, this will be the admin setting it up, but it can be other things that exist in rooms as well.

### DIFF
--- a/kod/object/passive/ornobj.kod
+++ b/kod/object/passive/ornobj.kod
@@ -721,9 +721,6 @@ properties:
    % This value overrides the standard resource description when set.
    psCustomDesc = $
 
-   % Optional timer used to delete the object upon expiring.
-   ptDecayTimer = $
-
 messages:
 
    Constructor(type=0, rName=$, rIcon=$, rDesc=$)
@@ -772,56 +769,6 @@ messages:
       }
 
       propagate;
-   }
-
-   StartDecayTimer(days=0, hours=0, minutes=0, seconds=0)
-   "Admin supported\n"
-   "Starts a timer for this object which will automatically delete it upon expiring.\n"
-   "Can specify the time using values for any or all of the parameters.\n"
-   "These parameters are: days, hours, minutes, and/or seconds."
-   {
-      local iTime;
-
-      iTime = (((days * 24 + hours) * 60 + minutes) * 60 + seconds) * 1000;
-      if iTime > 0
-      {
-         if days < 21
-         {
-            Send(self, @ClearDecayTimer);
-            ptDecayTimer = CreateTimer(self, @DecayTimerExpired, iTime);
-         }
-         else
-         {
-            Debug("Time must be 20 or fewer days to avoid undesirable timer results.");
-         }
-      }
-      else
-      {
-         Debug("Timer duration must be greater than zero.");
-      }
-
-      return;
-   }
-
-   ClearDecayTimer()
-   "Admin supported\n"
-   "Clears this object's active decay timer if it exists."
-   {
-      if ptDecayTimer <> $
-      {
-         DeleteTimer(ptDecayTimer);
-      }
-      ptDecayTimer = $;
-
-      return;
-   }
-
-   DecayTimerExpired()
-   {
-      ptDecayTimer = $;
-      Send(self, @Delete);
-
-      return;
    }
 
    PlaceAt(what=$)
@@ -959,9 +906,6 @@ messages:
 
       % Clear the custom description, if any
       Send(self, @ClearCustomDesc);
-
-      % Clear the current decay timer, if any
-      Send(self, @ClearDecayTimer);
 
       if type = OO_3_CLOTH_BOLTS
       {  vrName = oo_3bolts_name;  vrIcon = oo_3bolts_icon;  vrDesc = oo_3bolts_desc;  }
@@ -1438,7 +1382,6 @@ messages:
 
    Delete()
    {
-      Send(self, @ClearDecayTimer);
       Send(self, @ClearCustomDesc);
 
       propagate;
--- a/kod/object/passive/ornobj.kod
+++ b/kod/object/passive/ornobj.kod
@@ -789,7 +789,6 @@ messages:
          {
             Send(self, @ClearDecayTimer);
             ptDecayTimer = CreateTimer(self, @DecayTimerExpired, iTime);
-            Debug("iTime:", iTime, " | ptDecayTime:", GetTimeRemaining(ptDecayTimer));
          }
          else
          {
--- a/kod/object/passive/ornobj.kod
+++ b/kod/object/passive/ornobj.kod
@@ -717,7 +717,11 @@ properties:
    piXlat = $
    piGroup = 1
 
+   % Custom description string to be displayed for the object.
+   % This value overrides the standard resource description when set.
    psCustomDesc = $
+
+   % Optional timer used to delete the object upon expiring.
    ptDecayTimer = $
 
 messages:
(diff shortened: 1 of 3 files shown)


kod/object/passive/ornobj.kod
```diff
@@ -735,14 +749,143 @@ messages:
...
+   }
+
+   StartDecayTimer(d=0, h=0, m=0, s=0)
+   "Admin supported\n"
+   "Starts a timer for this object which will automatically delete it upon expiring.\n"
+   "Can specify the time using h for hours, m for minutes, and/or s for seconds."
+   {
+      local iTime;
+
+      iTime = d * 86400000 + h * 3600000 + m * 60000 + s * 1000;
```
akirmse, line 780: Easier to understand: 1000 * (s + 60 * (m + 60 * (h + 24 * d))).

I'm a bit concerned about rollover here.  2^28 / 86400000 = 3.1, i.e. if you set days to >= 4, the Blakod int value will rollover, and you won't get the time you want.  Should add a check for this.  And it might be only 2^27 depending how signed values work.. please test these large values.

kod/object/passive/ornobj.kod
```diff
@@ -735,14 +749,143 @@ messages:
...
+
+      if (iTime > 0)
+      {
+         ptDecayTimer = CreateTimer(self, @DecayTimerExpired, iTime);
+      }
+      else
+      {
+         Send(self, @ClearDecayTimer);
```
akirmse, line 788: This is a weird behavior.. pass in a negative value and it clears the existing timer?  I think instead, negative values should give an error.

kod/object/passive/ornobj.kod
```diff
@@ -735,14 +749,143 @@ messages:
...
+      {
+         AddPacket(4, oo_custom_template_rsc, 4, psCustomDesc);
+
+         return;
+      }
+
+      propagate;
+   }
+
+   StartDecayTimer(d=0, h=0, m=0, s=0)
```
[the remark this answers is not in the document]
Aesica: Done.

kod/object/passive/ornobj.kod
```diff
@@ -1286,6 +1448,14 @@ messages:
...
          send(oDuplicate,@SetNoMoveOn,#value=TRUE);
       }
+      if Send(self, @GetObjectFlags) & FLICKERING_YES
+      {
+         Send(oDuplicate, @SetFlickering, #value=TRUE);
+      }
+      if NOT Send(self, @GetObjectFlags) & LOOK_NO
+      {
+         Send(oDuplicate, @SetLookable, #value=TRUE);
+      }
```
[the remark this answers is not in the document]
Aesica: I wasn't too sure about even touching this function since it seems like only rented rooms use it, but I can see the value for admins who might want to make, say, multiple signs with the same special description or something for different inns/etc to advertise an event, so added custom description copying.

I'm unsure about timers though since all it would do is copy the other object's remaining time to make its own timer, and I'm not sure how valuable that'd even be.  It's easy enough for me to add though if you think it should be there.

kod/object/passive/ornobj.kod
```diff
@@ -1270,6 +1424,14 @@ messages:
       propagate;
    }
 
+   Delete()
+   {
+      Send(self, @ClearDecayTimer);
+      psCustomDesc = $;
```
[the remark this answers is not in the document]
Aesica: Done.

kod/object/passive/ornobj.kod
```diff
@@ -700,12 +710,16 @@ properties:
...
    vrDesc = $
 
+   pbLook = FALSE
    pbHanging = FALSE
    pbNoMoveOn = FALSE
    pbFlicker = FALSE
    piXlat = $
    piGroup = 1
 
+   psCustomDesc = $
```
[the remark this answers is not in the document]
Aesica: Done.

kod/object/passive/ornobj.kod
Aesica: Done, although I reversed the order to be days ... -> ... seconds for even more clarity.  I'd assumed (big mistake, I don't usually assume) timers could handler large numbers since raw time values can be quite big, but it seems things go south at 25+ days, so I set the cutoff to only allow 20 or fewer.  It's still a very rudimentary check since one could potentially do 20 days, then like 5000 hours and still cause the rollover.

Overall I'm a bit confused about something with this.  When I debug-output iTime after passing 25 days, it shows a negative value, however the iTime > 0 checks still pass?

kod/object/passive/ornobj.kod
Aesica: I agree, not sure why I even did it that way in the first place.  Changed to a debug message.

kod/object/passive/ornobj.kod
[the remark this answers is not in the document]
Aesica: Added note:  I'm not actually sure if I even need to de-reference strings on dispose, so I did it "just in case."

kod/object/passive/ornobj.kod
akirmse: Raw time values are stored as 64-bits in the server, but Blakod values are 28 bits.  (Or at least, they were until recently when I made some server changes.  Did you clean and rebuild the server recently?)  And I see that 2^28 / 86400000 = 3.1, so I think we don't understand exactly what's happening here.  Putting a cutoff at a random 20 days doesn't seem safe until we do.

Aesica: Yeah the server's been rebuilt fairly recently and I keep my repo synced, so everything should be up to date.  Things got weirder with this too--either I was losing my mind and tired, or it seemed that, when I skipped assigning the result to the local variable and instead put the equation directly into CreateTimer(), the days could go to 40 and possibly a bit higher?  I checked this with a Debug(GetTimeRemaining(timerHere)) after creating it and the output was an ever-increasing value all the way to 40, then I skipped to 50 and it failed.

Overall I don't feel confident in how this is behaving (results vs expectations) and I'm also not convinced this particular feature is even worth the trouble.  So it's nuked.

kod/object/passive/ornobj.kod
```diff
@@ -1286,6 +1403,20 @@ messages:
       {
          send(oDuplicate,@SetNoMoveOn,#value=TRUE);
       }
+      if Send(self, @GetObjectFlags) & FLICKERING_YES
+      {
+         Send(oDuplicate, @SetFlickering, #value=TRUE);
+      }
+      if NOT Send(self, @GetObjectFlags) & LOOK_NO
```
akirmse, line 1410: I think this needs parentheses.  NOT is higher priority than &.

kod/object/passive/ornobj.kod
```diff
@@ -735,14 +750,101 @@ messages:
...
+                  #fine_row=Nth(lPos, 3), #fine_col=Nth(lPos, 4),
+                  #new_angle=Send(what, @GetAngle));
+            }
+         }
+      }
+
+      return;
+   }
+
+   SetCustomDesc(text=$)
```
akirmse, line 806: Given that you found that dynamic strings are never garbage collected, do you think this is safe?  I'm not sure how often it will be used.

Aesica: I'm pretty confident in this, actually.  It's the same technique used by ballots, books, and various other items that allow users to write in them, all of which get used with more frequency than this probably will due to it being admin-only.  The testing I did showed that strings do seem to be garbage collected after a save once they were no longer associated with something.  These were my test steps, in case you're curious:

- Set custom description for 2 ornamental objects.
- Save to garbage collect/compact references.
- Check and note both string ids (string 5439 and 5440, as well as 5438 which is unrelated).
- Remove custom description for the object using 5439.
- Save again.
- Note that string 5440 has moved down to 5439, there is nothing in 5440, and 5438 is what it was unchanged.  So the original 5439 was garbage collected successfully due to no longer being used.

kod/object/passive/ornobj.kod
Aesica: Done.

PR #561, opened 2023-09-13
Allow advancement cap settings to be customized or disabled by admins

Aesica:
Currently the advancement cap (an antiquated anti-botting mechanic) is hardcoded and cannot be changed without editing player.kod.  This change allows for an admin to adjust the following aspects of the advancement cap:

- Min duration of advancement cap timer.
- Max duration of advancement cap timer.
- Number of improves allowed before being capped.
- Option to enable or disable advancement caps entirely.

Note:  This new setup disables the advancement cap by default for the following reasons:

- It confuses newbies who can't figure out why they're suddenly not improving anymore.
- It forces people who aren't botting to do weird rituals (run in and out of a room several times after 10 improves) in order to continue spell or skill practice.
- It doesn't actually deter botting.  Since the values are known, anyone determined to bot will just adjust their bot to stagger casts.
- Now that overall advancement is faster, and since PEs can be recast over and over, botting isn't even worth doing anymore in all but a few cases.

PS:  I tested this by:

- Cranking up the improvement rate on my test server to an absurd number, so that improves happened nearly every cast.
- Casting a spell with it off all the way from 11% to 50% in only a few minutes.
- Checking to make sure the test player's piAdvancement_points were 0 and that there was no ptAdvancement timer set.
- Casting several more times with the advancement cap on, and noting that now the player accumulates piAdvancement_points and that their ptAdvancement timer exists, was counting down, and that they ceased to improve after 10 casts (the current setting)
- Turning the advancement cap back off and noting that the player can improve again and that they do not accumulate further piAdvancement_points.  Their existing points/timer expire normally without issue with the advancement cap off, but have no effect on anything.

Added note:  The advancement cap timer is tied in with some weird old skill/spell atrophy system that I've never seen active in all my years of playing.  I don't know if it even works anymore, but if it does, it most likely requires the advancement cap system to be on in order to function properly.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -62,11 +62,6 @@ constants:
    % Number of ms to lose one boosted health or mana point.
    BOOST_DECAY_TIME = 30000 
 
-   % Currently: 10 improves every 15-22 minutes.
-   ADVANCE_TIMER_MIN =   900000
-   ADVANCE_TIMER_MAX =   1320000
-   ADVANCEMENT_LIMIT =   10
-
    PKPOINTER_TIME = 10*60*1000
 
    PLAYER_TRANSLATION_LEGS_MASK = 0xff
@@ -7450,15 +7445,18 @@ messages:
    "A player can only gain so many advancement points per hour, in spells"
    "And skills combined."
    {
-      local iTime;
+      local iTime, oSettings;
       
-      if ptAdvancement = $
+      oSettings = Send(SYS, @GetSettings);
+      if Send(oSettings, @IsAdvancementCapOn)
       {
-         iTime = random(ADVANCE_TIMER_MIN,ADVANCE_TIMER_MAX);
-         ptAdvancement = CreateTimer(self,@AdvancementTimer,iTime);
+         if ptAdvancement = $
+         {
+            iTime = Send(Send(SYS, @GetSettings), @GetAdvancementWait);
+            ptAdvancement = CreateTimer(self,@AdvancementTimer,iTime);
+         }
+         piAdvancement_points = piAdvancement_points + number;
       }
-      
-      piAdvancement_points = piAdvancement_points + number;
       
       return;
    }
@@ -7466,7 +7464,11 @@ messages:
    CheckAdvancementPoints()
    "Checks to be sure a player isn't at the limit yet."
    {
-      if piAdvancement_points < ADVANCEMENT_LIMIT
+      local oSettings;
+
+      oSettings = Send(SYS, @GetSettings);
+      if NOT Send(oSettings, @IsAdvancementCapOn) OR
+         piAdvancement_points < Send(oSettings, @GetAdvancementLimit)
       {
          return TRUE;
       }
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -7452,7 +7452,7 @@ messages:
       {
          if ptAdvancement = $
          {
-            iTime = Send(Send(SYS, @GetSettings), @GetAdvancementWait);
+            iTime = Send(oSettings, @GetAdvancementWait);
             ptAdvancement = CreateTimer(self,@AdvancementTimer,iTime);
          }
          piAdvancement_points = piAdvancement_points + number;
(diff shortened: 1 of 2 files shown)


CardiovascularMike:
There are two other ways to add advancement points we need to consider.  These aren't show-stoppers though.

Buying a new skill adds 4 advancement points per skill.  Buying a new spell adds 2 advancement points per spell.  With the change as it is currently drafted I believe the player's piAdvancement_points will increment up to some reasonably large integer (for example, a new player buys skills up to level three and gets to kranaan 3, they will hvae a piAdvancement_points of 70).  This isn't a problem in it's own right.  If the timers and limiters get turned on they will have to wait 15-22 minutes before they can get any improvements.  

A bad actor could try using forget potions and buying level 1 WC over and over again but our max integer is high enough it would take a year of work to even potentially cause any issues.

All that said it may be worth looking at these other times piAdvancement_points are incremented and setting those up to check the server settings.

CardiovascularMike:
> This seems like a major change that I guess will have the effect of making the game substantially easier? Need some feedback on this one from more people.

The game would only be easier for new players.  Experienced players know the system and work around it by zoning in and out of rooms a bunch.  Every time you zone into a room where fighting is allowed your current piAdvancement_points is reduced by 2.

The way I and most experienced players do this is literally we count improvements to 10, then zone 2-3 times, then start counting improvements again.  New players instead get 10 fast improvements, then beat their heads against the wall for 22 minutes, then get 10 more improvements.

akirmse:
Yes, that's what I meant by easier: people will advance faster.  Beating
heads against the wall sounds bad, but there could also be negative
consequences to people reaching the endgame sooner and getting bored.  As
an example, one might try to preserve the total "game length" by slowing
the improvement rate to balance out this change.  Or not, but it's worth
discussing.

Aesica:
> There are two other ways to add advancement points we need to consider. These aren't show-stoppers though.
> 
> Buying a new skill adds 4 advancement points per skill. Buying a new spell adds 2 advancement points per spell. With the change as it is currently drafted I believe the player's piAdvancement_points will increment up to some reasonably large integer (for example, a new player buys skills up to level three and gets to kranaan 3, they will hvae a piAdvancement_points of 70). This isn't a problem in it's own right. If the timers and limiters get turned on they will have to wait 15-22 minutes before they can get any improvements.
> 
> A bad actor could try using forget potions and buying level 1 WC over and over again but our max integer is high enough it would take a million repetitions to be a problem.
> 
> All that said it may be worth looking at these other times piAdvancement_points are incremented and setting those up to check the server settings.

All of those increases are done through the AddAdvancementPoints function, and if pbAdvancementCapOn is FALSE (disabled) the logic that increments piAdvancement_points is skipped entirely.  Also worth noting that the room transition logic that decrements piAdvancement_points down to 0 is still in effect AND the timer that inevitably resets piAdvancement_points is still counting down if the player has any advancement points prior to the setting being disabled.  It will do its job when it expires and reset their points, even if the advancement cap is disabled.  (no new points will be accumulate and no new timer will be started while it's off, however)  In short, our potential bad actor would have an easier time breaking the game through shilling accumulation than through piAdvancement_point accumulation, even in extreme circumstances, such as advancement cap on, massively long reset timer.


> This seems like a major change that I guess will have the effect of making the game substantially easier? Need some feedback on this one from more people.

As an active player, I don't think it'd actually make that game all that much easier.  Currently, with the cap in place, I just count my improves to 10, then run in/out of the room I'm in to reset the timer when I hit 10, then continue as I was until I have 10 more improves.  Having the advancement cap off would just mean I (and others) don't need to do the room in-and-out dance every 10 improves.  

Plus, it's worth noting that this change makes the whole thing admin-adjustable, so if it does turn out to cause some sort of balance upset, it can be flipped back on and the times and number of improves can be adjusted on the fly via the settings object.  It's also worth noting that the overall improve rate is currently adjustable and can be adjusted if needed.  Players also have additional limiters in place (which are fine) such as swing timers (chance to improve only after X swings) for weapon skills, mana/vigor downtime, reagent costs, etc.

Aesica:
After careful consideration, I think it's best to abandon this PR for now, as there seems to be two very different schools of thought on the topic of improvement caps.  While I don't think it's a healthy mechanic to have in a game like this, it seems to be seen as a sort of stopgap against certain spells or skills that improve too quickly.  Perhaps when improve rates themselves receive some sort of balance pass (some *do* improve too quickly, I'll agree) the option of just eliminating the improvement cap entirely can be explored.

Since that time isn't right now, I'm going to close this.

PR #562, opened 2023-09-13
Restorate update

CardiovascularMike:
This PR addresses a gameplay issue where the Jala spell Restorate has no effect for players in the instances where they are most likely to use the spell.  The reason for this is that the HP regeneration rate has a strict lower bound of 1000ms per tick.  Most players reach this lower limit without Restorate being active and therefore the spell has no effect.  For reference: a player with 45 stamina at 100hp or higher hits the limit at 162 vigor.

To solve this problem we lower the lower bound for HP regeneration if the spell Restorate is active.

This solution does create another small gameplay problem: Any reasonable lower limit for HP tick rates makes it so that the maximum effect of the spell tops out at a low spell power.  We solve this problem by reducing the spell power slope and introducing a small y-axis offset to the spell power line.  This means that the spell still has some marginal effect at very low spell powers and this effect gradually scales up to a spell power of 80% where we begin to flirt with the lower limit for HP regeneration tick-rates.

I've attached some plots to show the problem and solution since there are a number of design constraints to balance.
image
Here we want to select a lower limit that allows for spell power to matter as much as possible.  I've chosen 670 ms as the lower limit since spell power seems to matter up to about 80% and since this means at maximum spell power our players will gain about 50% more HP per unit of time than the otherwise would.  Notice that the chart includes the current lower limit of 1,000 ms at which point the spell has no effect at any spell power for our reference player (red rectangle).

image
Here the new spell power curve (expressed as a ratio of HP regeneration to baseline HP regeneration) is plotted with the proposed lower limit as a function of spell power.

image
Finally we plot the spell power slope in terms of ms per HP tick.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5511,7 +5511,7 @@ messages:
          }
       }
 
-      return bound(iTime,1000,40000);
+      return bound(iTime,1000,25000);
    }
 
    CalculateManaTime()
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5506,7 +5506,7 @@ messages:
                iTime = Send(oJalaSpell,@AdjustHealthTime,#time=iTime,
                             #iSpellPower=iJalaPower);
 
-               return bound(iTime,750,25000);
+               return bound(iTime,670,25000);
             }
          }
       }
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5506,7 +5506,7 @@ messages:
                iTime = Send(oJalaSpell,@AdjustHealthTime,#time=iTime,
                             #iSpellPower=iJalaPower);
 
-               return bound(iTime,625,25000);
+               return bound(iTime,700,25000);
             }
          }
       }
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5501,12 +5501,12 @@ messages:
             if IsClass(oJalaSpell,&Restorate)
             {
                iJalaPower = Nth(Nth(lJalaInfo,3),3);
-               
+
                iTime = bound(iTime,1000,40000);
                iTime = Send(oJalaSpell,@AdjustHealthTime,#time=iTime,
                             #iSpellPower=iJalaPower);
 
-               return = bound(iTime,650,40000);
+               return bound(iTime,650,40000);
             }
          }
       }
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5502,11 +5502,11 @@ messages:
             {
                iJalaPower = Nth(Nth(lJalaInfo,3),3);
 
-               iTime = bound(iTime,1000,40000);
+               iTime = bound(iTime,1000,25000);
                iTime = Send(oJalaSpell,@AdjustHealthTime,#time=iTime,
                             #iSpellPower=iJalaPower);
 
-               return bound(iTime,625,40000);
+               return bound(iTime,625,25000);
             }
          }
       }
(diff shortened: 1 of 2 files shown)


akirmse:
I have no idea if this improves balance.  We'll need feedback from people who actually play.

There sure are a lot of magic numbers flying around here.  Make some into symbolic constants?

CardiovascularMike:
> I have no idea if this improves balance. We'll need feedback from people who actually play.
> 
> There sure are a lot of magic numbers flying around here. Make some into symbolic constants?

In terms of magic numbers the 290 and 320 are admittedly that.  I tried to outline the design constraints in the original PR comment.  Essentially we want: 1) The spell to have an effect, 2) spell power to matter, and 3) the effect to be reasonably useful and balanced.

The 320 is the spell power slope for Restorate - every 1 increase in spell power reduces the time to earn an HP by a fraction of 1/320.  The old value of 200 was very steep - at max power players were getting double the HP per unit of time.  If we go back to this we would want to make the lower limit per HP tick 500-600ms.  I'm not sure a change that drastic is wise in terms of gameplay balance.

The 670 lower limit number looks like a magic number but it is simply a convenient round up from 667ms which would mean at max spell power for most players this spell allows you to gain HP at 1.5 times the normal rate.

In terms of symbolic constants are you suggesting something like piRestorateSlope = 300 and piRestorateOffset = 30 and reference that in Restorate.kod?  I can do that easily of course.

akirmse:
No, those are properties, I mean make them into constants.  Giving them a
name helps document what they mean.  (No one reading the code in the future
will go back and read the PR.)

kod/object/passive/spell/jala/restorat.kod
```diff
@@ -37,6 +37,8 @@ classvars:
    viMana = 5           % Mana is amount used upon inititiation
    viManaDrain = 3      % Drain is amount used every viDrainTime milliseconds
    viDrainTime = 5000   % Drain some mana every viDrainTime milliseconds
+   viSpellPowerIntercept = 40    % Hypothetical zero spell power restorate effect coefficient
+   viSpellPowerSlope = 400       % 1.12 to 1.51 times faster HP regen, linear effect on time per tick
```
akirmse, line 41: Please put constants in the "constants" section, not classvars.

CardiovascularMike: Sorry for confusion.  Done.  I copied constants formatting from another file.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -5501,13 +5501,17 @@ messages:
...
                iTime = Send(oJalaSpell,@AdjustHealthTime,#time=iTime,
                             #iSpellPower=iJalaPower);
+
+               return bound(iTime,670,25000);
             }
          }
       }
 
-      return bound(iTime,1000,60000);
+      return bound(iTime,1000,25000);
```
akirmse, line 5514: I see your comment about changing this line, but since it isn't related to Restorate, I'd like to see it reverted, and you can make a separate PR to discuss it if you want.

CardiovascularMike: Ok.  Upper limits for both with and without the Jala spell are returned to 60 seconds.  We can discuss fixing this in a separate PR.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -5501,8 +5501,12 @@ messages:
             if IsClass(oJalaSpell,&Restorate)
             {
                iJalaPower = Nth(Nth(lJalaInfo,3),3);
+
+               iTime = bound(iTime,1000,60000);
                iTime = Send(oJalaSpell,@AdjustHealthTime,#time=iTime,
                             #iSpellPower=iJalaPower);
+
+               return bound(iTime,670,60000);
```
akirmse, line 5509: Can these changes to player.kod be moved into Restorate instead?  Player is already enormous so it's good to keep new code out of it.  Plus these additional magic numbers seem to be specific to Restorate.

CardiovascularMike: I pushed both bounding statements over to Restorate.kod.  We can do this in fewer lines but I think this is readable.

I thought about going for return bound(Send(oJalaSpell,@AdjustHealthTime,#time=iTime,#iSpellPower=iJalaPower); or even bounding the iTime argument in the function call but I think clever code can be problematic down the road.  Although _return bound(send_ is used in one other spot in the database.

PR #563, opened 2023-09-14
Fix flip signs and floating lines in Barloque North and South (barln.roo and barls.roo)

Dirhai:
fffff
wwwww
sign1
sign2


akirmse:
This change looks empty?

I think it should be possible to have no transparent artifacts here though.

Dirhai:
Yes it really looks empty, but atleast easier to hide, and no more floating line. I will look again tomorrow those room, the only thing i can do to hide it using the room editor is creating another wall infront of the gap🤷

akirmse:
What I mean is, there appear to be no files in this PR.

Dirhai:
Sorry, i think the files are there now.

akirmse:
Hm, I don't know about this... I feel like a more complete fix is possible with more room editor experience.

Dirhai:
Makes sense i will return to those rooms later, i dont like to creating new sectors and walls to avoid new bugs, i feel like the room editor are very sensitive lol

davidklaw:
These floating lines can be resolved using the new "no horizontal tiling" option added in #583. Sorry @Dirhai, but I think this can be closed _or_ updated using the new room editor capabilities.

akirmse:
I'm going to close this, since the new flag is a better way to fix this.

PR #565, opened 2023-09-20
Update Containers: Withdraw selectable amount of number items and hierarchical sort

CardiovascularMike:
This PR updates storage boxes like guild hall chests, personal chests in rooms, and other chests that are available throughout the game (IE: Thrashers).

The goal is to make managing inventory easier for players. To that end we make two major changes:

We allow players to pickup a user-selectable number of items for number items
We sort all items in the list box. The sort is two steps: First number items are put at the top and sorted alphabetically then not-number items are sorted alphabetically against each other and appended to the bottom of the list.

This PR differs from draft PR #554 in an important way: We are looping over the list of selected items in C and sending the items to User.Kod one at a time.  This allows the server message to call UserGet and doesn't require a new message in User for getting items from a container.  A number argument is added to UserGet which can be NULL if needed.

Testing was done in RID_GUILDH9 and included:
- Withdrawing combinations of number items and not number items
- Grabbing combinations of items and number items from the ground
- Testing both situations where the user inventory is nearly full
- Testing both situations where the user inventory is completely full
- Opening the get menu and selecting items but removing those items with another user before the user can get the items

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -962,8 +962,16 @@ messages:
       if liClient_cmd = BP_REQ_GET
       {
          oWhat = Nth(client_msg,2);
-         Send(self,@UserGet,#what=oWhat);
+         Send(self,@UserGet,#what=oWhat,#number=$);
          
+         return;
+      }
+
+      if liClient_cmd = BP_REQ_GET_FROM_CONTAINER
+      {
+         oWhat = Nth(client_msg,2);
+         Send(self,@UserGet,#what=oWhat,#number=number_stuff);
+
          return;
       }
       
@@ -3537,10 +3545,10 @@ messages:
       return;
    }
 
-   UserGet(what = $)
+   UserGet(what = $, number = $)
    {
       local oOldOwner, i, lItem_pos, iRow, iCol, iRow_dist, iCol_dist,
-            iCan_hold, oSplit, oCorpse;
+            iCan_hold, iNumber, oSplit, oCorpse;
 
       % User took an action!  Wake any AIs in the room to the user's presence!
       % bitflag check instead of function for speed.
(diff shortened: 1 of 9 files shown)


kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3537,10 +3545,10 @@ messages:
       return;
    }
 
-   UserGet(what = $)
+   UserGet(what = $, number = $)
```
akirmse, line 3548: This number parameter needs to be documented since it's not obvious what it does.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3627,26 +3635,37 @@ messages:
               
          return;
       }
-        
-      if NOT Send(self,@ReqNewHold,#what=what) 
+      % Either we pass in a number or we didnt but cant hold it all
```
akirmse, line 3638: didn't, can't

akirmse (commented):
Yes, I definitely like this approach better.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: Done with doc string.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: Done.

CardiovascularMike:
I just wanted to update that I ran a mortal through the thrashers dungeon to make sure these changes don't adversely effect the loot generation in the chest that is there.  The test went smoothly and the sorted contents worked fine.

CardiovascularMike:
Improved the comment on what may be the most opaque part of the changes to User.Kod.  Now explains what iCan_hold = $ means.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3627,26 +3637,38 @@ messages:
...
+      % Either we pass in a number or we didn't but can't hold it all
+      if number <> $
+         OR NOT Send(self,@ReqNewHold,#what=what)
       {
          if IsClass(what,&NumberItem)
          {
             iCan_hold = Send(self,@GetNumberCanHold,#what=what);
-            if iCan_hold > 0
+            % iCan_hold of $ intended to signal infinite capacity
+            % Therefor set iCan_hold to number of items requested
```
akirmse, line 3648: "Therefore"

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -962,10 +962,18 @@ messages:
       if liClient_cmd = BP_REQ_GET
       {
          oWhat = Nth(client_msg,2);
-         Send(self,@UserGet,#what=oWhat);
+         Send(self,@UserGet,#what=oWhat,#number=$);
```
akirmse, line 965: This change isn't necessary.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3627,26 +3637,38 @@ messages:
               
          return;
       }
-        
-      if NOT Send(self,@ReqNewHold,#what=what) 
+      % Either we pass in a number or we didn't but can't hold it all
+      if number <> $
+         OR NOT Send(self,@ReqNewHold,#what=what)
```
akirmse, line 3642: Is there a security problem here?  I'm looking for where we check that the number received from the client is at most the number that actually exist.  It might fail at the call to Split with a bunch of errors, but it would be much better to have an explicit check.

CardiovascularMike: So I did test something along these lines: the game doesn't let you enter a number greater than the available quantity.  I am not sure where that code lives and will find it and see how robust it is.  

It may be possible that a bad actor could hack the client to circumvent this.  I will look into the Vaultman code and see if I can find some method for checking that the number of items exist as a secondary check.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: Done.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: Removed this change as it is not needed.

kod/object/active/holder/nomoveon/battler/player/user.kod
CardiovascularMike: Added an explicit check for if Split() returns $.  This also checks if somehow the user enters a number larger than the number of items available in the container.

PR #568, opened 2023-09-29
Ambient sound volume preference and music fast exit

sicoderealm:
This contains two main improvements/adjustments

- Looping ambient sound (such as the Tos fountain sounds) volume can now be controlled via a new preferences slider (just like sounds and music)

- We also stop music from playing more swiftly when closing the game client.


clientd3d/statgame.c
```diff
@@ -72,9 +72,10 @@ void GameExit(void)
    // Can't close modules here, because WM_DESTROY messages sent to dialogs causes crash.
    // Modules are freed elsewhere.
 
-   SoundClose();
    MusicAbort();
-
+   SoundClose();
+   MusicClose();
```
akirmse, line 77: I'm curious about this change.  It's not called only when exiting the client, but also when exiting back to the main menu.  With this change, do sounds in progress continue instead?

akirmse:
I don't know what that Sleep is doing there---that's terrible.  Getting rid of it would be nice but I'm sure it's hiding some other bug, so we should look at that separately.

I will assume you've tested this thoroughly in all cases: retail/non-retail, process exit vs. return to main menu, sound/music enabled vs. not enabled, etc.

PR #569, opened 2023-09-30
SpellItem bugfixes

Aesica:
Currently, the SpellItem subclasses, particularly potions, have some broken logic that causes them to expend charges vs invalid targets, fail due to karma when karma is flagged to not matter, or even not work at all.

- purify.kod:  Fixed some bad target logic that caused purify potions to fail 100% of the time in all situations when the purify spell was set to self-cast only.
- spell.kod:  Removed a weird check for made items (distill) vs normal potions that was causing normal potions to always consider the user's karma, even when flagged not to (remove curse potion, etc)
- spelitem.kod, potion.kod:  Fixed some buggy logic that caused spellitem charges to be consumed whether or not their effects could be utilized.  (Drinking a remove curse potion without a cursed item equipped, etc)

### DIFF
--- a/kod/object/item/passitem/spelitem.kod
+++ b/kod/object/item/passitem/spelitem.kod
@@ -193,7 +193,8 @@ messages:
    }
    
    CastSpell(what = $, apply_on = $)
-   "Casts the spell set by the spell object"
+   "Casts the spell set by the spell object, returning TRUE\n"
+   "If successful, or FALSE if the spell failed for whatever reason."
    {
       local oSpell,iSpellPower,iRand,iInt,lTargets;
 
@@ -206,18 +207,16 @@ messages:
          AND IsClass(apply_on,&Monster)
          AND NOT Send(apply_on,@CanMonsterFight,#who=poOwner,#oStroke=self)
       {
-         return;
+         return FALSE;
       }
 
       % Karma check?
       if pbCheckKarma AND NOT Send(oSpell, @KarmaCheck, #who = poOwner)
       {
-         Debug("user karma = ", Send(poOwner, @GetKarma), " spell needs = ",
-             Send(oSpell, @GetLevel));
          Post(what, @MsgSendUser, #message_rsc = SpellItem_bad_karma,
               #parm1 = Send(self, @GetDef),
               #parm2 = Send(self, @GetName));
-         return;
+         return FALSE;
       }
 
       if piSpellPower = $
@@ -228,10 +227,11 @@ messages:
       {
          iSpellpower = piSpellpower;
       }
-      
+
       if (vbFailSafe)
          OR ( (Send(oSpell,@CanPayCosts,#who=poOwner,#lTargets=lTargets,
-                    #iSpellPower=iSpellPower,#bItemCast=TRUE,#bMade=pbMade))
+                    #iSpellPower=iSpellPower,#bItemCast=TRUE,
+                    #bCheckKarma=pbCheckKarma))
               AND (Send(Send(what,@GetOwner),@ReqSpellCast,#who=what,#oSpell=oSpell))
               AND ((NOT Send(oSpell,@IsHarmful))
                    OR Send(Send(what,@GetOwner),@ReqSomethingAttack,#what=what,
@@ -243,9 +243,11 @@ messages:
       else 
       {
          Send(self,@DoFailure,#what=what,#lTargets=lTargets);
+
+         return FALSE;
       }
 
-      return;
+      return TRUE;
    }
 
    DoFailure(what = $,lTargets=$)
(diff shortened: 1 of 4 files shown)


kod/object/item/passitem/spelitem.kod
```diff
@@ -193,7 +193,8 @@ messages:
    }
    
    CastSpell(what = $, apply_on = $)
-   "Casts the spell set by the spell object"
+   "Casts the spell set by the spell object, returning TRUE\n"
+   "If successful, or FALSE if the spell failed for whatever reason."
```
akirmse, line 197: formatting is weird here: I is capital and there's no space before it.

kod/object/passive/spell.kod
```diff
@@ -795,29 +795,24 @@ messages:
       }
 
       % Moved from CanPayManaVigor, because it doesn't relate to paying vigor 
-      %  or mana!  Spells cast from items should follow Karma restrictions
-
-      % If the spell is being cast by a made item then there are no Karma
-      %  restrictions.
-      if NOT (bMade OR Send(self,@KarmaCheck,#who=who))
+      % or mana!  Spells cast from items should follow Karma restrictions
```
akirmse, line 798: This seems like a pretty big change!  Are people OK with it?

kod/object/passive/spell/purify.kod
```diff
@@ -79,20 +79,25 @@ messages:
...
       if NOT pbCan_Cast_on_others
       {
-         lTargets = Cons(who, lTargets);
+         target = who;
       }
       
-      % Can cast spell if the 1 target item is a user
-      if Length(lTargets) <> 1
+      % If there's no target by now, something weird is going on.
+      if target = $
```
akirmse, line 96: If this is a coding error (i.e. it should "never happen"), there should be a debug print here.

kod/object/passive/spell.kod
```diff
@@ -687,7 +687,7 @@ messages:
    }
 
    CanPayCosts(who = $, lTargets = $, iSpellPower = 0, bItemCast = FALSE,
-               bMade = FALSE)
+               bCheckKarma = TRUE)
```
akirmse, line 690: Document what bCheckKarma does.

kod/object/item/passitem/spelitem/potion.kod
```diff
@@ -135,13 +142,13 @@ messages:
                  #duration=(POISON_DURATION*Random(8,12)/10));
          }
 
-         send(self,@Delete);
```
akirmse, line 138: Potion no longer is deleted after using it?

akirmse (commented):
In the future, please make unrelated changes into separate PRs.  This is a mishmash of different things that would be impossible to roll back individually.

kod/object/passive/spell.kod
Aesica: It's not as big as you might think, given the way things behave on live.  Currently on live, spellitems have the following behaviors in relation to karma:

- All spellitems not created by distill care about karma, whether they are flagged to or not.  This is a major reason why remove curse, cure poison, and cure disease potions bought from NPCs or dropped by monsters fail when the expectation is that they shouldn't.
- Rescue potions are the only spellitem actually flagged to take karma into consideration for some reason.  (A line of text is appended to the description of items flagged with a karma requirement, so this behavior is transparent to players.)  Personally, I think more of these items should be flagged for karma restrictions and would be happy to adjust that in a separate PR.

So in short, all this change really does is allow spellitems to start working correctly, as dropped or bought potions will behave as expected.  The check for made (distilled) items honestly didn't make a lot of sense anyway.

kod/object/item/passitem/spelitem/potion.kod
Aesica: It gets deleted when its piHits value reaches 0, which is handled in the parent class's OnSuccessfulApply

kod/object/passive/spell.kod
Aesica: Done.

kod/object/passive/spell/purify.kod
Aesica: Yeah, it shouldn't happen unless something somehow is trying to cast the spell in an unconventional way.  Done.

kod/object/item/passitem/spelitem.kod
Aesica: My bad, all fixed.

Aesica:
> In the future, please make unrelated changes into separate PRs. This is a mishmash of different things that would be impossible to roll back individually.

Sorry about that.  The original intent was to fix karma checks not working right with spellitems (potions primarily) but I had no idea these things were so broken on so many levels.  If you prefer, the purify potion specific malfunctions and the prevent-consumption-without-use bits could be split off into separate PRs, but they all felt connected enough that I included them.

PR #570, opened 2023-10-03
Removing dead code: SoundClose

sicoderealm:
SoundClose() is not used and is now dead code - while unused this also remove the Sleep(1500);


akirmse:
Please also remove it from sound.h.

PR #571, opened 2023-10-05
Shaman update: Varied spells by clan/faction, increased difficulty, and removal of shatter

Aesica:
Currently, avar shamans are just an annoying pest that only serve to make the zones they inhabit a pain to do anything in.  They're nonthreatening overall, but they'll shatter equipment which...frankly sucks and nobody likes them because of it.  Their existence pretty much ruins most of the island zones.

This PR updates them to be less annoying and more interesting and formidable.  Their faction/clan now determines their spell lists.  (none of which includes shatter)

Peet-seeeep (promagic):
- Increased mana pool
- Increased cast frequency
- Increased mana regen time
- Can cast fireball, lightning bolt, heat, and earthquake

Bone faction:
- Slightly increased mana pool
- Slightly increased cast frequency
- Can cast hold, apparition, darkness, and sandstorm

Kyip-kyip-kreeet (antimagic):
- Reduced mana regen time
- Increased speed (tougher combatant)
- Can cast dement, forget, dazzle, and antimagic aura

All:
- Increased level to 130, up from 100
- Known spells are stored in a list which can be edited by creative DMs, although some spells (such as evil twin) don't work right when cast by monsters.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/avshaman.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/avshaman.kod
@@ -14,9 +14,7 @@ constants:
 
    include blakston.khd
 
-   SPELL_CHANCE = 12
    SPELL_RANGE_SQUARED = 666
-   MANA_REGEN_TIME = 1000
 
    ANIM_CAST = 2     %% cast spell animation.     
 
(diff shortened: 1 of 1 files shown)


Aesica:
I'm closing this because I believe the overall framework would be extremely beneficial in monster.kod rather than just the shaman class.  Thus, going forward, any new monsters with spellcasting ability could be implemented by just building a spell list and setting their mana/regen rates.  Currently there's quite a bit involved in setting up spellcasting monsters, and a lot of it is repeat code.

When (and if) that gets approved, I'll open another version of this which will take advantage of that framework.

PR #572, opened 2023-10-05
Rentable room update 4:  Secure key copy option

Aesica:
Silver room keys (key copies) are nice because you can pass them out and share your room with friends, alts, mules, etc, however the risk of losing these keys, either on purpose or accident, exists.

This PR adds an alternate type of key copy option, the bronze room key, which the owner of the room can drop or trade freely, but anyone else cannot drop it, nor can they give it to anyone who isn't the owner or the innkeeper.  

image

Edit:  Image updated to reflect latest changes

### DIFF
--- a/kod/util/rntmaint.kod
+++ b/kod/util/rntmaint.kod
@@ -34,6 +34,7 @@ resources:
    RoomMaintenance_command_rent = "rent"
    RoomMaintenance_command_cost = "cost"
    RoomMaintenance_command_copy_key = "copy key"
+   RoomMaintenance_command_secure = "secure"
    RoomMaintenance_command_change_lock = "change lock"
 
    RoomMaintenance_innkeeper_say_room_rented = \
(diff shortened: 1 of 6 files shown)


kod/util/rntmaint.kod
```diff
@@ -263,7 +265,7 @@ messages:
       % Is the item a key copy?  If so, destroy it.
       for oItem in itemList
       {
-         if GetClass(oItem) = &RoomKeyCopy
+         if IsClass(oItem, &RoomKeyCopy) AND Send(oItem, @IsKeyCopy)
```
akirmse, line 268: This IsKeyCopy looks pretty suspicious.  Is it not redundant with the IsClass check?

akirmse (commented):
There are almost no comments in this PR.  Given that "RoomKey", "RoomKeyCopy", and "RoomKeySecureCopy" are super confusing, a full explanation is needed in the code.

kod/util/rntmaint.kod
Aesica: Yeah it's a bit weird at a glance, but due to the way that the key classes were originally set up, RoomKeyCopy is the superclass of RoomKey and now RoomKeySecureCopy, so it'd be true for all 3 key types when it should be false for RoomKey.  So basically, that check is asking "are you a room key of any kind?  Okay great, now I can ask if you're a room key copy without causing errors."

Aesica:
I've added several comments to functions and other parts.  If there's anything in particular you'd like to see comments for that I missed, let me know and I'll make it happen.

Aesica:
Originally I was going to just make silver keys function this way, and while most I ran it by loved it, a few others didn't like the idea of it being "more work" to pass keys between alts.  (needing to use the offer window to return it to the owner vs just dropping the key to pass it between alts/mules/etc)  I guess in short, the bronze key is for peace of mind, and the silver key is for convenience.

PR #573, opened 2023-10-14
Rentable room invalid mail fix

Aesica:
This adds a check for the expiry warning resource (now intended to be a whisper and thus not formatted for user mails) to CleanBadMail in user.kod.  It should allow CleanBadMail to effectively remove all of the invalid innkeeper expiry warning mails while leaving any others untouched.

After being run once and saving, this change can be safely reverted as it will no longer be relevant.

Send o 0 CleanAllBadMail should allow it to run on every single user in one go.

Testing:  I tested this on a character with an invalid expiry mail plus a mail from another user, as well as a mail from the guardian angel--all unread.  It deleted the expiry mail while leaving the others untouched.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -7839,7 +7839,7 @@ messages:
             j = 1;
             while j <= Length(stuff)
             {
-               if (Nth(stuff,j+1) = $)
+               if (Nth(stuff,j+1) = $) OR Nth(stuff, j + 1) = RentableRoom_expiry_warning
 	       {
 	          Debug("found bad mail",self);
 		  plNew_mail = DelListElem(plNew_mail, lNew_mail);
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -7837,9 +7837,13 @@ messages:
 	 if (Length(stuff) > 1)
 	 {
             j = 1;
+            % Note:  The line containing a check for RentableRoom_expiry_warning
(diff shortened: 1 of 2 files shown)


kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -7839,7 +7839,7 @@ messages:
             j = 1;
             while j <= Length(stuff)
             {
-               if (Nth(stuff,j+1) = $)
+               if (Nth(stuff,j+1) = $) OR Nth(stuff, j + 1) = RentableRoom_expiry_warning
```
akirmse, line 7842: This sure looks random out of context.  Add a comment explaining why it's here, and the fact that it can be deleted after a certain date (say, October 2023).

kod/util/system.kod
```diff
@@ -6574,6 +6574,22 @@ messages:
       }
       return total;
    }
+
+   CleanAllBadMail()
+   "Repeatedly calls CleanBadMail on all users until no bad mails remain."
```
akirmse, line 6579: This is pretty much guaranteed to hit the Blakod interpreter limit on the number of instructions allowed in a single top-level call that prevents infinite loops.  I'll have to figure out the best way to do this, i.e. do I need to bring the server down and do it locally on my machine with a copy of the saved game?

kod/object/active/holder/nomoveon/battler/player/user.kod
Aesica: Done.

kod/util/system.kod
Aesica: Based on this, I removed that function since I really don't like the idea of the operation aborting itself partway.

PR #576, opened 2023-10-21
Animation intensity now based on time rather than frame rate

sicoderealm:
As animation intensity is based on frame rate, the players experience differs.
We use time in milliseconds over frame numbers to correct this variability.

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -20,9 +20,6 @@
 extern int main_viewport_width;
 extern int main_viewport_height;
 
-typedef std::chrono::time_point<std::chrono::steady_clock> steady_clock_time_point;
-static auto& chrono_time_now = std::chrono::steady_clock::now;
-
 // Define field of views with magic numbers for tuning
 inline float FovHorizontal(long width)
 {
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -29,11 +33,33 @@ inline float FovVertical(long height)
...
+{	
+	return (milliseconds & 3) / 256.0f;
+}
+
+// Calculates the intensity of an animation based on the time.
+void updateRenderChunkAnimationIntensity(d3d_render_chunk_new* pChunk)
 {
-	return (frameNumber & 3) / 256.0f;
+	int milliseconds = getMillisecondsLapsed();
+	pChunk->st1[0].s -= animationIntensity(milliseconds);
```
akirmse, line 55: Call animationIntensity once and reuse the value.  And then, once you're doing that, you probably no longer need animationIntensity as a separate function.

clientd3d/d3drender.c
```diff
@@ -29,11 +33,33 @@ inline float FovVertical(long height)
...
+
+// Calculates the intensity of an animation based on the time.
 // This can be used to cycle in a pattern e.g. for invisibility.
-float animationIntensity(int frameNumber)
+float animationIntensity(int milliseconds)
+{	
+	return (milliseconds & 3) / 256.0f;
+}
+
+// Calculates the intensity of an animation based on the time.
```
akirmse, line 51: This comment isn't quite accurate---the function modifies pChunk.

clientd3d/d3drender.c
```diff
@@ -29,11 +33,33 @@ inline float FovVertical(long height)
...
+void updateRenderChunkAnimationIntensity(d3d_render_chunk_new* pChunk)
```
akirmse, line 52: Make function static; it doesn't need to be visible outside this file.

clientd3d/d3drender.c
```diff
@@ -19,6 +20,9 @@
 extern int main_viewport_width;
 extern int main_viewport_height;
 
+typedef std::chrono::time_point<std::chrono::steady_clock> steady_clock_time_point;
```
akirmse, line 23: This typedef seems to be unused?

clientd3d/d3drender.c
```diff
@@ -19,6 +20,9 @@
 extern int main_viewport_width;
 extern int main_viewport_height;
 
+typedef std::chrono::time_point<std::chrono::steady_clock> steady_clock_time_point;
+static auto& chrono_time_now = std::chrono::steady_clock::now;
```
akirmse, line 24: This function renaming is a little odd, and only used once, so I'd suggest just using std::chrono::steady_clock::now in the one place it's called.

PR #582, opened 2023-12-09
Add hardware renderer performance mode toggle

sicoderealm:
Support for players to choose to use lower performance client option from the game configuration. This change includes a clean up of the keybind configuration resource file to meet the minimal visual studio expectations and remove bespoke integrations.

Standard:

GPU Efficiency mode:

Results likely to be more dramatic for low spec / resource machines.


clientd3d/graphics.c
```diff
@@ -52,12 +52,18 @@ static steady_clock_time_point lastEndFrame;
 // The clock to use for fps calculations - updating here will update throughout.
 static auto& chrono_time_now = std::chrono::steady_clock::now;
 
+static bool performance_mode = false;
```
akirmse, line 55: Needs a comment as to what true vs. false means here.  And maybe a better name.

clientd3d/graphics.c
```diff
@@ -330,13 +336,15 @@ void RedrawForce(void)
    auto elapsedMilliseconds = elapsedMicroseconds / 1000;
    msDrawFrame = elapsedMilliseconds;
 
+   auto maxFPS = performance_mode ? 60 : config.maxFPS;
```
akirmse, line 339: Make 60 into a named constant.

Issue #540, opened 2023-08-12
[Geometry] CV Basement vertical grey lines (cobweb related?)


sicoderealm:
The CV Basement has a few vertical lines on the stairs and around the boxes - possibly related to the cobwebs?

<img width="1293" alt="image" src="https://github.com/Meridian59/Meridian59/assets/7548210/1daa4298-5a68-417d-8a24-0dcf45d6dc29">

![image](https://github.com/Meridian59/Meridian59/assets/7548210/5d548eef-6166-4bb5-a220-47cd768f1ae9)

![image](https://github.com/Meridian59/Meridian59/assets/7548210/ac85719c-ab87-4866-b24a-cbf7b4c624aa)

![image](https://github.com/Meridian59/Meridian59/assets/7548210/1c3ebb24-44d3-4ff0-9573-e62044937aca)

and so on...

davidklaw:
So, this can be fixed by adjusting the offset of these textures by 1. However, doing so doesn't feel like the appropriate fix knowing that there are many more instances of this in this room and others. We've seen signs in Jasper fixed in the past, and the gates of Tos and Barloque also have this issue. I'm wondering if there's an adjustment that needs to be made in the code for texture drawing as it seems like a bit of an issue in many place single floating line textures are used. Perhaps everywhere they're used, but most notable when the texture's edge has artwork.

![image](https://github.com/Meridian59/Meridian59/assets/467443/9507b6a6-ac28-45d3-9e9c-e35f50fb9285)

akirmse:
Yeah, agreed.  In rendering there's the concept of wrapping versus clamping
textures.  I wonder if this wall isn't set correctly, or if the hardware
renderer itself isn't setting the mode right.

davidklaw:
```
Seems like it might be tied to this function in `d3drender.c`

Bool D3DMaterialWorldPool(d3d_render_pool_new *pPool)
{
	D3DRENDER_SET_COLOR_STAGE(gpD3DDevice, 0, D3DTOP_MODULATE, D3DTA_TEXTURE, D3DTA_DIFFUSE);
	D3DRENDER_SET_ALPHA_STAGE(gpD3DDevice, 0, D3DTOP_MODULATE, D3DTA_TEXTURE, D3DTA_DIFFUSE);

	IDirect3DDevice9_SetSamplerState(gpD3DDevice, 0,
                                         D3DSAMP_ADDRESSU, D3DTADDRESS_WRAP);
[...]

PR #583, opened 2023-12-10
roomedit: add support for disabling horizontal texture tiling

davidklaw:
## Description
This PR adds a new option flag to room sidedefs: No horizontal tile. When checked, the texture for a sidedef is clamped along the U axis rather than wrapped, which in some cases is causing a narrow errant slice of texture to be visible. This can be used to address issues like signs or webs, and other one-ended lines that aren't rendering as intended. This PR also adds clamping along the V axis when "No vertical tile" is applied.

You can learn more about texture clamping [here](https://learn.microsoft.com/en-us/windows/win32/direct3d9/clamp-texture-address-mode).

### Room Editor UI Changes
The editor's "LineDef Edit" and "Confirm copy of LineDef attributes" UI windows has been adjusted to accommodate this new flag.

### Future Work
There's still work to be done to update `.roo` files where this issue is present.

## Related Issues
#540

## Motivation and Context
There are one-sided linedefs throughout the world used for tavern signs, cob webs, and possibly more though I didn't do an extensive search. When in hardware rendering mode, the textures visibly wrap, ever so slightly. Clamping the sidedef cause the texture to be slightly stretched on one edge as opposed to show this edge wrap issue. See screenshots for examples.

## Screenshots:


davidklaw:
I tried and failed to cleanup the commit history, so re-opening.

roomedit/include/bsp.h
```diff
@@ -56,6 +58,7 @@
 #define WF_BELOW_TOPDOWN  0x00000080      // Draw lower texture top-down
 #define WF_NORMAL_TOPDOWN 0x00000100      // Draw normal texture top-down
 #define WF_NO_VTILE       0x00000200      // Don't tile texture vertically (must be transparent)
+#define WF_CLAMP          0x00000400      // Clamp texture
```
akirmse, line 61: Heads up that this value is not safe to use because 0x400 is already used for WF_HAS_ANIMATED in the client (whatever that is), in clientd3d/bsp.h.  Please copy over that client value so that this can be avoided in the future.

davidklaw:
I'm at a point where I could use some support.

Where I'm at now, is that the Clamp flag can be enabled, but not then disabled. When debugging, I can see that the flag values remains the same when attempting to disable even though the UI check is seeing the unchecked checkbox.

I'm testing the flags at `lineedit.cpp:834`. I've been testing with `a1.roo` and the flags are initially 263172. When I remove 'No vertical tile' for the first sidedef, I get 1028 and then successfully revert it to 263172 by re-enabling it. When I enable clamp for the first sidedef, the flags update to 1311748 and remain unchanged when I try to disable clamp.

My best guess is that it has something to do with my bit flag choices, but I haven't been smart enough to figure out why 😐

davidklaw:
Utilizing the two high order bits seems to do the trick, though I need to do more extensive testing. Something feels icky about using the last remaining bits for these particular flags, but I guess that just means some future bridge that might need to be crossed. I don't see any cases of the flags being used in a signed integer context but I could use other eyes for assurance. Assuming there isn't, should I go around and updated the flags to unsigned ints? Seems pointless, but if it's best practice...

Meridian59:
No need to go and change all the flag types.

davidklaw:
After discussing in Discord, it was decided to refactor this as a horizontal tile flag along the lines of the existing "No vertical tile" flag. When unchecked, the texture is wrapped. When checked, the texture is clamped.

clientd3d/d3drender.c
```diff
@@ -9280,6 +9292,16 @@ Bool D3DMaterialWorldStaticChunk(d3d_render_chunk_new *pChunk)
 		SetZBias(gpD3DDevice, ZBIAS_WORLD);
 	}
 
+	if (pChunk->flags & D3DRENDER_NO_VTILE)
```
akirmse, line 9295: In the spirit of repeating things less, I might write this instead as:

```
auto state = (pChunk->flags & D3DRENDER_NO_VTILE) ? D3DTADDRESS_CLAMP : D3DTADDRESS_WRAP;
IDirect3DDevice9_SetSamplerState(gpD3DDevice, 0, D3DSAMP_ADDRESSV, state);

```

Same thing for the other similar changes in this file.

akirmse (commented):
Looks pretty clean, just one suggestion.

clientd3d/d3drender.c
davidklaw: I'm not sure I love the way it turned out, though I did check in your suggestion. Setting it directly seemed more concise to me.

clientd3d/d3drender.c
```diff
@@ -9205,6 +9205,8 @@ Bool D3DMaterialWorldPacket(d3d_render_packet_new *pPacket, d3d_render_cache_sys
 
 Bool D3DMaterialWorldDynamicChunk(d3d_render_chunk_new *pChunk)
 {
+	auto state = NULL;
```
akirmse, line 9208: This is unexpected and leads to unnecessary if (state) checks later.  Try declaring state where it's first used.

davidklaw: Got it, thanks.

PR #588, opened 2023-12-31
Arena fix part 2: Reflections, Evil Twin, Apparition

CardiovascularMike:
This PR allows summoned monsters (Reflections, Apparitions, and Evil Twins) to attack in the Tos Arena.  Prior to this PR those monsters would get summoned and then just stand around in the arena.

ReqSomethingAttack is the relevant message that needed updating.  The order of the request checks was incorrect: we were checking if the attacker was a combatant and returning False before we ever got to the code that allows monsters to attack.

The ReqSomethingAttack message in Room used to handle arena specific checks.  Instead this PR moves those checks to the ReqSomethingAttack message in the Watchers.

In the Tos Arena this PR adds a few new dialogue options for the Tos Watcher.  This prevents the Tos Watcher from getting confused by killed reflections.  Without the new message resources the Watcher could say something like "John has killed the champion!" when John really only killed the champion's reflection.

The change to ReqSomethingAttack can, in principle, allow evil twins, reflections, and apparitions to attack in Goad's Grinder. 
However, the result of these spells is deeply unsatisfying.  This is because the room geometry of Goad's Grinder is such that summoned monsters immediately fall into the lava.  From the lava they can not attack due to no line of site.  The lava also doesn't kill the monsters at this time.

To get around these problems - we are using ReqSpellCast to disallow these spells for now.  This necessitated 4 simple changes to Spell.kod, SummonApparition, SummonTwin, and Summon Reflection.


kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
```diff
@@ -1208,6 +1225,16 @@ messages:
               return FALSE;
            }
 
+      %%% Summoned Monsters just seem to fall into the lava.
+      %%% For now Goad disallows their use - November 2023
```
akirmse, line 1229: "For now" is likely to be years or decades.. strengthen comment.

kod/object/active/holder/room.kod
```diff
@@ -1202,26 +1202,13 @@ messages:
       {
          oWatcher = Send(self,@GetWatcher);
 
-         % if a fight isnt in session, attack fails.
-         if NOT Send(oWatcher,@FightInSession)
+         % The watcher decides if the attack moves forward
+         if oWatcher <> $ AND NOT
```
akirmse, line 1206: Simplify to 

```
return oWatcher = $ OR Send(oWatcher,@ReqSomethingAttack,#what=what,#victim=victim);
```

CardiovascularMike: Done.  Thanks.

kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
CardiovascularMike: Ok - done.

PR #594, opened 2024-01-03
Fixes for `KD4.roo`

davidklaw:
This PR makes changes to `KD4.roo` that likely requires more detailed testing than previous roo file changes. It addresses 3 key areas that were either found by others or by me while editing.

1. **The guard tower:** This tower suffered from missing and floating textures. That and the southwest pillar seemed oddly out of place, though I'm sure that was intended as a workaround for something. I fixed the texture issues and updated the geometry to bring the pillar in alignment with the others.
2. **The chimney:** While the chimney looks alright on live, it appeared to be missing one of the side blocks in my environment. I rebuilt it and did my best to align the textures for the best possible look.
3. **The dormers:** At the backend of each dormer are notches of visible skybox, where the dormer doesn't completely line up with the roof slope. I filled these gaps with roof texture on all 4 dormers.

I regularly used the editors checks for closed sectors and to delete unused elements, then went around to verify that everything was still intact. I'm not aware of any special functionality in this room, and so I felt simply looking around was the best way to verify deleting these things didn't cause any issues. _I will note however that re-opening the file and running the cross-reference check would still tell me I have unused elements even though I had previously deleted them and saved the file._

### Screenshots
Screenshot 2024-01-02 090213
Screenshot 2024-01-02 090226
Screenshot 2024-01-02 090240
Screenshot 2024-01-02 090307
Screenshot 2024-01-03 100036
Screenshot 2024-01-03 100028


akirmse:
I walked around in there some, and I do still see a small glitch or two:
kd4

davidklaw:
That looks like the version currently on live. This is what I see in the latest version of the room file.

image

akirmse:
You're right, I forgot to copy the file into my local client.  Nice fixes!

PR #595, opened 2024-01-03
Adds PowerShell script to detect and copy updated .roo files to localclient

davidklaw:
This PR adds a Powershell script to `contrib/` that will detect new and modified .roo files in `/resource/rooms` and copy them to the localclient resources in real-time. A huge click-saver when editing room files.

### DIFF
--- a/contrib/powershell/Monitor-Room-Resources.ps1
+++ b/contrib/powershell/Monitor-Room-Resources.ps1
@@ -1,6 +1,30 @@
-# Set the folder paths
-$sourceFolder = "../../resource/rooms" # Change this to your source folder path
-$destinationFolder = "../../run/localclient/resource" # Change this to your destination folder path
+<#
+    .DESCRIPTION
+    Monitors `/resource/rooms` for new or updated `.roo` files, and when found, copies them to `run/localclient/resource`.
+
+    .EXAMPLE
+    From the project root:
+    ./contrib/powershell/Monitor-Room-Resources.ps1
+
+    .NOTES
+    Saves time and clicks when making and testing frequent room file changes.
+#>
+
+# Get the absolute path of the directory where the script is located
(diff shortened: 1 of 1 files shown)


contrib/powershell/Monitor-Room-Resources.ps1
```diff
@@ -0,0 +1,40 @@
+# Set the folder paths
```
akirmse, line 1: Add a big block comment at the top explaining what this script is for, and giving an example of how and where to run it.

contrib/powershell/Monitor-Room-Resources.ps1
```diff
@@ -0,0 +1,40 @@
...
+$sourceFolder = "../../resource/rooms" # Change this to your source folder path
```
akirmse, line 2: This is kind of a weird relative path.  Should you instead assume the top of the source tree?

contrib/powershell/Monitor-Room-Resources.ps1
```diff
@@ -0,0 +1,40 @@
...
+$destinationFolder = "../../run/localclient/resource" # Change this to your destination folder path
+
+# Create a FileSystemWatcher object to watch the source folder
+$watcher = New-Object System.IO.FileSystemWatcher
+$watcher.Path = $sourceFolder
+$watcher.IncludeSubdirectories = $false
+$watcher.EnableRaisingEvents = $true
+
+# Define the filter and types of changes to watch for
+$watcher.Filter = "*.roo" # Monitor all files, change this to something like "*.txt" to monitor only text files
```
akirmse, line 12: "Monitor all files" doesn't match the code.

contrib/powershell/Monitor-Room-Resources.ps1
davidklaw: I ended up adding some additional lines to determine the proper paths, so that it can be run from anywhere.

PR #596, opened 2024-01-05
Arena Fixes Part 3: Goad and Necro Arena clean ups

CardiovascularMike:
This PR fixes a few minor bugs and improves the logic for teleporting a victorious combatant from the arena to the shrine area.

Here is the list of changes - all relatively minor:

Cleanup (no gameplay effect):
1. Move the block comment about advertising in Goad so that it appears just prior to the advertising logic in Goad and not prior to the New Owner and Post New Owner logic.
2. Remove unused local variables in a few messages
3. Remove some dead code related to guild-vs-guild combat in Goad.  Goad never had any plans for Guild combat and this snippet is likely an old copy and paste error.
4. Delete a comment that doesn't seem to be referring to any of the lines of code near it.
5. Sometimes we needed to loop over the list of players in plGoToShrine and we were calling each item in the list 'iCntr'.  I'm  not sure what this means but I have changed it to each_obj for now as the elements of plGoToShrine are all player objects.

Bug Fixes:
1. Goad's postNewOwner message calls CheckWatcher in Necro Arena but was missing the 'What' argument so the check watcher message was non-functional
2. CheckWatcher in Necro Arena was warping the Goad to the wrong location in Necro Arena in the unlikely event that a DM moves him away.  Now he is moved back to his normal spot inside the arena.
3. ShrineUsed is either called by the shrine after a successful dedication or called by a timer after a player fails to dedicate the shrine in time.  Before this PR, in the case where the shrine was successfully used we were creating a delayed teleport timer and then setting plGoToShrine list to $.  This meant that when the delayed teleport timer expires the code had no way to identify which players to teleport back into the stands and the teleport call would fail - leaving the players in the shrine area.  Note that players could still Renege from this location to escape but Reneging here may cause other bugs.  Reneging related bugs will be fixed in a separate PR as there are a few such bugs in production as of this PR.
4. The delayed teleport function which teleports players away from the shrine after a dedication needs a call to reset plGoToShrine - otherwise the most recent player to be on the list will be teleported to the shrine any time a player is killed in future arena battles.  This couldn't happen before due to bug 3 above.

Gameplay improvement:
1. Before this PR a player can only be teleported to the shrine if they defeat an enemy player with 11HP more than themselves.  This is problematic for the case of two established and high-level characters who wish to work together to dedicate the shrine.  This PR makes it so that a player is warped to the shrine if they defeat a stronger enemy (10hp higher), OR if the defeated player has at least 140 HP.


CardiovascularMike:
I've updated this PR to also remove dead code related to a fight ending in a draw - it is not possible to end a fight in a draw inside Goad's Grinder.  The deleted message was never called.

kod/object/active/holder/room/necarena.kod
```diff
@@ -619,14 +622,13 @@ messages:
       }
       else
       {
-         for iCntr in plGoToShrine
+         for each_obj in plGoToShrine
          {
-            Send(iCntr,@Teleport,#what=iCntr,#OutOfPlay=TRUE);
+            Send(self,@Teleport,#what=each_obj,#OutOfPlay=TRUE);
```
akirmse, line 627: nice find!

But what about just calling DelayedTeleport instead?  If the code isn't duplicated, there's no way the two versions can get out of sync.

CardiovascularMike: Yes this is a good suggestion.  I will implement and do a quick test and then update this.

CardiovascularMike: Ok - I implemented this suggestion and tested it just to be safe.  It works well and reads more clearly too.

PR #605, opened 2024-01-11
Fix for animated monsters never disposing

Aesica:
Besides zombies which have special handling, monsters created by the animate spell don't properly get disposed in rooms because the effect (in this case, having a master--the caster) never wears off.  This fixes that in the following ways:

- Changed the animate duration tracking to properly make use of StartEnchantment and EndEnchantment instead of the timer-based system exclusive to zombie.kod
- Added a RevertControl function to monster.kod, which will allow any monster to "forget" its master and revert to normal when called.  (This is called in EndEnchantment)
- Removed the timer-based animate framework from zombie.kod, as it's no longer needed.
- Added a message to inform the caster that their control over a given monster has ended.
- Added a check to monster.kod to prevent summoned monsters (vbSummoned) from dropping loot.*

* Unsure if it would be better to keep it this way or add a boolean property (pbNoTreasure) for this purpose.

Known quirks:

- Skeletons of all types still drop their swords.  This is easily fixable but should probably be a separate PR since it also happens to skeletons created by apparition.
- Monsters won't drop loot when controlled, but will do so normally after animate wears off, but that seems fine to me because at this point they are normal, garden-variety monsters.  If I go the boolean (pbNoTreasure) route, they can be set to never drop treasure, whether controlled or after control breaks.

### DIFF
--- a/kod/object/passive/spell/animate.kod
+++ b/kod/object/passive/spell/animate.kod
@@ -213,7 +213,7 @@ messages:
       oMaster = Send(who, @GetMaster);
       if report AND IsClass(oMaster, &User)
       {
-         Send(oMaster, @MsgSendUser, #message_rsc=Animate_expire_rsc , #parm1=Send(who, @GetDef), #parm2=Send(who, @GetName));
+         Send(oMaster, @MsgSendUser, #message_rsc=Animate_expire_rsc, #parm1=Send(who, @GetDef), #parm2=Send(who, @GetName));
       }
 
       % Only reset behavior if the monster is still alive, so we don't


kod/object/active/holder/nomoveon/battler/monster.kod
```diff
@@ -5512,6 +5512,12 @@ messages:
       return;
    }
 
+   RevertControl()
```
akirmse, line 5515: A string comment on this function would be useful.

kod/object/passive/spell/animate.kod
```diff
@@ -204,6 +206,24 @@ messages:
...
    }
 
+   EndEnchantment(who=$, report=TRUE)
+   {
+      local oMaster;
+
+      oMaster = Send(who, @GetMaster);
+      if report AND IsClass(oMaster, &User)
+      {
+         Send(oMaster, @MsgSendUser, #message_rsc=Animate_expire_rsc , #parm1=Send(who, @GetDef), #parm2=Send(who, @GetName));
```
akirmse, line 216: Remove extra space.

kod/object/passive/spell/animate.kod
```diff
@@ -204,6 +206,24 @@ messages:
...
+      }
+
+      if Send(who, @GetHealth) > 0
+      {
+         Send(who, @RevertControl);
+         Post(who, @ResetBehaviorFlags);
```
akirmse, line 222: This is somewhat unexpected, as it wasn't in the previous implementation in Zombie.  Comment on why it's here?

akirmse (commented):
This is pretty cool, using existing enchantment code rather than a dedicated timer.

kod/object/passive/spell/animate.kod
Aesica: Feel free to correct my interpretation of the logic at any time because I could very well be wrong about it:  My understanding is that, with zombie.kod, RevertControl is never be called if the zombie is killed while affected by animate's control effect since Delete removes the timer.  Since all enchantments are removed when a monster dies (is deleted), this *would* call RevertControl, which is a new behavior for the effect that may or may not be ideal.

I've added a briefer explanation in the code comments.

kod/object/passive/spell/animate.kod
Aesica: Done.  Not sure how that even got there.

PR #606, opened 2024-01-12
Allow for equipping weapons, armor, robes, and clothing without breaking trance with Jala Necklace

CardiovascularMike:
This change improves the usability of the Jala Necklace.  The Necklace was designed to allow a warrior bard type player to fight and keep a Jala song running.  The problem is fighting very often requires changing weapons, equipping a shield or gauntlets, etc.  Prior to this PR those actions would break the trance.  

The change includes adding a usewhat argument to the break trance calls in player and user.  These calls live inside UserUseItem and UserUnuseItem.  We pass the usewhat argument into the breaktrance call that ends up in Jala.kod.  There we check to make sure the item used isn't one we would reasonably expect to break a trance: Instruments, Food, and Potions.

One loophole existed in changing instruments where a player could equip a jala necklace, start singing, and then equip a different necklace (perhaps a god gift).  The one-click swapping of items doesn't use UserUnuseItem and would have avoided the breakTrance call.  We prevent this issue by adding a new breaktrance call in NewUnused() inside instrument.  This insures that if an instrument is UnUsed for any reason the Jala spell trance will be broken.

Breaking trances involves messages to a lot of important files - User, Player, Room, and the Spell itself or the Spell's super-class.  As such I believe some extensive play testing is warranted.  I will update this PR to document that testing before converting from a draft.


CardiovascularMike:
Trace Message - UserUseItem

Because numerous messages are sent as part of a UserUseItem message and because User, Player, Room, Spell, and Jala.kod are all involved I have attached a trace of the messages to show the message-flow.

CardiovascularMike:
Adding a comment to document some testing:

1. Changing instruments always break trance - yes, tried offering, putting, dropping, unequipping, etc.
2. Tested singing a Jala spell and then using: Wands, potions, weapons, armor, gauntlets, helmets, masks, all foods and drinks, etc.  All worked as expected.
3. Tested with distill which isn't in the Jala super class.  Casting distill doesn't break a trance but when the potion is actually created the trance breaks.  I think this is perfect!  This way you could increase the spell power of your distilled potion with mana convergence but the song will end after the spell takes effect.  Other options would be to break the trance as the spell is cast (this is ok too) or not break the trance at all (this would be slightly unexpected).  Breaktrance() in distill uses Event_Steer which is used everywhere we really want to break a trance.  In the future this should be updated to something more sensible like: EVENT_BREAKTRANCE.  As in - this is an event that for sure breaks trances. Steering I think is the action of turning around so the current convention is weird.
4. Casting any regular Jala song breaks the trance of the current song and starts the new one.  For now this is only if the second song is a higher 'volume' than the first.  Future updates to the volume calculation and GetJalaInfo are needed to fix all the associated problems with song volume and multiple jala casts from multiple players.  I've left that outside of the scope of this change since everything in this PR works as intended.

kod/object/passive/spell/jala.kod
```diff
@@ -160,11 +160,22 @@ messages:
...
             OR event = EVENT_ATTACK
             OR event = EVENT_RUN
-            OR (event = EVENT_CAST AND Send(Send(SYS, @GetSettings), @CanRecastJalaWithNecklace))
+            OR (event = EVENT_CAST AND Send(Send(SYS, @GetSettings), @CanRecastJalaWithNecklace)
+               AND NOT IsClass(what,&Jala))
          {
             return FALSE;
          }
 
+         if event = EVENT_USE & usewhat <> $
```
akirmse, line 169: & is bitwise and.  You want AND instead.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -9902,7 +9902,7 @@ messages:
       return;
    }
 
-   BreakTrance(event = 0, what = $)
+   BreakTrance(event = 0, what = $, usewhat = $)
```
akirmse, line 9905: This is super hard to understand.  What is "what" and "usewhat"?  Add a string comment explaining the arguments.

kod/object/active/holder/room.kod
```diff
@@ -3251,7 +3251,7 @@ messages:
       return viTerrain_Type;
    }
 
-   BreakTrance(who=$, event=$)
+   BreakTrance(who=$, event=$, what=$, usewhat=$)
```
akirmse, line 3254: Explain "what" and "usewhat" here too.

kod/object/passive/spell/jala.kod
```diff
@@ -160,11 +160,22 @@ messages:
...
+         {
+            if NOT IsClass(usewhat,&Food)
```
akirmse, line 171: These class checks are unexpected.  There should be a comment here explaining why they're here.

kod/object/passive/spell/jala.kod
CardiovascularMike: Thanks - fixed.

kod/object/active/holder/nomoveon/battler/player.kod
CardiovascularMike: 'what' in this context was always associated with EVENT_CAST.  As such I have changed the 'what' argument to 'castwhat' everywhere.  This did necessitate changing some additional files because we check 'what' more than we check the other arguments.  The additional changes are very minor.

kod/object/active/holder/room.kod
CardiovascularMike: Fixed here too.  Thanks.

kod/object/passive/spell/jala.kod
CardiovascularMike: I added a comment on why food, and instruments, and potions are different then other items that get used.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -9902,7 +9902,7 @@ messages:
       return;
    }
 
-   BreakTrance(event = 0, what = $)
+   BreakTrance(event = 0, castwhat = $, usewhat = $)
```
akirmse, line 9905: I had asked for a comment here explaining "usewhat", and instead you renamed "what" :-)  Can you please add a comment here explaining what the arguments mean?

CardiovascularMike: ok - will do - sorry for the confusion.  I think castwhat is better than just 'what' but I will add a comment or a doc-string.

CardiovascularMike: I added documentation in the form of a doc string - it explains all the arguments and what the function does.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -9902,7 +9902,11 @@ messages:
...
+   "Sends a BreakTrance message to any trance-holding enchantments."
+   "The event is the action causing the broken trance."
+   "If the event is the player casting a spell then castwhat is the ID for that spell."
```
akirmse, line 9908: I'm not sure this comment is correct.  It says that castwhat is the ID for the spell, but I think it's actually the spell object itself, not the SID_ code.  Similarly for usewhat: it's the item, not the ID for an item.

Issue #601, opened 2024-01-09
Shadow Form makes names unreadable in who list dialog box


adrienlaws:
## What
When a user casts shadow form, they become invisible on the who list.  This is because the name color becomes black and is hard to see with the black wholist background color.

Shadow Form is only intended to shadow the name when viewing the player in person.

It's only a level 1 spell and provides a defensive bonus as well.

## Suggestion
Have shadow form continue to affect player names in the 3D view, but remove the name coloring effect from listboxes.  This should include look and offer dialog boxes along with the who dialog box.

PR #609, opened 2024-01-23
fix: shadowform name effect in "who list" dialog box becoming unreadable

adrienlaws:
# What
This fix addresses issue #601 "Shadow Form makes names unreadable in who list dialog box"

# Why
When a player has the shadowform personal enchantment cast on them, their name turns black.  This also makes their name appear in black font in the "who list" dialog box which has a default black background.  As a result, you can't read their name.  This was not intended when the spell was implemented but results because of a change to show player name colors.  That change was intended to make it easier to see murderers, outlaws, and administrator colors in the who list.

## Example Bug 🐛
meridian_dPP7uPMlb9
"test2" name is unreadable in "Who is in Meridian" list box.

## Example Fix ✨
meridian_ACoM81GQqI
"test2" name is readable but the "test2" text above the character remains black, as intended.

# Changes 🔨
Removes the `DRAWFX_BLACK` flag from the ObjectFlags before being sent from `user.kod`
This is a server side solution, which I believe is more elegant than manipulating the client side `color.c` functions, since there is not a use case where we want the name to be drawn in black font.

# Attribution 👏
I was inspired to use a server-side solution from this PR on OpenMeridian by cyberjunk
https://github.com/OpenMeridian/Meridian59/pull/234

# Warning / Doubts / Concerns⚠️
In the referenced PR from OpenMeridian, they use a bitwise operation on INVIS, stating that
> shadowformbits are subset of invis bits

This is true if you look at both flags for shadowform and invisibility.
```
DRAWFX_BLACK        = 0x400000   (binary: 0100 0000 0000 0000 0000 0000)
DRAWFX_INVISIBLE    = 0x500000   (binary: 0101 0000 0000 0000 0000 0000)
```
I used `DRAWFX_BLACK` because that seemed more precise.  I'm not sure if I should use `DRAWFX_INVISIBLE` because it is true that `DRAWFX_BLACK` is a subset of `DRAWFX_INVISIBLE`


akirmse:
It looks like DRAWFX_INVISIBLE is used, and it's different from DRAWFX_BLACK, so you do need to understand and test the difference.  e.g. what happens with invisible players?
 https://github.com/Meridian59/Meridian59/blob/df1873e0717e9011ce12ae3f31bc827a1fac8f6f/kod/object/active/holder/nomoveon/battler/player/user.kod#L7270

Also, please add a comment at the line you changed explaining why this is there, otherwise readers will be confused.

PR #619, opened 2024-02-07
fix: reveal dialog box correctly centered

adrienlaws:
# What
This PR correctly centers the dialog box "Object Description" when the client receives a `BP_LOOK` protocol command.

- All object description windows have the same behavior and centering, regardless of monitor count.
- They are all centered on their respective M59 application windows, not the desktop, and not on the wrong monitor.

# Why

1. **Longstanding Issue**: The `BP_LOOK` object description dialog boxes, when centered, were always centered relative to the desktop and not the client.  This created inconsistency when "looking" at an object vs "revealing" or "identifying" an object.  The "look" object description dialog boxes would be centered on the client, and the `BP_LOOK` dialog boxes would be centered on the desktop.  This is fixed. ✔️ 
2. **Longstanding Issue**: `BP_LOOK` dialog boxes now appear on the correct monitor. This is fixed.✔️ 
3. **Self-inflicted Issue**: After PR #608 the "revealed" object description dialog boxes would not appear centered even via the desktop after casting "reveal."  I had deleted code that used the desktop for centering thinking it was unneeded.  However, the `BP_LOOK` cases have no parent window so without some kind of catch-all, it wasn't working right.  This is fixed. ✔️ 

# Changes
- If a dialog box is created without a parent and `hwndParent` is not set, set the `hwndParent` = `hMain` (the M59 application) in the `CenterWindow()` function.  This is a modification of a previous catch-all that used the Desktop window as a catch-all.

# Example Bug 🐛
explorer_CiKzIhFFae
* When you cast reveal and the server sends a BP_LOOK command - the object description dialog box is centered per the primary monitor.  It is not centered on the M59 application.
* When you "look" at the same item, the object description is centered on the M59 application window.

# Example Fix ✨
meridian_jnJtxSFDaq
- Both revealed and client initiated look actions result "object description" dialog boxes centered on the generating client instance.  They are centered properly on the application and on the correct monitor.

# Attribution
Thanks for @Aesica for pointing out this bug


clientd3d/dialog.c
```diff
@@ -629,7 +629,7 @@ void SetDescParams(HWND hParent, int flags)
 * DisplayDescription:  Display given object's description.
 *   Caller must ensure that obj is not deleted while dialog is up.
 *   flags gives info on what user can do with dialog.
-*   hDescParent global is used as parent for this dialog.
```
akirmse, line 632: It looks like now hDescParent is no longer used, and further, the function SetDescParams no longer works.  SetDescParams is called from a lot of places in the client.  You'll need to test them all to make sure they still work.

akirmse:
Sure, since hDescParent will be set after an object is first looked at.

It's important to understand how existing code works before changing it.  Otherwise it's very difficult to fix a bug without introducing others.

clientd3d/util.c
```diff
@@ -188,6 +188,10 @@ void CenterWindow(HWND hwnd, HWND hwndParent)
    RECT rcDlg, rcParent, rcScreen;
    int x, y;
 
+   // If the window to center has no parent, set hwndParent to the main window
```
akirmse, line 191: This comment doesn't add anything useful---it just says in English exactly what the code does.  What the reader wants to know is *why* this check is here.

PR #622, opened 2024-02-10
Goad Fixes Part 4 - Fix Style_No_Combat bug and remove Kludge message setup

CardiovascularMike:
This PR fixes a bug that can happen only rarely.   The bug is if the Necro Arena is recreated and if the first player to attempt a fight says 'Champion' the player will get a window pop-up to update the game.  This is due to a message being sent that is missing parameters.

For clarity on this change here is the reason for the bug:

1. The default combat style was STYLE_NO_FIGHT
2. A player says Champion - TrigChampion gets called from SomeoneSaid.
3. Trig champion notices that Style is STYLE_NO_FIGHT and POSTS a message to ChooseCombat which will fix this and turn it into STYLE ONE ON ONE
4. Before the POST can run though, NewCombatant is called - this teleports the player into play and calls SetChampion()
5. Set Champion looks if the style is One-On-One or Last-Man-Standing, if it is it sets the player as the champion, if it is not it just returns - at this point because the style is still STYLE_NO_FIGHT the poChampion is still null.
6. Finally TrigChampion sends a say message with a resource which has a parameter defined by a message sent to poChampion (which is still null).  The game sees a message resource with a null arguments and this triggers the Update your client message.

7. After all that, the ChooseCombat message runs and sets the combat style to STYLE_ONE_ON_ONE.  This fixes the bug so it doesn't come up again until the next RecreateAllRooms.

I've fixed this bug in two ways.  First, I've set the default combat style to the only available combat style: one-on-one.  However, as an additional fix, even if an admin sets the combat style to some other style the game won't break.  This is because there was no reason to POST the ChooseCombat message.  Converting this message to a SEND fixes the bug as well.

TO TEST THE MESSAGE RESOURCE BUG:  
Use the admin console to set the piCombatStyle to 0 or any integer other than 1.  Then say champion and notice that the game correctly returns piCombatStyle to 1, sends a message to the client that Goad loves a good duel, and everything else works as intended.

Two additional small messaging changes are rolled into this PR.  Firstly, there was a really awkward (and self-described kludgy) set-up for conveying to the players that a player has challenged another.  We are slowly building up a string with the message we want to send and then using SomeoneSaid to send the message as a temporary string.  I can't figure out why this was done.  I've returned this set-up to the usual method.  While making that change I noticed that it was strange that we are essentially asking everyone in the room if they accept the challenge - not just the champion.  I have fixed this by changing the @ Say to an @ SayToOne(target=poChampion).  Now since we are only asking the champion if they accept we should also add a message for Goad to excitedly announce that the challenger has issued a challenge.  All players in the room see this message.  The champion sees this message and it is immediately followed up by Goad asking them if they will crush the challenger like a bug.

Finally two extra clean-ups are added to Goad.  First, ChooseCombat() had an unused actor argument.  I've removed this argument from the message definition and from every instance where the message is called with Goad as the target.  This is to avoid any bugs in the future if someone wanted to add some function for the actor argument without realizing that actor is always left as $ in the function calls.  Secondly, GetCombatName() was simplified.  We are never passing in a Style argument in Goad and so it doesn't make sense to have a bunch of repeated code for cases when Style is defined and for cases when it is null.  I've likewise removed the Style argument completely.

One note: There is lots of places where actions related to Last-man-standing have blocks of commented out code.  I've left these for now in order to keep the diffs clean and readable.  All of this dead code should be removed but doing so will probably also include simplifying numerous messages that check what the style is and then do some action.  As such it is outside of the scope of this PR.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
@@ -267,7 +267,7 @@ messages:
    {
       if piCombat_Style = STYLE_ONE_ON_ONE
          AND (send(poOwner,@ShrineInUse) = FALSE)
-         AND length(plCombatants = 2)
+         AND length(plCombatants) = 2
       {
          return TRUE;
       }
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
@@ -605,7 +605,7 @@ messages:
        local index, i;
 
        %% Tell room to lower maze or get rid of pests, whatever.
-       Send(poOwner,@EndFight,#lCombatants = plCombatants);
+       Send(poOwner,@endfight,#lCombatants = plCombatants);
 
        %% Assume that everyone left in plCombatant is a winner.
 
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
@@ -49,7 +49,8 @@ resources:
    Goad_already_champion = "So generous . . . but, you are already mine."
    Goad_already_have_champion = "So you would be champion?  Nothing could be easier, simply challenge and kill %s%s.  Would you dare to challenge %s?"
 
-   Goad_challenge_offered = "%s%s has challenged you.  Will you crush %s's like the bug %s is?"
+   Goad_challenge_offered = "%s%s has challenged you.  Will you crush %s like the bug %s is?"
+   Goad_challenge_excitement = "%s%s has challenged %s%s to bloody battle!"
    Goad_must_accept = "Hurry fool!! You only have %i seconds left to accept %s%s's challenge."
 
    Goad_cant_renege = "You cannot renege unless you are a combatant!"
@@ -325,6 +326,10 @@ messages:
 
       if piCombat_style = STYLE_ONE_ON_ONE
       {
+         % Tell the room and then ask the champion if they accept
+         Send(self,@say,#message_rsc=Goad_challenge_excitement,
+            #parm1=send(who,@getcapdef),#parm2=send(who,@getname),
+            #parm3=send(poChampion,@getcapdef),#parm4=send(poChampion,@getname));
          Send(self,@SayToOne,#target=poChampion,#message_rsc=Goad_challenge_offered,
             #parm1=send(who,@getcapdef),#parm2=send(who,@getname),
             #parm3=send(who,@gethimher),#parm4=send(who,@gethimher)); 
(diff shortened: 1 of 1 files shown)


kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
```diff
@@ -763,13 +764,13 @@ messages:
          
          if StringEqual(string,"duel")
          {
-            post(self,@ChooseCombat,#style = STYLE_ONE_ON_ONE,#actor=what);
+            post(self,@ChooseCombat,#style = STYLE_ONE_ON_ONE);
             return FALSE;
          }
          %%Rumble style cut due to insuficient test.
```
akirmse, line 770: Delete this commented-out section.

kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
```diff
@@ -791,12 +792,9 @@ messages:
...
    {
-      if piCombat_style = STYLE_NO_FIGHT
-         {
-           %%Only one style availible: Duel.
-           %send(self,@say,#message_rsc=Goad_must_choose_style);
-           %return;    
-           post(self,@ChooseCombat,#style = STYLE_ONE_ON_ONE,#actor=what); 
+      if piCombat_style <> STYLE_ONE_ON_ONE
+         {  
+           Send(self,@ChooseCombat,#style = STYLE_ONE_ON_ONE);
```
akirmse, line 797: This is confusing.  If the style isn't 1-on-1, make it 1-on-1?  Is the "if" needed, or it is safe to simply always set the style to 1-on-1?

kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
```diff
@@ -831,11 +829,9 @@ messages:
    "the match, but in general, this will fail if there is no champion, and "
    "will succeed only if there is a challenger spot open."
    {
-      if piCombat_style = STYLE_NO_FIGHT
+      if piCombat_style <> STYLE_ONE_ON_ONE
          {
-           post(self,@ChooseCombat,#style = STYLE_ONE_ON_ONE,#actor=what);
-           %send(self,@saytoOne,#target=what,#message_rsc=Goad_must_choose_style);
-           %return;     
+           Send(self,@ChooseCombat,#style = STYLE_ONE_ON_ONE);
```
akirmse, line 834: Why was this changed from Post to Send?

CardiovascularMike: The posting was the cause of the bug this is fixing.  Its double fixed now since the new default is always One_on_one anyhow.

For trigChampion() the posting means we try to say() with a parm=send(poChampion,@getname) when poChampion=$.

For trigChallenge() in the unlikely event that somehow the style isn't one-one-one but there is a champion to challenge this would mean the accept timer doesn't start and I don't know if the fight would even validate and run.  This one is much less likely than the TrigChampion version of the bug - which happens every first fight after a Recreate.  Essentially the Goad has to set the fight to One-on-One in order for everything to function and that has to happen before the rest of the work in each message and so the POST is a mistake.

Now that said - the entire setup is unnecessarily complex because Last-man-standing was cut just prior to production and so we shouldn't really have to set the combat style and check the combat style so often (or at all).  I can implement that type of fix if you think that would be better.  It would mean deleting 3-4 messages and re-writing 4-5 others.

The simplest possible fix to the message bug would be only changing the piCombat_style default property.  That alone does prevent the game update bug - assuming no admin ever changes the piCombat_style property using the admin console.

kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
CardiovascularMike: Yes we don't need an if or even a choose combat message.  I was trying to fix the potential problem without re-writing too much of the machinery.

Its possible that re-writing the machinery is the better path.  The simplest re-write would be removing ChooseCombat and rolling its function into the Trig functions. 

I also don't want to leave this particular game-update-bug in production even though it only comes once after each recreate.

CardiovascularMike:
I'm closing this in order to submit a few smaller PRs to the same file.  This will include some of these improvements while reducing the need to re-write any code in the future if a contributor wants to continue to fix this problematic file.

PR #623, opened 2024-02-10
fix: hidden DM/Admin should not be publicly greeted by watcher and goad

adrienlaws:
# What
This PR adjusts the flag checking for Admin and DM characters so that the Tos Arena Watcher and Goad do not publicly ("say") respond to their presence if certain flags are set.

# Why
The Tos Watcher and Goad should not be greeting characters with certain flags set.

# Example Fix ✨ 
meridian_QUhInE7HAI
- Admin logs in
- Goad greets them with "say"
- Admin sets `dm hidden` and logs out
- Admin logs back in and Goad greets them but only with "tell"

# Changes 🚧 
- Changed both Tos Watcher and Goad to have the same checks
- Added checks
  - Watcher/Goad does not greet Admins or DMs with a "say" if any of the following are true:
    - IsHidden() -> checks flag `DMFLAG_HIDDEN`
    - IsDMStealthed() -> checks flag `DMFLAG_INVISIBLE`
    - flag=`PFLAG_MORPHED`
    - flag=`PFLAG_ANONYMOUS`


kod/object/active/holder/nomoveon/battler/monster/towns/Goad.kod
```diff
@@ -1243,16 +1243,22 @@ messages:
    SomethingEntered(what=$)
    {
 
-      if IsClass(what,&admin) and not send(what,@checkplayerflag,#flag=PFLAG_MORPHED)
-         and not send(what,@checkplayerflag,#flag=PFLAG_ANONYMOUS)
+      if IsClass(what,&Admin)
```
akirmse, line 1246: This is kind of ridiculous to have the same set of checks in 4 different places.  Turn it into a function so that it's just in one place.

akirmse:
Very nice!

PR #630, opened 2024-02-12
Rentable Room Update 5:  Fix for room key copies not being garbage collected after being deleted

Aesica:
As long as the master (gold) room key that a key copy is created from exists, that key copy cannot be garbage collected after being deleted (given to an innkeeper for destruction) since it remains in the master key's plCopies list.

This should solve that problem.

Note:  Destruction of the gold room key via room expiration, as well as use of the "change lock" command are unaffected by this issue.


kod/object/item/passitem/roomkeyc.kod
```diff
@@ -254,6 +254,16 @@ messages:
       propagate;
    }
 
+   Delete()
+   {
+      if (Send(self, @IsKeyCopy))
+      {
+         Send(&RoomKey, @RemoveCopy, #what=self);
```
akirmse, line 261: Sends to every key in the entire game?  Is there no backreference to the key this is a copy of?

Aesica: I originally tried that and it worked great when handing the key to an innkeeper to destroy, but when a room expires, it spits out errors--presumably because the room and associated gold key are also deleted, so the lookup chain that eventually finds the gold key ended up breaking in various ways.

While perhaps not ideal, this is the most foolproof way to make sure a copy gets unlinked from whatever gold key is storing a reference to it, no matter where it or the gold key are (or aren't) in the world.  To note, something similar is done with riija swords and sword shards when a player suicides.

I *can* try to find an alternative again if you like, especially if it results in server lag.

Aesica:
Going to take a completely different approach to solving this problem that requires an additional implementation to be done, first.    As such, I'm closing this one.

PR #635, opened 2024-02-17
Vile qor axe spell, item and adjustments

sicoderealm:
Addition of a new spell and item for Qor (equivalent to spiritual hammer).
Includes slight resistance tweaks to higher end mobs to make viable.

image
image

### DIFF
--- a/kod/object/item/passitem/weapon/vileaxe.kod
+++ b/kod/object/item/passitem/weapon/vileaxe.kod
@@ -25,7 +25,7 @@ resources:
    vileaxe_expires = "Your vile axe disappears in a puff of foul-smelling smoke!"
 
    vileaxe_window_overlay_rsc = povaxe.bgf
-   vileaxe_player_overlay = axeov.bgf
+   vileaxe_player_overlay = vaxeov.bgf
 
 classvars:
    vrName = vileaxe_name_rsc
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqbart.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqbart.kod
@@ -27,6 +27,11 @@ resources:
    BaBar_entry_welcome = "Hail %s%s, and welcome to my humble tavern!  Praise Shal'ille!"
    BaBar_unwanted_keep = "~k%s%s tells you, \"I thank thee for thy offering.  I shall distribute it to the poor and needy.~n~k\""
 
+   bar_enter_qor_axe = "These ancient town sewers are quite literally flooded with odd looking amphibious creatures. "
+     "They say they are peaceful folk but they know how to fight! I don't trust them, stay clear of the deepest waters. "
+     "The Qor priestess is rumored to teach select disciples the secrets of a Vile Axe that is a powerful weapon against prey but you didn't hear it from me! "
+     "Now... how about a sampling of my fine wines from Jaspers sweet grass prairies region, where the sun always shines?"
+
 
 classvars:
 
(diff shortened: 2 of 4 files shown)


kod/object/item/passitem/weapon/vileaxe.kod
```diff
@@ -0,0 +1,171 @@
...
+
+   include blakston.khd
+
+resources:
+
+   vileaxe_name_rsc = "vile axe"
+   vileaxe_icon_rsc = axe.bgf
+   vileaxe_desc_rsc = \
+      "Inherently having unholy blessings, this axe is even more powerful "
+      "than the standard variation."
```
akirmse, line 23: "standard variation" sounds like an oxymoron.  "standard variety"?

kod/object/item/passitem/weapon/vileaxe.kod
```diff
@@ -0,0 +1,171 @@
...
+
+   vileaxe_expires = "Your vile axe disappears in a puff of foul smelling smoke!"
```
akirmse, line 25: "foul-smelling"

kod/object/active/holder/nomoveon/battler/monster/temples/qorprst.kod
```diff
@@ -125,6 +125,12 @@ messages:
          return FALSE;
       }
 
+      if sid = SID_VILE_AXE
```
akirmse, line 128: Looks like this can be combined with the previous check with an OR ?

kod/object/item/passitem/weapon/vileaxe.kod
```diff
@@ -0,0 +1,171 @@
...
+   vrWeapon_overlay = vileaxe_player_overlay
+
+   viGround_group = 1
+   viInventory_group = 3
+   viBroken_group = 2
+
+properties:
+
+   piAttack_type = ATCK_WEAP_MAGIC+ATCK_WEAP_SLASH
+   ptLongevity
```
akirmse, line 59: Add a comment describing what this timer is for.

Gar103:
I'd be very careful with altering mob resistances so heavily. Looks like it's essentially creating an 'easy mode' for neg karma leveling. However well intentioned, that can lead to making building feel less immersive and more like a chore, since this will make poggs and mollusks far and away 'the only choice,' and so you'll be in sewers, then 3 screens over in beach, essentially the entire process. The best build paths afais give the player multiple comparative options and keep them switching up and doing interesting things. So I'm not necessarily saying don't do this, but rather, if you do, it's not enough on its own. Need to add multiple other places to build and things to do to offset this. Otherwise I could absolutely build a character from ~40 hp to 150 in basically one single spot on what are among the most boring mobs and locations in the game. 

As a separate thing to bring up, since resistances are invisible, there's a high knowledge cost for the existing and returning playerbase. Everyone's got an identity for Poggs in their head, so finding out poggs are super weak to this axe (and any qor enchanted weapon actually) will lead to what I call a Knowledge Violation Moment. You can only afford a few KVMs before players lose their immersion and nostalgia, assume everything's just crazy now, and quit. This can be mitigated by doing the work of education. Globe posts and stuff help. NPC dialogue and quests are even better. If a quest was implemented that teaches players poggs and mollusks are now weak to unholy, that can handle new and returning players... mostly. For example, I can imagine us implementing a new quest line that triggers for characters that were inactive, but have just now logged on. Maybe for Qors, the Priestess sends them a tell asking them to come to the Temple, and they get involved in some sort of story or drama, during which they find out poggs / mollusks are now weak to unholy. That kind of thing.

PR #637, opened 2024-02-19
Update GetJalaInfo to consider multiple Jala spells in room

CardiovascularMike:
In current production multiple Jala spells can be active in any one room at once.  The room enchantment icons correctly reflect the active room enchantments.

PR606 removed the ability for any one bard to sing multiple songs at once but it is still very possible for two Jala spellcasters to sing two different songs simultaneously.

While users might expect that both Jala spells are effecting the room, for most spells this isn't true.  Everywhere that GetJalaInfo() is used to measure the effectiveness of a Jala spell we are currently only returning the information on the first Jala spell in plEchantments.  This means that if one player casts Restorate and a second player casts Rejuvinate only the Restorate spell has any effect.

In current production some spells do work as the second Jala spell: Jig, Mirth, and Melancholy currently work this way. Currently spell bane 'works' as the second spell but in a problematic way: If spellbane is the second Jala spell the room will count that Spellbane as active but will then use the first Jala spell's spellpower in any Spellbane related calculations.

This PR updates this behavior by changing GetJalaInfo() to return a list of lists of all Jala spell related info.  We then update everywhere in the database where GetJalaInfo is used and iterate over the list of Jala spells that we receive to calculate the Jala spell effects.

Casting a new Jala spell still requires the caster to have a higher volume than all other active Jala spells.  This could be updated in the future.  

This update does not allow for players to stack the effects of any one Jala spell.  For example, if a first player casts Restorate and a second player also casts Restorate - only the first player's spell has any effect.  This could be updated in the future with a harmonize skill or something similar but the total effect of each spell should still be capped to whatever the maximum spell power is for each spell.  This is to prevent power-gaming and for reasons of balance.  Each spell was only ever intended to have a maximum effect based on spell power.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5510,7 +5510,7 @@ messages:
    CalculateManaTime()
    "Calculate # of milliseconds until user will gain a mana point"
    {
-      local iTime, lJalaInfo, oJalaSpell, iJalaPower, oSpell, i;
+      local iTime, lJalaInfo, oJalaSpell, iJalaPower, oSpell, lJalaElement;
 
       if piMana > piMax_mana
       {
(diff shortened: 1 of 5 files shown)


CardiovascularMike:
I'm going to close this so I can re-write this PR now that PR #644 is merged.

PR #641, opened 2024-02-23
Wedding Rings: Refactor NewUsed() to fix first time use bug

CardiovascularMike:
This PR fixes a bug whereby on the first use of a wedding ring the defense modifier isn't added for that very first time it is equipped.  The game still tries to remove a defense modifier when the ring is UnUsed.  This results in a debug message to the server but no other problems.

I've refactored the NewUsed() message to always apply the Defense Modifier if the player is one of the two players who are bonded by the ring in matrimony (or actually if they are listed on the ring at all - it doesn't matter any more since we removed the defense buff itself).

Note that the defense modifier effect is removed from the wedding rings.  This was done by simply returning when we call Modify Defense Power or Modify Defense Damage.  That change was made a long while ago.

After looking into that change, I believe that this method was chosen to avoid any problems with players who already had a defense bonus applied and have kept their wedding rings equipped (perhaps for many years).  If we simply removed the defense modifier qualities of wedding rings these players would have a permanent modifier in their plDefenseModifier lists.  These players could then generate errors whenever the game checks them for defense modifiers.  For this reason I believe the choice of zeroing out the defense effect of wedding rings was done in a reasonable manner.

Summary: NewUsed was implemented wrong and the game can't remove a defense modifier that it never applied.  This fixes that even though there is no gameplay change.

To test:  Load up a current branch without this PR, create a new wedding ring, equip and un-equip it.  You will see a note in the debug log about the missing defense modifier.  Rebuild from this branch and do the same and there will be no debug log.  Furthermore equip the ring and examine your test character to see that the ring is added to plDefenseModifier on first use.

### DIFF
--- a/kod/object/item/passitem/ring/ringwedd.kod
+++ b/kod/object/item/passitem/ring/ringwedd.kod
@@ -115,8 +115,8 @@ messages:
 
    NewUnused(what = $)
    {
-      if what = poFirstHeld
-         OR what = poSecondHeld
+      if (what = poFirstHeld OR what = poSecondHeld)
+         AND poSecondHeld <> $
       {
          Send(what,@RemoveDefenseModifier,#what=self);
       }
--- a/kod/object/item/passitem/ring/ringwedd.kod
+++ b/kod/object/item/passitem/ring/ringwedd.kod
@@ -103,8 +103,8 @@ messages:
             }
          }
 
-         if what = poFirstHeld
-            OR what = poSecondHeld
+         if (what = poFirstHeld OR what = poSecondHeld)
+            AND poSecondHeld <> $
          {
             send(what,@AddDefenseModifier,#what=self);
          }
--- a/kod/object/item/passitem/ring/ringwedd.kod
+++ b/kod/object/item/passitem/ring/ringwedd.kod
@@ -86,27 +86,27 @@ messages:
 
       if what <> $ AND isClass(what,&Player)
       {
+
          if poFirstHeld = $
          {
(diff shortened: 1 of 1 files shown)


akirmse:
This would definitely fix the debug message, but I wonder if it's the original intention.   I'm guessing the defense modifier is meant to be applied once the second person picks the ring, i.e. the modifier should not be applied if only one person has ever used the ring.  What do you think about that?

CardiovascularMike:
You are correct about not applying on first use.  The defense modifier doesn't do anything currently but perhaps it might in the future.

I will add a check before applying or removing the defense modifier:  if poSecondHeld <> $.  This will work as it used to but not have the problem of removing a non-existent defense modifier.

In fact - the entire fix could be a check for poSecondHeld <> $ on removing the defense modifier.  I didn't like the old set-up but I will review once more - perhaps this change can touch much fewer lines of code.

Gar103:
Elsewhere we split the defense bonus off into a new 'ring of defense' sold by Rook that does nothing but preserve the 100 defense bonus. Wedding rings of course still existed in their old form. I'm surprised the wedding ring was simply disabled in this case, that's a pretty significant balance change. In terms of balance, the 100 defense ring is still in quite heavy competition with zerker rings, invis rings, resist rings, etc.

CardiovascularMike:
I'm going to close this an look at it again in the future as there are numerous more important PRs in the works.  The issue this resolves is only possible debug message spam on the server and the amount of spam is expected to be very small.

PR #644, opened 2024-02-27
Fix persistent Jala music after song ends

CardiovascularMike:
Currently Jala music is bugged such that when a song ends it keeps playing in the room.  Furthermore if the Jala player casts another Jala spell its possible for this bug to crash the client unexpectedly.

This PR aims to fix the issue with persistent Jala music in the hopes that doing so will fix the client crash bug.

To accomplish this goal we make a few minor changes in KOD:
1. Jala.kod was previously used to send a a StartMusic message with a null resource argument.  This was intended to tell the MIDI music handlers to abort the midi file. This has been changed such that EndSpell calls a new message to restart the room music after Jala.
2. The null argument eventually got to MidiSendUser() in User.KOD.  Here however the function just returns if the argument is NULL.  At some point these two functions fell out of synch with each other.
3. Room.kod now has a RestartMusicAfterJala() message which uses a variation of the machinery we use when a player enters a new room to restart the music in the room after the Jala spell ends.  This is sent to all users in the room.

The game is split into two types of music handlers - one for production and one for open-source users / servers.  The majority of this PR is fixing the open source implementation at this time as that implementation was lacking a few necessary features.

Testing the production music will require some collaboration and this PR will be updated based on that feedback.  We may be able to use the changes to the MIDI only music.c as a roadmap but I think the production music set-up isn't lacking any functions.  I would assume any problems remaining there is due to the handling of the various Boolean tracker variables.  

Music.C has numerous tracking variables.  I've tried to comment their uses when they are declared initially.  Of note: Playing_music tracks if we are supposed to be playing music, not if we are actively playing music.  The difference is that playing_music is true even when we are paused.

Paused_music is a new variable.  Previously we considered music paused if the music_pos variable was 0.  This isn't safe because the background music loops and it is very possible to pause the music at position 0.  In fact, entering a room and immediately casting a Jala spell frequently would pause the music at position 0.

Pausing and unpausing wasn't implemented in MIDI world.  We had the machinery to seek and find the music position but we never send an MCI_PLAY message.  This has been fixed.

Abort music was used inappropriately in PlayMusicRsc - this has been fixed.

Finally and most importantly - a second MIDI device has been added.  This allows us to pause the background music without removing the music file from memory.  Instead we open the second Midi device and load the Jala song into memory pointed to that second device.  The two midi devices are: Midi_element and midi_bg_music_element.  This change was made because of the significant lag caused by loading the background music into memory.  This lag is less noticeable when we are changing rooms but is very noticeable when interrupting a Jala song.

There are a few other changes and fixes.  I will comment those in-line in the as a review.

Finally - if needed I can make a flow-chart of the message flow in Music.c.  There are two starting points: PlayMidiRsc (this is always used for Jala songs) and PlayMusicRsc (this is used for background music - even if that music is a MIDI).

There is one outstanding issue: Tracking the current song while the music setting is toggled off works except in the case that a user turns off the music, sings a Jala song, then interrupts that song with a second Jala song and then toggles on the music setting.  In this event the user will get the room background music and not the second Jala song music.  I'm not yet sure why - but it certainly has to do with the latest_music tracking variable not being saved in the correct order.  I will either fix this issue or at least understand why it happens before switching this PR from a draft to a merge request.

### DIFF
--- a/clientd3d/music.c
+++ b/clientd3d/music.c
@@ -27,10 +27,16 @@ static Bool has_midi = False;    /* Can system play MIDI files? */
 
 static Bool playing_midi = False;  /* Is a MIDI file currently playing as an effect? */
 static Bool playing_music = False;  /* Is a MIDI file currently playing as background? */
+static Bool paused_music = False;   /* Is background music paused? */
+static Bool case_bg = False;   /* For loading midi files: which device to load to */
+static Bool badclosebool = False; /* When closing a midi and posting new 
+                     midi-music - did the close function error?*/
 
 static UINT midi_element;  /* Currently playing MIDI file ID */
+static UINT midi_bg_music_element;  /* Currently playing background music ID */
 
 static ID    bg_music = 0;     /* Resource id of background music MIDI file; 0 if none */
+static ID    latest_music = 0;   /* Resource id of most recent music file played - regardless of type */
 static DWORD music_pos = 0;    /* Current position in background music when paused */
 static DWORD time_format;      /* Time format of current music MIDI file */
 
@@ -132,6 +138,7 @@ void MusicClose(void)
 //	Not used by MSS version
 /*
  * OpenMidiFile:  Open midi file for playing.
+ *   If case_bg: Opens background music file; otherwise, opens jala music file.
  *   Returns 0 if successful; MCI error code otherwise.
  */
 DWORD OpenMidiFile(const char *lpszMIDIFileName)
@@ -146,22 +153,60 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
 
    sprintf(filename, "%s%s", current_dir, lpszMIDIFileName);
    debug(("music filename = %s\n", filename));
-
-   memset(&mciOpenParms, 0, sizeof(MCI_OPEN_PARMS));
-   mciOpenParms.lpstrDeviceType = "sequencer";
-   mciOpenParms.lpstrElementName = filename;
-   if (dwReturn = mciSendCommand(0, MCI_OPEN, MCI_OPEN_ELEMENT,
+   // Is it a background music or jala music file?
+   if (case_bg)
+   {
+      if (playing_music != 0)
+      {
+         mciSendCommand(midi_bg_music_element, MCI_CLOSE, 0, 0);
+      }
+      memset(&mciOpenParms, 0, sizeof(MCI_OPEN_PARMS));
+      mciOpenParms.lpstrDeviceType = "sequencer";
+      mciOpenParms.lpstrElementName = filename;
+      debug(("Loading background music into memory \n"));
+      if (dwReturn = mciSendCommand(0, MCI_OPEN, MCI_OPEN_ELEMENT,
                                  (DWORD_PTR)(LPVOID) &mciOpenParms)) 
-      return dwReturn;
-
-   midi_element = mciOpenParms.wDeviceID;
-   debug(("midi element = %d\n", midi_element));
-   return 0;
+         return dwReturn;
+      // Save element ID and set flag
+      midi_bg_music_element = mciOpenParms.wDeviceID;
+      playing_music = 1;
+      debug(("midi background element = %d\n", midi_bg_music_element));
+      return 0;
+   }
+   else
+   {
+      if (playing_midi != 0)
+      {
+         mciSendCommand(midi_element, MCI_STOP, 0, 0);
+         if (dwReturn = mciSendCommand(midi_element, MCI_CLOSE, 0, 0))
+         {
+            // Just debug don't return
+            debug(("Midi Element - mciClose problem\n"));
+         }
+      }
+      memset(&mciOpenParms, 0, sizeof(MCI_OPEN_PARMS));
+      mciOpenParms.lpstrDeviceType = "sequencer";
+      mciOpenParms.lpstrElementName = filename;
+      debug(("Loading Jala song into memory \n"));
+      if (dwReturn = mciSendCommand(0, MCI_OPEN, MCI_OPEN_ELEMENT,
+                                 (DWORD_PTR)(LPVOID) &mciOpenParms))
+      { 
+         debug(("can't open MCI OPEN new midi for jala \n"));
+         return dwReturn;
+      }
+      // Save element ID and set flag
+      midi_element = mciOpenParms.wDeviceID;
+      playing_midi = 1;
+      debug(("midi element = %d\n", midi_element));
+      return 0;
+   }
 }
 /******************************************************************************/
 /*
  * PlayMidiFile:  Play given MIDI file and notify given window when done.
  *   Returns 0 if successful, MCI error code otherwise.
+ *   Only used by Jala music files - never background music 
+ *   even if it is a midi file.
  */
 DWORD PlayMidiFile(HWND hWndNotify, char *fname)
 {
@@ -225,10 +270,12 @@ DWORD PlayMidiFile(HWND hWndNotify, char *fname)
    {
       DWORD dwReturn;
       MCI_PLAY_PARMS mciPlayParms;
+      case_bg = False;
 
       if ((dwReturn = OpenMidiFile(fname)) != 0)
       {
-        return dwReturn;
+         debug(("OpenMidiFile error - can't open \n"));
+         return dwReturn;
       }
 
       /*
@@ -237,10 +284,13 @@ DWORD PlayMidiFile(HWND hWndNotify, char *fname)
        * MM_MCINOTIFY message when playback is complete.
        * The window procedure then closes the device.
        */
+      memset(&mciPlayParms, 0, sizeof(MCI_PLAY_PARMS));
+
       mciPlayParms.dwCallback = (DWORD_PTR) hWndNotify;
       if (dwReturn = mciSendCommand(midi_element, MCI_PLAY,
                                     MCI_NOTIFY, (DWORD_PTR)(LPVOID) &mciPlayParms)) 
       {
+         debug(("Midi Element - mciPlay problem\n"));
          mciSendCommand(midi_element, MCI_CLOSE, 0, 0);
          return dwReturn;
       }
@@ -264,7 +314,7 @@ DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
 
 #ifdef M59_MSS
 	// If a sequence was paused, resume it
-	if (music_pos != 0)
+	if (paused_music)
 	{
 		UnpauseMusic();
 		return 0;
@@ -321,18 +371,23 @@ DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
    DWORD dwReturn;
    MCI_PLAY_PARMS mciPlayParms;
    char temp[81];
+   case_bg = True;
+
+   // If already playing music, pick up where we left off
+   if (paused_music)
+   {
+      UnpauseMusic();
+      return 0;
+   }
 
    if ((dwReturn = OpenMidiFile(fname)) != 0)
      {
        debug(("OpenMidiFile error code = %d\n", dwReturn));
        mciGetErrorString(dwReturn, temp, 80);
+       strcat(temp, " \n");
        debug((temp));
        return dwReturn;
      }
-
-   /* If already playing music, pick up where we left off */
-   if (music_pos != 0)
-      UnpauseMusic();
 
    /*
     * Begin playback. The window procedure function
@@ -341,18 +396,19 @@ DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
     * The window procedure then closes the device.
     */
    mciPlayParms.dwCallback = (DWORD_PTR) hWndNotify;
-   if (dwReturn = mciSendCommand(midi_element, MCI_PLAY,
+   if (dwReturn = mciSendCommand(midi_bg_music_element, MCI_PLAY,
                                  MCI_NOTIFY, (DWORD_PTR)(LPVOID) &mciPlayParms)) 
    {
       mciGetErrorString(dwReturn, temp, 80);
+      strcat(temp, " \n");
       debug((temp));
       
-      mciSendCommand(midi_element, MCI_CLOSE, 0, 0);
+      mciSendCommand(midi_bg_music_element, MCI_CLOSE, 0, 0);
       
       return dwReturn;
    }
 
-   debug(("Playing music file, element = %d\n", midi_element));
+   debug(("Playing music file, element = %d\n", midi_bg_music_element));
    playing_music = True;
    return 0;
 #endif
@@ -377,41 +433,46 @@ DWORD RestartMidiFile(DWORD device)
       mciSendCommand(device, MCI_CLOSE, 0, 0);
       
    }
+   debug(("Restarting MIDI file, element = %d\n", midi_element));
    return dwReturn;
 }
 /******************************************************************************/
 /* 
  * PauseMusic:  Store current position of background music.  Assumes that
  *   music is playing on MIDI device.
+ *   Only room background music can be paused.
  */
 void PauseMusic(void)
 {
-#ifdef M59_MSS
    if (!has_midi)
       return;
+
+#ifdef M59_MSS
 
 	if( AIL_sample_status( hseqBackground ) == SMP_PLAYING )
 		AIL_stop_sample( hseqBackground );
 	// indicate we are paused
-	music_pos = 1;
+	paused_music = True;
 	debug(( "Pausing music.\n" ));
 #else
    MCI_STATUS_PARMS mciStatusParms;
 
-   if (!has_midi)
-      return;
-
    mciStatusParms.dwItem = MCI_STATUS_POSITION;
-   mciSendCommand(midi_element, MCI_STATUS, 
+   mciSendCommand(midi_bg_music_element, MCI_STATUS, 
                   MCI_STATUS_ITEM, (DWORD_PTR)(LPVOID) &mciStatusParms);
    music_pos = mciStatusParms.dwReturn;
 
    /* Get time format */
    mciStatusParms.dwItem = MCI_STATUS_TIME_FORMAT;
-   mciSendCommand(midi_element, MCI_STATUS, 
+   mciSendCommand(midi_bg_music_element, MCI_STATUS, 
                   MCI_STATUS_ITEM, (DWORD_PTR)(LPVOID) &mciStatusParms);
    time_format = mciStatusParms.dwReturn;
 
+   // Pause the music!
+   MCI_GENERIC_PARMS mciPauseParms;
+   mciSendCommand(midi_bg_music_element, MCI_PAUSE, 0, (DWORD_PTR)(LPVOID) &mciPauseParms);
+
+   paused_music = True;
    debug(("Pausing, position = %ld\n", music_pos));
 #endif
 }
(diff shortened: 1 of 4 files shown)


clientd3d/music.c
```diff
@@ -27,10 +27,16 @@ static Bool has_midi = False;    /* Can system play MIDI files? */
 
 static Bool playing_midi = False;  /* Is a MIDI file currently playing as an effect? */
 static Bool playing_music = False;  /* Is a MIDI file currently playing as background? */
+static Bool paused_music = False;   /* Is background music paused? */
+static Bool case_bg = False;   /* For loading midi files: which device to load to */
```
akirmse, line 31: This is a device?  Then why is it named "case_bg"?

clientd3d/music.c
```diff
@@ -27,10 +27,16 @@ static Bool has_midi = False;    /* Can system play MIDI files? */
...
+static Bool badclosebool = False; /* When closing a midi and posting new
```
akirmse, line 32: Please find a better name.

clientd3d/music.c
```diff
@@ -492,43 +622,47 @@ void PlayMusicRsc(ID rsc)
...
  * NewMusic:  Start playing new music.  type gives type of sound (MIDI, 
  *   music).  rsc gives resource id of file.
+ * Here we convert the rsc to a filename and send a message to
+ * either PlayMusicFile or PlayMidiFile depending on the type.
  */
 void NewMusic(WPARAM type, ID rsc)
 {
    char *filename, fname[MAX_PATH + FILENAME_MAX];
-   
+   debug(("NewMusic Call \n"));
```
akirmse, line 663: Remove

clientd3d/music.c
```diff
@@ -492,43 +622,47 @@ void PlayMusicRsc(ID rsc)
...
 	// NULL rsc => abort midi in progress
+   // This is vestigial but could be used by KOD to stop music.
```
akirmse, line 665: If it's really vestigial, it should be removed.

clientd3d/music.c
```diff
@@ -647,7 +782,16 @@ void MusicAbort(void)
 void MusicStart(void)
 {
    if (state == STATE_GAME)
-      PlayMusicRsc(bg_music);
+   {
+      if (latest_music != bg_music){
```
akirmse, line 786: Change this to "latest_music == bg_music" and flip the then and else clauses.

kod/object/active/holder/room.kod
```diff
@@ -1790,6 +1790,20 @@ messages:
       return;
    }
 
+   RestartMusicAfterJala()
```
akirmse, line 1793: Functions should be named for what they do, not who calls them.  Notice that this function doesn't restart anything, and has nothing to do with Jala.  One clue that this is a problematic name is that a generic Room shouldn't know anything about particular spell schools.

The danger of a name like this is that someone later notices that it actually just sends the current music to all users in the room.  Then they put a call to "RestartMusicAfterJala" in some code that has nothing to do with Jala, and it becomes super confusing.

clientd3d/music.c
```diff
@@ -146,22 +153,60 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
...
-   memset(&mciOpenParms, 0, sizeof(MCI_OPEN_PARMS));
-   mciOpenParms.lpstrDeviceType = "sequencer";
-   mciOpenParms.lpstrElementName = filename;
-   if (dwReturn = mciSendCommand(0, MCI_OPEN, MCI_OPEN_ELEMENT,
+   // Is it a background music or jala music file?
+   if (case_bg)
+   {
+      if (playing_music != 0)
+      {
+         mciSendCommand(midi_bg_music_element, MCI_CLOSE, 0, 0);
```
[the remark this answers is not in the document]
CardiovascularMike: I'm not sure about mp3s.  I only just learned how MCI works this last week.

Yes we need to get all the midis into the database though.  Testing this change means you need a ton of mana because only Jig and Spellbane songs seem to exist in my resource folder.  I've given my tester guy every node, 70 mysticism and a few mints and he can only just barely hold up Jig long enough to make sure the music loops correctly.

clientd3d/music.c
```diff
@@ -146,22 +153,60 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
 
    sprintf(filename, "%s%s", current_dir, lpszMIDIFileName);
    debug(("music filename = %s\n", filename));
-
-   memset(&mciOpenParms, 0, sizeof(MCI_OPEN_PARMS));
-   mciOpenParms.lpstrDeviceType = "sequencer";
-   mciOpenParms.lpstrElementName = filename;
-   if (dwReturn = mciSendCommand(0, MCI_OPEN, MCI_OPEN_ELEMENT,
+   // Is it a background music or jala music file?
+   if (case_bg)
```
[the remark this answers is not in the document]
CardiovascularMike: This is how I had it working about 10 days ago.  The reason for the change: Loading a midi file into memory eats up a bunch of compute power.  Its probably not an issue on modern machines but my testing environment is a Gen 3 or 4 intel I5 Dell laptop.  For me I get at least 250ms of lag while the new midi is loaded into memory.  The result is that if I am singing and start running which breaks my trance, the player gets stuck for a good chunk of time before he can run away.

The old version with only one device did work.  I've got it saved on another branch just in case.

clientd3d/music.c
CardiovascularMike: This has been fixed.  The case_bg variable was removed completely and the OpenMidiFile() function was refactored.  An argument is used to identify which device to open the Midi File to and this allowed for reduction in the amount of code required in this function.

clientd3d/music.c
CardiovascularMike: badclosebool was removed as a variable.  Instead we are just using if (MCI Function) { Error handling } else { code predicated on the mci function }.  This part was poorly considered by me but I am slowly learning.

clientd3d/music.c
CardiovascularMike: Thanks. Done.

clientd3d/music.c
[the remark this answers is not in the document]
CardiovascularMike: I've greatly updated this section - thanks for the feedback.

clientd3d/music.c
CardiovascularMike: Thanks - done.

kod/object/active/holder/room.kod
CardiovascularMike: Fixed.

clientd3d/music.c
```diff
@@ -321,38 +371,55 @@ DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
...
        return dwReturn;
      }
 
-   /* If already playing music, pick up where we left off */
-   if (music_pos != 0)
-      UnpauseMusic();
-
+   if (playing_midi)
+   {
+      // We're trying to play BG music, but Midi music is still playging.
```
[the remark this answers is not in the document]
CardiovascularMike: Fixed.

kod/object/passive/spell/jala.kod
```diff
@@ -277,17 +277,17 @@ messages:
       local oCaster, lActive, each_obj, oRoom, i;
       oCaster = Nth(state,2);
 
-      if pbUserEffects
+      if where = $
```
akirmse, line 280: Can be simplified:

```
      oRoom = where;
      if oRoom = $
      {
         oRoom = Send(oCaster,@GetOwner);
      }
```

kod/object/active/holder/room.kod
```diff
@@ -1790,6 +1790,20 @@ messages:
       return;
    }
 
+   RestartRoomBackgroundMusic()
```
akirmse, line 1793: Maybe "ResendRoomBackgroundMusic" is a better name?  There's nothing in the code here about restarting; it's up to the client to decide what it wants to do with this message.

clientd3d/music.c
```diff
@@ -647,7 +801,16 @@ void MusicAbort(void)
 void MusicStart(void)
 {
    if (state == STATE_GAME)
-      PlayMusicRsc(bg_music);
+   {
+      if (latest_music == bg_music)
```
akirmse, line 805: I don't understand this change... a comment would be good here.

clientd3d/music.c
```diff
@@ -492,51 +654,53 @@ void PlayMusicRsc(ID rsc)
...
+		  debug(("DEBUG Already playing that music.\n" ));
         return;
+        }
+        /* Playing music is true, not paused, need new bg music */
+        /* so kill the current background music before continuing. */
+        #ifdef M59_MSS
+			   AIL_end_sample( hseqBackground );
+        #else
+			   mciSendCommand(midi_bg_music_element, MCI_CLOSE, 0, 0); 
+         #endif
```
akirmse, line 672: The indentation looks weird in this section.  It's not worth the trouble to fix everywhere, but maybe you can set your editor to use spaces instead of tabs so you can line things up here?

clientd3d/music.c
```diff
@@ -132,9 +136,11 @@ void MusicClose(void)
 //	Not used by MSS version
 /*
  * OpenMidiFile:  Open midi file for playing.
+ *   For background room music opens midi_bg_music_element device.
+ *   For any other music (jala) opens midi_element device.
```
akirmse, line 140: You should generally avoid referring to game-specific terms like "Jala" or even "room" in files like this.  Suppose someone adds another spell school one day... does this function not work, because it says it only works with Jala music?  You get the idea.  The client doesn't know about game details.

clientd3d/music.c
```diff
@@ -132,9 +136,11 @@ void MusicClose(void)
...
  *   Returns 0 if successful; MCI error code otherwise.
  */
-DWORD OpenMidiFile(const char *lpszMIDIFileName)
+DWORD OpenMidiFile(const char *lpszMIDIFileName, UINT device)
```
akirmse, line 143: I'm confused what the new "device" parameter means in this function.  The new documentation talks about "midi_bg_music_element" and "midi_element" devices.  Is "device" the same as one of these?  Does it mean something else?

akirmse (commented):
Haven't actually run this yet, but here are some small things to fix first.

clientd3d/music.c
CardiovascularMike: Sorry - I didn't realize how annoying this looks on Github since it looks correct on my machine.

The good news is I can use the Github code view to see where I've messed this up and I will fix it manually.

clientd3d/music.c
CardiovascularMike: I added a comment.  The idea is the user has turned on music in the config.  We need to determine which song type to play in that instance.

kod/object/passive/spell/jala.kod
CardiovascularMike: Done.

kod/object/active/holder/room.kod
CardiovascularMike: Done.

clientd3d/music.c
CardiovascularMike: Fixed.

clientd3d/music.c
CardiovascularMike: I've fixed this everywhere.  Jala is mentioned in one comment still where I think it is an appropriate example for future coders.

clientd3d/music.c
CardiovascularMike: I've added a comment on why there is a device argument.  OpenMidiFile is used for background music and for gameplay action related music.  The argument tells us what to name the device when we open a new midi file so that we can always juggle the two MCI devices correctly regardless of any convoluted user actions.

akirmse (commented):
I tried testing this locally, and while the debug (non-MSS) version seems to work pretty well, I'm not hearing any music at all in the MSS version.  Might need some help debugging this.

akirmse:
There's still one more string concatenation to take care of (the one I noted originally).

clientd3d/music.c
```diff
@@ -321,38 +393,63 @@ DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
    DWORD dwReturn;
    MCI_PLAY_PARMS mciPlayParms;
    char temp[81];
-
-   if ((dwReturn = OpenMidiFile(fname)) != 0)
+	debug(( "Play music file %s. LatestMusic=%d. playing_music=%d, playing_midi=%d \n", fname, latest_music, playing_music, playing_midi ));
```
akirmse, line 396: split long line

clientd3d/music.c
```diff
@@ -206,29 +258,27 @@ DWORD PlayMidiFile(HWND hWndNotify, char *fname)
...
+	debug(( "Playing midi music file %s. LatestMusic=%d \n", fname, latest_music ));
 	playing_midi = True;
 	return 0;
 #else
    {
       DWORD dwReturn;
       MCI_PLAY_PARMS mciPlayParms;
-
-      if ((dwReturn = OpenMidiFile(fname)) != 0)
+      // Open file with deviceflag = 2
```
akirmse, line 277: Where did this 2 magic number come from?  Make a named constant for this instead?

clientd3d/music.c
```diff
@@ -206,29 +258,27 @@ DWORD PlayMidiFile(HWND hWndNotify, char *fname)
...
+      if ((dwReturn = OpenMidiFile(fname, 2)) != 0)
       {
-        return dwReturn;
+         debug(("OpenMidiFile error - can't open \n"));
```
akirmse, line 280: This debug message wouldn't help very much--at least include the filename.

clientd3d/music.c
```diff
@@ -145,23 +154,66 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
       debug(("Unable to get current directory!\n"));
 
    sprintf(filename, "%s%s", current_dir, lpszMIDIFileName);
-   debug(("music filename = %s\n", filename));
-
+   debug(("music filename = %s \n", filename));
+   // Is it a background music or gameplay element music file?
+   if (deviceflag == 1)
```
akirmse, line 159: Make named constants for these 1 and 2 flags.  Also maybe rename the variable, since it's not a flag.

clientd3d/music.c
CardiovascularMike: I've split this in the code.  Do you want a line break in the actual debug message as well?

clientd3d/music.c
CardiovascularMike: Thanks - I've changed this to a named constant defined by enumeration and updated the comments and documentation string.

clientd3d/music.c
CardiovascularMike: Thanks, done.

clientd3d/music.c
```diff
@@ -43,9 +47,10 @@ static BYTE *pMIDIImmediate;
 #endif
 
 enum {SOUND_MIDI, SOUND_MUSIC};
+enum {LOAD_MUSIC, LOAD_MIDI}; /* Arguments for OpenMidiFile MCI device naming */
```
akirmse, line 50: The more readable way to do this is:
```
typedef enum {LOAD_MUSIC, LOAD_MIDI} LoadMusicType;

DWORD OpenMidiFile(const char *lpszMIDIFileName, LoadMusicType musicType)
```

This way, it's clear to the caller that they're supposed to pass in a value from the enum, and not any old integer.

Although you should use a better name than "LoadMusicType", since one of the values is "LOAD_MUSIC", which doesn't make much sense.

clientd3d/music.c
```diff
@@ -145,23 +154,66 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
...
+      midi_element = 0;
+      midi_bg_music_element = 0;
       return dwReturn;
-
-   midi_element = mciOpenParms.wDeviceID;
-   debug(("midi element = %d\n", midi_element));
-   return 0;
+   }
+   // Save element ID and set flag
+   if (loadWhat == LOAD_MUSIC)
```
akirmse, line 197: Switch here too.

clientd3d/music.c
```diff
@@ -145,23 +154,66 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
       debug(("Unable to get current directory!\n"));
 
    sprintf(filename, "%s%s", current_dir, lpszMIDIFileName);
-   debug(("music filename = %s\n", filename));
-
+   debug(("music filename = %s \n", filename));
+   // Is it a background music or gameplay element music file?
+   if (loadWhat == LOAD_MUSIC)
```
akirmse, line 159: A safer way is to use a switch statement here, with a default, with a default case that prints an error.  That way if the caller does pass in an illegal value, it will be noticed.

It would in general also make it easier to add more cases in the future (though that seems unlikely for this particular enum).

clientd3d/music.c
```diff
@@ -321,38 +393,64 @@ DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
...
+            fname, latest_music, playing_music, playing_midi ));
+   // If already playing music, pick up where we left off
+   if (isMusicPaused)
+   {
+      if (paused_music == bg_music && playing_music)
+      {
+         UnpauseMusic();
+         return 0;
+      }
+      else
```
akirmse, line 406: This doesn't need to be "else" since the "then" clause always returns.

CardiovascularMike: Thanks. Fixed.

clientd3d/music.c
CardiovascularMike: This is done.  It reads better too.

clientd3d/music.c
CardiovascularMike: Done.

clientd3d/music.c
CardiovascularMike: Thanks.  Done.

clientd3d/music.c
```diff
@@ -145,23 +155,74 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
...
+            mciSendCommand(midi_element, MCI_STOP, 0, 0);
+            if (dwReturn = mciSendCommand(midi_element, MCI_CLOSE, 0, 0))
+            {
+               // Just debug don't return
+               debug(("mciClose problem for midi_element\n"));
+            }
+         }
+         break;
+      default:
+         debug(("OpenMidiFile: Invalid musicType\n"));
```
akirmse, line 179: It's a good idea to log the value of musicType here to help in debugging.

clientd3d/music.c
```diff
@@ -145,23 +155,74 @@ DWORD OpenMidiFile(const char *lpszMIDIFileName)
...
+         playing_music = True;
+         debug(("midi_bg_music_element = %d\n", midi_bg_music_element));
+         break;
+      case GAMEPLAY_MUSIC:
+         midi_element = mciOpenParms.wDeviceID;
+         playing_midi = True;
+         debug(("midi element = %d\n", midi_element));
+         break;
+      default:
+         debug(("OpenMidiFile: Invalid musicType\n"));
```
akirmse, line 214: Log musicType here too.

clientd3d/music.c
```diff
@@ -132,9 +137,14 @@ void MusicClose(void)
 //	Not used by MSS version
 /*
  * OpenMidiFile:  Open midi file for playing.
- *   Returns 0 if successful; MCI error code otherwise.
+ *   The loadWhat argument instructs the client
```
akirmse, line 140: "loadWhat" is now out of date.

akirmse (commented):
Almost there, just a few small changes.

clientd3d/music.c
CardiovascularMike: ok thanks - done.

clientd3d/music.c
CardiovascularMike: done.

clientd3d/music.c
CardiovascularMike: Good catch - I've updated this.

PR #645, opened 2024-03-13
feature: change MoveOn type from bit flags to enum

adrienlaws:
# What 🔍 
This PR changes the type of MoveOn values from bit flags to an enum type.

# Why 🤔 
- more extensibility
  - clears bit space for other flags - we don't have any for other future changes
- if this method is approved, we could restructure other flags types like other servers have

# Changes 🚧 
There are changes to both the client and server, with many changes occurring in .kod files to reference type enum instead of flags.

There are 3 areas of change
- changes in the client to accomodate the new `moveon_type` field in objects' data structure
- protocol changes `proto.h`, `server.h`, `blakston.khd`, `user.kod`, and `player.kod`
- changing `.kod` objects to use a `viMoveOn_type` `classvar` instead of a flag within `viObject_flags`

The main change from which all others stem from is a change in `blakstone.khd`

### Before
```
   %%% Object flags for client
   MOVEON_YES        = 0
   MOVEON_NO         = 0x00000001
   MOVEON_TELEPORTER = 0x00000002
   MOVEON_NOTIFY     = 0x00000003
```

### After
```
  % MOVEON type enum
   MOVEON_YES        = 0
   MOVEON_NO         = 1
   MOVEON_TELEPORTER = 2
   MOVEON_NOTIFY     = 3
```

# Attribution ✨ 

This change was inspired by skittles1's PR here for server 103 https://github.com/OpenMeridian/Meridian59/pull/833
 
I believe these changes have been incorporated in other servers over time, including 105.

# Issues ❗ 
- This is obviously a big change.  I've done my best to test but I would need more time to evaluate each case given everything it's touched.  
- At first glance, looks ok.  
  - I was able to compile my dev environment and verify normal behavior of running into objects and stopping.
  - I was also able to test portals and running into monsters to ensure `MOVEON_TELEPORTER` and `MOVEON_NO` are both working for portals and battler classes.


adrienlaws:
Closing this PR since it only provides minimal benefit for a lot of refactoring.  I might submit a different one later with a full int for additional flags, as suggested.

PR #650, opened 2024-03-17
Adjustments to render depth defaults for player/non-player

sicoderealm:
Currently players can renderer strangely with other objects in the 3d world.  This is most visible on faction flag poles where soldiers can appear both inside and outside the mast. This also manifests in other ways (such as items appear in-between player bodies).

This change improves the default z bias used and also improves how stacked items z-fighting is managed (by using is gettable).

**Current:**
image
image
image

**With this change:**
image

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -7025,9 +7025,12 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 		pChunk->xLat0 = xLat0;
 		pChunk->xLat1 = xLat1;
 
-		pChunk->zBias = ZBIAS_BASE;
-		// Nodes with a bound height adjust are part of other players' upper bodies.
-		if (pRNode->boundingHeightAdjust == 0)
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -7025,9 +7025,9 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 		pChunk->xLat0 = xLat0;
 		pChunk->xLat1 = xLat1;
 
-		pChunk->zBias = ZBIAS_BASE;
-		// Nodes with a bound height adjust are part of other players' upper bodies.
-		if (pRNode->boundingHeightAdjust == 0)
+		pChunk->zBias = (pRNode->obj.flags & OF_PLAYER) || (pRNode->boundingHeightAdjust!=0) ? ZBIAS_BASE : ZBIAS_DEFAULT;
+
+		if (pRNode->obj.flags & OF_GETTABLE)
```
akirmse, line 7030: Please add a comment explaining what's going on in these two lines.  Without that, this looks like voodoo magic.

Why was the previous comment here about player bodies removed, was it wrong?

PR #657, opened 2024-03-21
Colhorr weapon att fix

Gar103:
The equations for Colhorr weapons are incorrect. The lack of a +1 on the proficiency value means proficiency maxes out at 99 from the skill, which later leads to a truncation that cuts off a point of damage. This is why Colhorr weapons never 'felt like they did anything' - in many cases they quite literally did not. E.g. +1 damage became +0, +2 became +1. This, combined with the fact that Colhorr weapons actually _subtract_ their bonuses from low proficiency wielders, meant the community simply assumed these don't work at all.

### DIFF
--- a/kod/object/passive/itematt/weapatt/waexpert.kod
+++ b/kod/object/passive/itematt/weapatt/waexpert.kod
@@ -88,10 +88,10 @@ messages:
          return damage;
       }
       
-      Damage = damage - iPower;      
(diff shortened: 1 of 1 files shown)


akirmse:
This looks to be working as designed: the comment says it does "proficiency percent" of the power, which is exactly what it does.  But having 99 round up to 100 for an extra point seems defensible.

PR #658, opened 2024-03-22
Fix Berserker Ring inconsistency

Gar103:
Berserker Rings only add damage to melee attacks. However, attacking with a ranged weapon will still incur the hit roll penalty and drain durability. This change makes it so that the berserker ring only inflicts a hitroll penalty and drains durability on melee attacks, simply to make things consistent and less confusing.

I have also added a text line that is sent the first time a zerker ring is used, informing the player zerker rings don't work for ranged attacks. This 'first use' does cost 1 durability, if only to avoid spam on subsequent uses. Lore wise, I'd say this is the player channeling away one actual activation of the berserker ring while they focus their mind.

"You push the berserker ring's bloodlust out of your mind as you calm and focus on your distant target."

### DIFF
--- a/kod/object/item/passitem/ring/besring.kod
+++ b/kod/object/item/passitem/ring/besring.kod
@@ -35,6 +35,9 @@ resources:
 
    berserkerring_break_rsc = \
       "The berserker ring, having drunk its fill of battle, simply disappears!"
+
+   berserkerring_no_ranged_rsc = \
+      "You push the berserker ring's bloodlust out of your mind as you calm and focus on your distant target."
       
 classvars:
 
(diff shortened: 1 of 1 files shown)


Aesica:
It doesn't look like your message will show if the user attacks with a melee weapon first, then uses the bow.  It might be best to just not bother with the extra message since the rest of the fix is good.

Gar103:
True, but as long as there's some indication, it'll eventually filter into community consciousness. I could move the text message to any time you equip a bow while you have a berserker ring equipped, and any time you equip a berserker ring while you have a bow equipped. Might be a bit spammy, but just a bit.

davidklaw:
I understand where you're coming from with the new message, but I think the ring says _just enough_ to give the user room to deduce the rest. I'd like to see this fix without the extra message.

Berserker ring description:
> This ring seems to glow a warm red. It bears a simple inscription, "Lustful **blows** are strong but not always true."

davidklaw:
Was thinking over this one some and wondered if the berserker ring should indeed "punish" players for choosing ranged over melee in an effort to encourage close combat. This change would make the ring something you could wear more casually, which isn't necessarily a bad thing, but seems quite the opposite of "you want only to **KILL**", per its description. Or perhaps ranged attacks should do half the damage?

Aesica:
I don't see any reason to punish players for using it.  In practice, this change would be unwelcome and seen as an unneeded nerf, since it would mean more gear swapping to switch between ranged and melee attack modes:  equip bow + unequip berserker ring vs equip melee weapon and reequip berserker ring.  Having it go dormant on ranged attacks keeps things more true to the original while remedying the bug of charges hemorrhaging for no reason during ranged attacks.

PR #662, opened 2024-03-23
Remove defunct Gaze of the Basilisk code

Gar103:
I stumbled upon a piece of code in AssessDamage in player.kod that, when a player is damaged, removes a type of paralysis we no longer use, from the old version of Gaze of the Basilisk. There is no other reference to this anywhere else now, meaning PFLAG2_PARALYZED is a player flag that could be used by something else.

### DIFF
--- a/kod/include/blakston.khd
+++ b/kod/include/blakston.khd
@@ -2017,7 +2017,6 @@
    SID_SHALBANE = 118                  % use SID_SHALLILEBANE
    SID_WARPTIME = 122                  % use SID_WARP_TIME
    SID_DETECT_INVIS = 123              % use SID_DETECT_INVISIBILITY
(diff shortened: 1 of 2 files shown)


kod/include/blakston.khd
```diff
@@ -2154,7 +2154,7 @@
    PFLAG2_DETECT_GOOD      = 0x000004
    PFLAG2_DETECT_EVIL      = 0x000008
    PFLAG2_DETECT_INVIS     = 0x000010
-   PFLAG2_PARALYZED        = 0x000020
+   PFLAG2_PARALYZED        = 0x000020 % Not currently used by anything
```
akirmse, line 2157: Delete this flag, which will prove that it really isn't used anywhere else.

It would be a good idea to leave a comment here identifying that the value 0x00020 is currently unused.

PR #664, opened 2024-03-23
Remove too-low bounding of Jala instrument spellpower

Gar103:
Jala's notes in GetSpellpower explicitly state Jala's primary bonus is supposed to go up to 40. True Lutes also offer 40 instrument level. However, there was a bound after this that limited it to 30, meaning all this time fine lutes and true lutes were functionally identical.

### DIFF
--- a/kod/object/passive/spell.kod
+++ b/kod/object/passive/spell.kod
@@ -2167,6 +2167,13 @@ messages:
             iItemModifier = iItemModifier/2;
          }
       }
+
+      iPowerBonus = Send(Send(SYS,@GetParliament),@GetFactionSpellPowerBonus,
+                         #who=who,#theSpell=self);
+
+      % Bind the bonuses to reasonable values.
+      iPrimaryBonus = bound(iPrimaryBonus,0,30);
+      iSecondaryBonus = bound(iSecondaryBonus,0,10);
       
       if viSchool = SS_JALA
       {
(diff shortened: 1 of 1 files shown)


akirmse:
This is a balance issue; please revert that change.

Gar103:
ok done

akirmse:
Please pick a more meaningful title than "Fix bug in X".

akirmse:
This title is way too long, a good sign that this should be broken up into two changes.

Gar103:
this was changed to just one change

davidklaw:
The PR doesn't mention anything about the secondary bonus bound you added. Is that a change leftover from when this PR had a larger scope? If so, it should be reverted and handled in whatever separate PR addresses the BaseMaxHealth vs. MaxHealth change.

I'm otherwise a +1 on this change based on the simple fact that 40 was the original intention, at least as far back as this codebase goes. I'd be looking for players wiser to me to understand if Jala is already powerful enough and we should instead just revise the comment for today's state of Jala.

Gar103:
Ah, the secondary bonus bound that was added is maintaining the existing bound - the original bounding was moved just before Jala instead of after, so the Jala secondary bound now needs a 10 bound or it'll go up a couple points on the top end.

davidklaw:
Gotcha. I wonder if that higher secondary bonus was intentional, to go with Jala's higher primary bonus. But, I see what you did -- thanks.

PR #665, opened 2024-03-24
Use BaxMaseHealth instead of MaxHealth in Distill calculations

Gar103:
The calculation for potion power in Distill uses Max Health, which invisibly lowers your power if you don a circlet, and also allows you to cheese the restriction by morphing into something with high health. Switching to Base Max Health fixes that issue.


akirmse:
Please split this into two changes.

For each one, pick a descriptive title, more useful than "Fix bug in X".

PR #666, opened 2024-03-24
Set Jala potions to never expire

Gar103:
Set the Go Bad Time for player-made potions to -1 so that they never expire and don't have Go Bad timers.

Formerly, Jala potions never went bad because they are 'labeled,' and the game does not start go bad timers on labeled spell items upon creation. However, the go bad timer could be potentially be restarted if the potion was sold to an NPC and repurchased. Not that such a thing would matter, because the time calculation was also massively off, resulting in go bad times that would have been over a year in length. Jala potions have not had this mechanic for over 20 years at this point, so I figured it's best just to officially set their go bad time to -1. I also don't think it's a good idea to add such a massive number of potential timers to the game, as every single Jala potion would have its own go bad timer of a week... while people are building, that could easily mean hundreds or thousands of timers just for a mechanic that wouldn't really add anything.

### DIFF
--- a/kod/object/passive/spell/distill.kod
+++ b/kod/object/passive/spell/distill.kod
@@ -287,10 +287,8 @@ messages:
       iPotionPower = iPotionPower + Random(-20,(iSpellPower/10));
       iPotionPower = Bound(iPotionPower,5,SPELLPOWER_MAXIMUM);
 
-      % Potion has duration (In days) determined by distill spellpower and
-      %  spell level.
-      iGoBadTime = Bound((iSpellPower / iLevel) / 3,1,7);
-      iGoBadTime = iGoBadTime * 24 * 60 * 60 * 60 * 1000;
+      % Potion never goes bad.
+      iGoBadTime = -1;
 
       oPotion = Create(cPotionClass,#labelled=TRUE,#iSpellPower=iPotionPower,
                        #gobadtime=igobadtime,#maker=caster,#ability=iSkill);


akirmse:
I am going to keep hammering you on this one until it's right, and it will probably seem like ridiculous nitpicking for a tiny change.  But if you learn to do things right, your skills will improve, slowly but surely.  Most programming is 10% code and 90% communication.

Five or ten years from now, someone may find a bug with Jala potions, and go back through the change history to try to figure out what changes affected it. 

Previously, this change was buried inside a different one that said "Fix Distill oversights", so it wouldn't have stood out.  This one says "Set Jala potion..." which is good.

However, it then says "GoBadTime to -1".  That is not a good summary of the change.  The name "GoBadTime" may change in the future, and the number -1 (which is meaningless to a reader) may change.  What this change *actually* does is set Jala potions to never expire.  *That* should be the title of the change.

The description of the change repeats the title, which is redundant and should be removed.

The description contains an in-depth explanation for *why* the change was made, which is excellent and is the most important thing to include.  That's what future readers want to know.

Now whether this change is a good idea or not I don't know, and I will have to confer on Discord with those who know game mechanics better than I do.  It would be wise to do that before submitting design-related changes in the future.

Gar103:
Fair point on the titles, I have experienced that myself countless times over the years trying to figure out what the heck anyone did on Open Source and the commits are impossible to search.

I would absolutely never submit anything that actually changes design without discussion. I guess I just figured in this case it's not actually _changing_ anything, just removing doubly broken code. If I was aiming to 'fix the bugs' so to speak, and therefore completely change how jala potions handle and play, then I would have definitely started a Discord discussion first

davidklaw:
Here to say thanks, since I'm a guy searching for PRs and trying to quickly understand what changes were made 👍🏻

PR #669, opened 2024-03-26
Block Kraanan Enchantment from applying to Nerudite Swords

Gar103:
When Kraanan Enchantments wear off, they apply a ATCK_WEAP_NONMAGIC flag that permanently messes up a nerudite sword's attack type. Striking a troll, for example, will no longer make the troll shiver and suffer.

I would just fix that part, but the Kraan enchant itself applies ATCK_WEAP_MAGIC to a weapon, which gives nerudite swords both ATCK_WEAP_NERUDITE and ATCK_WEAP_MAGIC at the same time. This probably shouldn't happen, given that they are treated exclusively elsewhere, and lore-wise nerudite is an anti-magic metal.

Prior to a Kraan enchantment, a nerudite sword is neither magical nor non-magical (0x10040). The enchantment makes this 0x10044, and then 0x10042 when it wears off, with no way to fix that. A magical nerudite sword with 0x10044 will probably screw up or bypass calculations somewhere unintended. In the past, I have found nerudite swords that deal nonmagic slashing damage (0x10082) which leads me to believe either this issue compounds somewhere/somehow, or there's another problem with nerudite attack types we need to find.

In terms of gameplay, I don't think this commit changes anything. There are no situations that I could find where a Kraan-enchanted nerudite sword was necessary or advantageous for something. The ATCK_WEAP_MAGIC flag has no function in PvP at all, and in PvE, there is no monster resistance combination where having both nerudite and magic benefits in some non-trivial way.

Qor and Shal enchants still work on nerudite swords without issue as far as I can tell, and those do have uses in PvP and PvE. Lore-wise, I would suggest these shouldn't work on nerudite swords either due to nerudite's anti-magic properties, but that's a separate discussion.

### DIFF
--- a/kod/object/passive/itematt/weapatt/waench.kod
+++ b/kod/object/passive/itematt/weapatt/waench.kod
@@ -77,6 +77,11 @@ messages:
    {
       if send(oItem,@getattackspell) > 0
       { return FALSE; }
+      
+      if Send(oItem,@CheckTypeFlag,#flag=ATCK_WEAP_NERUDITE)
+      {
+         return FALSE;
+      }
 
       return TRUE;
    }
--- a/kod/object/passive/spell/enchwp.kod
+++ b/kod/object/passive/spell/enchwp.kod
@@ -21,6 +21,8 @@ resources:
    enchantweapon_fails = "The %s is strangely unaffected by the spell."
    enchantweapon_already_done = \
       "This weapon is already dedicated to Kraanan."
+   enchantweapon_cannot_enchant_nerudite = \
+      "Nerudite cannot be enchanted in such a manner."
 
    enchantweapon_name_rsc = "enchant weapon"
    enchantweapon_icon_rsc = ienchant.bgf
@@ -93,6 +95,11 @@ messages:
       {
 	 Send(who, @MsgSendUser, #message_rsc=enchantweapon_already_done);
 	 return False;
+      }
+      if Send(oWeapon,@CheckTypeFlag,#flag=ATCK_WEAP_NERUDITE)
+      {
+         Send(who,@MsgSendUser,#message_rsc=enchantweapon_cannot_enchant_nerudite);
+         return False;
       }
       propagate;   % Check other things higher up
    }


jrotunda:
Does the troll not get effected even moreso by a Kraanan enchanted nerudite sword? I could've sworn using Kraanan enchantment made these weapons stronger.

For historical context - nerudite weapons when first developed, and for most of their history, used to do this, "Your enchantment bounces back - !" if you tried to enchant them to Qor, Shal'ille or Kraanan. There was a  patch sometime between NDS to now that allowed them to be enchanted.

Gar103:
Good question, Meridian's resistance system only takes 'best of' for weakness and vulnerability. So two vulnerabilities never combine into a bigger vulnerability. Interestingly enough, the troll does not have a weakness to magic weapons. It is simply resistant to all nonmagic weapons and -80 weak to nerudite. This is why it becomes simply a normal weapon against trolls once the Kraan enchant wears off - Nerudite + Nonmagic = 80% vulnerability + 80% resistance = 0% effect.

You know, I was thinking nerudite weapons used to be impossible to enchant. I was surprised this was currently possible. I personally would like to see them made impossible to enchant with any type, because I'd like to see more 'unevenness' in the weapon system. If nerudite swords could not be Qor or Shal enchanted, then the only thrust weapons that could actually be qor or shal enchanted would be short swords and long swords. Obviously in PvP people are not going to be using short swords and long swords, so this would make high end thrust weapons exclusively physical, in a meta where it's so common to always be attacking with a physical + spell type through one of the two enchants.

Aesica:
Just chiming in to say that they should probably just go back to being 100% unenchantable, as that's how I've always remembered them and prior to reading this, didn't know you *could* enchant them.  I'm not sure why this was changed, but as mentioned above, having them be unenchantable again gives players a reason to use a long sword or short sword if a player wants a qor or shal thrusting weapon.  Unfortunately, thrust as a damage type is pretty lousy, but that might not always be the case.

Gar103:
I'd be happy with unenchantable, too

jrotunda:
I don't think we should take this away, personally. I'm not convinced it's hurting anyone and makes the weapon have a little more versatility; why it was added to begin with.

davidklaw:
I'm +1 for unenchantable. I don't know how many players look at changes like this the way @Gar103 is describing, but it's helpful to understand it in those terms and therefore see it as an effort to balance weapon choices. Makes me wish we had in-depth stats for PVP/PVE weapon attacks, but I'll keep dreaming :)

Gar103:
What kind of in-depth stats would you like? I've got whole spreadsheets of calculations 😅

PR #676, opened 2024-04-02
Show the damage caused with the hit message

rotshine:
Added a parameter, `battler_attacker_hit`, to store the damage value. Updated the `AssessHit` function to display the damage to the user.

### Why?

The purpose is to inform the user about the amount of damage being dealt to the monster, enhancing the player's awareness of their own attacks.


akirmse:
It's intentional that we don't show numbers in most cases.  It's just the style of the game.

PR #677, opened 2024-04-02
Move the camera 90° Up/Down

rotshine:
Updated the `HEIGHT_MAX_OFFSET` to enable the camera to tilt up or down by 90°.

### Why?

This enhancement allows players to freely explore their surroundings, providing them with additional options to navigate the game world.


akirmse:
The current limits are intentional.

Issue #643, opened 2024-02-27
System: Message for first time user appeal [Req.]


jrotunda:
**Request:**

As a DM or ADMIN, we see " ### has just logged on for the first time." when a new user logs in.

More often than not it's a newbie, although it may also be an established player on an alternate account - but when in doubt, help! 

To continue building the social nature of the game - helped by the additional setting to adjust broadcast mana, this would be a beneficial message to be seen by the players themselves, for veterans who should be meeting and greeting the newbies of Meridian, and for newbies to receive a warm welcome.

current code is below:

   % Returns a running count of player's logged in time, across sessions
   %  Used by quest engine to check how much time a player has spent online
   %  since performing a quest
   GetTimeLoggedIn()
   {
      if piLastLoginTime = 0
      {
         return 0;
      }

      if pbLogged_on
      {
         return (GetTime() - piLastLoginTime + piTimeLoggedIn);
      }

      return piTimeLoggedIn;
   }

   SendFirstTimeAppeal()
   {
      Send(self,@UserAppeal,#first_time=TRUE);
      
      return;
   }

   UserLogon(session_id = $)
   {
      local i;
      
      poSession = session_id;
      pbLogged_on = TRUE;

      if piLastLoginTime = 0 AND piLast_Restart_Time = 0
      {
         Debug("Sending first time appeal");
         Send(self,@SendFirstTimeAppeal);
      }

      piLastLoginTime = GetTime();
      % Gotta load interface early!
      Send(self,@UserLoadModule,#module=user_interface_module);

      Send(self,@MsgSendUser,#message_rsc=user_logon);

      Send(SYS,@SystemUserLogon,#what=self);

      if NOT (IsClass(self,&DM) AND Send(self,@IsHidden))                            ################### GUT IT?
      {
         % Advertise as normally.
         Send(SYS,@SystemUserLogonAdvertise,#what=self);
      }
      else
      {
         % We don't want to be seen, except by                                     #################
         for i in Send(SYS,@GetUsersLoggedOn)
         {
            % Only do this for admins
            if IsClass(i,&Admin)
            {
               Send(i,@SomeoneLogon,#what=self);   

user_first_time_appeal = "### %s has just logged on for the first time."****

From: 
**Meridian59/kod/object/active/holder/nomoveon/battler/player/user.kod**

davidklaw:
@Meridian59 I think this can be closed based on the results of #684

PR #684, opened 2024-04-18
feature: make first logon notification global

adrienlaws:
# What
- This change makes the `"### [player name] has just logged on for the first time.` message visible to all logged on players.
image

# Why
- Request from issue #643
- This lets users know when a new player has logged on for the first time so they can provide support or greetings

# Changes
- Created new function `SendFirstTimeLogonMessage()` in `user.kod` to handle new character advertisements
  - Filtered for DMs (do not advertise new DM characters)
- Simplified `UserAppeal()`
  - Removed boolean `first_time` paramter from `UserAppeal()`
- Removed `SendFirstTimeAppeal()` from `guest.kod`


kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -3159,18 +3159,19 @@ messages:
       lUsers = Send(SYS,@GetUsersLoggedOn);
       for i in lUsers
       {
+         if first_time
```
akirmse, line 3162: As hinted at by the name, this is meant to be a mechanism to inform DMs that a new user has joined, kind of like an "appeal" command.  If this goes to all users, it's no longer an appeal, and thus this broadcast doesn't belong here.  You should go up higher in the stack and modify/rename the "Appeal" functions where appropriate.

Also, I don't think that DMs logging in for the first time should be broadcast.  e.g. event-specific characters that are meant to be kept secret.

Do most people want this?  As a player I think I would consider it pretty spammy.

adrienlaws:
Good feedback, thank you both.  I tried to incorporate the technical feedback and went up the stack to create new function `SendFirstTimeLogonMessage()`

In regards to
> Do most people want this? As a player I think I would consider it pretty spammy.

It could indeed be used to spam people with an un-ignorable message.  Valid point.

However, I think it wouldn't be more spammy than death messages since this message should only occur rarely.  There are mitigations for rerolling and spamming this message (reroll timer, admin intervention, etc).

Perhaps @jrotunda can comment here since he created the issue this PR was trying to address.

The intent was to welcome new players and provide them guidance in their first moments online, especially during our recent population surge.  But if it's too controversial we can just close this and focus on less controversial changes (IMO).

jrotunda:
As a social experiment, for the last *6 years* I've been copying and pasting the automatic DM logon appeal then sending it out to users. It hasn't been reported as spammy, nor has it had any detrimental effect on the atmosphere, and no one has ever replied as it being "annoying" to be welcomed, whether or not they can reply - the effect of knowing there are others out there and aware of their existence is a positive one in an online game.

The players who may not 'like this' are ones who are trying to hide and avoid attention: basically, throwaway characters, mules, sock puppets, pk alts, etc - but this is also conjecture and not anything that's been reported or witnessed in 6 years.

Without fail, a user, usually multiple, will reply a greeting in broadcast, as well as an offer to help the user. Whether or not this goes to tells or goes further, the player base then knows "this user is new and is going to be rough around on the edges" and knows to give them a greeting, whether they can reply or not, and a hand later. 

It's pretty well understood that replying in broadcast or tell won't happen immediately (except from some veterans who forgot what it was like).

- **These messages create a burst of social activity in broadcasts and lead to crucial new player & experienced player interactions. 
- 
- These messages help a DM or Admin in promoting and maintaining a healthy and welcoming environment, as well as ease work flow during the onboarding process when Meridian has population influxes.**

jrotunda:
I hear what Gar is saying and agree...but more for the experienced player's response. I have iterated this message to veterans - let the newbie figure out how to reply and move around, be patient, have empathy and not have a quick-to-judge "Well I guess they're not friendly!" thought process.

It is confusing when you first log on, and are trying to figure out how to chat and leave the room.  The issue there is more on the veteran/experienced player in their patience and expectation. I do not think the new player is getting irritated and have seen more feedback about the welcoming environment and appreciation of both a greeting and assistance from others - one of the strong points of our community.

davidklaw:
Dropping a comment here to +1 this based on @jrotunda 's points. The chat window is often quite quiet aside from combat messages. The occasional "new player logged in" notification, conversation it might create, and potential upside to the new player experience all lead me to 👍🏻

davidklaw:
Bump!

adrienlaws:
Closing this PR after poll results on the 101/102 Discord.

It looks like there's a significant population of voters against this change.  Even though "Yes" is dominant, I think it isn't enough.

```mermaid
pie title Poll Results on 101/102 Discord
         "Yes" : 19
         "No" : 10
         "Indifferent" : 6
```

PR #686, opened 2024-04-20
Add GetTickCount function for greater timing accuracy

Gar103:
This is my first pull in a step by step attempt to fix 'melee lag' as people call it, by reverse engineering other pull requests and things done on open source. 

This pull intends to add GetTickCount, which will allow kod to access timing in ms.

However, I can't build it, and I need help figuring out why.  This is the error I get when trying to build it:

Making in .\blakserv
sendmsg.c
ccode.c
.\sendmsg.c(160): error C2440: '=': cannot convert from 'int (__cdecl *)(int,local_var_type *,int,parm_node [],int,parm_node [])' to 'ccall_proc'
.\sendmsg.c(160): note: This conversion requires a reinterpret_cast, a C-style cast or parenthesized function-style cast
.\ccode.c(1971): error C2220: the following warning is treated as an error
.\ccode.c(1971): warning C4244: 'return': conversion from 'blak_int' to 'int', possible loss of data
NMAKE : fatal error U1077: 'cl -nologo -DBLAK_PLATFORM_WINDOWS -DWIN32  -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE  -D_WINSOCK_DEPRECATED_NO_WARNINGS /wd4996 /wd4312  -TP -WX -GR- -EHsc- -MP -Zi -W3 -DBLAKDEBUG /MT /FpCpch /Fddebug\vc90.pdb /Fodebug/ -c .\sendmsg.c .\ccode.c ' : return code '0x2'
Stop.
NMAKE : fatal error U1077: 'nmake -nologo /S                  ' : return code '0x2'
Stop.

I don't know what this means, can't look it up anywhere, and it looks identical to GetTime() in every way but the function's internal code. What am I missing?

### DIFF
--- a/blakserv/ccode.c
+++ b/blakserv/ccode.c
@@ -1939,6 +1939,38 @@ blak_int C_GetTime(int object_id,local_var_type *local_vars,
 	return ret_val.int_val;
 }
 
+blak_int C_GetTickCount(int object_id,local_var_type *local_vars,
+			  int num_normal_parms,parm_node normal_parm_array[],
+			  int num_name_parms,parm_node name_parm_array[])
+{
+	val_type ret_val;
+	
+	// GetMilliCount is from blakerv/time.c. 
+	// Its return is in ms and with a precision of 1ms.
+	// It also provides Windows & Linux implementations.
+	UINT64 tick = GetMilliCount();
+	
+	// but tick is unsigned 64-bit integer
+	// and blakserv integers are signed with only 28-bits
+	// the high-bit is the sign at bit-index 27/31
+	// recapitulate:
+	// 0x00000000 = 0000 0000 0000 0000 0000 0000 0000 = 0
+	// 0x07FFFFFF = 0111 1111 1111 1111 1111 1111 1111 = 134217727
+	// 0x08000000 = 1000 0000 0000 0000 0000 0000 0000 = -134217728
+	// 0x0FFFFFFF = 1111 1111 1111 1111 1111 1111 1111 = -1
+	
+	// convert:
+	// 1) We grab the low 32-bits by casting to unsigned int (so next & can easily be done in 32-bit registers)
+	// 2) We grab the value within the positive blakserv-integer mask by &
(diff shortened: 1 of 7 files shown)


blakserv/ccode.c
```diff
@@ -1939,6 +1939,38 @@ blak_int C_GetTime(int object_id,local_var_type *local_vars,
 	return ret_val.int_val;
 }
 
+int C_GetTickCount(int object_id,local_var_type *local_vars,
```
adrienlaws, line 1942: change the return type to `blak_int`

`blak_int C_GetTickCount` instead of `int C_GetTickCount`

blakserv/ccode.h
```diff
@@ -193,6 +193,10 @@ blak_int C_GetTime(int object_id,local_var_type *local_vars,
 	      int num_normal_parms,parm_node normal_parm_array[],
 	      int num_name_parms,parm_node name_parm_array[]);
 
+int C_GetTickCount(int object_id,local_var_type *local_vars,
```
adrienlaws, line 196: Here also, change `int` to `blak_int`

adrienlaws:
Change your return type (I put it in other comments) from `int` to `blak_int` and it will build.  I don't know if will work (after building) or much about this change otherwise but that will give you a starting point.

See here as to why:
https://github.com/Meridian59/Meridian59/blob/40ba7c014195d21d4938a83173facd3717c9bf0b/blakserv/blakserv.h#L153

Gar103:
Awesome, that worked. Honestly I should have guessed 😓

Gar103:
Ok, so testing locally, this does work, even with the changes to values described in that link. My current understanding is that this, combined with a couple other pull requests (one where i'll allow a higher packet rate and improve spam detection, and one where i'll actually increase the update rate) will drastically cut 'melee lag' and make movement updates much more responsive.

Gar103:
Hmm, Open Source must have changed something to do with integers elsewhere in the code then. I don't know why this still works (I tested it locally and found it to returning ms accurate timing), or how to proceed from here.

PR #687, opened 2024-04-21
Change Hunt to only work on outlaws and murderers

Gar103:
Change Hunt to only work on outlaws and murderers. As predicted in [this pull from 11 years ago](https://github.com/Meridian59/Meridian59/pull/204), Hunt has been used by veteran PKs to harass specific newbies to an egregious degree. This should really be a tool to hunt PKs, not to be used by them. We did this on OS and it generally improved the overall situation. Although the PK / hunter / builder overall system still needs tons of work in other ways.

I'm putting up this change again because, after doing some research, Hunt has caused a much bigger headache for everyone. Looking back at Discord, the IP limit for all players was permanently lowered to 3 because of one PK using Hunt to harass a player last year. That's extremely inconvenient to the rest of us, so we need a fundamental fix for this so we can remove that bandaid. To elaborate a little more on that frustration, an IP limit this low cuts off entire pieces of content - prism spells - which apparently it was actually intended to do at the time I guess - but on top of that, some of us are trying to play w/ family members, and juggling accounts has become onerous and frustrating. I'm really hoping fixing Hunt can let us revert to classic settings.

For some context, I actually came back to Meridian after a recent server crash where the IP limit was accidentally reverted to 6. So that's all I knew. That does still cut off the use of 7-sided prisms, but it felt fair. It didn't feel frustrating.

### DIFF
--- a/kod/object/passive/spell/multicst/hunt.kod
+++ b/kod/object/passive/spell/multicst/hunt.kod
@@ -44,6 +44,9 @@ resources:
    hunt_lost_trackers = "You seem to have managed to lose your trackers!"
    hunt_gain_trackers = "You sense that someone knows where you are..."
 
+   hunt_must_be_outlaw_or_murderer = \
+      "The soul of an innocent has no negative energies to track. The spell fails."
+
 classvars:
 
    vrName = Hunt_name_rsc
@@ -103,6 +106,16 @@ messages:
       local i, oTarget;
 
       oTarget = First(lTargets);
+
+      if NOT Send(oTarget,@CheckPlayerFlag,#flag=PFLAG_MURDERER)
+         AND NOT Send(oTarget,@CheckPlayerFlag,#flag=PFLAG_OUTLAW)
+      {
+         for i in Send(oPrism,@GetCasters)
+         {
+            Send(i,@MsgSendUser,#message_rsc=hunt_must_be_outlaw_or_murderer);
+         }
+         return;
+      }
 
       for i in Send(oPrism,@GetCasters)
       {


jrotunda:
Q: Does this fix the issue with the hunt not 'leaving' the targets for a long period of time?

Barring any further revisions to Hunt's design, this is the simplest fix for a high-cost advanced spell, used mainly by players with an advantage, to hurt others that they shouldn't be able to, usually by a PK, and not used by a group of players trying to hunt or avoid one.

jrotunda:
This change is simple and effective. In Internal development discussions with @sicoderealm , who's a proponent on keeping the 'edge' intact and small changes to these systems, we agreed that Hunt should:

* require that you have safety off, turning orange to cast 

I am not sure about removing the ability for reds to be involved in the cast.

I think a little nuance here is needed that makes it less easy for outlaws and murderers to use - if not disabled for now as that seems to be controversial.

* scale cast time/power based on karma and PvP flags: lower cast time if white, higher orange, higher red 
* lower cast time it player has active kill based on kill count decay

Gar103:
I'm confused, why should I have to go orange to hunt outlaws & murderers?

davidklaw:
I've given myself a bit of a crash course in Hunt here today and think there's more to consider (and fix), but I'm in favor of this change if Hunt is being abused in this way.

Perhaps this description should be updated with something about outlaws (and murderers):
> "Allows the casters to use a prism to focus on **_a player's_** trail and track him or her for a limited time.  The hunted player's name must be spoken during the initial trance.  When the enchantment is completed, for the duration of the spell the hunters will receive periodic notification of the hunted player's whereabouts.  Requires all of the claws from one kriipa paw, and the casters' mystical energies must be focused through a prism.

davidklaw:
Related issue #807 - Hunt is now disabled on account of implementation issues. We can certainly still have this conversation and bring this change in, but maybe we want to punt until it gets reworked.

Gar103:
I saw the disabling - instead, for the time being, we could just put a check for the target being online at the time of firing? In PrismCast in hunt.kod, If Send(oTarget,@IsLoggedOn). Fail the spell, with a failure message for the casters, if the target is not online. It does cut some of the functionality, because it was great to cast it on offline targets to be warned the moment they logged on, but it shouldn't be any more problematic than that because targets won't know a Hunt cast is being built up, so they won't know to log off to fizzle it

davidklaw:
The thought crossed my mind, but I assumed we'd then have a situation where a few folks would know how to take beat it -- e.g. if you've got a target on your back, just log off and on between fights or every so often. Then I thought, dial the duration down so that's less likely to happen, but figured that'd kind of run counter to all the time and energy that goes into casting it.

It'd be nice to solve the offline/online timer issue instead 🤷🏻

PR #688, opened 2024-04-22
Prevent karma changes in the arena

Gar103:
Since reflections and evil twins will soon be functioning in the arena, I got in there and tested them out. Turns out, if you and your opponent are both high karma in the same direction, you will lose the ability to cast Karma-based spells above level 3 after only two reflection kills. Killing reflections in the arena should not affect karma, so I put in a check to prevent that. I put in the broadest prevention possible to shortcut any problems in the future - the Arena isn't real death combat, so Karma should never be changed by actions inside it.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -6357,6 +6357,12 @@ messages:
       {
          return 0;
       }
+      
+      % No karma change in arenas.
+      if Send(oRoom,@IsArena)
+      {
+         return 0;
+      }
 
       % No karma changes from killing neutral monsters.
       if bIsMob AND karma_victim = NEUTRAL
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/tostwn/tswatch.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/tostwn/tswatch.kod
@@ -199,7 +199,7 @@ resources:
 
    TosWatcher_summoned_reflection_killed = "%s spoils an illusion, laying bare the deception of %s!"
    TosWatcher_summoned_twin_killed_other = "%s vanquishes the evil twin of %s!"
-   TosWatcher_summoned_twin_self_kill = "%s overpowers his evil twin!"
+   TosWatcher_summoned_twin_self_kill = "%s overpowers their evil twin!"
 
    TosWatcher_arena_help = \
       "Proceed all subcommands with the DM command.  'Start Tournament' "


CardiovascularMike:
Thanks for finding this oversight.  A few comments:

1. Reflections probably shouldn't effect karma anywhere.  That said I still want to see this change implemented at this current scope because reflections in the arena would be especially troublesome for a tournament if you are trying to keep your karma.
2. Both arenas have an IsArena() message that returns true which we could use instead of roomSafeDeath.  I'm not sure which is better but I thought I would offer the option.  IsArena returns true for the Tos Arena and Goad's Grinder and FALSE for any other room.

Thanks!

adrienlaws:
I think having reflections affect your karma is undesirable, generally.

However I think there might be an edge case with `earthquake `and `blind`/`dazzle` where it is desirable for balancing purposes.

As it stands now
- If you have someone invisible casting `earthquake` on the map (especially a large or complicated map) they will have their karma affected when their `earthquake` casts kill the reflections.
- Since reflections are "0" karma, the caster will lose their ability to `purge`, `dazzle`, or `blind`.

If you change reflections not to affect karma
- You could have a scenario of `purge`/`dazzle`/`earthquake` and it might be a bit cheesy.
  - Similarly with `blind` and `earthquake`.  
- It could make things like arena tournaments strange where you hide for most of the match and then `dazzle`/`purge`/`eq`/`eq`/`eq`/`eq` while not having to fight the person directly.

**_Not a strong objection from my side_** but something to consider.  I suppose in "real world" or "non-arena" PVP, you would be killing the monsters on the map unless it's a town, so your karma would end up adjusted anyway.

Aesica:
This is a solid PR, imo.  Seal of approval.

> I think having reflections affect your karma is undesirable, generally.
> 
> However I think there might be an edge case with `earthquake` and `blind`/`dazzle` where it is desirable for balancing purposes.

I don't think this is anything to worry about for a few reasons:

- Earthquake's damage diminishes with distance, so if A and B are on opposite ends of the same large map, A's earthquake isn't going to be doing much to B.  A will run out of mana long before B dies.
- Blind, dazzle, and earthquake as a combination is fairly hard to maintain on one, or especially more than one person since earthquake has a cast time, and all 3 have fairly significant mana and vigor costs.

If in fact this is some sort of problematic gameplay combination, I don't think allowing reflections to gimp the blind/dazzle/earthquake user's karma is the right approach to solving it.

jrotunda:
Since this discussion came up - and this should be addressed in another PR...

Another scenario that does occur:

Player A is ghosted. Player B casts a damaging wall spell on the ghost.
Player A on a 2nd character (A2) brings a Riija character that is outside of the PvP scenario between Player A and B.
Player A2 casts their reflections and has them die on the firewall, ruining the karma of Player 2.

Personally, I do not think Evil Twins, Apparitions or Reflections should affect karma. I do not think "ruining the other player's karma" should be a meta-strategy to promote.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -6357,6 +6357,12 @@ messages:
       {
          return 0;
       }
+      
+      % No karma change in arenas.
+      if Send(oRoom,@CheckRoomFlag,#flag=ROOM_SAFE_DEATH)
```
akirmse, line 6362: To check if a room is an arena, use IsArena()

Gar103:
Changed to @IsArena

PR #689, opened 2024-04-24
Add base items for Neo Ao3s & SotHs

Gar103:
Per enduring and popular demand, I'd like to essentially build an entirely new Necromancer & Hunter scenario (as I've done before elsewhere), but that's a rather complex process. In this first step, I'm adding internally named 'Neo' Amulets of the Three and Swords of the Hunt. These are separated from the Necromancer Balance utility functions and have had almost all of their code removed - for now.

I've tried for many years to decipher the unspecified 'problem' with the Necromancer scenario that necessitated its deactivation. Ultimately, it's just inscrutable, with mechanics we don't want to keep (including many factors that are reliant to and tied to the number of PKs per day). We really need to just ditch the NecBal utility and create something new, with a keen eye on the lessons learned from Soldier Shields.

Broad overview of the intended scenario:

* No limit on the number of Ao3s and SotHs in the game
* No spawning on the ground, instead you ask the Goad for an ao3 or Tendrath for a SotH
* HP requirement remains, or adjusted higher (100?)
* No karma requirements for either - there are Qor hunters and Shal PKs, karma does not equate to good or evil, a theme which is supported in other places in the game as well
* No relation to innocent, outlaw, or murderer status anymore
* Rebalanced stat boosts (if any)
* A much simpler hunger system that basically just requires wielders to go out and kill mobs daily
* Most importantly, Ao3 and SotH users will be able to attack each other legally under certain circumstances. In the previous Ao3/SotH scenario I built, this was simply always, but I think 101's style would be more suited to a situation where Ao3 wearers are attackable when dispel illusion reveals that they are undead - and we'll make it so that only a Hunter's dispel illusion _in a combat area_ can reveal this. The Hunter that dispelled then also becomes attackable by the Necromancer(s) he revealed. So the 'hunt' would involve traveling to places where people are building and dispelling to discover whether they're secretly undead.

I would also like to rebuild the Necromancer Guild, a subset of Ao3 users that raises and serves the Lich and uses the Brax guild hall. When she is killed, only Necromancer Guild members die and have their loot dropped there, rather than all ao3 users. I'd also fix all the bugs, so killing her is not something that can be cheesed.

I believe we can also come up with special and interesting powers and reasons to take these items to participate in a new form of 'legal' and thematic PvP, since Soldier Shields are great, but have become rather massively lopsided toward Jonas faction.

There should be some overarching goal that the Necromancers are trying to accomplish, and the the Hunters are trying to prevent, similar to the flagpole faction game. And the Necromancer Guild itself can have its own nefarious goals.

### DIFF
--- a/kod/object/item/passitem/necklace/makefile
+++ b/kod/object/item/passitem/necklace/makefile
@@ -5,6 +5,6 @@
 !include $(TOPDIR)\common.mak
 
 DEPEND = ..\necklace.bof
-BOFS = shadowam.bof qrcharm.bof krcharm.bof shcharm.bof rjcharm.bof frcharm.bof
+BOFS = shadowam.bof qrcharm.bof krcharm.bof shcharm.bof rjcharm.bof frcharm.bof neonecroam.bof
 
 !include $(KODDIR)\kod.mak
--- a/kod/object/item/passitem/necklace/neonecroam.kod
+++ b/kod/object/item/passitem/necklace/neonecroam.kod
@@ -1,0 +1,154 @@
+% Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
+% the LICENSE file that accompanies it.
+%
+% Meridian is a registered trademark.
+
+
+NeoNecromancerAmulet is Necklace
+
+constants:
+
+   include blakston.khd
+
+    % Four hours
+    HUNGER_TIME = 240000000
+%   HUNGER_1 = 2   %one session (assuming 3-4 hours of play per day)
+%   HUNGER_2 = 4   %one to two sessions
+%   HUNGER_3 = 6   %two sessions
+%   HUNGER_4 = 8   %two to three sessions
+%   HUNGER_5 = 10  %three sessions
+
+   %BASE_HUNGER_GAIN = 2
+
+resources:
+
+   NeoNecromancerAmulet_name_rsc = "Amulet of the Three"
+   NeoNecromancerAmulet_icon_rsc = necamlet.bgf
+   NeoNecromancerAmulet_desc_rsc = \
+   "There is something deeply disturbing about this amulet.  When you look at it in the corner of your eye, "
+   "you can almost see the evil pouring forth from it."
+
+   NeoNecromancerAmulet_newbie_pickup_good = "Your guardian angel says, \"That amulet is a thing of great "
+      "evil.  Even holding it in your possession would corrupt your soul, and eventually destroy you.\""
+   NeoNecromancerAmulet_newbie_pickup_evil = "Your guardian angel says, \"That amulet holds far too much "
+      "power for you to contain.  You must stay away from it, for now.\""
+
+   NeoNecromancerAmulet_holding_damage_wearing = \
+   "The unholy energy streaming from the Amulet of the Three seeps into you, racking your body with pain as it "
+   "corrupts your soul.  You must do something before it destroys you!"
+   NeoNecromancerAmulet_holding_damage = \
+   "The unholy energy streaming from the Amulet of the Three seeps into you, racking your body with pain as it "
+   "corrupts your soul."
+   NeoNecromancerAmulet_holding_unworthy_wearing = \
+   "The unholy energy streaming from the Amulet of the Three seeps into you, filling you with pleasure even as it "
+   "racks you with pain.  You must somehow remove it before it destroys you!"
+   NeoNecromancerAmulet_holding_unworthy = \
+   "The unholy energy streaming from the Amulet of the Three seeps into you, filling you with pleasure even as it "
+   "racks you with pain.  You feel a strong compulsion to put the amulet on and channel its full power, but you "
+   "know somehow that it would surely destroy you if you did."
+   NeoNecromancerAmulet_holding_tempt = \
+   "The unholy energy streaming from the Amulet of the Three seeps into you, filling you with pleasure.  You "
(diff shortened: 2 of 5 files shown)


Aesica:
Several thoughts, questions, and concerns about this:

- What happens when the lich dies and the necromaners die with her?  Do all the soths and ao3s vanish like in the original scenario?  In the past, the major failing of the hunter/necromancer scenario was that holding onto the soths and ao3s had more value than killing the lich.  On the servers I played on, lich kills were mostly "for the lulz" to cause trouble and drama. People just wanted the soths and ao3s.
- You just ask Tendrath for these and he pops one into your inventory?  I don't think I like the idea of another loot dispenser for a really good item.  Soldier shields are kind of bad enough in that regard, as they're the best shield in the game, but only require minor effort to obtain once you have the right HP.
- The stat effects were migrated to soldier shields.  If these come back, there should be zero stat changes on them.
- What was wrong with the existing hunger system?  It was basically that, except not daily if I recall.  Now that many m59 players are older and busier, "kill stuff daily" is kind of a big ask.
- Why create two new items that are just like the originals instead of fixing the originals?
- With no karma or murderer/outlaw restrictions on the sword of the hunt, why would anyone choose the ao3?  The amulet was an inferior item to the soth for a number of reasons:  vulnerability to silver arrows, god gift/amulet slot permanently occupied, dispel betrayed that you had one, and it didn't give you a free weapon to use or not use at your discretion. 

Overall, there's several balance related issues plaguing these two items and the overall execution of the scenario before these should be brought back into the game.  Several changes have already been made to the game to facilitate a post hunter/necromancer scenario meridian that need too be taken into account as well.

Gar103:
**What happens when the lich dies and the necromaners die with her? Do all the soths and ao3s vanish like in the original scenario?**

That's a good question. I think the 'value' of simply having these items will be gone, since anyone can get one from Goad or Tendrath, so we don't need to remove them all when the lich dies.  I think the lich, the actual Necromancer Guild, and their objective should be different from just general Ao3 use. For example, the Goad may give Ao3s freely to eventually gather enough energy to raise Brax back into the world and/or wake the Ancient One, while the Lich might have specific political agendas (infiltrating the factions perhaps, trying to secretly take over those that are vying to become king/queen)

**You just ask Tendrath for these and he pops one into your inventory? I don't think I like the idea of another loot dispenser for a really good item. Soldier shields are kind of bad enough in that regard, as they're the best shield in the game, but only require minor effort to obtain once you have the right HP.**

We can have players complete a quest, but I don't think these should necessarily be viewed as 'items' per se. They're a visible method of opting in to a PvP scenario, so their powers are part of your role in that scenario.

**The stat effects were migrated to soldier shields. If these come back, there should be zero stat changes on them.**

I don't necessarily want stat changes on these things either. Previously, SotHs gave 'fighter' stats and reduced int and myst, and Ao3s gave 'caster' stats and reduced might and aim... I don't like that pigeonholing. That was made even worse by Ao3s being intended for negative karma, and for orange/red, and soths being positive karma and only useable by innocents. So you basically had to be a Shal fighter type for soths, and a qor mage type for Ao3s. 

I think spells and abilities would be more interesting. For instance, Hunters could get a single-target Greater Dispel Illusion, the only spell capable of exposing a Necromancer. Necromancers could have nefarious abilities, like a buff that lets them secretly steal XP from those nearby. That way, you get people paranoid and potentially accusing each other, exactly the way Necromancers want... "Man, my tougher's taking awhile. Is this guy building with me secretly a Necromancer and stealing my XP??" Stuff like that... hunters being intrusive and overbearing bringers of 'order,' and Necromancers trying to secretly sow chaos.

**What was wrong with the existing hunger system? It was basically that, except not daily if I recall. Now that many m59 players are older and busier, "kill stuff daily" is kind of a big ask.**

It's an inscrutable mess. By daily hunger I meant online time. Every 4 hours of online time (basically 2 meridian days) it goes up one hunger stage. That should lead to quite a long time frame that isn't too annoying (20 hours online time = 1 full hunger refill, which could be any number of kills we want). I don't really like a hunger system at all, but it's part of the theme.

**Why create two new items that are just like the originals instead of fixing the originals?**

They files are a gigantic mess linked to objects and timers that already exist on the server (NecBal and any old soths and ao3s from yesteryear). I feel like it's best not to touch that, especially since a pull request that has dozens of pages of deletions and changes is just going to get rejected for being too complex.

**With no karma or murderer/outlaw restrictions on the sword of the hunt, why would anyone choose the ao3? The amulet was an inferior item to the soth for a number of reasons: vulnerability to silver arrows, god gift/amulet slot permanently occupied, dispel betrayed that you had one, and it didn't give you a free weapon to use or not use at your discretion.**

Well sure, the soth was better, but we'll fix that. The bonuses of both need to be vastly different than before. We can definitely make them equal.

**Overall, there's several balance related issues plaguing these two items and the overall execution of the scenario before these should be brought back into the game. Several changes have already been made to the game to facilitate a post hunter/necromancer scenario meridian that need too be taken into account as well.**

I completely agree, which is why I aim to take this step by step so people can be on board with each brick of the foundation. I envision a series of pull requests, such as 'Ok, the items exist. Now, what abilities do we want them to give?' and we'll debate on what exactly they'll do.

akirmse:
Design should come before code.  I'd be willing to look at this again once there's an agreed-upon design.

Gar103:
How do we begin that process? Who must agree? Discord doesn't seem to be working for complex design discussion topics, and I don't think any two people will fully agree on a single design when discussed as large hypotheticals. It seems like the kind of thing where the person in charge needs to say 'let's try design X,' we actually put it in even if might not be perfect, then every so often, after some live experiences and player feedback, we keep iterating it. The community would probably be pretty excited for a process like that. Everyone loses their minds over soths and ao3s 😅

PR #691, opened 2024-04-25
Revert xlat on mug of stout, goblet of wine, and bowl of soup

Aesica:
Not long ago, these items (mug of stout, goblet of wine, and bowl of soup) were recolored to set them apart from items using the same image (mug of brew, goblet of ale, bowl of stew).  After seeing them in game for awhile with their new colors, it's become apparent that this isn't the best way to set them apart from their look-alikes.

This PR restores them to their original colors.  It would still be nice to give these 3 items a distinct look of their own, however it should be done through a unique image rather than recoloring images that weren't set up to use the xlat system in the first place.

image
Recolors on the left, original colors to be restored on the right.

### DIFF
--- a/kod/object/item/passitem/numbitem/food/gobletwn.kod
+++ b/kod/object/item/passitem/numbitem/food/gobletwn.kod
@@ -44,7 +44,6 @@ properties:
    viFilling = 8
    viNutrition = 6
    piNumber = 1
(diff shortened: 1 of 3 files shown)


Gar103:
Yeah have to agree, I'd personally like to keep everything as close to classic as possible. We remember and treasure these little details. The off color mugs of stout really bug me more than they should!

Aesica:
Yeah me too almost immediately after seeing them on live.  I was hoping it was just me being weird, but nope! :D

PR #693, opened 2024-04-27
Server settings for greatly diminishing throwaway PK effectiveness, and for protecting low hp players from them

Gar103:
Added a setting to scale damage dealt to innocents by murderers based on their hp below 100.

Added a setting to scale damage by low hp murderers based on their hp below 100.

So this is something that absolutely needed to happen 20 years ago, and 10 years ago, and today. Meridian's design has a fundamental problem at low levels. Completely unstoppable throwaway PKs can run in and kill newbies in 1-3 hits before they have time to react or even understand what's happening. This has led to a parasitic subculture of a select group of players who are absolutely unwilling to actually invest any time or energy into the game, but absolutely willing to roll up dozens of 30 hp tofers to grief newbies until the server empties out. This is a cycle we've seen over and over for most of our lives at this point. I could explain at great lengths how lopsided the time investment is for hunters to try to stop these guys, or how easy and effortless it is for the PKs because they're taking zero risks, but let's just take that as a given because it's been discussed to death. We're already slowly losing the new influx of players to this behavior.

This pull shifts the damage at low levels to a far more reasonable balance. I'm not preventing PK behaviors with this, I'm cutting their damage to something newbies can actually react to. A 40 hp player being attacked by a 30 hp ToFer will essentially take 1 or 2 damage instead of 12 or 13. The damage penalty softens the more hp the victim has, and the more hp the PKer has. A 100 hp+ murderer attacking a 100 hp+ victim will not suffer any penalty. In essence, this forces our parasitic subculture to either put in some actual time investment (and therefore risk) or stop cheese griefing our newbies.

To be honest, these PKs will still get some kills. Newbies are bad at the game, that's why they're newbies. However, they might actually have a chance to react, understand what's going on, and possibly even defend themselves. The murderers aren't protected the same way, so throwaway PKs will die in a couple hits if the newbie actually manages to turn around and hit them with their mace.

I surely expect this to be contentious, and it's not the only idea out there, but something's got to change. Nothing feels more awful than watching an influx of new excited players slowly stop logging in because they're being attacked constantly by ridiculous naked throwaway characters nobody can do anything about.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -4429,7 +4429,8 @@ messages:
    AssessDamage(what = $,damage = $, atype = 0, aspell = 0, stroke_obj = $,bonus = 0,
               report = True, report_resistance = TRUE, absolute = FALSE)
    {
-      local i, iResistance, oSoldierShield, color_rsc, iDuration, iLimit;
+      local i, iResistance, oSoldierShield, color_rsc, iDuration, iLimit, iMyMaxHP,
+            iPKMaxHp;
 
       color_rsc = player_hit_color_none;
       
@@ -4498,6 +4499,27 @@ messages:
             AND Send(oSoldierShield,@IsEnemyAttack,#what=self,#damage=TRUE)
          {
             damage = (damage * 115)/100;
+         }
+         
+         % Diminish throwaway PK damage output, and diminish PK damage to low hp innocents        
+         if NOT Send(self,@CheckPlayerFlag,#flag=PFLAG_OUTLAW)
+            AND NOT Send(self,@CheckPlayerFlag,#flag=PFLAG_MURDERER)
+            AND (Send(what,@CheckPlayerFlag,#flag=PFLAG_OUTLAW)
+               OR Send(what,@CheckPlayerFlag,#flag=PFLAG_MURDERER))
+         {
+            iMyMaxHP = Send(self,@GetBaseMaxHealth);
+            iPKMaxHP = Send(what,@GetBaseMaxHealth);
+            
+            if Send(Send(SYS, @GetSettings),@DiminishPKDamageToNewbiesEnabled)
+               AND iMyMaxHP < 100
+            {
+               damage = (damage * iMyMaxHP)/100;
+            }
+            if Send(Send(SYS, @GetSettings),@DiminishPKMuleDamageEnabled)
+               AND iPKMaxHP < 100
+            {
+               damage = (damage * iPKMaxHP)/100;
+            }
          }
       }
 
--- a/kod/util/settings.kod
+++ b/kod/util/settings.kod
@@ -146,6 +146,11 @@ properties:
    % ability to ask questions and socialize.  (Puts a damper on trivia
    % contests too!)  This value can be changed to tweak the penalty in realtime
    piBroadcastManaCostPercent = 50
+   
+   % Greatly diminish the damage output of throwaway outlaws and murderers,
+   % and the damage of PKs against low-hp innocents.
+   pbDiminishPKDamageToNewbies = TRUE
+   pbDiminishPKMuleDamage = TRUE
 
 messages:
 
@@ -352,6 +357,16 @@ messages:
    GetBroadcastManaCostPercent()
    {
       return piBroadcastManaCostPercent;
+   }
+
+   DiminishPKDamageToNewbiesEnabled()
+   {
+      return pbDiminishPKDamageToNewbies;
+   }
+
+   DiminishPKMuleDamageEnabled()
+   {
+      return pbDiminishPKMuleDamage;
    }
 end
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%


davidklaw:
Great idea! 

- Did you consider whether this ought to apply when a white player attacks another? E.g. a throwaway pkers first kill.
- Might be useful to allow admins to adjust that 100 hp threshold as well.

Gar103:
**Did you consider whether this ought to apply when a white player attacks another? E.g. a throwaway pkers first kill.**
Yes, outlaw status is lumped in with murderer for this, so it should be covered

**I’m not convinced the issue to be as broad as suggested. **

So, I hear you, but I think at the end of the day I fundamentally disagree with the basis of the argument. I don't see any valid reason why a character with under 100 hp should be PKing and expecting to be welcomed or favored by the game system. It may feel to you like it took significant effort to build a Faren character with 60 hp, I'm not disagreeing with your lived experience, but anyone hunting that character or killed by that character is only going to feel annoyed and disheartened. The 60 hp character has nothing on the line. No pride, no vast array of skills and spells losing %s, no real loot needed to operate. By design, it's a character intended to avoid taking real risk. After a death, that 60 hp Faren character will currently function exactly the same. Its Blast of Fire will not be diminished in a noticeable way, it will have only lost some robes, orc teeth, reds, and elders. In comparison, someone like Vinsfeld is kitted out with 4 schools whose spells/skills would all lose 2%, weapons, armor, reagents for a vast array of abilities, a hard-won HP off the top end of the game's building spots, and a recognisable name you'll get to put in your bio. He's going to actually fight with a variety of tactics - while a Faren-only 60 hp character cannot do much but get blinded, take a few arrows, and log. I can't see any intention there to actually fight fairly. Expecting the community to invest countless hours fruitlessly chasing down these kinds of PKs has empirically never worked. It's not even a question of hypotheticals, the community has died repeatedly over the years in no small part to this fatal flaw in the game's design.

The damage change should be broad. Newbies need time to react, understand, and figure out what's going on. If we only diminish touch spells, PKs will just switch to splash spells or wall spells. Low HP characters should take minimal damage from all sources so that they have a significant amount of time to process what's happening. I fully believe that even with 30 HP tofers doing 1 or 2 damage, they'll still manage to kill newbies. The difference will be vast, though, in that the newbies will have had time to understand the attack and choose a course of action - run or fight. If they die, they at least got to participate. Currently, they're dead before they can even react.

Meridian's best PvP is between mid-builts and above. Meridian's worst PvP involves massive imbalances, such as a fully built PK attacking a 30 to 50 hper, or a throwaway PK with nothing on the line attacking newbies. Both of these settings mitigate that somewhat without hard preventing the behavior. The newbies are still likely helpless and going to lose the fight, but they'll get _time_, and time will give them a chance to think and learn, rather than just shooting straight to the Underworld having no idea what happened.

I see a few possible outcomes from choices made by PKs in response to this.

* They may choose to invest time and build real characters with 100 hp or more to PK with
* They may choose to avoid attacking low hp newbies because it isn't such a quick and easy kill anymore
* They may choose to seek other more fair types of combat

In any of these cases the community is much better off. Meanwhile, classic forms of PKing that are actually engaging, like attacking farmers at OOK or builders in brax / the avar village / the mollusk beach, are completely unchanged. These will generally be players who know what's going on and can actually fight back. That's what we should be encouraging... all of this punishment heaped on new players has historically never worked out for us.

davidklaw:
I don't have an argument against what you've laid out here. Seems well thought out and a solution to a periodic problem that I believe does plague the game as it is today: a place where new players enter and quickly depart because of friction of all types, this easily being one of the most frustrating imo.

adrienlaws:
Interesting PR and I think the spirit is in the right place.  It's a very provocative idea.  Toko's ramblings and suggestions are below.

Thinking about it, you might consider flipping the damage issue around.  Instead of lowering damage for the Pker, increase damage taken for the Pker.

I thought of 2 ways you could do this.
- Remove the 1/3 max HP damage cap
- Change soldiers behavior to HATE outlaws and murderers

###  1/3 damage cap problem
One of the more well-intentioned changes that perhaps has led to this situation of 20-30 HP disposable Pkers is the 1/3 max HP as damage per hit.  This is set in `player.kod` right at the top in constants.

```
% Damage is capped at piBase_Max_Health divided by this number.  Reduces newbie slaughter.
MAX_HEALTH_DAMAGE_FRACTION = 3
```

The problem with this is that it artificially boosts the "real" HP of a 20 HP touch of flame Pkers to 60 HP.  I believe most "max damage" is around 20 HP, and the PKER is 20 HP, but there's no way to kill them in one hit.  It will always take 3.  If they took full damage, they will probably die en route or die in literally one hit due to being so weak.

### Soldier behavior
An opportunity here is to ... actually make the soldiers do something useful.

In `troop.kod` there is an amazingly named function called `SpecialHatredBehavior` (I love it)

```
SpecialHatredBehavior(what=$)
{
   if isClass(what,&User) or isClass(what,&FactionTroop)
```
I suggest you change this to check if the player is an outlaw or murderer and have them attack on site.  And not just one, but all, so that walking near towns as a 20-30 HP Pker almost guarantees death.  Soldiers do spawn near most newbie areas, and if there's a gap, add new spawns.  Then you have a sort of automatic militia.  Right now they're just standing around... thinking I guess.

Toko to Faction Soldiers currently:
image

Final thought:
I think your idea is good, just it would be better to hurt the PKer more or add more risk, rather than nerf damage output which I consider somewhat sacred.  Meridian has always been a brutal game, so we can increase the brutality for the ones trying to kill other players.  It's also just so much more amusing to have those disposable Pkers die almost immediately, repeatedly.

Gar103:
I'm a big fan of faction troops attacking oranges and reds, making them actual 'outlaws.' Unfortunately I think that has to go on top of everything else, if it goes in. Doing more damage to throwaway PKs doesn't really solve the problem, as they already log or die in 1-2 hits. The problem is catching them in the first place, or the fact they can die, pop out of the Underworld across the mainland, and rush right back out at newbies. I want to deal massive damage to the entire concept of running around on a naked 0 year old character with a dumb name. This just shouldn't even be a viable thing, especially when the brunt of it is borne by new players, who will assume the rest of the game is as poorly designed and silly.

**, rather than nerf damage output which I consider somewhat sacred. Meridian has always been a brutal game**

You know, I was thinking about this exact tenet recently. Who exactly are we preserving this brutality _for?_ The only people who benefit from being able to essentially instakill newbies with riskless throwaway characters are the absolute worst elements - a small handful of people who don't contribute, don't actually play, and don't even act as meaningful antagonists. Not to keep using Vinsfeld as an example, but there is a legendary name that you can die to early, vow to get revenge, build up, and one day hunt him down. That's good conflict. The same can't be said of ROfL RoFl ToFL, the 34 hp tof mule. Even if you do manage to turn the tables and kill him, he just laughs at you. He's back at it thirty seconds later.

I argue the brutality is preserved with the PR's change. Newbies can still *be* attacked by all these PK types. They'll just have slightly more time to understand what's happening.

For some math:

Let's say Yang is a 43 HP newbie is at CV, he's at 28 hp because he's fighting a skeleton. Currently, he can take 14 damage a hit maximum because of the 1/3rd max hp damage cap. Yang's first opportunity to find out he's under attack is when a ToF hits him for 14. He has roughly one second to realize what's happening and log before the second hit for 14 kills him. Essentially, from his point of view, this was basically an instant death. Running or fighting back were both impossible options.

With the PR's changes, if Vinsfeld attacks Yang, Yang can only take up to 6 damage per hit (6.02 truncated). With 28 hp, Yang can survive 5 hits. Again, he'll only find out he's under attack when the first one hits him. So, instead of 1 more hit, he'll have 4 more hits. We go from 1 second to react, to 4 seconds to react. Yang is still **absolutely** going to lose this fight, but he has some crucial time to try to run or log. He could hypothetically extend his survival time by running. Sure, he'll get held or blind, but at least Yang gets to do _something._

Now, if ROfL RoFl ToFL, the 34 hp tof mule tries the same thing, Yang can only take a maximum of 2 damage. In this case, Yang can survive 14 hits (and maybe more with a little regen). Suddenly, Yang's in a position to potentially fight back! He has 15-30 seconds to realize what's happening, run around a bit, weigh his options, then start swinging his mace. He could hypothetically actually win this fight, even if he's terrible. Of course, he probably won't win against a skilled vet, but this kind of fight will feel vastly more fair to a newbie than the current instant death. And if ROfL RoFl ToFL wants to deal more damage, he's always got the option of actually playing the damn game and getting some HPs. There's the investment. There's the risk and putting something real on the line.

adrienlaws:
Good points.

I like the idea that this will cause PVP to last longer, because once someone is defeated it's just over usually.

One thing I was thinking about too is that most servers (and please correct me if I'm wrong) do not do full loot drops anymore when someone is killed.
  - 101/102 still have that very brutal element in place.
  - It's probably super disheartening to lose all your progress from a moment's inattention while grinding so this change would help.

If I was still trying to PK with these nerfs, then I would try and increase the speed of attacks if I can't increase damage output.  
- I would probably try and cause someone to panic and run into a firewall since those allow rapid tick damage if someone runs through them.  
- My understanding is those ticks would be nerfed as well (which is good)
- That makes for a more interesting PVP engagement.

The only abuse I can think of from this change is if you have a low HP healing mule or dazzle mule fighting a guilded murderer in guild PVP or something.  The mule can just purge and dazzle and will be more difficult to kill since incoming damage will be reduced significantly, and their teammates can be hitting the outlaw/murderer hard.  Maybe having a check to see if they are enemy guilds would make sense, or people will just never join a guild war if their outlaw or murderer flag is set.  Kind of an edge case and maybe not worth worrying about.

Overall I think this is a step in the right direction since it makes it harder for throwaway 20 HP Pkers.

jrotunda:
A lot of discussion, don't want this to get bogged down. Si, please contact me in private as I have relevant information regarding this change, and I did discuss this with Gar before he began coding it. I'm personally impressed by how well-thought out and designed this is. **When in doubt - Go Boolean.**

This is a good PR in its original form. This PR doesn't contain changes to any other systems, no added angels, no huge gameplay change for users to get familiar with, no changes to soldiers or the landscape.

Gar, can you address Si's comment here "Perhaps add these conditionals to a new function with a suitable name (e.g. attackingInnocent)" or contact him regarding any concerns?

**I think this is an important and current issue critical to the health of the live server and future play, it that needs to be addressed. Having to warn people about their behavior when the game allows them to act this way, requiring moderation - is not conducive to my time, personally, or Ouija's.** 

It won't stop the behavior either - but it will make it harder for a newbie to get slaughtered by some asshat they never met, quitting the game before a DM can warn the Career PK _(who never quit despite many promises to the contrary over the last 20 years - they're dedicated fans...)_

We have a community of players. To log on and never interact with them, and use throwaaway characters and built characters together to hurt newbies (Vinsfeld just killed a 31hp character today - brand new, that we all spent time helping - who logged off, probably forever earlier). _**This change would have saved that newbie, or at least given them a chance.**_

This gives newbies a little extra time. A lot of these players also have very poor connections. We need to fix this out of respect for our new, supportive and non-destructive player base that really love this game, want to keep playing, leave good reviews, have patience, help others, and are creating a whole new fan scene the likes of which we haven't seen since the 90's. 

We can monitor the effect on gameplay thereafter.

PR #697, opened 2024-04-30
feature: add client support for item rarity grade system

adrienlaws:
# What
- Adds support for item rarity grades in the client
- Client parses a full INT from server to populate a new object member called `rarity`
- `obj.rarity` is used to determine how to color item text and whether to apply suffixes
- This change also allows for item text to be colored in dialog boxes (look, get, etc).
- There are 5 new rarity grades with associated colors in dialog and list boxes:
  - magic
  - rare
  - legendary
  - unrevealed
  - cursed
- Extensibility: the rarity object member could later be used for other presentations if people have better ideas
  - Minimap blip?
  - Customized object description boxes?
  - Things I haven't thought of
- Important: a client object is only assigned ONE rarity grade value at a time
- Note: this change only implements rarity presentation capability in the client, it requires the server to send the proper rarity grade information for any visible changes

# Why
- To make it easier to decide what items to pick up
- It can be difficult for players to identify items with magical properties
- Many items on the floor make it tedious for the user to click on each one and read the description
- Rare items should be highlighted

# Other changes
- Added wrapper to `ExtractObject `to reduce code duplication with `ExtractObjectNoLight`
- Configured `player.kod` and `user.kod` to send a blank INT so this code is shippable even if the server side is not configured

# Testing
- No errors or debug messages observed in limited testing

# Attribution / Inspiration
- skittles1's PR here https://github.com/OpenMeridian/Meridian59/pull/502

# Notes
- This PR is a refactoring of #490 which contained too many changes.
  - Several more, smaller PRs will be submitted to capture the changes of that original PR


clientd3d/dialog.c
```diff
@@ -651,7 +651,18 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
...
+	info.rarity       = obj->rarity;
 	
+	std::string raritySuffix = GetRaritySuffix(info.rarity);
+
+	if (!raritySuffix.empty())
+	{
+        char nameWithSuffix[MAXNAME + 15];
+		snprintf(nameWithSuffix, sizeof(nameWithSuffix),
+		   "%s (%s)", info.name, raritySuffix.c_str());
+        info.name = nameWithSuffix;
```
akirmse, line 663: This is a serious bug.  info.name is pointing to nameWithSuffix, which is on the stack and immediately goes out of scope.  You will get garbage text and the game will sometimes crash when this memory is later overwritten.

One fix would be to make info,name a std::string, but I'm not sure how big a change that is.

clientd3d/dialog.c
```diff
@@ -663,6 +674,43 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
 }
 /************************************************************************/
 /*
+* GetRaritySuffix:  Provide a rarity suffix based on the given rarity type.
+*/
+std::string GetRaritySuffix(int rarity)
+{
+	std::string suffix;
```
akirmse, line 681: You don't really need a variable here.

clientd3d/lagbox.c
```diff
@@ -310,7 +310,8 @@ void Lagbox_Command(HWND hWnd, int id, HWND hwndCtrl, UINT uNotify)
 		0/*flags*/,
 		pszLagboxDescription, 
 		szTooltip,
-		pszLagboxURL);
+		pszLagboxURL,
+		ITEM_RARITY_GRADE_NORMAL);  // Required rarity grade parameter
```
akirmse, line 314: This comment doesn't add anything.

clientd3d/ownerdrw.c
```diff
@@ -68,23 +68,39 @@ void WindowEndUpdate(HWND hwnd)
  */
 int OwnerListAddItem(HWND hwnd, object_node *obj, int index, Bool combo, Bool quan)
 {
-   char *name, desc[MAXNAME + 15];
-   int pos;
+   std::string raritySuffix;
+   char nameWithSuffix[MAXNAME + 15];
+   char nameWithQuantity[MAXNAME + 15];
+   char nameWithItemInUse[MAXNAME + 15];
```
akirmse, line 74: Weird mix of std::string and char[] here.  Can you make them all strings?  Then you don't need to worry about the size of the buffer, etc.

clientd3d/dialog.h
```diff
@@ -55,6 +55,7 @@ typedef struct {
    char        *description;
    char        *fixed_string;
    char        *url;
+   int          rarity;      /* Item rarity grade */
```
akirmse, line 58: Shouldn't the type of this be item_rarity_grade instead of int?

clientd3d/object.h
```diff
@@ -79,6 +79,7 @@ typedef struct {
 		                     this field gives amount of object */
    DWORD     temp_amount;         /* Scratch field used when user is selecting amount of object */
    int       flags;               /* Flags describing various properties of objects */
+   int       rarity;              /* Used to draw object text differently based on rarity tiers */
```
akirmse, line 82: Type should be item_rarity_grade?

clientd3d/dialog.c
adrienlaws: Removed the suffix variable and changed the return to a GetString call.

clientd3d/dialog.h
adrienlaws: Yes, correct.  I changed the type from int to item_rarity_grade.

This was a learning experience for me.  I figured since enums are implemented as integers for practical purposes, this was fine.

I did some research after your comment.  I now understand that enums are distinct types from integers for type-checking purposes, even though they are implemented as integers.

clientd3d/object.h
adrienlaws: Yes, same as the previous instance. I have changed it accordingly.

clientd3d/ownerdrw.c
adrienlaws: Changed all `char` to `(std::string)` as recommended and it seems to work well now.  

However I had to convert the final name string to a `LPCTSTR` `char*` since the ListBox API requires that type of parameter.

akirmse:
It looks like DescDialogStruct is used only in dialog.c.  So changing the name field to a std::string should involve very few changes, and only to that file.   Want to give that a try?

adrienlaws:
> It looks like DescDialogStruct is used only in dialog.c. So changing the name field to a std::string should involve very few changes, and only to that file. Want to give that a try?

Yes, I can try that.  

I did try it once in another commit but I wasn't successful.  This is the commit where I attempted to change it.
https://github.com/Meridian59/Meridian59/pull/697/commits/58ccd79411f5146e7376c91dd0c917be6a81011d

I will try again, slower this time.

akirmse:
I see.. that's actually not too bad.  Those "consts" should arguably be there anyway.

clientd3d/dialog.c
adrienlaws: Converted info.name to std::string as suggested.

clientd3d/dialog.c
```diff
@@ -647,22 +647,66 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
 	ZeroMemory(&info,sizeof(info));
 	info.obj          = obj;
 	info.flags        = flags;
-	info.name         = LookupNameRsc(obj->name_res);
+	// Load the name into a temporary C-style string
+    char* nameCStr = LookupNameRsc(obj->name_res);
+    info.name = std::string(nameCStr);; // Convert C-style string to std::string
```
akirmse, line 652: Extra semicolon.
Not a useful comment.

clientd3d/dialog.c
```diff
@@ -647,22 +647,66 @@ void DisplayDescription(object_node *obj, BYTE flags, char *description,
 	ZeroMemory(&info,sizeof(info));
 	info.obj          = obj;
 	info.flags        = flags;
-	info.name         = LookupNameRsc(obj->name_res);
+	// Load the name into a temporary C-style string
```
akirmse, line 650: Not a useful comment.

clientd3d/dialog.c
adrienlaws: Removed comment.

clientd3d/dialog.c
adrienlaws: Removed comment.

PR #698, opened 2024-05-02
Zero bound on Princess Soldier Shield mana cost reduction

Gar103:
At low Faction Rank levels, the Princess Soldier Shield is actually *increasing* mana costs by 1. Given that nothing else related to Soldier Shields is ever a negative, this must be an oversight. In this PR, I put a zero bound on it so it will no longer actually make you perform worse.

### DIFF
--- a/kod/object/item/passitem/defmod/shield/soldshld/prinshld.kod
+++ b/kod/object/item/passitem/defmod/shield/soldshld/prinshld.kod
@@ -153,7 +153,7 @@ messages:
    {
       local iReduction;
 
-      iReduction = ((piFactionRank/3) - 1);
+      iReduction = Bound(((piFactionRank/3) - 1),0,$);
 
       return iReduction;
    }


akirmse:
This change doesn't do what its description says.  Where is the 2 coming from?

Gar103:
2 is the highest the mana reduction can be. piFactionRank maxes at 10, /3 = truncated to 3, -1 = 2

akirmse:
Why is the princess shield enforcing the maximum faction rank?  Remove the 2.

Gar103:
I changed it to no upper bound, but to clarify the information, the piFactionRank there is a property of the soldier shield, it's the thing that determines how much bonus the shield gives, it's not enforcing a separate faction related value

akirmse:
I think you are not seeing my point.  Let's say one day someone decides to make factions stronger, so that piFactionRank could take on larger values.  Months or years later, someone has the feeling that the princess shield isn't quite as strong as it should be when the faction rank is high.  It turns out this is because its influence was capped at 2 for no particular reason.  And good luck to the person debugging the problem, because the PR that introduced the bug doesn't mention the cap in its title or description, nor is there a comment in the code describing where this magic 2 came from.  Is it important?  Will changing it break something else? 
 They'd have no way to know.

I am trying to help you to foresee the consequences of your changes so they are not so onerous to review.

Gar103:
I didn't miss the point. I'm aware of it. I disagree - from my perspective, an equation like ((piFactionRank/3) - 1) tells the prospective programmer nothing. I'm constantly having to decipher these things by digging deep into other files and functions. A bound like 0 to 2 tells me at a glance that this value is between 0 and 2 - this is consistently done in many other places, like in spell values, and when the code is written this way, I appreciate it, because I don't have to go on an archaeological dig to figure out what I'm looking at. So that's what I was thinking with a 'let's update the formatting of this' mindset initially.

If someone later wants to raise piFactionRank's maximum limit on soldier shields, they should absolutely parse through where it's used and see what effect that would actually have. If the code needs to be commented more, I guess, sure, that could be done too.

akirmse:
You're entitled to your opinion of course, but if you code that way in this repo, your reviews are going to take a long, long time.

PR #699, opened 2024-05-03
feature: add Blakod support for item rarity grades

adrienlaws:
# What

- Adds logic to evaluate an item's rarity
  - magic
  - rare
  - legendary
  - unrevealed
  - cursed
- Follow on to #697 and the 2nd of 3 refactored PRs from #490
- Adds an outbound int which determines the rarity grade in the client through `ToCliObject` in `user.kod`
- Note: does not contain individual item attribute rarity configuration or item class configuration.
  - A follow on PR will contain the baseline configuration of rarity assignment
  - This PR gives Blakod the capability to assign, but does not assign rarity grades besides `cursed`,`unidentified`, and the default of `normal`

# Why
- Rarity grade assignment should follow a structured process and be extensible

# Notes
- In the future (assuming this PR is merged) we can assign rarities in 3 ways
  - Statically setting a class variable `viRarity` to something other than `NORMAL` for specific items like MSH
  - Modifying the `GetRarity `message to return a specific class assignment based on conditions
    - Unidentified rings or potions
  - Assigning a positive or negative value to `GetRarityCountModifier` to an item attribute (for example `WA_Blinder `or blind weapons can add +1)
  
# Logic Diagram
See the diagram below for the flow chart
- It first checks to see if the object is of class &Item to avoid needless checks
- `item.kod` has a class variable of `viRarity` set by default to `NORMAL` or 0
- Only one rarity grade is assigned at a time
- First checks for unidentified, then cursed, then counts up desirable item attributes, compares that to the class assignment, and then returns a rarity grade
  - The class assignment of rarity grades or `viRarity` will supersedes the count of rare item attributes
    - The reason for this is you don't want a Magic Spirit Helm to appear as "magic" if it is bonded with 1 desirable item attribute count when the class viRarity will be set to Legendary
    - If the count of item attributes exceeds the class `viRarity` count then we go with that.
  
```mermaid
flowchart TD
   A[ToCliObject]--> B{Object = &Item?}
   B -->|Yes| C{Unidentified?}
   B ---->|No| N[Send Rarity Grade Normal]
   C -->|No| D{Cursed?}
   C -->|Yes| U[Send Rarity Grade Unidentified]
   D-->|Yes| M[Send Rarity Grade Cursed]
   D-->|No| E{Rare Item Attribute Count > Class Rarity?}
   E-->|No| O[Send Class Rarity Grade]
   E---->|Yes| F{Count >0?}

   F-->|Yes, Count =1| G[Send Rarity Grade Magic]
   F-->|Yes, Count =2| H[Send Rarity Grade Rare]
   F-->|Yes, Count >=3| I[Send Rarity Grade Legendary]
```


kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -2414,7 +2414,15 @@ messages:
       }
 
       AddPacket(4,iFlags);
-      AddPacket(4,0); % Item rarity grade
+
+      if IsClass(what,&Item)
```
akirmse, line 2418: A little less fragile in the face of future changes would be this:
```
    rarity = 0;
    if IsClass(what,&Item) {
      rarity = Send(what,@GetRarity);
    }
    AddPacket(4, rarity);
```

i.e. there is only one path that adds bytes, so there's nothing to get out of sync

kod/object/item.kod
```diff
@@ -633,6 +690,46 @@ messages:
       return FALSE;
    }
 
+   GetMagicRarityGradeCount()
+   "Checks plItem_attributes for HasRarity() = TRUE."
+   "For each attribute with HasRarity = TRUE,  add 1 to count"
+   "Return either viRarity or count, whichever is larger"
+   "This ensures the higher rarity is chosen if in the case "
+   "of competing rarities for an item."
```
akirmse, line 698: Put periods at the ends of sentences or this string will look like a mess in admin commands.

kod/object/item.kod
```diff
@@ -633,6 +690,46 @@ messages:
...
+
+            if Send(oItemAtt,@HasRarity)
+            {
+               count = count + 1;
+            }
+         }
+      }
+
+      % Use assigned class rarity grade if it is higher than the count
+      if viRarity > count
```
akirmse, line 719: I'd use Bound() here and then have only one return statement.  Less fragile if code is added in the future.

kod/object/item.kod
```diff
@@ -633,6 +690,46 @@ messages:
...
+      count = 0;
+
+      if Send(self,@HasAnyAttribute)
+      {
+         for i in plItem_attributes
+         {
+            iNum = Send(self,@GetNumFromCompound,#compound=First(i));
+            oItemAtt = Send(SYS,@FindItemAttByNum,#num=iNum);
+
+            if Send(oItemAtt,@HasRarity)
```
akirmse, line 711: "HasRarity" is kind of an odd construction.  I can think of two possibilities:

1. IsRare() or AddsToRarity() would be clearer if this is going to be a boolean.
2. You could have this function return an int, and sum up all the rarity values.  That would allow an attribute to contribute more than 1 if that's ever wanted.  But if that's overkill, a boolean is also fine.

kod/object/item.kod
```diff
@@ -613,6 +615,61 @@ messages:
       return 1;
    }
 
+   GetRarity()
+   {
+      local count;
+      count = 0;
```
akirmse, line 621: It doesn't look like this value is ever used.

kod/object/active/holder/nomoveon/battler/player/user.kod
adrienlaws: Modified to be less fragile as suggested.

kod/object/item.kod
adrienlaws: Fixed.  I did not know the impact of this docstring on admin commands.  I have adjusted the spacing and periods accordingly.

kod/object/item.kod
adrienlaws: Done.  

However, I had to use the format of `bound(count, viRarity, $)`

From what I read in the manual, it looks like it should not be in this format.  It should be:
`bound(n ,a, b) = bound(count, $, viRarity) = max( min(count, $), viRarity)`

I do want `max(n, b)` which is `max(count, viRarity)` so I am slightly bothered why that doesn't work.

But it seems to work from my change above in testing.

>Bound: Takes an integer (call it n) and two values which can be integers or nil. Call the
first of these a and the second b. If a and b are both nil, then this function returns n.
If a is nil and b is an integer, then it returns max(n,b). If a is an integer and b is nil,
then it returns min(n,a). If both a and b are integers, returns max(min(n,a),b).

kod/object/item.kod
adrienlaws: Refactored `HasRarity` as suggested in option 2 to `GetRarityCountModifier`

It now returns a int of local variable count.  By default it returns count=0 unless there are modifiers in the future for an item attribute.

I also changed the description to say it is a modifier and could be either positive or negative.  Perhaps at some point in the future we will want to add an inferior item category.

kod/object/item.kod
adrienlaws: Removed `count = 0` since it was pointed out it is never used.

adrienlaws:
I have added a commit to address all the comments.

There is one thing I would like to highlight about the current state of this PR.  Since `IsCursed` always shows if an item is cursed and `IsIdentified` only checks item attributes, rings and potions are not showing up as unidentified.  In the case of a cursed ring, even though it is unidentified it is assigned the CURSED item rarity grade.

I was going to modify `ring.kod` to have the following code in the 3rd and final PR (and perhaps something similar for unidentified potions)

```
   GetRarity()
   {
      if vrName <> vrRealName
      {
         return ITEM_RARITY_GRADE_UNIDENTIFIED;
      }

      if Send(self,@IsCursed)
      {
         return ITEM_RARITY_GRADE_CURSED;
      }

      return ITEM_RARITY_GRADE_NORMAL;
   }
```

My question: should I add this in to this PR so that if we don't reach the 3rd and final PR where we assign rarity grades to classes, that cursed items are not always visible even if they should be?  Or should I wait to handle all the classes and desirable item attribute assignment in that final PR?

TLDR; unidentified cursed rings show up as cursed and visible to the player, should I add code to show them as unidentified in this PR?  Or wait until final PR where we handle rarity assignment for classes and attributes?

kod/object/item.kod
akirmse: Cool!

kod/object/item.kod
akirmse: You are correct and the manual is wrong.  The manual source has been fixed, but I haven't gotten around to regenerating the PDF of it (which is a hassle since it's LaTeX).

akirmse:
I don't think you want to make this change only for rings---isn't it true that any unidentified cursed item should show up as unidentified?  You can certainly override IsIdentified (rather than rewriting GetRarity) for particular subclasses if necessary.

I think you should do that in this PR, since otherwise the state of the checked-in code would be something we wouldn't want users to see.

adrienlaws:
Ok I think this is fixed.  Rings and Potions will correctly show up as unidentified if they are, and will show unidentified if they are cursed (until revealed or identified).  

So in short, I think this PR works in showing anything unidentified (rings, potions, or item attributes), along with providing the logic to assign rarity grades based on class assignment through viRarity, item attribute rarity calculations, and whether the item is cursed.

image



Toko rambling and nuance below:

Overrode `IsIdentified` in both `ring.kod` and `potion.kod`
- For rings I added a check to see if the real name matches the displayed name.
  - I used `propagate` because it is still possible a ring may have an item attribute and that attribute may be unrevealed.
- For potions I used an existing message `IsLabelled` to determine if it is identified
  - Used a simple `return FALSE` if `IsLabelled` is FALSE for potions

Note on potions
- My change above seems to work well for any unidentified potions.
- There are some potions that appear generic or are just not "unidentified" but appear very simple
- Here is the list of potions that are not identifiable, revealable and do not show up as unidentified
  - pungent elexir - nodeburst
  - brown potion - illusionary form
  - all potions of forgetfulness have the name "potion of forgetfulness" but have more information in their descriptions
  - "flask" of class flask
  - "flask" of class arsenic
  - There's no problem here - this all works as intended.  Anything unidentified shows as so.  Just some potions are always identified by nature.

Amulets of Shadows are not unidentified and are cursed.
- This means the "mystery" is gone for the poor new player who picks one up
- This might be okay... but if we want to maintain the "Yikes" factor we could consider making them unidentified in another PR

Ring of Lethargy (cursed ring)
- There's something wonky with this ring and the spell identify.
  - It automatically reveals itself when used (this is normal) so it will show up as cursed if you pick it up and it is equipped
  - It shows an unidentified if you haven't equipped it (this is normal too)
  - However, when you cast identify on it, it will show the true name "Ring of Lethargy" but still show the unidentified description (this is not correct)
    - I don't think I caused this but it should be fixed
    - When you cast reveal on it, the description shows up properly
image

adrienlaws:
I need to check scrolls and wands to see if they need to show as unidentified - I think they do.  Will follow up on it.

adrienlaws:
This PR is ready for another review, when possible.

Rings, wands, potions, and scrolls all correctly show their unidentified states when so.

I moved the IsIdentified potion override to its parent class of  `SpellItem` / `spelitem.kod`.
- wands, scrolls, and potions are all children of the `SpellItem` class
- They can all use the same IsIdentified function that uses an existing Blakod message `IsLabelled`
- This future proofs in case there are any new wands, scrolls, potions, or new children of `SpellItem`

Rings have their own `IsIdentified` message that propagates so it checks if the rings have other item attributes that affect ring rarity grades.

Example screenshot showing rings and wands in various state of identification
image

kod/object/item.kod
```diff
@@ -633,6 +689,41 @@ messages:
...
+      return Bound(count, viRarity, $);
+   }
+
+   GetRarityCountModifier()
+   "Returns a modifier count to an item's rarity grade.  "
+   "Used to modify the rarity grade for an item based on "
+   "notable item attributes.  Positive means increase the "
+   "rarity grade, negative means decrease the rarity grade."
+   {
+      local count;
```
akirmse, line 722: Soo.. 
```
return 0;
```
?

kod/object/item/passitem/spelitem.kod
```diff
@@ -442,6 +442,15 @@ messages:
       propagate;
    }
 
+      IsIdentified()
+   {
+      if Send(self,@IsLabelled)
```
akirmse, line 447:
```
return Send(self, @IsLabelled);
```

akirmse (commented):
Much nicer!

kod/object/item.kod
adrienlaws: Changed to return 0 as suggested.  I definitely overthink things or leave relics of my thought process sometimes.  We can modify individual item attributes' `GetRarityCountModifier` messages to affect rarities in the future, but it makes sense for the superclass message of item to default to return 0 (no change).

kod/object/item/passitem/spelitem.kod
adrienlaws: Changed as suggested.  More efficient.

PR #700, opened 2024-05-03
Set correct graphics for scrolls

Gar103:
Set the correct school graphics for scrolls, per Mayhem's issue.

However, the Kraanan icon changes color with the scroll. Maybe that's not a huge issue, or we could try to find some other solution for that.


Aesica:
Best solution definitely would be to change the seal to a color that isn't vulnerable to getting recolored along with the ribbon.  Otherwise, this is pretty nice.

PR #704, opened 2024-05-06
fix: adjust rarity grade logic when casting identify and make amulets of shadows appear unidentified

adrienlaws:
# What

There are 2 changes in this PR that fix issues related to #699
- This change makes Amulets of Shadows appear with a rarity grade of `UNIDENTIFIED` until they are either picked up or revealed.  
- It also fixes some rarity grade logic so that when using the spell identify, the proper object description with the proper rarity suffix is assigned.  
  - Previously it would show as `UNIDENTIFIED` when identifying an item.  
    - This is obviously not helpful.  
    - The new logic skips the `UNIDENTIFIED` assignment for "identify".

# Why
- Amulets of Shadows are currently cursed but also identified.  
- Unlike other cursed items like the ring of lethargy, their true cursed nature is directly apparent.
- So far this hasn't mattered much, but now that the item rarity grade system is implemented, amulets of shadows will show as cursed before users use them.
- Regarding the identify bug, we must fix the rarity grade assignment so it is correct and doesn't cause confusion.

# Changes
- This change makes the amulet of shadows appear as unidentified.
- If the victim picks up the amulet it will reveal itself automatically (same as the ring of lethargy).
- GetRarity modified to accept parameter `bSkip_Identify`
  - this returns the proper rarity grade if the spell identify is being used
  - returns the proper rarity grade

### unidentified amulet in the look dialog box
image

### self-revealed amulet of shadows in the victim's drop dialog box
image

### unidentified amulet title and description
image

### revealed amulet title and description
image

# Notes
- Casting reveal will strip the `UNIDENTIFIED` rarity suffix and color from the item
- Casting identify will show the description but as `UNIDENTIFIED` which is perhaps not desirable.
  - Identify will show the true name of an object normally but since the true name and the unidentified name are the same there's no difference
  - The object is not stripped of it's unidentified state by identify, so it will show up as unidentified until the `pbIdentified` property is flipped from 0 to 1.


akirmse:
Hm, I don't know that the intent was to change the name or description of an unidentified AoS.  I thought the idea was simply to avoid having it show as "(cursed)" before it's identified.

Maybe you could instead simply add a boolean property saying whether it's been identified, and don't change the name and description?

kod/object/item/passitem/necklace/shadowam.kod
```diff
@@ -236,28 +230,21 @@ messages:
...
       {
-         vrName = vrRealName;
-         vrDesc = vrRealDesc;
-         bDone = TRUE;
+         pbIdentified = TRUE;
+         Send(poOwner, @SomethingChanged, #what=self);
       }
-      
+
       return bDone;
```
akirmse, line 241: bDone is now always FALSE here.  I don't think that was intentional.

kod/object/item/passitem/necklace/shadowam.kod
```diff
@@ -236,28 +230,21 @@ messages:
...
    }
 
-   GetTrueName()
-   {
-      return vrRealName;
-   }
 
    IsIdentified()
    {
```
akirmse, line 246: This function can be written simply
```
    return pbIdentified;
```

kod/object/item/passitem/necklace/shadowam.kod
adrienlaws: Fixed.  Removed `bDone` completely since it was redundant.

kod/object/item/passitem/necklace/shadowam.kod
adrienlaws: Done.  Embarrassing.  I was going to fix this but you beat me to it.  Much better.

adrienlaws:
Thank you for the feedback.  I have fixed the issues, hopefully.

I also identified (no pun intended) an edge case with identify.

Since identify would force the player's client to open a window with the current rarity grade of an unidentified item, it would always show as unidentified.  That is not desirable.  It should be showing the actual rarity grade after being identified.

I added in a check for `bShow_All `to `GetRarity `as a parameter.  This seems to fix it.  When I cast identify on an amulet, the object description now has the correct suffix of "cursed" but the item in all subsequent dialog boxes (drop/look/description) appears as unidentified.

kod/object/item/passitem/necklace/shadowam.kod
```diff
@@ -215,6 +219,27 @@ messages:
...
+
+   RevealHiddenAttributes()
+   {
+      if pbIdentified = FALSE
+      {
+         pbIdentified = TRUE;
+         Send(poOwner, @SomethingChanged, #what=self);
+      }
+
+      return pbIdentified;
```
akirmse, line 235: pbIdentified is always TRUE here

kod/object/item.kod
```diff
@@ -615,11 +615,12 @@ messages:
       return 1;
    }
 
-   GetRarity()
+   GetRarity(bShow_all = FALSE)
```
akirmse, line 618: The "show_all" behavior isn't really related to the rarity of the item, so this check feels like it belongs in the caller, not in Item.

This parameter is also undocumented, which is a no-no, but no need to document it if you're removing it of course.

kod/object/item/passitem/necklace/shadowam.kod
adrienlaws: Fixed to only return TRUE if `pbIdentified` was changed from FALSE to TRUE, otherwise returns FALSE.

kod/object/item.kod
adrienlaws: This issue is confusing me.  Maybe I'm thinking about it wrong, but I don't see how I can intercept the rarity or modify it without sending a parameter to `GetRarity`.

When a player casts identify, it will go `CastSpell` in `identify.kod`
`Send(who,@SendLook,#what=oTarget,#bShow_All=TRUE);`

`SendLook `in turn will call `ToCliObject`
`Send(self,@ToCliObject,#what=what,#show_type=SHOW_LOOK,#bShow_All=bShow_All);`

Then I have my call to `GetRarity`
`rarity = Send(what,@GetRarity);`

I figured I could do a check to jump past the `IsIdentified `check in `GetRarity `if the original parameter of `bShow_All = TRUE`.  

I could create some kind of other function that doesn't have the unidentified check but that feels wrong and steps on the duplication of code issue.

I still need to run through all my checks to see what the item is (CURSED, MAGIC, RARE, LEGENDARY, NORMAL) and I can't think of a better place than the `GetRarity `function. I must skip the Identified section somehow.

I guess my question is: what would be the caller in your comment?  Is that `ToCliObject `that calls `GetRarity`? Or is that the original caller `CastSpell` in `identify.kod`?

Aesica:
Wouldn't it also make sense to just call it "Amulet" and give it a more vague description until it's identified/picked up, while we're at it?

kod/object/item.kod
akirmse: The caller of GetRarity is ToCliObject.  I think it makes more logical sense to do the "show all" check there, even though it's already a super enormous function.  Passing it as an argument to GetRarity muddles the behavior of that function.

adrienlaws:
> Wouldn't it also make sense to just call it "Amulet" and give it a more vague description until it's identified/picked up, while we're at it?

I did try this at first, but I think the intent of this PR is to just made Amulets of Shadows retain their normal behavior after the rarity grades are implemented.

I think this might be a good idea in a follow on PR, perhaps to add some "good" amulets too.  Perhaps something like 1 in 10 does something decent and shows as magic when revealed instead of cursed.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -2418,8 +2419,39 @@ messages:
       rarity = 0;
       if IsClass(what,&Item)
       {
-         rarity = Send(what,@GetRarity);
+         if bShow_All  % Used for identify and reveal "look" rarity assignment
```
akirmse, line 2422: Yikes, this isn't what I expected.  This is equivalent to the previous version?  I thought you could put a single check here for bShow_All, not reproducing the entire list of options (which we clearly don't want to do).

Can you explain the intent here?

adrienlaws: Sorry.  My intent was to skip the "unidentified" section of the rarity grade check.

The reason is that if I cast identify on an unidentified item, it will go through the GetRarity function and return "unidentified" because ... it is.

So what I want to do is skip that check, to evaluate everything else except unidentified rarity.  With the change I just put in, if bShow_all is true, it does all the checks except unidentified.

I don't really know how to fix this without modifying GetRarity to accept the parameter of bShow_all or create a second function like this where the unidentified check isn't done.

I backed myself into a corner without knowing it, because if we were doing flags instead of int, then I could pass the full rarity value with (unidentified + cursed or + itematts) and then just parse it on the client.

I hope I'm explaining my intent correctly - it is complicated.  But in short, I want to check for everything except IsIdentified if bShow_All is TRUE.

akirmse: I see.. I guess I didn't understand this before.  The previous version was much better than this level of code duplication.  But perhaps you can come up with a better name for the parameter than the meaningless "bShow_All", and also add documentation for it.

Issue #705, opened 2024-05-08
Item: Heal Wand: Interface: Update use sound and add heal-cast visual effect to target player


jrotunda:
Current Behavior:

When a user casts heal on themselves or another using the Heal Wand, there's a text response to indicate if the heal went through or not, as well as a specific heal wand sound.wav (audio effect from 2011-13 period on github by @adrienlaws ).

The player who gets healed will see their HP rise (anywhere form 2-8 HP, randomly).

This sound effect isn't used anywhere else, and while unique, does cause some confusion.
In addition, in PvP when heal wands are used, the items aren't held, so their use is already well-obfuscated. 

There were recent interface updates to enhance visual-audio feedback, such as the Chalice of the Rain sounds.  Current behavior for the SpellWand items is audio consistent with the spells cast.

Expected/Suggested Behavior:

* Play Shal'ille's generic audio sound effect (shalille.wav) when the heal wand is used and have it be heard by all players in the room (like minor heal, etc)
* Visual 'Heal' Flash special fx to the graphic display window for the person being healed (self or other user)

PR #709, opened 2024-05-09
feature: modify heal wand sound and add heal flash on recipient

adrienlaws:
# What

- changes heal wand sound to the default Shal'ille sound
  - the sound is public and heard by others in the room
  - the sound originates from the user of the wand
- adds a "good spell" flash to the recipient's screen

# Why

- changes requested in #705 
- does perhaps make more sense to have the well known heal sound play for a healing action
- gives a little bit more obvious feedback to the recipient that they are being healed

# Notes

- removed a third space in the description (normal seems to be 2 spaces)


akirmse:
Not related to the code, but check out your commit summary.  Something is
not great with your git branches... maybe you're not pulling, or you're
reusing branches, or something like that?

PR #710, opened 2024-05-09
feature: make heal wand use more obvious

adrienlaws:
# What

- this change adds several queues to make heal wand use more obvious to the recipient and 3rd parties in the same room

# Why

- changes requested in #705 
- gives a more obvious feedback to the recipient that they are being healed

# Changes

- adds a "good spell" flash to the recipient's screen
- adds a projectile upon using a heal wand from the user to the target (if user != target)
- adds a third person message so others in the room know that a heal wand is being used by person X on person Y
- makes the user of the wand "point"

### Example of `GoodSpellFlashEffect` on successful heal wand cast
meridian_odjuPvcsUL

# Notes

- also removed a third space in the heal wand description (normal seems to be 2 spaces)

### DIFF
--- a/kod/object/item/passitem/specwand/healwand.kod
+++ b/kod/object/item/passitem/specwand/healwand.kod
@@ -21,7 +21,7 @@ resources:
    healWand_desc_rsc = "This wand is almost comically decorated by obviously fake gems.  "
        "Still, the wand hums nearly inaudibly, and when you put your hand on it, it is "
        "warm to the touch."
-   healWand_success_wav_rsc = shalille.wav
+   healWand_success_wav_rsc = healwand.wav
    healed_rsc = "You feel better."
    wand_used_rsc = "The wand pulses once in your hand."
    wand_fail_snd = spelfail.wav
--- a/kod/object/item/passitem/specwand/healwand.kod
+++ b/kod/object/item/passitem/specwand/healwand.kod
@@ -18,10 +18,10 @@ resources:
 
    healWand_name_rsc = "wand of healing"
    healWand_icon_rsc = healwand.bgf
-   healWand_desc_rsc = "This wand is almost comically decorated by obviously fake gems.   "
+   healWand_desc_rsc = "This wand is almost comically decorated by obviously fake gems.  "
        "Still, the wand hums nearly inaudibly, and when you put your hand on it, it is "
        "warm to the touch."
-   healWand_success_wav_rsc = healwand.wav
+   healWand_success_wav_rsc = shalille.wav
    healed_rsc = "You feel better."
    wand_used_rsc = "The wand pulses once in your hand."
    wand_fail_snd = spelfail.wav
@@ -101,7 +101,13 @@ messages:
       oRoom = Send(poOwner, @GetOwner);
       Send(oRoom, @SomethingWaveRoom, #what = poOwner, 
            #wave_rsc = healWand_success_wav_rsc);
-      
+
+      if apply_on <> $
+         AND IsClass(apply_on,&User)
+      {
+         Send(apply_on,@GoodSpellFlashEffect);
+      }
+
       return;
    }
 


Gar103:
I tried this out locally and the current shal sound is too long, it overlaps quite awkwardly when 5 play at once, and I began to get sound crackling

adrienlaws:
Thanks for sharing.  I see what you mean.  It starts crackling when you use the wand very fast.

3 options
- cancel this change and stay with the current short sound
- implement a throttling mechanism like with gnarled staffs (they have a 3 second timer before you can cast again)
  - heal wands have no throttling; they cast as fast as you can tab-space-click
- find some way to avoid the same sound playing over itself (max same sound config somewhere?)
Or find some

Gar103:
Do we have anyone with sound editing skills to make a shorter riff on the shal sound?

adrienlaws:
I tried to edit the sound with elastic audio (so the pitch and sound stays the same) but due to the low bitrate it starts to have distortion.  I can look into trying to recreate a similar Shal'ille sound or to cut the sound length but I'm not sure if this is the right strategy.

jrotunda:
Per discussion. The issue was only pertaining to sound and visual - not intended to include functionality change, the title needs to be updated if delay is added.

If including delay, I support this. Facts:

* Staff of jolting was given a ms timer delay 
* Minor heal; major heal, hospice spells have cast delay 

There are other behavioral reasons we could get into but I think the longer sound + delay is a good way to go.

Gar103:
I would not touch the functionality of heal wands without an in-depth systemic review. I don't _love_ heal wands the way they are, but they are a major foundation of interesting PvP. By that I mean, any delay will probably make them ineffective, and if they were made ineffective, all combat would meta shift closer to simply blinding and bowing your opponent until they die in less than one mana bar, with no chance for them to do anything. Heal wands are the only thing that gets any of us past the 'gank' stage and into the interesting part of combat... as sad as that might be. Imagine the Official Tournament being held and all the 1v1s ending in like 30 seconds each. No time to do anything cool, unexpected, or fun.

At the very least, if a delay was added, I'd suggest keeping the total healing delivered the same. For example, if heal wands become single use, multiply the healing of that single tap by 5. But really with any delay, we lose the manual dexterity and the frantic nature of using these things that is honestly kind of fun, and, ofc, literally the only thing we can do while blind. This might be one of Meridian's many mechanics that _feels_ like it sucks, but there's no better alternative (without massive systemic overhauls)

davidklaw:
Since other actions, like eating, create this same overlap, and because it seems like it only affects an extreme minority of players (who already experience this effect), I think it's acceptable in favor of the overall improvement.

adrienlaws:
I looked at this a bit more.  The Shal'ille sound appears to be a variant of F#m and lasts slightly over 1 second.  I don't think you can "rapid fire" healing wands without sound distortion since you can probably do it several times a second and this will create distortion.  Even if you don't have distortion it's going to sound very strange to have 4-5 Shal'ille sounds playing at once.

Maybe keeping the "new" (now 10 years old) heal sound is best since it is a single note.

Along with that, add 2 things:
 - a text message to all users in the room
   - > Toko points his heal wand at TokoFriend
 - make the user of the heal want point

This combined with the `GoodSpellFlashEffect` would inform players in the area that someone is being healed, and would make it more obvious to the recipient that they are being healed as well (due to the flash).  None of this would affect the current use of heal wands.

davidklaw:
I like it. Keep it simple and move it in a positive direction.

davidklaw:
@adrienlaws Does your description need updating now that the sound has changed?
@Meridian59 This looks good to go!

adrienlaws:
Thanks for your comment!  I have a bit more work to do.  I will make this a draft PR for the moment to avoid confusion and then promote it back to ready-for-review.

adrienlaws:
Closing this PR.  I will keep the branch and reopen it when/if I can.

PR #713, opened 2024-05-10
feature: extend BGF format to support 24-bit color

adrienlaws:
🚧 Warning: Draft PR 🚧 

# What

- This change extends the BGF format to support 24-bit color.
- A command line argument is added to `makebgf.exe` with a `-b` to specify 8-bit or 32-bit color

# Why
- Currently all bitmaps that go through the makebgf process are reduced in color depth to Meridian 59's 256 color palette.
- Extending the palette could make the game look a lot better and artists could make use of AI upscaling tools on existing bitmaps

# Changes

- Added a new command line argument `-b`
- Added a new function `WriteBitmap32` which is called if the command line argument for `-b` is `32`

# Notes

- This is still in draft

### DIFF
--- a/makebgf/dibutil.c
+++ b/makebgf/dibutil.c
@@ -187,7 +187,8 @@ PDIB DibReadBitmapInfo(int fh)
 
     if (nNumColors != NUM_COLORS)
     {
-       printf("Expecting %d color bitmap; found %d colors\n", NUM_COLORS, nNumColors);
+       printf("Expecting %llu color bitmap; found %d colors\n",
+          (unsigned long long)NUM_COLORS, nNumColors);
        return NULL;
     }
        
--- a/makebgf/makebgf.c
+++ b/makebgf/makebgf.c
@@ -366,15 +366,15 @@ int main(int argc, char **argv)
 	 Usage();
 	 exit(0);
 
-      case 'b': // used for 8 bit or 32 bit color depth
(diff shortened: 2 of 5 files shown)


akirmse:
Before you go too far down this road, here are some things to consider:

- You're adding support for 24 bit color, not 32, since alpha is always being written as 0xff.  No need writing it at all I'd think.
- Changing the client to use these textures in the hardware renderer might not be too hard.  But adding it to the software renderer might be quite difficult.  Any new added textures need to work in both.
- You'd also need to add support to the room editor to display wall/floor/ceiling textures in one of the dialogs.
- Anytime you change a file format, you also need to change the version number in the file, and further verify that existing customers won't break with the new version of the file format (they should simply fail to load them)
- Does it make sense to invent yet another graphics format, or what about using PNG instead for the images inside a BGF?  The PNG library is already linked with the client.

akirmse:
Is there action here or should I close this?

adrienlaws:
Sorry for the lack of action on this PR.  Unfortunately, I haven’t had a chance to work on this for some time, but I expect to have more availability now.

If it's possible to keep the PR open a little longer, I’d like to try and see if I can get get this moving a bit this week.

adrienlaws:
Closing this PR. I will keep the branch and reopen it when/if I can.

PR #714, opened 2024-05-11
Alter Shadow Rift to only work in dark areas

Gar103:
To answer a few concerns with edge case Shadow Rift abuses, I added a thematic limitation to the spell. A portal cannot be created in areas brighter than LIGHT_DARK, and portals in an area that becomes brighter than LIGHT_DARK will start to close themselves according to the spell's existing closing procedures. This means that Shadow Rift portals are by and large temporary, as many areas are only dark enough at night, and when daylight comes, the portals will close of their own accord - even inside buildings, like inns, or private rooms. All these areas do in fact experience the day/night cycle of changing ambient light, often bringing it just dark enough at night, then too bright during the day. In many cases, players will need to cast Darkness to summon a portal at all, in which case the portal will only last as long as the Darkness spell left in your wake.

This also allows players to actively close any portals in public spaces by simply casting Light, or firewall, or lightning wall.

Towns and public areas are generally quite brightly lit, and this will basically make it impossible or pointless to open portals in Tos, Barloque, etc.

As a bit of extra theming that also doubles as edge case protection, if a player tries to go through while they have a torch equipped, the portal will instantly close and fail to teleport them. This will allow players to close problematic portals safely if someone puts portals in door landing spots or something similarly annoying where you might not actually get a chance to cast Light. It will also, in some cases, protect newbies - who are most likely to have torches equipped in dark places already.

I don't think that public space portal abuse was generally happening in any real capacity, as it's quite risky to try to open portals in public spaces - you're more liable to get your highly valuable prism stolen than anything else - however, it _has_ happened at times, so it's best to design out that silliness so admins don't have to manually handle it.

### DIFF
--- a/kod/object/active/portal/necport.kod
+++ b/kod/object/active/portal/necport.kod
@@ -22,6 +22,8 @@ resources:
    NecropolisPortal_icon_rsc = teleport.bgf
 
    NecropolisPortal_closing = "The portal flickers and begins to draw imperceptibly in on itself."
+   NecropolisPortal_stabilized = "As the shadows deepen, the portal stops flickering and slowly grows stable."
+   NecropolisPortal_torch_failure = "The light of your torch causes the portal to dissipate around you!"
    NecropolisPortal_closed = "The portal disappears."
 
 classvars:
@@ -67,6 +69,19 @@ messages:
       % Don't allow monsters to escape through these portals
       if (NOT IsClass(what, &Player)) 
       {
+         return;
+      }
+
+      propagate;
+   }
+
+   TeleportSomething(what=$)
+   "Called when something walks on top of us."
+   {
+      if IsClass(what,&Player) AND Send(what,@FindUsing,#class=&Torch) <> $
+      {
+         Send(what,@MsgSendUser,#message_rsc=NecropolisPortal_torch_failure);
+         Send(self,@ClosePortal);
          return;
       }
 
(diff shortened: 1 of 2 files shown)


Gar103:
This is a fairly conservative and highly thematic change that shifts the use of this spell (which is extremely rare already) to the shadows, so to speak, where it rightfully belongs.

I actually think Shadow Rift should allow the opening of portals in guildhalls - not TO guildhalls, but FROM guildhalls, because if you keep a portal up in your hall, it's private and out of the way, so no one else is bothered, _and_ it requires you to open the door of your hall every single time you use it. We had huge success with this on 103 - putting fixed portals in halls turned out to be a fantastic change, because it both upgrades halls and makes guildhall combat far more likely. Portals in halls was the best thing we ever did for getting people excited about halls, and fighting over them. Nothing ever made people open their front doors and take a risk more than that. I didn't put a change like that here, though, since I figured it was best to keep this simple at first.

I also think the requirement to keep a character standing in the room is problematic. On the one hand, with such a low IP limit, it's actually a _huge_ burden. If you want to keep just one portal open for awhile, you're already limited to only 2 other accounts online. Personally, I am trying to play with my partner, and we found that Shadow Rift for farming was quite literally not worth it, because we couldn't even run a character to sell things. You can completely forget about a portal loop between the selling and farming spot. That would require 2 accounts that can't move, then you've only got your main, and you can't sell anything at all.

On the other hand, any mechanic that requires an AFK mule or alt standing around is a bit silly. I would like to see if the 'darkness requirement' in this pull request is enough to match shadow rift portal times more closely to a player's 'play session.' Because of the onerously low IP limit, any shadow rift use is limited to quite literally the start of your play session, because you need to use all 3 of your available connections to open the portal. _Then_, you have to keep one character there, and only then can you get on your main character to build or farm. So the ideal target for a portal duration is a loosely defined 'play session.' Like, when I log for the night, that's it for those portals. The alt standing there does kind of accomplish that, because you can't really keep a char on indefinitely without them timing out or eventually losing connection to the vagaries of the internet, but I just feel like we can do better.

Gar103:
Thing is, I don't think that's actually true. As far as I've seen, trying to use Shadow Rift eventually just gets annoying and cumbersome, and Elusion turns out to be vastly more reliable and powerful. What I am seeing in terms of claims about shadow rift are not accurate. That recent Issues post, for example, is drastically wrong on most counts. I think it's a case of outside perspective (admins) only seeing pieces and not the whole picture. Playing with SR and trying to use it is a vastly different thing than simply assuming the worst case scenarios. Almost always, Elusion is just better.

I would love to know - so we can dissect - what the ridiculous advantage being taken is, and see if it's really a problem.

PR #719, opened 2024-05-12
Flip negative bludgeon resistance on SimpleHelm

Gar103:
Helmets make you take more damage from bludgeon. Judging by other extremely old resistances being incorrect in the same way, I'm guessing this was a mistake. Flipped it to a positive resistance in this commit.

### DIFF
--- a/kod/object/item/passitem/defmod/helmet/simphelm.kod
+++ b/kod/object/item/passitem/defmod/helmet/simphelm.kod
@@ -66,7 +66,7 @@ messages:
 
    GetResistanceModifiers()
    {
-      return [ [ATCK_WEAP_BLUDGEON,-10]
+      return [ [ATCK_WEAP_BLUDGEON,10]
              ];
    }
 


Gar103:
As an aside, due to the way defense modifiers work, helmets do almost literally nothing. Against a purely physical attack, they will prevent 0 or 1 damage (50/50). Against an enchanted attack, they will prevent no damage. Their 25 defense will sometimes have no effect on actual evade chances, and sometimes offer 1% higher chance to evade an attack (ex: 35% chance to evade -> 36% chance to evade). This is as close to no effect as a functional item can truly be - circlets and ivy circlets vastly outclass helmets.

Helmets are ripe for someone doing something interesting with them. A bit of slash, thrust, or pierce resistance would make a huge difference - or perhaps any of the variety of monster damages (bite / claw / etc resistance)

adrienlaws:
I like the idea of some slight modifications to other resistance types for SimpleHelm.

### Increase Pierce and Slash Resistance 🏹 ⚔️ 
I think light pierce, slash, and thrust resistance might need higher resistance values than than bludgeon resistance.  The simple helm has an exposed face so it would not be as complete as a medieval [Great Helm](https://en.wikipedia.org/wiki/Great_helm)  but it still should provide far superior cutting or piercing resistance than an ivy circlet or no head protection at all.  The ability to at least partially deflect arrows or slashing hits would suggest that additional resistance (albeit slight) might be warranted.

### Helms and Bludgeon Damage
Without going down the rabbit hole too much - it seems that some medieval helmets can **increase** the risk of concussion.  So while they might have been wrong in **why** they assigned additional bludgeoning damage, it might be sort of correct.  Any hindrance of movement could mean a more direct impact.  Padding would mitigate it somewhat.  Maybe just zero-ing it out to not provide any additional resistance is an option.

### Fire and Electrical Damage 🔥 ⚡ 
I feel that fire and electrical damage types would be increased by wearing a metal helm.

TLDR; helms should resist physical (especially piercing/arrows) more, maybe fire and electrical less

Aesica:
I'd probably leave out the fire/electrical vulnerability since plate armor and knight shields are bad enough already.  Definitely agree on just zeroing out the bludgeoning modifier though.

PR #721, opened 2024-05-15
Fix for karma-specific spellitems

Aesica:
In a pervious PR intended to fix potions and karma restrictions, karma  restrictions for all spellitems was unintentionally disabled.   This PR:

- Re-enables karma restrictions for wands and scrolls.  (Potions have it disabled with the exception of the rescue potion)
- Will now properly show text in the description about the item being karma restricted if it is.
- Adds a function/message to spell.kod to query a given spell about its required karma--surprising that this didn't exist already.

### DIFF
--- a/kod/object/passive/spell.kod
+++ b/kod/object/passive/spell.kod
@@ -455,29 +455,27 @@ messages:
    
    KarmaCheck(who=$)
    {
+      local iSpellKarma, iUserKarma;
+
       if Send(who,@PlayerIsImmortal)
       {
          return TRUE;
       }
-   
-      if viSchool = SS_SHALILLE
-      {
-         if Send(who,@GetKarma) < (10*viSpell_level)
-         {
-            return FALSE;
-         }
-      }
-      else
-      {
-         if viSchool = SS_QOR
-         {
-            if Send(who,@GetKarma) > -(10*viSpell_level)
-            {
-               return FALSE;
-            }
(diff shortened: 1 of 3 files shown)


kod/object/passive/spell.kod
```diff
@@ -481,6 +481,24 @@ messages:
       return TRUE;
    }
 
+   GetRequiredKarma()
+   "Returns the karma value required to use the spell, or 0 if there's no karma requirement."
+   {
+      local iKarma;
+
+      iKarma = 0;
+      if (viSchool = SS_QOR)
```
akirmse, line 490: This is pretty much a duplicate of the code in KarmaCheck.  It should exist only in one place.  I think you can change KarmaCheck to call this function.

Aesica: KarmaCheck checks the spell's karma against the player's karma and returns whether or not they can cast it with their current karma.  I thought about using this approach, but doing so would only show a karma message in the descriptions if the player lacked the karma to use the item, and only if they were holding it.

The two functions are similar though, and while I also thought about adding a call to it in KarmaCheck (so that the if school is qor or shal'ille check only existed in one place) but felt that went outside the scope of the PR.

I'm open to recommendations on all this.

Aesica: KarmaCheck updated. :)

PR #724, opened 2024-05-15
Consuming a mana crystal breaks the user's trance

CardiovascularMike:
This PR prevents a Jala singer who is using a Jala necklace from looping Crystallize mana songs.  Now the use of the Crystal will break the singer's trance and end the Crystallize mana song.

EVENT_STEER is used for this purpose to avoid the oddities of EVENT_USE or other events.  This convention is used in other places in the database for ensuring a trance is broken.  A comment to this effect is included in the change.

To test:  Use a Jala Necklace and cast Crystallize mana.  Prior to this change you could consume the mana crystal, leaving the song up.  A new crystal would appear and get filled with the mana that you generated from the first crystal.

After applying this change: Using the crystal breaks the caster's trance and ends the song.  The need for this change was introduced by the changes included in PR #606.

### DIFF
--- a/kod/object/item/passitem/manacrys.kod
+++ b/kod/object/item/passitem/manacrys.kod
@@ -117,6 +117,8 @@ messages:
 
       Send(poOwner,@GainMana,#amount=piStoredMana);
       Post(self,@Delete);
+      % Always break a trance when using a mana crystal for balance reasons.
+      Send(poOwner,@BreakTrance,#event=EVENT_STEER);
 
       return;
    }
--- a/kod/object/item/passitem/manacrys.kod
+++ b/kod/object/item/passitem/manacrys.kod
@@ -118,6 +118,7 @@ messages:
       Send(poOwner,@GainMana,#amount=piStoredMana);
       Post(self,@Delete);
       % Always break a trance when using a mana crystal for balance reasons.
+      % Event_Steer is used to ensure the trance is broken.
       Send(poOwner,@BreakTrance,#event=EVENT_STEER);
 
       return;


Gar103:
Wait, what? That's not the full extent of what this does. This breaks _any_ song when you use a mana crystal, doesn't it? Now Jala mages will lose all their songs mid-combat when they go to refill. That's a significant nerf to the entire point of the mana crystal, just to fix a non-issue... what was the problem with crystalize mana and filling a crystal repeatedly? It didn't allow any sort of exploit as far as I'm aware.

CardiovascularMike:
> Wait, what? That's not the full extent of what this does. This breaks _any_ song when you use a mana crystal, doesn't it? Now Jala mages will lose all their songs mid-combat when they go to refill. That's a significant nerf to the entire point of the mana crystal, just to fix a non-issue... what was the problem with crystalize mana and filling a crystal repeatedly? It didn't allow any sort of exploit as far as I'm aware. The new crystal gets filled with the mana from the old crystal.

This PR brings the Crystallize Mana functionality back to where it is on current (as of 5/15/2024) production on server 101 and 102.  PR #606 added the ability for Jala mages to use items without breaking a trance, except for food, potions, and instruments.  Prior to that using the mana crystal has always broken the caster's trance.

#606 unintentionally changed this behavior and this PR reverts that.  If we want we can consider improving Crystallize Mana but I would like to do it intentionally / thoughtfully.

As far as losing 'all the songs' remember that only the first song actually has any effect for the most part.  The exceptions are Jig, Spellbane, and possibly Mana Convergence work as the second song on production right now (this was also changed by #606 in the sense that a caster can only ever sing one song at a time).

The effect of two or more songs existing in a room will be looked at in the near future.  The groundwork was laid out roughly in  closed PR #637.  This PR was closed to allow for a bit more consensus building as there is an opportunity to make this a little sleeker in terms of gameplay design.

Gar103:
We talked a bit about this on Discord, so I'm just repeating that here for Github mostly. I think the 'prior functionality' thing is not something we need to worry about. All item use broke songs before, there was no specific consideration for mana crystals. If we're fixing the item use, we should let crystals be included, too.

By losing all songs I meant that every song will be broken by this, not just Crystalize Mana. If I get low on mana using Profane Resonance, for example, I'll have to drop the song mid-combat by using a mana crystal, then recast it. That's really clunky. The moment I drop the song, I'll be getting Dazzled. The mana crystal is therefore quite risky to use.

As far as I've seen, there is no exploit with Crystalize Mana being able to refill crystals under the same song. Mana regeneration stops when it is above max, so the next crystal can only be filled with the mana from the first. There is a cap of 130% caster's max mana or 250, whichever is lower, for the crystal as well. If testers found they could keep their mana near these caps, they needed to go longer - they'd eventually find out they were living on 'borrowed mana,' so to speak, from their original mana bar that filled the first crystal.

Ultimately, what I think Crystalize Mana is trying to replicate is a mana second wind. Imagine Crystalize Mana as something that automatically triggers on low mana, giving you a second burst, then there's a couple minutes before you can do it again. That's almost exactly what the song and crystal format is doing - and second wind doesn't break concentration or end songs. That's my ultimate point, though, to be honest, I would absolutely just shift Crystalize Mana into a mana second wind variant. Forget the crystal thing. Trying to create a crystal in Fams between PvP bouts, but having an enemy discordance your CM song, is just the most annoying damn thing in the world. If you wanted to keep CM as a song, you could have it do something nominal related to mana, but have its %s _also_ function as the basis for a mana second wind. I'd even include up to +20 max mana from 99%.

PR #731, opened 2024-05-20
fix: multiple Mark of Dishonor issues

adrienlaws:
# What

There were multiple issues in Mark of Dishonor that are fixed in this change
- removed an unused string resource
- made spell work even if the target player has positive karma
  - it used to consume resources on the caster but not cast on the target due to returning false if karma > 0
- adjusted monster behavior
  - previously it would try and adjust the behavior to ignore the caster but had no effect (monsters would continue attacking even though the code was trying to make them stop attacking)
  - now enrages the target monster and causes them to attack the caster (makes more sense I think)
- added a vague string regarding the target appearing lethargic if the spell successfully punishes the target player's vigor
  - > X appears more lethargic.

# Why

- the spell would fail when casted on any positive karma target but would consume vigor, mana, and reagents on the caster
- per the description, the spell should do 2 things
  - cap a target player's HP to their natural maximum (regardless of the target's karma)
  - punish evil or negative Karma players with a loss of vigor

# Testing

- now seems to work properly on both monsters and players

### DIFF
--- a/kod/object/passive/spell/dishonor.kod
+++ b/kod/object/passive/spell/dishonor.kod
@@ -38,7 +38,7 @@ resources:
    MarkOfDishonor_no_self = \
       "You cannot cast mark of dishonor upon yourself; if you were truly "
       "dishonorable you would not be able to cast Shal'ille's spells."
-   MarkOfDishonor_no_target = "You may only cast %s upon the dishonorable."
+   MarkOfDishonor_vigor_affected = "%s%s appears more lethargic."
 
 classvars:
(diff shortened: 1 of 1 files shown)


adrienlaws:
Understood.  These problems have been around for years.  Some for 2 years and some for much longer.  So it's fair to say they're not recent.

There are no errors on the server in the current state.  This PR fixes minor things and those fixes can be done at a later time.

akirmse:
OK.  Let's keep this open, but I won't merge it right now.

PR #740, opened 2024-05-28
fix: adjust rarity grade logic to sum (vs bound) and change magic grade to uncommon

adrienlaws:
# What
- several changes based on feedback to calibrate the rarity grade system

# Changes
- changed MAGIC rarity grade to be UNCOMMON both on server and client
- changed rarity grade logic to add class based rarity `viRarity` and item attributes
  - formerly picked the higher value of the two
- changed `viRarity` class based assignment for several items
  - removed MAGIC/UNCOMMON from Riija Sword (now is NORMAL)
  - moved the following to RARE
    - nerudite bow
    - magic bow
    - gold sword
- changed `item_rarity_grade` enums to be a little more spaced out
  - changed LEGENDARY to 4 from 3 so the values line up better with the rarity calculation (mostly just so things are consistent and easier to understand) 
  - UNIDENTIFIED is now 100
  - CURSED is now 200

# Logic Diagram
``` mermaid
flowchart TD
    I(Class Rarity
     default = 0)  -->|viRarity| C
    J(Item Att Count
    default = 0) --> C
    C(Sum of viRarity + Item Att Count) -->D
    D(NORMAL = 0
    UNCOMMON = 1
    RARE = 2
    LEGENDARY = 4+)
    style D text-align:right
 ```

# Reference documentation
Below are tables for class rarity assignments and item attributes that contribute to the rarity calculation of an item.  These are mostly unchanged by this PR and are here for reference.

## Class Rarity Assignments `viRarity`
All items have a default class rarity or `viRarity` of NORMAL.  Below are the exceptions.

| Name                      | Item Class    | viRarity Assignment |
| --- | --- | ---|
| Magic Spirit Helmet (MSH) | Helm          | Rare |
| Brass Medallion           | FarenCharm    | Rare                |
| Steel Torc                | KraananCharm  | Rare                |
| Monacle                   | QorCharm      | Rare                |
| Feathered Choker          | RiijaCharm    | Rare                |
| Diamond Pendant           | ShalilleCharm | Rare                |
| Gold Sword                | GoldSword     | Rare                |
| Magic Bow                 | MagicBow      | Rare                |
| Nerudite Bow              | NeruditeBow   | Rare                |

## Item and Weapon Attribute Modifiers
These are the item attributes that contribute to a higher rarity grade.  They are added together along with the class based rarity `viRarity` to determine the rarity grade for a revealed or identified item.

| Name | Attribute | Rarity Modifier | DM word | Description and Notes|
| --- | --- | --- | --- | --- |
| Bonded | IA_BONDED | +1 | N/A | This Item Attribute bonds an item to a specific player. Not sure if it is in use.                                                                                                     |
| Bonded | IA_TRANSCENDANT | +1 | transcendant | It glows with a soft, white light. This is what is commonly known as "bonded" where the item will remain in your inventory when you die.|
| Durable | IA_DURABLE             | +1 | durable  | The craftsmanship of this X is admirable, obviously the result of hundreds of hours of work by expert hands.  Default modifier is 300% bonus to durability.|
| Artifice                 | IA_MISDIRECTION        | +1 | N/A | An illusion has been cast on this item, hiding it's true abilities. This is used by the Riija spell Artifice to make an item appear to have attributes that it does not have, and (according to the description) hide true attributes the item has.|
| Engraved Item | IA_ENGRAVED | + 1 | use dm engrave spell | The words x are delicately engraved on the edge.|
| Blind Weapon             | WA_BLINDER             | +1 | blinder | Q's insignia is emblazoned on the metal of the weapon.|
| Bonk Weapon              | WA_BONK                | +1 | bonker | Mocker's signature flamboyantly emblazons the pommel.|
| Arena Weapon             | WA_CEREMONIAL          | +1 | ceremonial | Beautiful and ornate, this weapon is obviously  unsuitable for blood combat, though it might provide X edge in games and tourneys. |
| Expert Weapon            | WA_EXPERT              | +1 | expert | Colhorr's signature is engraved on the handle of this weapon. This "weapon looks to provide X advantage - in the right hands. |
| Faction Weapon           | WA_FACTION | +1 | duke, princess, rebel ||
| Guildhall Weapon         | WA_GUILD_HALL_DEFENDER | +1 | guild hall defender | The crest of X is inscribed on the hilt of the weapon, suggesting Y bonus in the defense of that hall. |
| Hold Weapon              | WA_Paralyzer | +1 | Paralyzer | A glow that can only be described as Zjiriaesqe surrounds the weapon. |
| Justice Weapon           | WA_PUNISHER | +1 | punisher | The emblem of a fist holding an ankh rests in the pommel of this weapon, promising X advantage in the fight for justice. |
| Purge Weapon             | WA_PURGER | +1 | purger | The sign of Psychochild is engraved on the weapon.|
| Spellcasting Weapon      | WA_SPELLCASTER | +1 | spellcaster | Mystical energy flits about this weapon.                                                                                               Possible spells<br>[SID_DEMENT, 12], [SID_ENFEEBLE, 13], [SID_VERTIGO, 13], [SID_SWAP, 12], [SID_INVISIBILITY, 12], [SID_FORGET, 13], [SID_EVIL_TWIN, 12],[SID_MARK_OF_DISHONOR, 13] |
| Attack spell type Weapon | WA_ATTACKSPELLTYPE     | +1 | fire, cold, shock, holy, unholy, acid | This weapon attribute merely flags a weapon as having the spell attack type that corresponds with it. It also has a timer, which, when it expires, flips the Attack type back.      |
| Twisting Weapon          | WA_TWISTER | +1 | twister | A swirling green 'GMT' is stamped into the pommel.|
| Vampiric Weapon          | WA_Vamper | +1 | Vamper | An unholy glow seems to suck all life from the air around the weapon.|

### DIFF
--- a/include/proto.h
+++ b/include/proto.h
@@ -496,9 +496,9 @@ typedef enum {
    ITEM_RARITY_GRADE_NORMAL       = 0,
    ITEM_RARITY_GRADE_MAGIC        = 1,
    ITEM_RARITY_GRADE_RARE         = 2,
-   ITEM_RARITY_GRADE_LEGENDARY    = 3,
-   ITEM_RARITY_GRADE_UNIDENTIFIED = 4,
-   ITEM_RARITY_GRADE_CURSED       = 5,
+   ITEM_RARITY_GRADE_LEGENDARY    = 4,
+   ITEM_RARITY_GRADE_UNIDENTIFIED = 100,
+   ITEM_RARITY_GRADE_CURSED       = 200,
 } item_rarity_grade;
 
 /* Size in bytes of numbers in protocol */
--- a/include/proto.h
+++ b/include/proto.h
@@ -494,7 +494,7 @@ enum {
 /* Item rarity designations used for different object presentations */
 typedef enum {
    ITEM_RARITY_GRADE_NORMAL       = 0,
-   ITEM_RARITY_GRADE_MAGIC        = 1,
+   ITEM_RARITY_GRADE_UNCOMMON     = 1,
    ITEM_RARITY_GRADE_RARE         = 2,
    ITEM_RARITY_GRADE_LEGENDARY    = 4,
    ITEM_RARITY_GRADE_UNIDENTIFIED = 100,
--- a/clientd3d/client.rc
+++ b/clientd3d/client.rc
@@ -1787,7 +1787,7 @@ END
 
 STRINGTABLE
 BEGIN
-    IDS_RARITY_GRADE_MAGIC         "magic"
+    IDS_RARITY_GRADE_UNCOMMON      "uncommon"
     IDS_RARITY_GRADE_RARE          "rare"
     IDS_RARITY_GRADE_LEGENDARY     "legendary"
     IDS_RARITY_GRADE_UNIDENTIFIED  "unidentified"
--- a/clientd3d/color.c
+++ b/clientd3d/color.c
@@ -56,7 +56,7 @@ static char colorinfo[][15] = {
 	{ "255,255,255"},   /* COLOR_BAR4 */
 	{ "192,192,192"},   /* COLOR_INVNUMFGD */
 	{ "0,0,0"},         /* COLOR_INVNUMBGD */
-	{ "141,242,242"},	/* COLOR_ITEM_TEXT_MAGIC        - cyan   */
+	{ "141,242,242"},	/* COLOR_ITEM_TEXT_UNCOMMON     - cyan   */
 	{ "0,255,0"},	    /* COLOR_ITEM_TEXT_RARE         - lime   */
 	{ "255,0,255"},     /* COLOR_ITEM_TEXT_LEGENDARY    - purple */
 	{ "252,128,0"},	    /* COLOR_ITEM_TEXT_UNIDENTIFIED - orange */
@@ -416,8 +416,8 @@ WORD GetItemListColor(HWND hwnd, int type, item_rarity_grade text_color_value)
 	{
 		switch (text_color_value)
 		{
-			case ITEM_RARITY_GRADE_MAGIC:
-				return COLOR_ITEM_TEXT_MAGIC;
+			case ITEM_RARITY_GRADE_UNCOMMON:
+				return COLOR_ITEM_TEXT_UNCOMMON;
 			case ITEM_RARITY_GRADE_RARE:
 				return COLOR_ITEM_TEXT_RARE;
 			case ITEM_RARITY_GRADE_LEGENDARY:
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: include/proto.h [hunk 1]
```diff
@@ -494,11 +494,11 @@ enum {
...
    ITEM_RARITY_GRADE_NORMAL       = 0,
-   ITEM_RARITY_GRADE_MAGIC        = 1,
+   ITEM_RARITY_GRADE_UNCOMMON     = 1,
    ITEM_RARITY_GRADE_RARE         = 2,
-   ITEM_RARITY_GRADE_LEGENDARY    = 3,
-   ITEM_RARITY_GRADE_UNIDENTIFIED = 4,
-   ITEM_RARITY_GRADE_CURSED       = 5,
+   ITEM_RARITY_GRADE_LEGENDARY    = 4,
+   ITEM_RARITY_GRADE_UNIDENTIFIED = 100,
+   ITEM_RARITY_GRADE_CURSED       = 200,
```
akirmse, line 501: I'm not sure why you changed these enum values.  They're basically opaque integers with no other meaning as far as I can tell.

adrienlaws: I can change them back but I figured it might be doing for 2 reasons
- after this PR, `viRarity` + count of attributes will determine the rarity of an item with attributes
  - Since enum legendary was set to 3 and the threshold for an item to be legendary was 4 I thought this might introduce confusion.  By making legendary always equal to 4 it just kind of standardizes it.  Not necessary.
 - if someone adds more grades or levels of rarity, the next number would be after cursed (5) but the current highest is legendary(4) so that would be (6).  
   - I felt like people might trip over this and perhaps to limit confusion we could jump to the next hundredth for categories since it is sending a full INT.

I don't know if this is a bad practice or not, I was just worried about confusing myself and locking myself into a structure that seemed logical (higher number = more rare) but then switched to unidentified, cursed, and then more rare.

akirmse: Changing values that are used in a network protocol is dangerous in general, because there can be existing clients out there that you break when you change the values.  It's not an issue in this particular case, since it hasn't launched yet, but it's something to keep in mind.  Let me know your preference.

adrienlaws: Understood.  I guess since we haven't done anything yet, we could do the current revision.  The only issue is if we add more grades in between rare and legendary or uncommon and rare it will fall apart again.  Maybe I should do by 10s? I get a IP addressing vibe from this when people design their networks to have space for expansion or "categories" of IP ranges (printers are 10-20, etc).

I guess as is works, or I can do every 10 so then 0 is normal, 10 is uncommon, 20 rare, etc.

From what I've done is brief 1:1 questioning with current players they are mostly interested in the unidentified rarity grade since that seems to be a huge pain point at the moment.  They really don't seem to have many questions about the higher rarity grades, at the moment anyway.

adrienlaws: Ugh, sorry for all the comments.

I think if I use bigger numbers the calculation for rarity will be messed up anyway.  As it stands is probably fine, and then if someone changes things they will have to adjust the count calculation anyway.

For example if rare is 20 then adding 3 item counts to it will not reach the next level.

In short, I don't think my idea will really scale well anyways.  We will need to modify the calculation function regardless.  So I guess the current state is fine as it's pseudo-intuitive?

PR #747, opened 2024-05-30
Correctly scale ui elements in gpu efficiency mode

sicoderealm:
Correct the ui scaling when in gpu efficiency mode
image
becomes:
image

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -3252,14 +3252,15 @@ Bool D3DObjectLightingCalc(room_type *room, room_contents_node *pRNode, custom_b
 
 Bool D3DComputePlayerOverlayArea(PDIB pdib, char hotspot, AREA *obj_area)
 {
-	float	screenW, screenH;
-
    // Scaling factor for UI elements (Scimtar/shield etc) using original magic number scaling.
    // The original magic numbers used here were 1.75f (width) and 2.25f (height) for 800 by 600.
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -3252,14 +3252,23 @@ Bool D3DObjectLightingCalc(room_type *room, room_contents_node *pRNode, custom_b
...
-
-   // Scaling factor for UI elements (Scimtar/shield etc) using original magic number scaling.
-   // The original magic numbers used here were 1.75f (width) and 2.25f (height) for 800 by 600.
-   // We have now scaled both of these for 1080p from 800 by 600 (2.4 and 1.8 respectively).
-   // Giving us the final scaling factors of 4.15f and 4.05f.
-   screenW = (float)(gD3DRect.right - gD3DRect.left) / (float)(main_viewport_width * 4.15f);
-   screenH = (float)(gD3DRect.bottom - gD3DRect.top) / (float)(main_viewport_height * 4.05f);
+   // Scaling factor for UI elements (Scimtar/shield etc).
+   // Reference resolution and scaling factors for classic 800x600 resolution.
+   const float REFERENCE_WIDTH = 800.0f;
```
akirmse, line 3257: Yes, but I'd like to actually share the values that are here:

https://github.com/Meridian59/Meridian59/blob/3ea980dd47e7676c889e23cf39433fc02a8f1e78/clientd3d/d3ddriver.c#L77

so that it can be changed in one place and work everywhere.

PR #749, opened 2024-05-31
Fix for DM-added attributes not being revealed

Aesica:
This PR causes item attributes added through the "dm create itematt <attribute>" command to always be revealed.  Previously, an admin using this command would have to use the reveal spell, which is problematic if one of the added attributes was shroud.

If the old behavior of adding an unrevealed attribute is preferred, the command "dm create unrevealed itematt <attribute>" has been added.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
@@ -32,6 +32,7 @@ resources:
    admin_currently_in = "The ROO for `b%s`n is `k`B%s`n."
 
    admin_create_itematt_command = "create item attribute"
+   admin_create_unrevealed_itematt_command = "create unrevealed itematt"
    admin_anonymous_command = "anonymous"
    admin_shadow_command = "shadow"
    admin_relic_command = "relic"
@@ -138,11 +139,12 @@ messages:
 
    UserSay(string = $,type = $)
    {
-      local oWeapon, i, oObject;
+      local oWeapon, i, oObject, lItemAttributes, lAttribute;
 
       if type = SAY_DM
          AND (StringContain(string,admin_create_itematt_command)
-              OR StringContain(string,"create itematt"))
+              OR StringContain(string,"create itematt")
+              OR StringContain(string, admin_create_unrevealed_itematt_command))
       {
          %% Only works on weapons for now.
          oWeapon = Send(self,@LookupPlayerWeapon);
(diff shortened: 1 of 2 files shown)


kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
```diff
@@ -159,6 +161,17 @@ messages:
             if Send(i,@DMCreateItemAtt,#who=self,#string=string,
                     #oWeapon=oWeapon)
             {
+               % Identify the newly-added attribute
+               if (NOT StringContain(string, admin_create_unrevealed_itematt_command))
```
akirmse, line 165: It feels weird to check the user's speech again here.  How about setting a boolean flag earlier, saying whether the attribute should be revealed?  That will also be less fragile in the future.

kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
```diff
@@ -159,6 +161,17 @@ messages:
...
+               {
+                  lItemAttributes = Send(oWeapon, @GetItemAttributes);
+                  lAttribute = First(lItemAttributes);
+                  if (lAttribute <> $ AND First(lAttribute) mod 2 = 0)
```
akirmse, line 169: What wizardry is this?  Definitely deserves a comment.

kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
Aesica: Done.

kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
Aesica: LOL I guess when you put it that way, the whole item attribute compound number isn't exactly user friendly at a glance.  Wrote up what I hope explains what's going on to ultimately result in the attribute being revealed.

kod/object/active/holder/nomoveon/battler/player/user/dm/admin.kod
```diff
@@ -138,12 +139,15 @@ messages:
...
-      local oWeapon, i, oObject;
+      local oWeapon, i, oObject, lItemAttributes, lAttribute, bShouldRevealAttribute;
 
       if type = SAY_DM
          AND (StringContain(string,admin_create_itematt_command)
-              OR StringContain(string,"create itematt"))
+              OR StringContain(string,"create itematt")
+              OR StringContain(string, admin_create_unrevealed_itematt_command))
       {
+         bShouldRevealAttribute = NOT StringContain(string, admin_create_unrevealed_itematt_command);
```
akirmse, line 149: This is still also being checked in the previous line, too.  The idea was to check it once.

Aesica: Done, should be good now.

PR #751, opened 2024-05-31
fix: sanitize target URLs in client before using

adrienlaws:
# What

- checks and sanitizes URL strings before opening them from the client with a browser

# Changes

- checks for
  - empty URL string
    - returns if so
  - spaces in the URL
    - returns if so
  - lack of `HTTP/HTTPS` prefix
    - adds `HTTPS://` if there is no prefix
- adds relevant debug messages in client

### DIFF
--- a/clientd3d/web.c
+++ b/clientd3d/web.c
@@ -42,7 +42,7 @@ void WebLaunchBrowser(const char *url)
         urlStr = "https://" + urlStr;
     }
 
-    HINSTANCE result = ShellExecute(NULL, "open", urlStr.c_str(), NULL, NULL, SW_SHOWNORMAL);
+    ShellExecute(NULL, "open", urlStr.c_str(), NULL, NULL, SW_SHOWNORMAL);
 }
 
 // convert c string to wstring
--- a/clientd3d/web.c
+++ b/clientd3d/web.c
@@ -19,9 +19,34 @@ using namespace std;
 /*
  * WebLaunchBrowser:  Attempt to run browser on given URL.
  */
-void WebLaunchBrowser(char *url)
+void WebLaunchBrowser(const char *url)
 {
-  ShellExecute(NULL, "open", url, NULL, NULL, SW_SHOWNORMAL);
+    if (url == nullptr || strlen(url) == 0) {
+        debug(("URL string empty\n"));
+        return;
+    }
+
+    std::string urlStr(url);
+
+    if (ContainsSpaces(urlStr)) {
+        debug(("URL contains spaces\n"));
+        return;
+    }
+
+    if (urlStr.find("https://") != 0 && urlStr.find("https://") != 0) {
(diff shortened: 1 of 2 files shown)


clientd3d/web.c
```diff
@@ -19,9 +19,34 @@ using namespace std;
...
+        debug(("URL contains spaces\n"));
+        return;
+    }
+
+    if (urlStr.find("https://") != 0 && urlStr.find("https://") != 0) {
+        urlStr = "https://" + urlStr;
+    }
+
+    HINSTANCE result = ShellExecute(NULL, "open", urlStr.c_str(), NULL, NULL, SW_SHOWNORMAL);
+    if ((INT_PTR)result <= 32) {
```
akirmse, line 41: This check doesn't do anything useful; remove.

clientd3d/web.h
```diff
@@ -39,7 +39,8 @@ enum AccountStatusWebResponse
     FAILURE = 1 << 2
 };
 
-void WebLaunchBrowser(char *url);
+void WebLaunchBrowser(const char *url);
+bool ContainsSpaces(const std::string& str);
```
akirmse, line 43: This doesn't need to be in the header; we don't want other files calling ContainsSpaces.  Move it to the .c file and make it static.

clientd3d/web.c
```diff
@@ -19,9 +19,34 @@ using namespace std;
...
+    }
+
+    std::string urlStr(url);
+
+    if (ContainsSpaces(urlStr)) {
+        debug(("URL contains spaces\n"));
+        return;
+    }
+
+    if (urlStr.find("https://") != 0 && urlStr.find("https://") != 0) {
```
akirmse, line 36: Doing the same check twice?  Is one of these supposed to be http?

clientd3d/web.c
adrienlaws: Removed

clientd3d/web.h
adrienlaws: Understood.  Made static and moved to before its WebLaunchBrowser caller.  Removed from header file.

clientd3d/web.c
adrienlaws: Yes, mistake.  Changed to 2 separate checks (one should be HTTP, the other should be HTTPS).

clientd3d/web.c
```diff
@@ -15,13 +15,34 @@
...
+    if (ContainsSpaces(urlStr)) {
+        debug(("URL contains spaces\n"));
+        return;
+    }
+
+    if (urlStr.find("http://") != 0 && urlStr.find("https://") != 0) {
+        urlStr = "https://" + urlStr;
+    }
+
+    HINSTANCE result = ShellExecute(NULL, "open", urlStr.c_str(), NULL, NULL, SW_SHOWNORMAL);
```
akirmse, line 45: result unused; remove

PR #752, opened 2024-05-31
fix: make hotkey aliases refresh without restarting client 

adrienlaws:
# What

- this change reloads the hotkey aliases into memory after they are saved

# Why

- currently players must reload their client for changes in the hotkey aliases "automatic enter" to take effect
  - example: players change f1 from `help` to `tg <guild banner>` and uncheck automatic enter
    - when they close the hotkey aliases edit screen, it will still automatically enter
    - the only workaround is to restart the client
- it's a small quality of life change but I think will be welcomed by players
  - today while on 101 someone was complaining about this issue after telling a guildmate to log on and off for their aliases to be updated properly

# Changes

- created new function `UpdateKeyTables` to update saved in-memory key configuration
  - used by both `AliasInit` and `AliasSave`

### DIFF
--- a/module/merintr/alias.c
+++ b/module/merintr/alias.c
@@ -87,6 +87,21 @@ extern player_info *GetPlayer(void);
 
 /****************************************************************************/
 /*
+ * UpdateKeyTables:  Update either the classic or custom keys interfaces table
+ * depending on the client configuration, then also update the inventory key
+ * table so that the hotkey aliases work when the user is in that child window
+ */
+static void UpdateKeyTables(int key, WORD command, char *text)
+{
+   if (gbClassicKeys) 
+      AliasSetKey(interface_key_table, key, KEY_NONE, command, text);
+   else
+      AliasSetKey(gCustomKeys, key, KEY_NONE, command, text);
+   AliasSetKey(inventory_key_table, key, KEY_NONE, command, text);
+}
+
+/****************************************************************************/
+/*
  * AliasInit:  Read aliases from INI file and set them up.
  */
 void AliasInit(void)
@@ -147,12 +162,7 @@ void AliasInit(void)
 	 aliases[i].cr = True;
       }
 
-	  if (gbClassicKeys)
-		AliasSetKey(interface_key_table, aliases[i].key, KEY_NONE, command, aliases[i].text);
-	  else
-		AliasSetKey(gCustomKeys, aliases[i].key, KEY_NONE, command, aliases[i].text);
-
-      AliasSetKey(inventory_key_table, aliases[i].key, KEY_NONE, command, aliases[i].text);
+   UpdateKeyTables(aliases[i].key, command, aliases[i].text);
    }
 }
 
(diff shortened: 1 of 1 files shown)


module/merintr/alias.c
```diff
@@ -277,6 +277,8 @@ void AliasSave(void)
 	 }
       }
    }
+
+   AliasInit();
```
akirmse, line 281: I don't know what the underlying problem is, but this doesn't strike me as the right answer.  A function called "Save" should save, not load.

Whoever is calling AliasSave should also be adjusting the in-memory structures.  Can you figure out which caller isn't doing that correctly?

adrienlaws: Understood.

It looks like the in memory state was not saved when AliasSave was called to write to the .ini file, and that appears to be the cause of the issue.

I added in 2 in-memory saves when closing the `hotkey aliases` dialog box
1. either `interface_key_table` or `gCustomKeys` depending on whether the client is using classic keys or not
2. `inventory_key_table`

This is the same logic used when `AliasInit` is called.

From limited testing it now seems to work properly and the client does not require a restart to switch a hotkey alias from text command to text insert (or vice versa)

module/merintr/alias.c
```diff
@@ -251,12 +252,25 @@ void AliasSave(void)
...
       else
-	 sprintf(text, "%s~", aliases[i].text);
-
+      {
+         sprintf(text, "%s~", aliases[i].text);
+         command = A_TEXTINSERT;
+      }
       WritePrivateProfileString(fullSection, temp, text, cinfo->ini_file);
+
+      if (gbClassicKeys)
```
akirmse, line 268: This exactly duplicates existing code.  Make a shared function instead.

adrienlaws: Created new function `UpdateKeyTables` as suggested and called in within `AliasInit` and `AliasSave`

module/merintr/alias.c
```diff
@@ -85,6 +85,21 @@ static INT_PTR CALLBACK VerbAliasDialogProc(HWND hDlg, UINT message, WPARAM wPar
 extern Bool	gbClassicKeys;
 extern player_info *GetPlayer(void);
 
+/****************************************************************************/
+/*
+ * UpdateKeyTables:  Update either the classic or custom keys interfaces table
+ * depending on the client configuration, then also update the inventory key
+ * table so that the hotkey aliases work when the user is in that child window
+ */
+static void UpdateKeyTables(int key, WORD command, char *text)
```
akirmse, line 94: This is minor, but it's better to put helper functions like this closer to the end of the file, so that important functions like "AliasInit" come first, as they're more likely to be what a reader is looking for.

In order to get that to compile, you'll need to forward declare UpdateKeyTables.  Line 84 would be a good spot.

adrienlaws: Thanks.  I didn't know about forward declarations and that was helpful.  I have done as you suggested.

PR #753, opened 2024-06-02
Reinstate animation sleep to lower CPU load while in background

sicoderealm:
This change reintroduces the original sleep used to lower CPU load while the game is running in the background which was previously [removed](https://github.com/Meridian59/Meridian59/pull/452).

In addition to this change we now also improve app background detection for sleeping, reduce maximum frame rate of 120 FPS and finally prevent the animation timer triggering on non game screens (was happening on at least the character selection window).

A background sleep of 1ms quickly produces a 0% cpu when the game is in the background without significant game activity (such as numerous hostile monsters animating). Using a value of 0ms does not achieve 0% cpu and actually *significantly* raised background CPU usage due returning immediately (https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep).

### DIFF
--- a/clientd3d/statgame.c
+++ b/clientd3d/statgame.c
@@ -27,6 +27,8 @@ static int window_width = 0;
 
 static POINT previous_mouse_position;
 
+#define ANIMATE_BACKGROUND_SLEEP_MS 50
+
 /****************************************************************************/
 void GameInit(void)
 {
@@ -358,6 +360,11 @@ void GameCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
    }
 }
 /****************************************************************************/
+void AnimationSleep(void)
+{
+    Sleep(ANIMATE_BACKGROUND_SLEEP_MS);
+}
+/****************************************************************************/
 void GameTimer(HWND hwnd, UINT id)
 {
    switch (id)
@@ -377,6 +384,12 @@ void GameIdle(void)
 
    // Are we the active foreground application?
    AnimationTimerAbort();
+
+   // If we're in the background, sleep to lower CPU load
+   if (GetWindowLong(GetActiveWindow(), GWL_HINSTANCE) != (LONG)hInst)
+   {
+       AnimationSleep();
+   }
 }
 /****************************************************************************/
 void GameEnterIdle(HWND hwnd, UINT source, HWND hwndSource)
--- a/clientd3d/statgame.c
+++ b/clientd3d/statgame.c
@@ -27,7 +27,7 @@ static int window_width = 0;
 
 static POINT previous_mouse_position;
 
-#define ANIMATE_BACKGROUND_SLEEP_MS 50
+#define ANIMATE_BACKGROUND_SLEEP_MS 1
 
 /****************************************************************************/
 void GameInit(void)
@@ -360,11 +360,6 @@ void GameCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
    }
 }
 /****************************************************************************/
-void AnimationSleep(void)
-{
-    Sleep(ANIMATE_BACKGROUND_SLEEP_MS);
-}
-/****************************************************************************/
 void GameTimer(HWND hwnd, UINT id)
 {
    switch (id)
@@ -374,6 +369,18 @@ void GameTimer(HWND hwnd, UINT id)
       break;
    }
 }
+
+void BackgroundSleep()
+{
+    // If we're in the background, sleep to lower CPU load.
+    HWND hWnd = GetForegroundWindow();
+    bool backgroundSleep = (hWnd == NULL) || (GetWindowLongPtr(hWnd, GWLP_HINSTANCE) != (LONG_PTR)hInst);
+
+    if (backgroundSleep)
+    {
+        Sleep(ANIMATE_BACKGROUND_SLEEP_MS);
+    }
+}
 /****************************************************************************/
 void GameIdle(void)
 {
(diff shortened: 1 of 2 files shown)


davidklaw:
The original sleep code was removed intentionally, so that the animation speed would be unaffected by the game's various dialogs. If the background animation speed change alone is an improvement, it'd be nice to take that step and not reintroduce background sleeping. Or, if we do reintroduce, don't sleep in the case of dialogs whose parent is current game application.

Aesica:
This seems like it'd be another thing that'd work well as a selectable option.  Disabling background sleep does make the gameplay experience smoother, however there are merits for having Meridian run in "eco mode" so to speak.  I'd likely have background sleep enabled if presented with the choice.

davidklaw:
I think that "in the background" means multiple things here. The original fix was meant to address sleeping and animation speed when dialogs are present.

Here's one of the videos from the original PR. I forget exactly what state it's representative of, but it shows the original issue. Looking up an item in the world and opening a buy dialog (at Izzio for example) would greatly slow down the game, which still remains largely visible. That felt undesirable from a user experience perspective.

https://private-user-images.githubusercontent.com/467443/228704161-6f1a45a3-7dc9-434a-a2a2-924886d20cba.mp4?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTc0Mzg0NDYsIm5iZiI6MTcxNzQzODE0NiwicGF0aCI6Ii80Njc0NDMvMjI4NzA0MTYxLTZmMWE0NWEzLTdkYzktNDM0YS1hMmEyLTkyNDg4NmQyMGNiYS5tcDQ_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNjAzJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDYwM1QxODA5MDZaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT05NzFkMzZhMGQxOWM3MGM3ODY0YWU0OGM0NjFkNjViZGJhOTAxYmFhZDQzODU5ZWVlY2I2ZGM5ZGViNTc2Zjk4JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.epXig8cxnM7mCw3pMnaISyVLah_M9p5kXzmDZbcPb9c

clientd3d/statgame.c
```diff
@@ -370,13 +381,31 @@ void GameTimer(HWND hwnd, UINT id)
 /****************************************************************************/
 void GameIdle(void)
 {
+   static const std::chrono::duration<double> minInterval(1.0 / 60.0); // 1/60th of a second
+   static auto lastCallTime = std::chrono::high_resolution_clock::now();
```
akirmse, line 385: This is not the last call time, it's the last time the function was allowed to run to completion.

clientd3d/statgame.c
```diff
@@ -370,13 +381,31 @@ void GameTimer(HWND hwnd, UINT id)
...
    MoveUpdateServer();       // Update our position on server, if necessary
 
    AnimationTimerProc(hMain, 0);
    HandleKeys();
 
    // Are we the active foreground application?
    AnimationTimerAbort();
+
+   // If we're in the background, sleep to lower CPU load
+   if (GetWindowLong(GetActiveWindow(), GWLP_HINSTANCE) != (LONG_PTR)hInst)
```
akirmse, line 405: I can't tell whether this is the right check.  See e.g.
https://stackoverflow.com/questions/3940346/foreground-vs-active-window

If the app is not in the foreground, I think GetActiveWindow() is returning NULL, which probably "works" but only by accident.  Can you do some more research on this, and add error checking?

clientd3d/statgame.c
```diff
@@ -370,13 +381,31 @@ void GameTimer(HWND hwnd, UINT id)
...
+   {
+       AnimationSleep();
```
akirmse, line 407: A separate function for this seems like overkill.

Also, should this sleep be for 1 or 0 ms?  A 0ms sleep yields to all other threads (gives up the rest of this thread's time slice), which might be enough to reduce CPU usage, but I'm not sure.

clientd3d/animate.c
```diff
@@ -27,7 +27,7 @@
 
 #include "client.h"
 
-#define ANIMATE_INTERVAL 8  // ms between background animation updates (8.3333 = 120fps)
+#define ANIMATE_INTERVAL 8 // ms interval between background animation updates
```
akirmse, line 30: Revert this line.  I believe the new comment is incorrect; this interval has nothing to do with being in the background.

clientd3d/statgame.c
```diff
@@ -370,13 +381,31 @@ void GameTimer(HWND hwnd, UINT id)
 /****************************************************************************/
 void GameIdle(void)
 {
+   static const std::chrono::duration<double> minInterval(1.0 / 60.0); // 1/60th of a second
+   static auto lastCallTime = std::chrono::high_resolution_clock::now();
+   auto currentTime = std::chrono::high_resolution_clock::now();
+   std::chrono::duration<double> timeSinceLastCall = currentTime - lastCallTime;
+   
+   // Guard to ensure the function is not called more than 60 times per second
+   if (timeSinceLastCall < minInterval) {
```
akirmse, line 390: Doesn't this lock graphics updates to 60Hz?

clientd3d/statgame.c
```diff
@@ -375,14 +389,15 @@ void GameIdle(void)
...
    AnimationTimerAbort();
+   BackgroundSleep();
 }
 /****************************************************************************/
 void GameEnterIdle(HWND hwnd, UINT source, HWND hwndSource)
 {
-   // When menu or dialog goes up, start animation timer in the background
-   AnimationTimerStart();
+   // When menu or dialog goes up during gameplay, start animation timer in the background
+   if (game_state == GAME_PLAY)
```
akirmse, line 399: With this change, does the two knights bonking animation in the About dialog still work when the client is not yet logged in?  Maybe not a huge deal if not but it would be good to know.

Issue #466, opened 2023-04-16
Gameplay: Monsters: Frogman loot is imbalanced, shares same treasure.kod as Stone Troll


jrotunda:
The Frogman (level 70) drops the same type and amount of loot as the opposite karma Stone Troll (level 120).

Loot dropped is a large amount of expensive reagents found on the Island.

For context: The stone troll was specifically made to carry this loot for the Revelations update, so players could dispel illusion the caves after no way to get those reagents on the mainland.

The frogman's original loot table viewable in the Ancient monster loot table document is more in line with what it's expected to drop, but it could use improvement to be more specific andpractical.

It's expected the Frogman drop reagents more appropriate to its level, for the new players and negative karma characters fighting it,and that make sense with its lore -negative karma reagents and things the frog may be eating around Meridian, such as:

- Edible Mushrooms
- Red or Purple Mushroom (not both)
- Dragonfly Eyes 👁 
- Orc Teeth 🦷 
- Entroot
- Fairy Wing 😈🧚‍♂️

--------

(Note: I'll work on this, logging here for reference)

jrotunda:
Needs:

Unique and new treasure template for Frogman & Stone Troll, remove generic template.

PR #758, opened 2024-06-04
Create a Frogman treasure table

Gar103:
Per https://github.com/Meridian59/Meridian59/issues/466

The treasure of Frogmen was the same as the Stone Troll, likely because the Frogmen were not actually placed in the world until recently, so it was just an ancient placeholder. This pull creates a treasure table based on some notes in the Ancient monster loot table. It's generally berries and mushrooms, with a tiny sprinkling of DAFs and useful equipment, to match Skeletons (the opposite karma path to this one) dropping BDS now and again.

The diff seed is also similar to Skeletons, but with an att chance similar to Shamans and other sentients, since Frogmen are intelligent beings who might seek out and wield special weapons. The special weapons they drop can only be the semi-useful intro stuff, like ceremonial. They cannot drop the powerful mods like Hold or Blind. I'm hoping this will help differentiate the karma paths and give a different feel for building as a Qor or Faren. Also, as I've posted in the Issues section, I believe the RNG for special weapons is overly punishing, so giving a few ways to get some drops is a good thing in and of itself.

I tried dragonfly eyes (per the issue suggestion) but they just didn't feel right when I was killing frogmen and looking at the drops. The general vibe for drops from this treasure table is intended to be 'dark' and 'natural', so it's mainly berries, mushrooms, etc, and wings and feathers the frogmen might have scavenged. I couldn't justify the eyes, because there are no dragonflies on the mainland. Plus, low level early builders don't really have a use for them.

There is also, super rarely, some equipment the frogmen themselves might hypothetically use. Chain armor is great against frogmen because they do pierce damage, and magic shield potions are something frogmen might have sorted in order to use themselves. Also a small shield, to help builders equip themselves.

### DIFF
--- a/kod/include/blakston.khd
+++ b/kod/include/blakston.khd
@@ -1591,6 +1591,8 @@
    TID_DARK_ANGEL    = 51
    TID_MUMMY         = 52
    TID_THRASHER      = 53
+   
+   TID_FROGMAN      = 54
 
    %%% English articles
 
--- a/kod/object/active/holder/nomoveon/battler/monster/frogman.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/frogman.kod
@@ -42,7 +42,7 @@ classvars:
    vrDead_icon = frogman_dead_icon_rsc
    vrDead_name = frogman_dead_name_rsc
 
-   viTreasure_type = TID_MEDIUM_HUMAN
+   viTreasure_type = TID_FROGMAN
 
    viSpeed = SPEED_AVERAGE
    viAttack_type = ATCK_WEAP_PIERCE
--- a/kod/object/passive/trestype/frogmant.kod
+++ b/kod/object/passive/trestype/frogmant.kod
@@ -1,0 +1,50 @@
+% Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
+% the LICENSE file that accompanies it.
+%
+% Meridian is a registered trademark.
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+FrogmanTreasure is TreasureType
+
+constants:
+   
+   include blakston.khd
+   
+classvars:
+
+
+properties:
+   
+   piTreasure_num = TID_FROGMAN
+   
+   %% frogman drops stuff good for qor and faren builders below 80 hp
+   % Compare to Skeletons - similar level/diff, opposite karma path
+   piDiff_seed = 5
+   piItem_att_chance = 8
+
+messages:
+   
+   constructed()
+   {
+      plTreasure = [ [ &EntrootBerry, 24 ],
+                     [ &PurpleMushroom, 24 ],
+                     [ &BlueMushroom, 15 ],
+                     [ &Mushroom, 10 ],
+                     [ &ElderBerry, 10 ],
+                     [ &InkyCap, 4],
+                     [ &FairyWing, 4 ],
+                     [ &DarkAngelFeather, 6 ],
+                     [ &ChainArmor,1 ],
+                     [ &MetalShield, 1],
+                     [ &MagicShieldPotion, 1 ]
+                   ];
+
+      propagate;
+   }
+
+end
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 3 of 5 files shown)


Aesica:
That treasure table is *really* good.  And by good, I mean lucrative.  Frogmen aren't particularly powerful foes, so they probably shouldn't be dropping things like dark angel feathers with a drop weight of 6, or highly sought-after reagents like entroot berries and purple mushrooms as their primary drops (24 each).

Gar103:
That's a valid concern, but I'd like to say I've done the work here to make sure this works. Skeletons, the good karma equivalent, drop BDS and inkies at the same 6% rate. Living Trees are much weaker and drop entroot berries at a rate of 40%. This treasure table was built using numerous comparisons, and I tested it by building on these frogmen as well on my local server, seeking a point where it felt decent. The main thing that's not directly on that table is that frogmen drop few items per kill and have 65% nonmagic weapon resistance, making them actually kind of annoying to fight unless you maintain repeated weapon enchantments. The areas they appear in (like King's Way) also have them generally spread out, rather than appearing in a concentrated fast spawning situation like CV.

The other thing is - entroot berries and purple mushrooms themselves are not directly lucrative. Their value is virtual, it's just great to loot them compared to paying 200 each from Tova or 160 from Zhieu B'hob. They can't be sold by the player for massive profit. They're actually going to use them to build spells. & that's the idea here. Build on frogmen, stockpile to build your spells.

I am definitely on the side of keeping drops as punishing and unrewarding as they currently are. My experience with OS showed that boosting drops and other improvements can quickly get out of hand and ruin the feel... so I've carefully selected the numbers and drops here to give Qor builders a path they don't currently have. Remember, there's also 52% chance here of getting mushrooms, elderberries, etc. Many frogmen kills will yield essentially nothing for you, unless you're a Faren actively using those elders and blues to cast offensive spells.

Aesica:
I see what you're getting at, although:

- I'm not sure it's a good idea to balance drop rates and treasure tables around how convenient a monster's current locations are, since that can always change, and spawns on 101 get revised periodically to keep things fresh.
- I'd consider purples (24), entroots (24), dafs (6), and inkies (4) the "high value" drops from this table.  With skeletons, it's a more narrow range with blue dragon scales (6), ruby (4), and inky cap (4) -- diamond (6) could also be argued as valuable I suppose.  If I were making this table, I'd probably increase the rates for mushrooms, blue mushrooms, elderberries, and fairy wings while scaling back purples and entroots.
- Trees have a higher drop rate, but also are a lot lower in power so they tend to drop less of everything.  Plus they come with the risk of killing weaker farmers with surprise back-to-back lightning bolts.  Frogmen are less risky in that regard.

davidklaw:
I like the discussion as it's helping me see sides of this I otherwise wouldn't have considered. I'm leaning towards being conservative. Does removing 1-2 of these high-value drops from the loot table make this more palatable? I see that as giving "evil" players a reason to go after frogmen, but not also making them farming grounds and forcing the player to move around the map from time to time for the other regs they need.

Aesica:
I don't think anything needs to be removed, just the weighting needs some adjustment so that high value items (purple mushrooms and entroot berries) aren't so common.

Gar103:
> I'm not sure it's a good idea to balance drop rates and treasure tables around how convenient a monster's current locations are, since that can always change, and spawns on 101 get revised periodically to keep things fresh.

Drops should definitely be revised if locations are changed. Or there could even be a room-specific loot property to boost or diminish based on location. Monster locations are integral to the risk you take and the reward you feel when building in an area - as I understand it, 101's spawns have basically been drastically changed without any sort of team review or record of what was done. The new spawns should definitely be posted, at least in dev discussion, for an analysis. Frogmen treasure is currently absolutely awful, dropping essentially nothing while also being in a high traffic area that will _definitely_ get you attacked sooner or later. I don't think we can ignore that part of the balance.

I'll refrain from jumping into a point-by-point debate kind of thing, though. I know I can be long-winded. Personally I'd wager, just put it in and try it. It's not lucrative, it's just different. We have never had a good source of purps, or a good source of ents that isn't a single screen of living trees. It's time Qor had that. The experience is very different from the way the numbers look - if you drop the purps and ents rates even more, and replace that with browns and elders, that might as well be nothing. It should absolutely be a source of purps and ents, rather than purps and ents incidentally dropping.

I wouldn't up fairy wings though. That primary source should definitely still be the vale. The fairy wings are in there as a rare thing to ease enchanting your weapon to Qor (tbh I'd be fine with removing fairy wings completely, they are sold in CN anyway)

davidklaw:
> Seems sensible to have a unique loot table dedicated to the mob. Perhaps we can add the new loot table with exactly the same loot drops inherited from tid_medium_human. Then follow up with the amendments in a separate pr?

Loot tables can be changed on the fly by server admins, right? We have plenty of time to debate this, but the point I took from this comment is that we can keep this moving by adding the unique frogman loot table. That we know we need/want. The server admins can adjust from there given and maybe it then make sense to codify it in the future, but we can keep it rather baseline until then.

PR #760, opened 2024-06-05
fix: make reveal and identify work on shrouded items (IA_SHROUD)

adrienlaws:
# What

- this change makes reveal and identify work on the `IA_SHROUD` item attribute

# Why

- when shroud is cast by a player, the `IA_SHROUD` item attribute is added as an identified attribute
- however, in some cases an admin can manually add the `IA_SHROUD` item attribute to an item as an unidentified attribute
    - with the new rarity grade system these items show up in the client as `UNIDENTIFIED`
- this change "fixes" the issue by allowing players to reveal and identify shrouded items
  - since `IA_SHROUD` does have a unique description string, it should be identifiable or reveal-able
    - > A cloak of magic surrounds it.

# Testing

- in limited testing I was able to reveal and identify items with the `IA_SHROUD` attribute that were applied before this change
- no errors seen in the server log

# Notes

- there is another change submitted to by default have item attributes be revealed when applied through the `dm create itematt` command #749 
- `IA_SHROUD` has a timer and should not last more than a day, even when applied by admins
  - the timer doesn't seem to work properly when applied by admins
    - players casting shroud seems to apply the timer correctly
  - should be addressed in another PR/Issue

### DIFF
--- a/kod/object/passive/itematt/iashroud.kod
+++ b/kod/object/passive/itematt/iashroud.kod
@@ -80,11 +80,6 @@ messages:
 		return FALSE;
 	}
 
-	ItemCanIdentify()
-	{
-		return FALSE;
-	}
-  
 	%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 	%%
 	%%	Maintenance Routines


akirmse:
The behavior of a few old admin-created objects seems minor.  I'm more concerned about the behavior change to player-shrouded items, especially if the timer is broken.  Can you say more about how this changes those items?

adrienlaws:
I must apologize. I was wrong.

The timers for shroud seem to work properly (both admin and player use)
- I tested this by applying them in my dev environment both with the spell and as an admin.  
- They work fine and the shroud enchantment `IA_SHROUD` is removed when the timer expires.

In terms of the unidentified admin-created shrouded items on the live server, it appears that the value where the timer should be is `$ 0` which doesn't seem normal.
- It might be these attributes were created in a unique way that didn't matter until the rarity grade system was implemented
- No timer is involved in this as far as I can tell

Here is an example object on the live server, where `INT 1000` is an unidentified shroud attribute in the `plItem_attributes` list with a timer of `$ 0`.
```
:<
: [
: [
: INT 1101
: $ 0
: STRING 49172
: ]
: [
: INT 1000
: $ 0
: ]
: [
: INT 501
: ]
: ]
:>
```

I will update the description of this PR.

I guess the question now is - should shrouded items be able to be revealed or identified?
- I would argue yes because there is a description for revealed/identified shrouded items.
  - this implies that they should be revealable
  - the shroud spell also applies itself as revealed (skipping the unidentified setting)
  - in normal cases shroud is already revealed, in abnormal cases we should allow for it to be revealed (IMO)
- Shal'ille is not what I would consider "polluting" magic (as defined by shroud's description) when it is evaluating the item and not modifying it.
- You also cannot cast reveal or identify on another player's item without them dropping the item or giving it to you

akirmse:
OK, seems fine then.

Issue #738, opened 2024-05-28
Graphics: Translucency effects aren't showing in the Hardware Renderer 


jrotunda:
This is a new issue that occurred after the last patch updated the rendering. 
The Draw FX translucency effects aren't showing anymore in the Hardware Renderer 
_(I saw that there's a fix for the Invisibility effect, unsure if it will fix this too)_

This effect is used for various classes:
 - ghosts, slimes, shadowbeasts, spectral mummies
 - NPC's like Tendrath (who's a ghost)
 - portals
 - walls of lightning, walls of fire, fog, poison fog
 
Affects the following Draw FX definitions f
![Meridian 59 - Translucent Draw FX - Hardware Rendering - 1](https://github.com/Meridian59/Meridian59/assets/2608213/15977b18-c917-444f-a5bb-aabce8d68545)
rom the code:
#define OF_TRANSLUCENT25 0x00100000    // Set if object should be drawn at 25% opacity
#define OF_TRANSLUCENT50 0x00200000    // Set if object should be drawn at 50% opacity
#define OF_TRANSLUCENT75 0x00300000    // Set if object should be drawn at 75% opacity

PR #763, opened 2024-06-07
Reinstate translucent effects in hardware renderer

sicoderealm:
Reinstate translucent effects (shadow beast / slimes) in the hardware renderer.

This change also prevents translucent objects (or any object) from being rendered twice. This would manifest as a slight visible reduction of translucency from certain distances. This change also supports improved translucency as previously you couldn't see other world objects at times behind a translucent object and now this should generally work correctly (second screenshot was not possible in the past)

image

image

Resolves: #738

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -7,6 +7,7 @@
 // Meridian is a registered trademark.
 #include "client.h"
 #include <unordered_map>
+#include <unordered_set>
 #include <chrono>
 
 #define	TEX_CACHE_MAX_OBJECT	8000000
@@ -29,6 +30,8 @@ inline float FovVertical(long height)
 {
 	return height / (float)(main_viewport_height) * (PI / 5.6f);
 }
+
+static const auto TRANSLUCENT_FLAGS = OF_TRANSLUCENT25 | OF_TRANSLUCENT50 | OF_TRANSLUCENT75 | OF_DITHERTRANS;
 
 // Update the pChunks animation values as a function of time.
 static void updateRenderChunkAnimationIntensity(d3d_render_chunk_new* pChunk)
@@ -322,7 +325,6 @@ d3d_render_chunk_new	*D3DRenderChunkNew(d3d_render_packet_new *pPacket);
 void					D3DRenderChunkInit(d3d_render_chunk_new *pChunk);
 d3d_render_packet_new	*D3DRenderPacketFindMatch(d3d_render_pool_new *pPool, LPDIRECT3DTEXTURE9 pTexture,
 												PDIB pDib, BYTE xLat0, BYTE xLat1, int effect);
-float					D3DRenderObjectLightGetNearest(room_contents_node *pRNode);
 void					D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 							Draw3DParams *params, int flags);
 void					D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DParams *params,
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -7064,16 +7059,35 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
...
 			bgra.a = D3DRENDER_TRANS50;
+			translucent = true;
+		}
 		if (GetDrawingEffect(pRNode->obj.flags) == OF_DITHERINVIS)
+		{
 			bgra.a = D3DRENDER_TRANS50;
+			translucent = true;
+		}
+
+		if(translucent)
```
akirmse, line 7089: I think you can check bgra.a here directly rather than introducing the "translucent" variable.

clientd3d/d3drender.c
```diff
@@ -7720,16 +7734,35 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
...
 						bgra.a = D3DRENDER_TRANS50;
+						translucent = true;
+					}
 					if (GetDrawingEffectIndex(pRNode->obj.flags) == (OF_DITHERINVIS >> 20))
+					{
 						bgra.a = D3DRENDER_TRANS50;
+						translucent = true;
+					}
+
+					if (translucent)
```
akirmse, line 7764: here too

clientd3d/d3drender.c
```diff
@@ -926,14 +924,12 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
...
-		IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_ZENABLE, D3DZB_TRUE);
+		IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_ZENABLE, TRUE);
 		if (gD3DDriverProfile.bFogEnable)
 			IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_FOGENABLE, TRUE);
 
-		// restore the view and world matrices
-		MatrixIdentity(&mat);
+		// restore the view matrix
 		IDirect3DDevice9_SetTransform(gpD3DDevice, D3DTS_VIEW, &view);
-		IDirect3DDevice9_SetTransform(gpD3DDevice, D3DTS_WORLD, &mat);
```
akirmse, line 936: I don't understand why this was removed---what does the world matrix have to do with translucency?

clientd3d/d3drender.c
```diff
@@ -6909,14 +6879,34 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
...
+		{
+			if ((GetDrawingEffect(pRNode->obj.flags) & OF_INVISIBLE) != OF_INVISIBLE)
 				continue;
 		}
 		else
 		{
-			if (GetDrawingEffect(pRNode->obj.flags) == OF_INVISIBLE)
+			if ((GetDrawingEffect(pRNode->obj.flags) & OF_INVISIBLE) == OF_INVISIBLE)
+				continue;
+		}
```
akirmse, line 6900: Weird logic here.. looks like checking GetDrawingEffect is not needed.

clientd3d/d3drender.c
```diff
@@ -6909,14 +6879,34 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
...
+
+		// Check for translucent objects
+		if ((flags & TRANSLUCENT_FLAGS) != 0) {
+			if ((GetDrawingEffect(pRNode->obj.flags) & TRANSLUCENT_FLAGS) == 0)
+				continue;
+		}
+		else 
+		{
+			if ((GetDrawingEffect(pRNode->obj.flags) & TRANSLUCENT_FLAGS) != 0)
```
akirmse, line 6909: here too

clientd3d/d3drender.c
akirmse: Hard to read this.. suggestion:
```
bool objInvisible = (pRNode->obj.flags & OF_INVISIBLE) == OF_INVISIBLE);
```
and use that in the checks.  The logic could further be simplified, I think to an XOR, but up to you whether that's a good idea.

clientd3d/d3drender.c
```diff
@@ -6909,14 +6874,31 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
...
 		}
 		else
 		{
-			if (GetDrawingEffect(pRNode->obj.flags) == OF_INVISIBLE)
+			if ((pRNode->obj.flags & OF_INVISIBLE) == OF_INVISIBLE)
+				continue;
+		}
+
+		// Check for translucent objects
+		if ((flags & TRANSLUCENT_FLAGS) != 0) {
```
akirmse, line 6895: same thing here

clientd3d/d3drender.c
```diff
@@ -6909,14 +6874,31 @@ void D3DRenderObjectsDraw(d3d_render_pool_new *pPool, room_type *room,
 		if (pRNode->obj.id == player.id)
 			continue;
 
-		if (flags & OF_INVISIBLE)
+		if (processedIds.find(pRNode->obj.id) == processedIds.end())
```
akirmse, line 6877: Another option would be to insert all the objects in a set before the loop to remove duplicates, then iterate over the set. 
 Then you wouldn't need to check each time through the loop.

But even if not, rewrite to the clearer:
```
if (processedIds.find(pRNode->obj.id) != processedIds.end())
  continue;
processedIds.insert(pRNode->obj.id);
```

PR #765, opened 2024-06-09
fix: remove GetRarityCountModifier from WeapAttSpellType

adrienlaws:
# What

- this change adjusts the item rarity grade calculation to ignore the `WeapAttSpellType` item attribute

# Why

- the `WeapAttSpellType` should never had been assigned a rarity count modifier
  - I was mistaken in what this item attribute did
    -  it always has a timer and appears to be different types of enchantments for weapon magic attack types (fire, shock, holy, evil, etc).
    - this was causing problems because holy weapon and unholy weapon were showing as rare.
 - In any case, it should always have a timer, and since it does then we do not want to count it towards rarity.

# Changes

- remove `GetRarityCountModifier` from `WeapAttSpellType` so it by default returns `0` instead of +`1`


adrienlaws:
I just realized we have to consider artifice. That should have a timer but we don't want to exclude it because of the nature of the spell. I will follow up on this.

kod/object/item.kod
```diff
@@ -719,6 +720,13 @@ messages:
          for i in plItem_attributes
          {
             iNum = Send(self,@GetNumFromCompound,#compound=First(i));
+            if ((Length(i) >= 2)   % Timers are 2nd in the sublist
+               AND (Nth(i,2) <> $) % If timer has a value
+               AND NOT(iNum = 12)) % If NOT IA_MISDIRECTION / Artifice
```
adrienlaws, line 725: Not sure if hardcoding `blakston.khd` constant of `IA_MISDIRECTION = 12` is okay or if I should use the result of the lookup from the @FindItemAttByNum value below somehow to make it less fragile if the number is changed.

akirmse: - Use the <> operator instead of NOT =
- Use IA_MISDIRECTION instead of 12
- This is gross, having Item, one of the highest-level classes, know about a particular spell. See if you can find a way where Item asks the oItemAtt instead whether it should be included in the count. e.g. could ItemAttMisdirection simply return 0 from GetRarityCountModifier, or something like that?

adrienlaws: Thank you for the feedback.  In short, I've removed a rarity grade count modifier for a problematic attribute that always has a timer.  This seems to be much more elegant and consistent than my previous ill-conceived solution.  I've updated the PR name and description to reflect this.

More detail below if you are interested.

- I thought about this a lot, and the existing `GetRarityCountModifier` is probably the best way to go.
- An item attribute will also have a message called `HasTimer` that by default returns `False`

The intent of my change was to have a catch-all that would always find the timer in position 2 within the `plItem_Attributes` list that is assigned to an item.
  - this is different than checking the item attribute class to see if it has a timer
  - the use case was that perhaps an admin assigned an item attribute that normally had a timer (or vice versa) then manually edited the timer to be `NIL` or have a value when it normally doesn't.
    - I'm going to just go with the normal flow as designed by the code without trying to do a catchall
    - part of the problem is that the timer is not always in the same location and something else may be in position 2 so the method I was working on produces inconsistent results

PR #767, opened 2024-06-11
WA_Vamper steal life from target

SioraSidhe:
Take life from the target when triggering WA_Vamper. Clean up code formatting.


davidklaw:
@SioraSidhe Please share more detail for everyone to understand what prompted this change and what you expected outcome is, so that we can not only discuss and verify it meets those expectations, but also have context to work with when referencing this later. Thanks.

SioraSidhe:
@davidklaw Happily, WA_Vamper weapon attribute on triggering heals the wielder, but it does not take that life from the target despite messages to bother wielder and target about sucking life away. It should take life from the target.

kod/object/passive/itematt/weapatt/wavamper.kod
```diff
@@ -58,31 +57,37 @@ messages:
...
-	    #parm1=send(target,@getcapdef),#parm2=send(target,@GetName),
-	    #parm3=send(target,@GetHisHer));
-    	 send(target, @MsgSendUser, #message_rsc = vamper_worked_target,
-	    #parm1=send(wielder,@getcapdef),#parm2=send(wielder,@GetName),
-	    #parm3=send(wielder,@GetHisHer));
-
-
+         iStolen = bound(damage,0,8);
+         iStolen = bound(iStolen,0,Send(target,@GetHealth)-1);
+         iStolen = Send(target,@AssessDamage,#what=wielder,#damage=iStolen,
```
akirmse, line 68: ModifyDamage is just supposed to return a modified damage number.  Turning one attack into two like this seems very strange and possibly quite unsafe, since it bypasses any other modifiers on the original attack.  I would have expected to see just ```damage = damage + iStolen;``` here.

kod/object/passive/itematt/weapatt/wavamper.kod
```diff
@@ -58,31 +57,37 @@ messages:
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
       
    ModifyDamage(damage = 0, target = $, wielder = $, lData= $)
-   "Wielder has a 1% chance of stealing life from target with any hit."
+   "Wielder has a 10% chance of stealing life from target with any hit."
```
Aesica, line 60: If somebody else changes piEffect_percent from 10 to something else sometime in the future, this comment becomes incorrect again, just like it was by stating a value of 1%.  It might be better to state that the chance is based on piEffect_percent.

kod/object/passive/itematt/weapatt/wavamper.kod
SioraSidhe: @akirmse Are you okay with the damage not matching the life that is taken? I made this change on my test, armor and resist can lower how much damage the target takes, but I cannot tell later how much came from the addition so the wielder will always heal 8

davidklaw: Interesting. Perhaps there should be a smaller life steal more frequently, if we can't take armor and resistances into account.

kod/object/passive/itematt/weapatt/wavamper.kod
SioraSidhe: Good thinking ok I made the change

kod/object/passive/itematt/weapatt/wavamper.kod
SioraSidhe: Ok I added the damage to the attack.

davidklaw:
@SioraSidhe I know that it's been a while... but would you mind updating the PR description?

Something to the effect of:

> This PR updates the vamper weapon attribute to drain and apply hp from the target to the attacker in addition to doing weapon damage. Currently, the attacked gains hp when the vamp effect procs, but an equal amount is not drained from the target.

Any tips you have for testing would be great, too. Through testing this myself I learned that you can hold a weapon in your hands and do `dm create itematt vamper` to make it a vamper weapon.

SioraSidhe:
ok I force pushed an update, get 2 characters, boost health and get skills, use a WA_Vamper short sword on one, when it triggers should see a damage spike. Need at least 8 normal dmg so it adds another 8 and see a 16 dmg hit when it triggers.

PR #775, opened 2024-06-22
Reinstate UserEntered npc triggering

sicoderealm:
Allow NPCs to be informed of new users entering their zones to trigger things such as welcome messages. 

Moods are adjusted and dampened via the hourly `RecalcLightAndWeather` which in turn adjusts `AmbientLightChanged` thus `SetMood` (mood dampening)!

This will be a phase one that enables these welcome messages - after that we can add new messages and logic.

image

image

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -3008,9 +3008,9 @@ messages:
    "A User just entered the room."
    {
 
-      if GetTime() > (lastWelcomeTime + 30)
+      if GetTime() > (piLastWelcomeTime + 30)
       {
-         lastWelcomeTime = GetTime();
+         piLastWelcomeTime = GetTime();
          propagate;
       }
  
@@ -5838,12 +5838,6 @@ messages:
    {
       return;
    }
-
-   GetLastWelcomeTime()
-   {
-      return piLastWelcomeTime;
-   }
-
 
 end
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -2660,7 +2660,6 @@ messages:
          }
       }
 
-      % here!
       % Maybe it was just a simple speech trigger. So, we check and parse.
       lActionList = $;
       lActionList = Send(oLibrary,@SearchSpeechLib,#mob=self,#string=string);
@@ -2976,9 +2975,6 @@ messages:
    {
       local iHour, iMoodMod;
       iHour = Send(SYS,@GetHour);
-
-
-      Debug("Damping mood via AmbientLightChanged() for ",Send(self,@GetTrueName),"");
 
       if iHour = 6
       {
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -3007,8 +3007,14 @@ messages:
    UserEntered(who = $)
    "A User just entered the room."
    {
-      piLastWelcomeTime = GetTime();
-      propagate;
+
+      if GetTime() > (lastWelcomeTime + 30)
+      {
+         lastWelcomeTime = GetTime();
+         propagate;
+      }
+ 
+      return;
    }
 
    MonsterAttack(what = $)
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -5619,6 +5619,7 @@ messages:
    SetMood(new_mood = 0)
    {
       piMood = Bound(new_mood,-100,100);
+
       return;
    }
 
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -3009,7 +3009,6 @@ messages:
    UserEntered(who = $)
    "A User just entered the room."
    {
-      piLastWelcomeTime = GetTime();
       return;
    }
 
@@ -5837,8 +5836,13 @@ messages:
 
    ReadyForWelcome()
    {
+      if GetTime() > (piLastWelcomeTime + WELCOME_MESSAGE_INTERVAL)
+      {
+         piLastWelcomeTime = GetTime();
+         return TRUE;
+      }
       
-      return GetTime() > (piLastWelcomeTime + WELCOME_MESSAGE_INTERVAL);
+      return FALSE;
    }
 
 end
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -2586,7 +2586,6 @@ messages:
    {
       local i,lActionList, oLibrary,iQuestID,iQuestType, oQuestNode;
 
-      debug("Someone said Monster!!\n");
       % Are we supposed to hear things?
       if (NOT (viAttributes & MOB_LISTEN))
          OR string = $
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -22,6 +22,8 @@ constants:
    %  learn to qualify for a new spell.
    MONSTER_LEARN_GRADIENT = 15
 
+   % Default interval in seconds between welcome messages from npcs when a 
+   %  player enters their room.
    WELCOME_MESSAGE_INTERVAL = 30
 
 resources:
@@ -5834,6 +5836,8 @@ messages:
       return;
    }
 
+   % True if the monster is ready to say a welcoming message to players
+   %  who have entered their room.
    ReadyForWelcome()
    {
       if GetTime() > (piLastWelcomeTime + WELCOME_MESSAGE_INTERVAL)
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -22,6 +22,8 @@ constants:
    %  learn to qualify for a new spell.
    MONSTER_LEARN_GRADIENT = 15
 
+   WELCOME_MESSAGE_INTERVAL = 30
+
 resources:
 
    monster_percent_q = "%q"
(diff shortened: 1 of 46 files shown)


Aesica:
Just my thoughts on the bullet points:

- Triggering:  The ideal *implementation* IMO would has some sort of short timer to avoid spamming.  You'd enter Familiars and Paddock would greet you, but if you left and came back in a few minutes later, you'd get nothing.  I'm not sure how ideal this would be in terms game data and code structure, though.  Each NPC would need a list of user/timer pairs which would be...unpleasant to work with.  Maybe best to just go with random *and* mood-based.  More likely to greet if the npc is in a good mood, for example.
- Mood adjustments:  As mentioned above, it could determine the rate of greetings.  It could also determine *how* the npc responds.  *Jonas greets you* vs *Jonas scowls at you.*  The level of detail could be per-npc, as not every npc needs to have detailed responses based on mood, and some don't care about outlaws/murderer status, factions, etc. 
- We'd definitely need to, especially if we want things like faction, outlaw, mood, etc to change responses.

Another thing to make note of is that some npc speech is in black text while other npc speech is in white text.  We should probably establish and adhere to a standard using just one or the other.

kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqSmith.kod
```diff
@@ -28,7 +28,7 @@ resources:
       "get the impression that he is uncomfortable with strangers, but his great size and clear "
       "devotion to Kraanan makes him afraid of no one."
 
-   BQsmith_entry_welcome = "Hello %s%s, and welcome."
+   BQsmith_entry_welcome = "~kFehr'loi Qan says, \"Hello %s%s, and welcome.\""
```
akirmse, line 31: I don't understand why in general these were all changed to include the name of the NPC here, duplicating it from elsewhere.  This makes the string more complicated and would also make translation more difficult in the future.  The previous way seems a lot better.

kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzinnk.kod
```diff
@@ -46,5 +44,11 @@ properties:
 messages:
 
 
+   UserEntered(who = $)
+   {
+      % Marcus says hello as part of the initial welcoming text: guest1.kod
```
akirmse, line 49: If someone changes guest1.kod, how will they know to update this file?

kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/MrElder.kod
```diff
@@ -87,9 +87,9 @@ messages:
       return;
    }
    
-   UserEntered(who=$)
+   UserEntered(who=$, what = $)
    {
-      Post(poOwner,@someoneSaid,#what=self,#string=MrEl_entry_welcome,
+      Post(who,@someoneSaid,#what=self,#string=MrEl_entry_welcome,
```
akirmse, line 92: Why is this one posted to the user instead of the owner, unlike the others?

kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/mrbart.kod
```diff
@@ -31,7 +31,7 @@ resources:
 
    tova_unwanted_keep = "~k%s%s tells you, \"Well, how nice of you.  A gift.~n~k\""
    tova_unwanted_pay = "~k%s%s tells you, \"What an interesting item!~n~k\""
-   tova_unwanted_give = "~k%s%s tells you, \"Ha, ha, you've got to be kidding me, stranger! This is a tavern, not a gypsy�s tent!~n~k\""
+   tova_unwanted_give = "~k%s%s tells you, \"Ha, ha, you've got to be kidding me, stranger! This is a tavern, not a gypsy�s tent!~n~k\""
```
akirmse, line 34: Not sure what the change is here.  Revert this file?

kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/mrhealer.kod
```diff
@@ -53,9 +53,8 @@ messages:
 
    UserEntered(who = $)
    {
-      Post(poOwner,@SomeoneSaid,#what=self,#string=MrHealer_entry_welcome,
-          #type = SAY_RESOURCE,
-          #parm1=Send(who,@GetDef),#parm2=Send(who,@GetName));
+
+      Post(poOwner,@SomeoneSaid,#what=self,#type=SAY_MESSAGE,#string=Mrhealer_entry_welcome);
```
akirmse, line 57: Does this work?  It looks like the substitution parameters have been removed.

kod/object/active/holder/nomoveon/battler/monster/towns/tostwn/TsSmith.kod
```diff
@@ -93,9 +93,9 @@ messages:
 
    UserEntered(who = $)
    {
-      Post(poOwner,@SomeoneSaid,#what = self, #string=entry_welcome,
-         #type = SAY_RESOURCE, 
-         #parm1=Send(who,@GetDef),#parm2=Send(who,@GetName));
+
+      Post(poOwner,@SomeoneSaid,#what=self,#type=SAY_MESSAGE,#string=entry_welcome);
```
akirmse, line 97: Does this work?  It looks like the substitution parameters have been removed.

kod/object/active/holder/room.kod
```diff
@@ -2419,6 +2421,35 @@ messages:
       }
 
       propagate;
+   }  
+   
+   UserEntered(who = $)
+   {
+      local i,each_obj,lastWelcomeTime;
+
+      if isClass(who,&Player)
```
akirmse, line 2430: Is this IsClass check necessary?  It's called "UserEntered"

kod/object/active/holder/room.kod
```diff
@@ -2419,6 +2421,35 @@ messages:
...
+      }
+
+      for i in plActive
+      {
+         each_obj = First(i);
+         % q: is there a better class than %Monster for all NPCs
+         if IsClass(each_obj, &Monster)
+         {
+            lastWelcomeTime = Send(each_obj, @GetLastWelcomeTime);
+            if GetTime() > (lastWelcomeTime + 30)
```
akirmse, line 2446: I think this check belongs in Monster, not Room, since it's a behavior of the monster.

akirmse (commented):
In general I think trying to restore this chatter is OK, but I don't like how the strings have been changed as a result.  Why was that necessary at all?

akirmse:
Please revert all files where there's been no change, which I think is most of them.

Sometimes you've changed Send to Post, not sure why.

kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcvaultm.kod
```diff
@@ -70,11 +70,11 @@ messages:
    UserEntered(who = $)
    {
       local iFlags;
-      if random(1,10)
```
akirmse, line 73: Why was this removed?

kod/object/active/holder/nomoveon/battler/monster.kod
```diff
@@ -22,6 +22,8 @@ constants:
    %  learn to qualify for a new spell.
    MONSTER_LEARN_GRADIENT = 15
 
+   WELCOME_MESSAGE_INTERVAL = 30
```
akirmse, line 25: Add a comment including purpose and units.

kod/object/active/holder/nomoveon/battler/monster.kod
```diff
@@ -5828,6 +5834,16 @@ messages:
       return;
    }
 
+   ReadyForWelcome()
```
akirmse, line 5837: Add a comment

kod/object/active/holder/room.kod
```diff
@@ -1773,6 +1773,22 @@ messages:
...
 
+         for i in plActive
+         {
+            each_obj = First(i);
+            if IsClass(each_obj,&Monster)
+            {
+               if NOT Send(what,@CheckPlayerFlag,#flag=PFLAG_ANONYMOUS)
+                     AND NOT Send(what,@CheckPlayerFlag,#flag=PFLAG_INVISIBLE)
+               {
+                  if Send(each_obj, @ReadyForWelcome)
```
akirmse, line 1784: How often does this result in multiple welcome messages upon one user entering the room?

kod/object/active/holder/room.kod
```diff
@@ -1773,6 +1773,22 @@ messages:
                  #new_col=Nth(new_pos,2));
          }
 
+         for i in plActive
+         {
+            each_obj = First(i);
+            if IsClass(each_obj,&Monster)
+            {
+               if NOT Send(what,@CheckPlayerFlag,#flag=PFLAG_ANONYMOUS)
```
akirmse, line 1781: Seems like these checks should be done outside the outer loop, not on each iteration.

kod/object/active/holder/nomoveon/battler/monster/towns/marntwn/MrSmith.kod
```diff
@@ -92,7 +92,8 @@ messages:
 
    UserEntered(who = $)
    {
-      Post(poOwner,@SomeoneSaid,#what=self,#string=MrBs_entry_welcome,
+
+      Send(poOwner,@SomeoneSaid,#what=self,#string=MrBs_entry_welcome,
```
akirmse, line 96: Just verifying that you have tested every one of these instances where you changed Post to Send.  It's possible some were added intentionally to get the order of messages correct upon entering a room.  Please revert any you haven't tested.

kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzinnk.kod
```diff
@@ -46,5 +44,12 @@ properties:
 messages:
 
 
+   UserEntered(who = $)
```
akirmse, line 47: This function has no effect.  Remove?

kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcsmith.kod
```diff
@@ -107,7 +107,7 @@ messages:
 
    UserEntered(who = $)
    {
-      Post(poOwner,@SomeoneSaid,#what = self, #string=kcSmith_entry_welcome,
+      Send(poOwner,@SomeoneSaid,#what = self, #string=kcSmith_entry_welcome,
```
akirmse, line 110: You still need to revert all of these changed files that are not related to this PR.

PR #788, opened 2024-07-07
fix: make colored listbox text more robust in client

adrienlaws:
# What

- this change attempts to make wholist handling more robust to reduce client crashes related to `DrawOwnerListItem`
 
| function | line | filename|
| - | - | - |
`meridian!DrawOwnerListItem` | 377 | ownerdrw.c
`meridian!OwnerListDrawItem`  | 288 | ownerdrw.c
`meridian!WhoDialogProc` | 153 | msgfiltr.c

# Why

- there are several bug reports on different dates related to a crash in `DrawOwnerListItem` on this line
```
if (style & (OD_DRAWOBJ | OD_DRAWICON) && obj != NULL)
   item_rarity_value = (item_rarity_grade)obj->rarity;
```
- It is suspected that the compiler might be reordering things within the `DrawOwnerListItem` function
  - if so, this line could be the culprit (as pointed out to me)
  - [Meridian59/clientd3d/ownerdrw.c](https://github.com/Meridian59/Meridian59/blob/bf65a1a5677e277379662456ca42f96106045b5b/clientd3d/ownerdrw.c#L461)
    - `crColorText = GetPlayerNameColor(obj->flags,NULL);` in this line if the obj is null then there could be a crash

# Changes
- added a check in `clientd3d/ownerdrw.c` `DrawOwnerListItem`
  - check for a null pointer and if it is, then use default black text for the obj
- added a check in `clientd3d/msgfiltr.c` within the `WHOM_UPDATE` case
  - validate the `i = ItemListAddItem(hList, pUser, -1, False);` call so that i does not return an error code
    - added check that adding the list item does not return `LB_ERR` which is the `WinUser.h` listbox error value return (`-1`)

### DIFF
--- a/clientd3d/msgfiltr.c
+++ b/clientd3d/msgfiltr.c
@@ -127,11 +127,16 @@ INT_PTR CALLBACK WhoDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lP
        debug(("WHOM_UPDATE NULL user\n"));
        return FALSE;
      }
+     if (!hList)
+     { 
+       debug(("Failed to get handle for IDC_USERLIST\n"));
+       return FALSE;
+     }
      debug(("WHOM_UPDATE, %s user %i\n", (LPCTSTR)(bAdded?"adding":"removing"), pUser->id));
      if (bAdded)
      {
        i = ItemListAddItem(hList, pUser, -1, False);
-       if (!IsUserInIgnoreList(pUser->name_res))
+       if (i != LB_ERR && !IsUserInIgnoreList(pUser->name_res))
 	       ListBox_SetSel(hList, TRUE, i);
      }
      else
--- a/clientd3d/msgfiltr.c
+++ b/clientd3d/msgfiltr.c
@@ -127,11 +127,6 @@ INT_PTR CALLBACK WhoDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lP
        debug(("WHOM_UPDATE NULL user\n"));
        return FALSE;
      }
-     if (!hList)
-     { 
-       debug(("Failed to get handle for IDC_USERLIST\n"));
-       return FALSE;
-     }
      debug(("WHOM_UPDATE, %s user %i\n", (LPCTSTR)(bAdded?"adding":"removing"), pUser->id));
      if (bAdded)
      {
--- a/clientd3d/ownerdrw.c
+++ b/clientd3d/ownerdrw.c
@@ -374,8 +374,7 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    SetBkMode(lpdis->hDC, TRANSPARENT);
 
    if (style & (OD_DRAWOBJ | OD_DRAWICON) && obj != NULL)
-      if (IsValidRarity(obj->rarity)) 
-         item_rarity_value = (item_rarity_grade)obj->rarity;
+      item_rarity_value = (item_rarity_grade)obj->rarity;
 
       crColorText = GetColor(
                         GetItemListColor(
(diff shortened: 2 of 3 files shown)


clientd3d/ownerdrw.c
```diff
@@ -485,3 +486,21 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
...
+ * IsValidRarity:  checks whether a given rarity value is valid
+ */
+int IsValidRarity(int rarity) {
+   switch (rarity) {
+      case ITEM_RARITY_GRADE_NORMAL:
+      case ITEM_RARITY_GRADE_UNCOMMON:
+      case ITEM_RARITY_GRADE_RARE:
+      case ITEM_RARITY_GRADE_LEGENDARY:
+      case ITEM_RARITY_GRADE_UNIDENTIFIED:
+      case ITEM_RARITY_GRADE_CURSED:
```
adrienlaws, line 500: I don't know if this is bad, because we will need to update rarity grades in both proto.h and in this function which feels bad.  But I'm not sure how to validate the input more efficiently.

clientd3d/ownerdrw.c
```diff
@@ -374,6 +374,7 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    SetBkMode(lpdis->hDC, TRANSPARENT);
 
    if (style & (OD_DRAWOBJ | OD_DRAWICON) && obj != NULL)
+      if (IsValidRarity(obj->rarity))
```
akirmse, line 377: I don't think this is necessary.  In C, a cast can't crash.

An access violation means referencing bad memory, which strongly suggests that obj is NULL.  Of course, there's a check immediately preceding this line, so that can't be the problem.  But it could be a problem here:

https://github.com/Meridian59/Meridian59/blob/bf65a1a5677e277379662456ca42f96106045b5b/clientd3d/ownerdrw.c#L461

I wonder if the compiler has drastically moved things around and done the dereferencing here in the compiled code.  It would be worth checking for nullptr before that line,

clientd3d/msgfiltr.c
```diff
@@ -127,11 +127,16 @@ INT_PTR CALLBACK WhoDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lP
...
+     { 
+       debug(("Failed to get handle for IDC_USERLIST\n"));
+       return FALSE;
+     }
      debug(("WHOM_UPDATE, %s user %i\n", (LPCTSTR)(bAdded?"adding":"removing"), pUser->id));
      if (bAdded)
      {
        i = ItemListAddItem(hList, pUser, -1, False);
-       if (!IsUserInIgnoreList(pUser->name_res))
+       if (i != LB_ERR && !IsUserInIgnoreList(pUser->name_res))
```
akirmse, line 139: Maybe worth it?  I don't think this can crash but it might not hurt.

clientd3d/ownerdrw.c
adrienlaws: Removed

clientd3d/msgfiltr.c
```diff
@@ -127,11 +127,16 @@ INT_PTR CALLBACK WhoDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lP
        debug(("WHOM_UPDATE NULL user\n"));
        return FALSE;
      }
+     if (!hList)
```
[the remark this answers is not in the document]
adrienlaws: Removed.

clientd3d/ownerdrw.c
adrienlaws: Added a null pointer check right before the call to `GetPlayerNameColor`.  If the obj is null then we use default black text.

clientd3d/msgfiltr.c
adrienlaws: Kept in unless you feel I should remove this.

clientd3d/ownerdrw.c
```diff
@@ -458,7 +458,15 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    if (style & OD_COLORTEXT)
    {
      // get the color we'd prefer for this particular obj
-     crColorText = GetPlayerNameColor(obj->flags,NULL);
+     if (obj != NULL)
```
akirmse, line 461: use == and flip the then and else clauses.  One less "not" to understand.

clientd3d/ownerdrw.c
```diff
@@ -458,7 +458,15 @@ void DrawOwnerListItem(const DRAWITEMSTRUCT *lpdis, Bool selected, Bool combo)
    if (style & OD_COLORTEXT)
    {
      // get the color we'd prefer for this particular obj
-     crColorText = GetPlayerNameColor(obj->flags,NULL);
+     if (obj == NULL) 
+     {
+        crColorText = PALETTERGB(255, 255, 255);
```
akirmse, line 463: Just double-checking that this is what you want.  In the PR commentary, you talk about using default black text, but this is white.

adrienlaws: Yes I think it should be white because in this case we are talking about player colored text or a catchall vs the rarity colors.  I got confused, I think it makes sense to stick with white because a few lines below it talks about having a black halo around the text to make it more visible, so black on black wouldn't work.

PR #778, opened 2024-07-01
Dedicated passes for transparent world objects with render state rework

sicoderealm:
Updates to the 3d hardware graphics to improve a few long standing issues, including: borders around Barloque archway doors, sign posts, and the cobwebs in cv basement. This also prevents small lines appearing around some of the smaller moon phases.

We add separate sweeps for translucent and opaque world objects and rework the rendering states throughout.

**These changes should to be fully beta tested before being released to ensure its working across a range of setups and configurations by a number of different users.** 

image
image
Please note north bq room requires an [update to fix another issue](https://github.com/Meridian59/Meridian59/pull/789) to see it like this:
image
image
also here (north bq room update required)
image
image
image
image
image

PR #789, opened 2024-07-07
Fix North Barloque gate to vaults reverse walls

sicoderealm:
As part of the [graphics enhancements ](https://github.com/Meridian59/Meridian59/pull/778) an issue with North Barloque gate was identified.

With the current live hardware renderer, you can see the thin lines of these bad sections of the map above the gate.
image
image

In the software renderer, these sections of the map are not rendered at all -- they are simply not supported:
image
image
Acting as if they are not actually part of the map.

When we improve/fix the renderer these can be fully identified (not just the thing lines above the gate house) as full textures.
image
image

As these are wrong, removing this texture fixes this issue and makes it as the original software renderer.


davidklaw:
Huh, interesting. I wonder how many of these errant sidedef textures exist. I'm sweating just thinking about it.

PR #790, opened 2024-07-07
Sort characters names alphabetically with new users at the end

sicoderealm:
Recently found instances where new characters are not listed at the end of the character selection screen:
image
This change uses std::sort and the outcome here being:
image

### DIFF
--- a/module/char/charpick.c
+++ b/module/char/charpick.c
@@ -110,6 +110,21 @@ void ChooseCharacter(Character *characters, WORD num_characters, char *motd,
    }
 }
 /********************************************************************/
+/*
+ * compareCharacters:  Sorting alphabetically with new character slots always at the end.
+ */
+bool compareCharacters(const Character& a, const Character& b) {
+    // Place characters with flags != 1 (new users) at the end
+    if (a.flags != 1 && b.flags == 1) {
+        return true;
+    }
+    if (a.flags == 1 && b.flags != 1) {
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


module/char/charpick.c
```diff
@@ -110,6 +110,21 @@ void ChooseCharacter(Character *characters, WORD num_characters, char *motd,
...
+ */
+bool compareCharacters(const Character& a, const Character& b) {
+    // Place characters with flags != 1 (new users) at the end
+    if (a.flags != 1 && b.flags == 1) {
+        return true;
+    }
+    if (a.flags == 1 && b.flags != 1) {
+        return false;
+    }
+    // If flags are the same and both have flags == 1, compare names case-insensitively
```
akirmse, line 124: I think this comment isn't right.  We reach here if the flags are the same, not if they're 1.

module/char/charpick.c
```diff
@@ -110,6 +110,21 @@ void ChooseCharacter(Character *characters, WORD num_characters, char *motd,
...
+ */
+bool compareCharacters(const Character& a, const Character& b) {
+    // Place characters with flags != 1 (new users) at the end
+    if (a.flags != 1 && b.flags == 1) {
+        return true;
+    }
+    if (a.flags == 1 && b.flags != 1) {
+        return false;
+    }
+    // Neither character is new so compare names case-insensitively
```
akirmse, line 124: I don't think this is right.  We get here if both flags are 0, or both flags are 1.  Both characters could be new.

PR #794, opened 2024-07-07
Adjust the viewing height of ViewpointGlobe's so as not to be affected by ceilings

davidklaw:
This PR seeks to resolve an issue seen in the Lich maze (`lichmaze`) when using the globes/cameras. When a player uses a globe their viewpoint becomes the top of the globe and is above the height of the ceiling, causing a visual bug. This PR uses `REMOTE_VIEW_BOTTOM` instead of `REMOTE_VIEW_BOTTOM` to set their viewpoint to the bottom of the globe, rather than the top.

### Before
image

### After
image

### Before
https://github.com/Meridian59/Meridian59/assets/467443/5266cb8c-55e2-4f49-9ed1-e8b044b67a03

### After
https://github.com/Meridian59/Meridian59/assets/467443/e6db9d5e-0510-4c0c-a728-5c964b18f043

### DIFF
--- a/clientd3d/game.c
+++ b/clientd3d/game.c
@@ -230,7 +230,7 @@ void SetPlayerRemoteView(ID objID, int flags, int height, BYTE light)
       if (flags & REMOTE_VIEW_VALID_HEIGHT)
 	 player.viewHeight = floor + height;
       else if (flags & REMOTE_VIEW_TOP)
-	 player.viewHeight = viewObject->motion.z - 10;
+	 player.viewHeight = viewObject->motion.z + objHeight;
       else if (flags & REMOTE_VIEW_BOTTOM)
 	 player.viewHeight = viewObject->motion.z;
       else if (flags & REMOTE_VIEW_MID)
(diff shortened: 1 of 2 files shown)


clientd3d/game.c
```diff
@@ -230,7 +230,7 @@ void SetPlayerRemoteView(ID objID, int flags, int height, BYTE light)
       if (flags & REMOTE_VIEW_VALID_HEIGHT)
 	 player.viewHeight = floor + height;
       else if (flags & REMOTE_VIEW_TOP)
-	 player.viewHeight = viewObject->motion.z + objHeight;
+	 player.viewHeight = viewObject->motion.z - 10;
```
akirmse, line 233: This can't be the right fix, as it puts REMOTE_VIEW_TOP below both REMOTE_VIEW_BOTTOM and REMOTE_VIEW_MID.  One alternative would be to go to the Blakod for the globe and have it send REMOTE_VIEW_BOTTOM or MID instead.  But in any case you'll need to check all the other globes in the game to make sure this doesn't break them.

akirmse: Or, maybe move the globes down a little bit.

davidklaw: My mistake. I think my brain was mush in this moment. My new approach is to use `REMOTE_VIEW_BOTTOM` rather than `REMOTE_VIEW_TOP`, setting the players viewpoint to be the bottom of the globe rather than the very top.

PR #796, opened 2024-07-07
Adds ability to sort newsglobes by column

davidklaw:
It's baaaack! This builds on the legacy of #398 to add sorting functionality to newsglobes. The previous PR suffered from a couple of problems and was reverted by #406. These were the problems noted at that time:

- Clicking on an article in the top half doesn't load the article content into the bottom half.
- Clicking on a header sometimes sorts that column, but sometimes it does nothing, and sometimes it appears to sort a different column.

Both of these should work as expected now. I've also added visual sort indicators.

This still needs to be tested, ideally with scale.

https://github.com/Meridian59/Meridian59/assets/467443/889ae4f1-948b-4f12-b343-770508d5c146


module/mailnews/newsread.c
```diff
@@ -394,3 +412,100 @@ Bool DateFromSeconds(long seconds, char *str)
...
+   case COL_TIME:
+      if (bSortAscending)
+         return article1->time - article2->time;
+      return article2->time - article1->time;
+   }
+
+   return 0;
+}
+
+void UpdateColumnHeaders(HWND hListView, int sortedColumn, BOOL sortAscending)
```
davidklaw, line 477: This isn't a strong suit of mine so I turned to AI to assist with this code. It seems a bit much and can perhaps be optimized.

module/mailnews/newsread.c
```diff
@@ -394,3 +412,100 @@ Bool DateFromSeconds(long seconds, char *str)
...
+{
+   BOOL bSortAscending = (lParamSort > 0);
+   int nColumn = abs(lParamSort) - 1;
+
+   NewsArticle *article1 = (NewsArticle *)lParam1;
+   NewsArticle *article2 = (NewsArticle *)lParam2;
+
+   // `ListView_SortItems` is unstable, so we append the article index for stability
+   char title1[MAX_SUBJECT + 10];
+   sprintf(title1, "%s %d", article1->title, article1->num);
```
akirmse, line 449: Use snprintf; sprintf is a known security risk.

module/mailnews/newsread.c
```diff
@@ -394,3 +412,100 @@ Bool DateFromSeconds(long seconds, char *str)
...
+
+   char poster1[MAXUSERNAME + 10];
+   sprintf(poster1, "%s %d", article1->poster, article1->num);
+   char poster2[MAXUSERNAME + 10];
+   sprintf(poster2, "%s %d", article2->poster, article2->num);
+
+   switch (nColumn)
+   {
+   case COL_TITLE:
+      if (bSortAscending)
```
akirmse, line 461: Might be a good place for the ternary operator: 
```
return bSortAscending ? stricmp(title1, title2) : stricmp(title2, title1);
```

module/mailnews/newsread.c
```diff
@@ -25,6 +26,15 @@ static list_type new_articles = NULL;   /* Build up new articles arriving from s
 
 static RECT  dlg_rect;        // Screen position of dialog
 
+// Constants for news dialog columns
+#define COL_TITLE 0x0000
```
akirmse, line 30: use static const int rather than macros for constants

module/mailnews/newsread.c
```diff
@@ -394,3 +412,100 @@ Bool DateFromSeconds(long seconds, char *str)
...
+}
+
+void UpdateColumnHeaders(HWND hListView, int sortedColumn, BOOL sortAscending)
+{
+    HWND hHeader = ListView_GetHeader(hListView);
+    int columnCount = Header_GetItemCount(hHeader);
+
+    for (int i = 0; i < columnCount; i++)
+    {
+        TCHAR text[256];
```
akirmse, line 484: TCHAR is a hideous Windows-ism.  Use char if the later code will still work with it.

Where did 256 come from?  Use a named constant.

module/mailnews/newsread.c
```diff
@@ -394,3 +412,100 @@ Bool DateFromSeconds(long seconds, char *str)
...
+        HDITEM hdi = {0};
+        hdi.mask = HDI_TEXT;
+        hdi.pszText = text;
+        hdi.cchTextMax = sizeof(text)/sizeof(text[0]);
+
+        Header_GetItem(hHeader, i, &hdi);
+
+        // remove any existing '+' or '-' from the text
+        TCHAR *plusPos = _tcschr(text, '+');
```
akirmse, line 493: If possible, use char instead of TCHAR

module/mailnews/newsread.c
```diff
@@ -394,3 +412,100 @@ Bool DateFromSeconds(long seconds, char *str)
...
+    {
+        TCHAR text[256];
+        HDITEM hdi = {0};
+        hdi.mask = HDI_TEXT;
+        hdi.pszText = text;
+        hdi.cchTextMax = sizeof(text)/sizeof(text[0]);
+
+        Header_GetItem(hHeader, i, &hdi);
+
+        // remove any existing '+' or '-' from the text
```
akirmse, line 492: I don't understand what this function does, and there's no comment.

module/mailnews/newsread.c
```diff
@@ -34,6 +44,10 @@ static ChildPlacement newsread_controls[] = {
 /* local function prototypes */
 static INT_PTR CALLBACK ReadNewsDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam);
 static void UserReplyNewsMail(NewsArticle *article);
+void OnColumnClick(LPNMLISTVIEW pLVInfo);
+int CALLBACK CompareListItems(LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort);
+void UpdateColumnHeaders(HWND hListView, int sortedColumn, BOOL sortAscending);
```
akirmse, line 49: Make these static if they're not called from outside this function.

module/mailnews/newsread.c
davidklaw: This compiles fine, but my editor is telling me this about the case value:
> expression must have a constant valueC/C++(28)
Using an enum solves this and compiles. Do you have a preference to keep these `static const int` or would an enum work?

davidklaw:
Revisions based on feedback are in.

module/mailnews/newsread.c
```diff
@@ -11,6 +11,9 @@
 
 #include "client.h"
 #include "mailnews.h"
+#include "string.h"
+
+#define MAX_HEADER_TEXT 256 // Named constant for buffer size
```
akirmse, line 16: This should be a static const int.

You say your editor is complaining?  It's possible it's treating this as a C language file (given its extension), while this is actually C++ code.  See if there's a setting for this.  Alternately we could rename every source file in the repo to .cpp, which probably will have to happen one day anyway.

module/mailnews/newsread.c
```diff
@@ -389,8 +411,137 @@ Bool DateFromSeconds(long seconds, char *str)
    if (t == NULL)
       return False;
 
-   sprintf(str, "%s %s %.2ld, %.4ld %.2ld:%.2ld", 
+   snprintf(str, 50, "%s %s %.2ld, %.4ld %.2ld:%.2ld",
```
akirmse, line 414: "str must have length at least MAXDATE", so use MAXDATE rather than a magic number.

module/mailnews/newsread.c
```diff
@@ -389,8 +411,137 @@ Bool DateFromSeconds(long seconds, char *str)
...
+   snprintf(str, 50, "%s %s %.2ld, %.4ld %.2ld:%.2ld", 
 	   GetString(hInst, weekdays[t->tm_wday]), GetString(hInst, months[t->tm_mon]), 
 	   t->tm_mday, t->tm_year + 1900, t->tm_hour, t->tm_min);
    return True;
 }
+
+void OnColumnClick(LPNMLISTVIEW pLVInfo)
+{
+    int nSortColumn = pLVInfo->iSubItem;
+    BOOL bSortAscending = TRUE;
```
akirmse, line 423: Use bool (C++ built-in type) instead of BOOL (horrible Windows-ism).

module/mailnews/newsread.c
```diff
@@ -389,8 +411,137 @@ Bool DateFromSeconds(long seconds, char *str)
...
+ *   lParam2 - Pointer to the second item (cast to `LPARAM`).
+ *   lParamSort - Sort parameter which indicates the column to sort by and 
+ *     the sort direction. A positive value indicates ascending order, and a 
+ *     negative value indicates descending order. The absolute value of 
+ *     `lParamSort` indicates the column index (1-based).
+ *
+ */
+int CALLBACK CompareListItems(LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort)
+{
+   BOOL bSortAscending = (lParamSort > 0);
```
akirmse, line 470: bool

module/mailnews/newsread.c
```diff
@@ -389,8 +411,137 @@ Bool DateFromSeconds(long seconds, char *str)
...
+   HWND hHeader = ListView_GetHeader(hListView);
+   // Get the number of columns in the header
+   int columnCount = Header_GetItemCount(hHeader);
+
+   for (int i = 0; i < columnCount; i++)
+   {
+      char text[MAX_HEADER_TEXT]; // Buffer to hold the header text
+      HDITEM hdi = {0}; // Initialize HDITEM structure
+      hdi.mask = HDI_TEXT; // Indicate interest in the text of the header item
+      hdi.pszText = text; // Set the buffer to receive the text
```
akirmse, line 525: Remove these AI-generated comments.

module/mailnews/newsread.c
```diff
@@ -389,8 +411,137 @@ Bool DateFromSeconds(long seconds, char *str)
...
+ * CompareListItems:  Comparison function for sorting items in a list view
+ *   control. This function is used by `ListView_SortItems` to compare two 
+ *   items and determine their sort order based on the specified column and 
+ *   sort direction. It:
+ * 
+ *   - Determines the sort direction and column index from `lParamSort`.
+ *   - Casts `lParam1` and `lParam2` to `NewsArticle` pointers.
+ *   - Constructs stable comparison keys by appending the article index to 
+ *     the title and poster strings.
+ *   - Compares the items based on the specified column and sort direction:
```
akirmse, line 454: Function comments should describe what the function does, not how it works.  As it is this doubles the amount of work for anyone in the future, since any internal change they make also needs to change the comment.  I bet you can guess what the likely outcome of that would be.

module/mailnews/newsread.c
davidklaw: Aha, okay. I'm able to select the language mode for .c files. If we ever wanted to settle on Visual Studio Code as the IDE of choice, we could set up some project settings to default this.

module/mailnews/newsread.c
davidklaw: _All_ instances of `BOOL` or `Bool` to `bool`? There were a few not highlighted in your comments.

akirmse: Use BOOL where you have to, such as parameters to Windows functions, bool everywhere else.  Bool and BOOL are from a time before C++ existed.

davidklaw:
Rethought my approach a bit and found this article that helped me implement bitmap sort indicators, so `UpdateColumnHeaders` is all new. This along with revisions based on feedback.

https://github.com/Meridian59/Meridian59/assets/467443/1af2a1d0-0a09-473f-b37a-456a7905f816

module/mailnews/newsread.c
```diff
@@ -389,8 +412,132 @@ Bool DateFromSeconds(long seconds, char *str)
...
+      if (i == sortedColumn)
+      {
+         UINT bitmapID = sortAscending ? IDB_UPARROW : IDB_DOWNARROW;
+
+         hdi.fmt |= HDF_BITMAP | HDF_BITMAP_ON_RIGHT;
+
+         HBITMAP hBitmap = (HBITMAP)LoadImage(GetModuleHandle(NULL), MAKEINTRESOURCE(bitmapID), IMAGE_BITMAP, 0, 0, LR_LOADTRANSPARENT | LR_LOADMAP3DCOLORS);
+         if (hBitmap == NULL)
+         {
+            MessageBox(NULL, "Failed to load bitmap!", "Error", MB_OK | MB_ICONERROR);
```
akirmse, line 523: This error message is kind of a "PC LOAD LETTER".  What is the user supposed to do?

Just silently fail and don't draw a bitmap in this case.

module/mailnews/newsread.c
```diff
@@ -389,8 +412,132 @@ Bool DateFromSeconds(long seconds, char *str)
...
+   {
+   case COL_TITLE:
+      return bSortAscending ? stricmp(title1, title2) : stricmp(title2, title1);
+   case COL_POSTER:
+      return bSortAscending ? stricmp(poster1, poster2) : stricmp(poster2, poster1);
+   case COL_TIME:
+      return bSortAscending ? article1->time - article2->time : article2->time - article1->time;
+   }
+
+   return 0;
```
akirmse, line 481: This would be a good place for an assertion or error print, as the switch should always handle all cases, so this code is unreachable.

davidklaw: If it is reached, and it's an assertion or error print, the callback continues repeatedly down the list and produces either many assertion dialogs or redundant error messages for each comparison that takes place. From what I understand, we can't break from the comparison prematurely, so maybe this is what it is? This is what I have currently:

```
switch (nColumn)
{
   case COL_TITLE:
      return bSortAscending ? stricmp(title1, title2) : stricmp(title2, title1);
   case COL_POSTER:
      return bSortAscending ? stricmp(poster1, poster2) : stricmp(poster2, poster1);
   case COL_TIME:
      return bSortAscending ? article1->time - article2->time : article2->time - article1->time;
   default:
      debug(("Unhandled column (column #%d) in CompareListItems\n", nColumn));
      return 0;
}
```

akirmse: That's fine.  Hitting that print should really "never happen", as it would mean a coding error, not a runtime error.

module/mailnews/newsread.c
```diff
@@ -389,8 +421,107 @@ Bool DateFromSeconds(long seconds, char *str)
...
+}
+
+void ListView_SetHeaderSortImage(HWND hListView, int sortedColumn, BOOL sortAscending)
+{
+   // Get the handle to the header control associated with the list view
+   HWND hHeader = ListView_GetHeader(hListView);
+   if (!hHeader)
+       return;  // Ensure we have a valid header handle
```
akirmse, line 500: Useless comment

module/mailnews/newsread.c
```diff
@@ -389,8 +421,107 @@ Bool DateFromSeconds(long seconds, char *str)
...
+
+   // Get the number of columns in the header
```
akirmse, line 502: also useless

module/mailnews/newsread.c
```diff
@@ -11,9 +11,17 @@
...
 #include "mailnews.h"
+#include "string.h"
+
+// Named constant for buffer size
+static const int MAX_HEADER_TEXT = 256;
 
 HWND hReadNewsDialog = NULL;
 
+// Declare global HBITMAP objects
+HBITMAP hbmUpArrow = NULL;
```
akirmse, line 22: Make static; these shouldn't be visible outside this file.

module/mailnews/newsread.c
```diff
@@ -118,7 +141,12 @@ INT_PTR CALLBACK ReadNewsDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPAR
    NM_LISTVIEW *nm;
    
    char date[MAXDATE], title[MAX_SUBJECT + MAXDATE + MAXNAME + 10];
-   
+
+    // Load sort arrow bitmaps
+   hbmUpArrow = (HBITMAP)LoadImage(GetModuleHandle(NULL), MAKEINTRESOURCE(IDB_UPARROW), IMAGE_BITMAP, 0, 0, LR_LOADTRANSPARENT | LR_LOADMAP3DCOLORS);
+
+   hbmDownArrow = (HBITMAP)LoadImage(GetModuleHandle(NULL), MAKEINTRESOURCE(IDB_DOWNARROW), IMAGE_BITMAP, 0, 0, LR_LOADTRANSPARENT | LR_LOADMAP3DCOLORS);
```
akirmse, line 148: lines too long

davidklaw: What's you preference here? How about this?

```
// Load sort arrow bitmaps
hbmUpArrow = (HBITMAP)LoadImage(GetModuleHandle(NULL), 
    MAKEINTRESOURCE(IDB_UPARROW), IMAGE_BITMAP, 0, 0, 
    LR_LOADTRANSPARENT | LR_LOADMAP3DCOLORS);

hbmDownArrow = (HBITMAP)LoadImage(GetModuleHandle(NULL), 
    MAKEINTRESOURCE(IDB_DOWNARROW), IMAGE_BITMAP, 0, 0, 
    LR_LOADTRANSPARENT | LR_LOADMAP3DCOLORS);
```

akirmse: Sure, that's fine.  Long lines are hard to read when they wrap in github.

davidklaw:
Moving out of draft. Thanks for your patience with this one!

akirmse:
I tried this locally, and while the columns sort, I don't see the bitmaps.  The problem might be that the bitmaps are in the main client.rc, where's I'd expect them to be in mailnews.rc.  The bitmap files themselves should also be moved into the mailnews directory.

davidklaw:
Huh... I'm been banging my head on this for a while now and get `ERROR_RESOURCE_NAME_NOT_FOUND` when trying to load them via `mailnews.rc` with definitions in `mailnwrc.h`, which all makes sense to me as I track through the includes. However, the resources only load on my end when I handle it the way it's written currently 🤔

module/mailnews/newsread.c
```diff
@@ -142,10 +142,14 @@ INT_PTR CALLBACK ReadNewsDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPAR
    
    char date[MAXDATE], title[MAX_SUBJECT + MAXDATE + MAXNAME + 10];
 
-    // Load sort arrow bitmaps
-   hbmUpArrow = (HBITMAP)LoadImage(GetModuleHandle(NULL), MAKEINTRESOURCE(IDB_UPARROW), IMAGE_BITMAP, 0, 0, LR_LOADTRANSPARENT | LR_LOADMAP3DCOLORS);
+   // Load sort arrow bitmaps
```
davidklaw, line 145: This was a big mistake I'm correcting. These bitmaps should not be loaded in this callback and instead will be moved to `WM_INITDIALOG`.

PR #797, opened 2024-07-09
Follow-up fixes for newsglobe sorting

davidklaw:
This PR is a follow-up to #796 that addresses issues with the placement and loading of the sorting bitmaps.


akirmse:
You should also remove the bmp files from the client directory so that we don't have two copies.

Something looks weird with your branches; github shows a ton of extraneous changes in this PR.  It's a good idea to make a new branch for each new PR, after first syncing master.

PR #798, opened 2024-07-09
Fixes for newsglobe sort bitmap placement and loading

davidklaw:
This PR is a follow-up to #796 that addresses issues with the placement and loading of the sorting bitmaps.


akirmse:
Works great here!  

It might be a good idea to show a down arrow on the Date column when the dialog loads, to indicate the default sort order.

Also, if you initially click on the Date column, nothing happens,  You have to click twice.

davidklaw:
Excellent. I'll take a look at those couple of things and start thinking about mail sorting.

PR #801, opened 2024-07-12
Add sorting to mail dialog

### LINKED ISSUE #26: Mail and news columns should be sortable

davidklaw:
The purpose of this PR is to bring sorting functionality to the mail reader dialog. Support for the German client is included and I'll attach a sample of mail messages below with German date strings in their headers.

Because of it's proximity to the newsglobe sorting, this PR refactors the shared code into `mailnews`.

https://github.com/user-attachments/assets/596f7c97-f258-49f4-8162-93bedcab5ebf

https://github.com/user-attachments/assets/edd6d798-8f73-427a-80ee-973050cec472

[GermanMailSamples.zip](https://github.com/user-attachments/files/16202507/GermanMailSamples.zip)
### 
This closes #26 in combination with #398


module/mailnews/mailread.c
```diff
@@ -353,3 +404,174 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+      time2 = mail_date_map[msg_num2];
+
+      return bSortAscending ? (time1 - time2) : (time2 - time1);
+   default:
+      debug(("Unhandled column (column #%d) in CompareListItems\n", nColumn));
+      return 0;
+   }
+}
+
+void ListView_SetHeaderSortImage(HWND hListView, int sortedColumn, BOOL sortAscending)
```
akirmse, line 479: I'm super confused.. there's also a function with the same name declared in the header, so that it's visible everywhere?

- Only put function signatures in headers if they're shared.  Otherwise declare them static in the .c file.
- If the functions are the same, have only one copy.
- Otherwise, name them differently so as not to confuse the reader.

module/mailnews/mailread.c
```diff
@@ -353,3 +404,174 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+      // Handle parse error
+      debug(("Failed to parse date string: %s\n", dateStr));
+      return 0;
+   }
+
+   // Adjust the struct tm fields
+   tm.tm_year -= 1900; // Year since 1900
+
+   // Convert month abbreviation to month number
+   const char *months[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
```
akirmse, line 541: This will break in German or any other language---you definitely don't want to be doing this yourself.  Look for a system routine that does this.  Maybe https://stackoverflow.com/questions/321849/strptime-equivalent-on-windows

davidklaw: I've checked in what I think might be a viable solution, but have ventured into unfamiliar territory yet again. The solution I dropped in is the most succinct solution I could come up with, this late in the evening at least. I may return to solution in this SO post again tomorrow and see about integrating locale into the mix there.

module/mailnews/mailread.c
```diff
@@ -353,3 +400,145 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+ */
+/***************************************************************************/
+int StringToTimestamp(const char *dateStr)
+{
+   struct tm tm = {0};
+
+   const char *localeName = setlocale(LC_ALL, "");
+
+   // Skip the day of the week
+   const char *remainingStr = dateStr + 4;
```
akirmse, line 487: Check the length before doing this or it can crash.

module/mailnews/mailread.c
```diff
@@ -353,3 +400,145 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+
+   std::istringstream input(remainingStr);
+
+   input.imbue(std::locale(localeName));
```
akirmse, line 491: I've never seen some of these functions before.  Hopefully you can test this by forcing your locale to German and receiving some mail.

module/mailnews/mailnews.c
```diff
@@ -402,3 +416,41 @@ bool IsNameInIgnoreList(const char *name)
   }
   return false;
 }
+
+/***************************************************************************/
+/*
+ * ListView_SetHeaderSortImage sets sort indicator image for a given ListView provided a column index and sort direction.
```
akirmse, line 422: long line

module/mailnews/mailread.c
```diff
@@ -11,12 +11,34 @@
...
+   COL_SUBJECT = 2,
+   COL_TIME = 3
+} MailColumns;
+
+// Constants used to track sorted column
+static int currentSortColumn = -1;
+static bool currentSortAscending = TRUE;
+
+// Hashtable to store converted mail timestamps
+static std::unordered_map<int, time_t> mail_date_map;
```
akirmse, line 40: It would be good hygiene to clear() this map when the dialog exits.

module/mailnews/mailread.c
```diff
@@ -99,6 +127,19 @@ INT_PTR CALLBACK ReadMailDialogProc(HWND hDlg, UINT message, WPARAM wParam, LPAR
 
       MailGetMessageList();
       RequestReadMail();
+
+      // Load sort arrow bitmaps
+      hbmUpArrow = (HBITMAP)LoadImage(hInst,
```
akirmse, line 132: Loading the bitmaps is already done elsewhere, so this is a memory leak?

module/mailnews/mailread.c
```diff
@@ -353,3 +400,145 @@ void UserMailReply(int msg_num, Bool reply_all)
    MakeReplySubject(reply->subject, MAX_SUBJECT);
    MailSendReply(hReadMailDlg, reply);
 }
+
+void OnColumnClick(LPNMLISTVIEW pLVInfo)
+{
+   int nSortColumn = pLVInfo->iSubItem;
+   bool bSortAscending = TRUE;
```
akirmse, line 407: true, not TRUE

module/mailnews/mailread.c
```diff
@@ -353,3 +400,145 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+   if (nSortColumn == currentSortColumn)
+   {
+      // Toggle sort direction if the same column is clicked again
+      currentSortAscending = !currentSortAscending;
+   }
+   else
+   {
+      // Set ascending order for new column
+      currentSortColumn = nSortColumn;
+      currentSortAscending = TRUE;
```
akirmse, line 418: true

module/mailnews/mailread.c
```diff
@@ -11,12 +11,34 @@
...
+#include <iomanip>
+#include <iostream>
+#include <sstream>
 
 extern HWND hReadMailDlg; /* Non-NULL if Read Mail dialog is up */
 extern HWND hSendMailDlg; /* Non-NULL if Send Mail dialog is up */
 
 static RECT dlg_rect; // Screen position of dialog
 
+// Constants for news dialog columns
```
akirmse, line 26: "mail dialog columns"?

module/mailnews/mailread.c
```diff
@@ -11,12 +11,34 @@
...
+{
+   COL_ORDER = 0,
+   COL_FROM = 1,
+   COL_SUBJECT = 2,
+   COL_TIME = 3
+} MailColumns;
+
+// Constants used to track sorted column
+static int currentSortColumn = -1;
+static bool currentSortAscending = TRUE;
```
akirmse, line 37: true

module/mailnews/mailread.c
davidklaw: Ah, I let the German client slip from my mind. Preliminary tests are all looking good (had to remove the NOSORT flag for the German resources) and I now know how to fire up the German client 👍🏻

module/mailnews/mailread.c
```diff
@@ -353,3 +395,153 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+   bool bSortAscending = (lParamSort > 0);
+   int nColumn = abs(lParamSort) - 1;
+   char szBuf1[MAX_SUBJECT], szBuf2[MAX_SUBJECT];
+   int time1 = 0;
+   int time2 = 0;
+
+   HWND hListView = GetDlgItem(hReadMailDlg, IDC_MAILLIST);
+
+   // Extract the msg_num values from the list view items
+   ListView_GetItemText(hListView, lParam1, COL_ORDER, szBuf1, sizeof(szBuf1));
```
davidklaw, line 443: > Note: During the sorting process, the list-view contents are unstable. If the callback function sends any messages to the list-view control aside from LVM_GETITEM (ListView_GetItem), the results are unpredictable.

Noting this here in case it raises any red flags. It would seem to me like rescanning would be the only existing functionality that _could_ interfere, but doesn't.

module/mailnews/mailread.c
```diff
@@ -353,3 +395,153 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+
+   LPARAM lParamSort = (currentSortAscending ? 1 : -1) * (currentSortColumn + 1);
+
+   if (hbmUpArrow != NULL && hbmDownArrow != NULL)
+   {
+      ListView_SetHeaderSortImage(pLVInfo->hdr.hwndFrom, currentSortColumn, currentSortAscending);
+   }
+
+   // Sort list
+   ListView_SortItemsEx(pLVInfo->hdr.hwndFrom, CompareMailListItems, lParamSort);
```
davidklaw, line 424: I want to call out that mail uses `ListView_SortItemsEx` while news uses `ListView_SortItems`. Down the line `ListView_GetItemText` (uses to retrieve text from the mail list) takes the item index.

> With ListView_SortItemsEx, the item's index is passed instead of its lparam value.

[Source](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-listview_sortitemsex)

module/mailnews/mailread.c
```diff
@@ -353,3 +395,153 @@ void UserMailReply(int msg_num, Bool reply_all)
...
+   }
+
+   // Skip past the space character
+   remainingStr++;
+
+   // Use a string stream to parse the date and time, leveraging locale
+   // to support the translation of month strings.
+   std::istringstream input(remainingStr);
+   input.imbue(std::locale(localeName));
+   input >> std::get_time(&tm, "%b %d, %Y %H:%M");
```
akirmse, line 495: Is it possible to use %a in get_time rather than manually skipping the day of the week in the string?

https://en.cppreference.com/w/cpp/io/manip/get_time

davidklaw: Oops! I thought it would contribute to the timestamp generation but that's not the case.

> Renormalizes local calendar time expressed as a struct tm object and also converts it to time since epoch as a time_t object. **arg->tm_wday** and arg->tm_yday are ignored.

PR #804, opened 2024-07-15
Brings back atmospheric lighting to 'Underground Lake'

### LINKED ISSUE #407: Graphics: Room : The Underground Lake roof texture and light are wrong
When @marksturm added the entrance to the Winding Caverns from the Underground Lake in 2012, the roof was covered up.

The Underground Lake is supposed to show the open sky and background sky graphics in the center of the ceiling, with light (not dynamic but via room editor) coming from it onto the Jewel of Froz pillars.

1) The ceiling is covered by a stone texture where the hole would've been

2) There is a misplaced light sector on another section of the ceiling visible when you go past the center of the map towards the north side and look up. 

The texture will not change lighting based on the room and is lit at maximum.

In the center of the ceiling,The open sky light that came down to the center platform is possibly what is causing this inside the .roo .

It's suggested that the entire ceiling be looked at again to get it back to its original state. 

While being edited, whoever fixes the roof could probably take a look at adding some more space (and depth) to the secondary door.

davidklaw:
This PR is a combination of `kod` and `roo` updates for `cave3`, the Underground Lake.

I cleaned up a number of texture issues, replaced wall textures around the perimeter to give it some depth at a distance, returned the roof to sky, made the waterfall a little more interesting, and made big changes to lighting. Now, the area under the cave opening should be lit at all times. The rest of the cave is dark and affected by ambient light, so players will either need a torch, light, night vision, or their map to navigate their way around the perimeter.

There's quite a lot going on and it's best to jump in and look around yourself in both hardware and software modes.

Screenshot 2024-07-14 195354
Screenshot 2024-07-14 195420
Screenshot 2024-07-14 195444

**Notes:**
- There were many sectors with animation values that were unnecessary. I set those to 0.
- `cave2` suffers from a similar lighting problem. It's almost entirely unaffected by light level spells when it could be quite moody and interesting for newer players.
- The changes made to the original `cave3` to allow for access to Brax impacted the visual depth of the room. Previously, the sectors around the perimeter rose up to varying heights and were visible to the player. It created an interesting backdrop to the centerpiece that is more or less non-existent today. This PR fixes a few things, but I'm also experimenting with a) the perimeter geometry and b) access to Brax.

This closes #407.


akirmse:
Tried it locally, looks good here.

PR #807, opened 2024-07-18
Disables Hunt until it's implementation issues are resolved

davidklaw:
This PR disables Hunt. When used on offline targets who log on during it's duration, the target will be tracked indefinitely. Offline players' timers are "frozen" by `FreezeAllEnchantments()` and converted into integer times, but Hunt introduces an active Timer that decouples from the enchantment during reactivation of enchantments in `ReactivateAllEnchantments()`. This should be addressed or the spell reworked before re-enabling.

### DIFF
--- a/kod/object/passive/spell/multicst/hunt.kod
+++ b/kod/object/passive/spell/multicst/hunt.kod
@@ -59,7 +59,6 @@ classvars:
    viSpell_level = 5
 
    viSpellExertion = 50
-   viCast_time = 20000
    viChance_To_Increase = 50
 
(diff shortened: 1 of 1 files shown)


kod/object/passive/spell/multicst/hunt.kod
```diff
@@ -59,7 +59,6 @@ classvars:
    viSpell_level = 5
 
    viSpellExertion = 50
-   viCast_time = 20000
```
davidklaw, line 62: I opted to remove this in this PR because it's defined again below.

PR #850, opened 2024-07-23
Set static height and width view of the world for bsp walk

sicoderealm:
In busier rooms (number of items/lights/walls) and at higher resolutions the number of draw items may exceed the maximum allowed. On further investigation the greater the resolution, the greater the number of items being identified as visible in the world - which is wrong. This spawns from the original implementation of the full screen for software renderer: https://github.com/Meridian59/Meridian59/pull/911

When this happens out of memory messages would appear in the debug logs as the `drawdata` was full and as a result objects were missing and not rendered in the 3d world.

As the world is being built from the bsp tree using a static world size and the width is given as a parameter - regardless of the screen resolution the same objects in the world should be visible. This suggests that the width of the players view on the screen directly affects the bounds of the viewing frustum - it should ideally be independent of the actual size of the viewport on the screen.

With this change the number of items in `drawdata` remains static regardless of the resolution of the viewport and the out of memory debug messages no longer appear. Overall this will be an optimization for all resolutions using the hardware renderer.

We now use a static height / width, with a small buffer,  for the purposes of traversing the bsp while using the hardware renderer.

This is how the issue manifests:
https://github.com/user-attachments/assets/994abf14-f54a-4c87-8ebd-4b3f66986e45

### DIFF
--- a/clientd3d/drawdefs.h
+++ b/clientd3d/drawdefs.h
@@ -33,7 +33,7 @@
 /* Scaler used to increase the classic view port x,y values */
 #define VIEW_SCALER 5
 
-/* The multiplier ensures that objects at the edges of the viewport are still considered visible. */
+/* Multiplier for hardware renderer view port area  */
 #define SIZE_MULTIPLIER 1.1
 
 /* Maximum values used to create view textures and buffers */


clientd3d/draw3d.c
```diff
@@ -387,8 +387,8 @@ void UpdateRoom3D(room_type *room, Draw3DParams *params)
 
    // Size of offscreen bitmap.
    area.x = area.y = 0;
-   area.cx = main_viewport_width;
-   area.cy = main_viewport_height;
+   area.cx = CLASSIC_WIDTH * 1.5;
```
akirmse, line 390: We have a lot of magic numbers like this in the renderer now.  Is it possible to derive this from other numbers?  Or at least document why 1.5 and not some other value.

akirmse: Also I'll add that the comment says that this is the size of the offscreen bitmap, not the size of a viewport used for culling.  Are you sure you're not affecting the rendered resolution with this change?  I would expect the offscreen bitmap to be the same size as the viewport, especially for software mode.

PR #851, opened 2024-07-24
Eliminate wasted defense improvement rolls to fix Stuck Parry problem

SioraSidhe:
There is code when a player kills a monster, it randomly sends an improve chance to block, dodge, or parry. This still happens when the skills are maxed, so lots of improvement chances are wasted. This leads to a situation with Parry that is not obvious. It is why players get Stuck Parry. When a player has other schools, their secondary improvement chance on Parry drops to minimum, as low as 1 in 100. Their primary improvement chance is between 20 and 30 based on Intellect, with +0-15 from Jonas faction power. So players must beat a very tough roll often 1 in 100 for the secondary. Then 20-45 in 100 for the primary.

Then we see the problem above, where 70 of 100 improvement chances are going to Block and Dodge even when are they are maxed, so it is 1 in 100, 20-45 in 100, 30 in 100 all must be beat to improve. This combined with only getting 1 attempt per monster kill is very slow, but then also you must have avoided a monster attack. If you buff a lot to kill mobs faster to get more attempts to improve, you will start killing monsters too fast and not avoid one of their attacks before they die. This makes a trap where players who try to improve Parry are unable to, because they don't know about this, and that reduces their chances even more. A player may kill more than a thousand mobs  never see an improve and assume they are stuck. Even an administrator cannot see or fix this, there is no clear problem to see, it's hidden.

This change sends all improvement chances to skills that are not maxed, this should help. The right place to set Parry chance to imp is viChance_to_Increase, currently 20, same as Dodge, yet Parry is much harder due to these hidden knots.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -7653,6 +7653,8 @@ messages:
             AND Send(self,@HasSkill,#num=SKID_PARRY)
          {
             if random(1,100) < 30
+               OR (Send(self,@GetSkillAbility,#Skill_num=SKID_BLOCK) >= 99
+                  AND Send(self,@GetSkillAbility,#Skill_num=SKID_DODGE) >= 99)
             {
                dodgeskill=SKID_PARRY;
             }
@@ -7663,6 +7665,7 @@ messages:
             AND Send(self,@HasSkill,#num=SKID_BLOCK)
          {
             if random(1,100) < 30
+               OR Send(self,@GetSkillAbility,#Skill_num=SKID_DODGE) >= 99
             {
                dodgeskill=SKID_BLOCK;
             }


akirmse:
I'm not sure I follow your explanation.  When block and dodge are maxed, this changes the "30 in 100" to "100 in 100" but does not address the other parts, which seems to be what you're unhappy with.

I'll also add that this in some sense makes the game shorter by reducing the time until a player is completely maxed out.  When changing the balance like this, it's a good idea to propose a counterbalance of equal strength that makes the game longer.  Otherwise, over time, the game becomes easier and easier as these changes pile up.

davidklaw:
If the intent is to have parry and block be challenging to level up, I think the current code accomplishes  that.

As a bit of an aside, if I'm not mistaken, this area of the code is saying that _level 1_ Block can't be improved without having _level 2_ Dodge. That doesn't seem right 🤔

SioraSidhe:
I suggest Parry viChance_to_Increase change to a lower number instead. I suggest 10 or 8 for this value. If the difficulty is in that roll, Intellect and Faction Power can be used to mitigate the difficulty. Right now, nothing can help with the wasted rolls problem. You can't take off a shield or stop dodging to focus on Parry. You can take off a shield and weapon to focus on Dodge, it only works one way.

The player doesn't need Dodge to level up Block, PFLAG_DODGED is given on any avoid, so that's okay.

SioraSidhe:
I have found more that describes the problem. Spells have a School Casts system that accumulates progress toward improvement. Just keep casting, your chance to improve slowly rises. Skills do not have this system. There are 2 skills that get Stuck if done as the last school with low stats, Parry and Disarm. They both cannot be targeted for improvement, both have problems that bring a player to the point they cannot advance without exponential grind. Second Wind may also suffer this wall. A player will not know this until it is too late and they will have to forget their schools and start over. What should be done?

PR #856, opened 2024-07-24
Adjustments to fix missing rooftop in East Jasper (`jas_east`)

davidklaw:
This PR fixes a broken rooftap in Jasper, seemingly caused by an intersecting line. There are several lines surrounding this building and which I believe might have been used to keep players away from troublesome notches or angles of the building. I didn't remove them -- only nudged a vertex for one out of the way of the building linedef. With that and a few adjustments to the texture rotation, the roof is fixed.

Note: This PR also deletes `jasper.roo` which appears to be a copy of `jas-east.roo` and is unused.

### Before & After
Screenshot 2024-07-24 080742
Screenshot 2024-07-24 080931
Screenshot 2024-07-24 080948
Screenshot 2024-07-24 083132
Screenshot 2024-07-24 083554


davidklaw:
I just realized that `jas-east.roo` is the roo file in use. I'll apply these some fixes to that and update the PR.

PR #872, opened 2024-07-27
Various fixes in Goad's Grinder (`necarena`)

davidklaw:
- Adds a missing texture to the surface of the arena floor
- Increases the height of the shrine "room"

When opening the roo file in roomedit, the following error is seen because the arena floor and floor under Goad had no texture assigned to them.

Untitled

### Before & After
Screenshot 2024-07-27 085722
Screenshot 2024-07-27 090827
Screenshot 2024-07-27 090948
Screenshot 2024-07-27 091511


jrotunda:
Wild, I always thought the black texture was an intentional effect here! Now it looks like a proper abattoir.

PR #898, opened 2024-07-29
Only check for account activation after 3 failed logins

sicoderealm:
Checking account activation is relatively slow - we now only do it after three failed login attempts.


clientd3d/login.c
```diff
@@ -173,7 +170,6 @@ void CheckAccountActivation(void)
             switch (webResponse)
             {
             default:
```
akirmse, line 172: This default case no longer does anything; remove it.

PR #901, opened 2024-08-03
Make npc text consistently black

sicoderealm:
image


davidklaw:
Here's a good example of what happens today. These two Frisconar messages popped up at the same time:

image

This fix looks good to me!

PR #903, opened 2024-08-04
Updates mollusk creature to include initial underground state

### LINKED ISSUE #429: Monster: Mollusk Creature is missing its dormant > standing animation frames
The Mollusk Creature is supposed to be dormant and hiding underground, looking like a rock. When aggravated, it comes out of the ground.

The current code for it, MolluskMonster.kod, isn't using the animation frames correctly and the monster has never been seen with its full animations which were created for it and are there in the files.

Note: To see what the full animation looks like, you can set a portal to use molluskmonster_icon_rsc if not in BGF editor.

davidklaw:
This PR enhances mollusk creatures in a couple of ways:
1. It updates the code and bgf to include a new "underground" state from which mollusks emerge once engaged. This state was in a separate `molrock.bgf` and it looks like some support might have been written into `mollusk.kod`, was incomplete.
2. I extracted the bitmaps from both existing bgfs, cleaned up alpha channel color bleed, then repackaged them in a new mollusk.bgf using `makebgf` and an updated `bbg`.

The most challenging aspect of this was disallowing the monster from doing anything until the popup animation completed. Some of my earlier solutions results in out of sync animation frames. I settled on a timer that paralyzes the monster during the animation.

#### mollusk.bgf source assets
[mollusk_assets.zip](https://github.com/user-attachments/files/16921998/mollusk_assets.zip)

## Testing
This new behavior should be tested using various triggers, including up close melee and ranged attacks. In all cases, the mollusk should remain in position, emerge, then start chasing and attacking the player.

### Commands
`dm monster mollusk creature` to generate a Mollusk

### The new behavior when attacked up close and from afar

https://github.com/user-attachments/assets/c8835bf7-0c46-4ba0-bff4-82168bebea11

https://github.com/user-attachments/assets/e2747d02-b765-494c-abe8-fc83cbf7a0b0

**Misc:**
Something that bothers me about the artwork is how the mollusk appears to hover through many of its animations. This is because its club arm extends beyond its feet. It seems, the right fix for that is updating the artwork.

Closes #429


kod/object/active/holder/nomoveon/battler/monster/mollusk.kod
```diff
@@ -98,53 +98,13 @@ messages:
...
-   }
-
-   Delete()
-   {
-      if ptActive <> $
-      {
-         DeleteTimer(ptActive);
-         ptActive = $;
-      }
+      Send(self,@SetBehaviorFlag,#flag=AI_NOMOVE,#value=TRUE);
```
davidklaw, line 101: I'm not convinced that this is the right place to set behavior.

davidklaw:
This is now ready for review!

akirmse:
This looks OK, but the large amount of binary artwork files brings up the question of how to handle them in source control.  We haven't typically checked in artwork but maybe we should.  Should we use git-lfs?

davidklaw:
I don't know much about `git-lfs`, but it'd be nice if it supported pushing files to Drive when a PR is merged. That way, public read access could allow for anyone to retrieve source assets while limiting write access. If all that is a bit much, maybe we just ask that contributors include a zip of assets with their PRs to be manually uploaded (to a Drive).

akirmse:
git-lfs is just an efficient way to store binary files in git.  It doesn't store previous versions.  I don't follow the Drive idea, but anyway I don't think it's necessary if the artwork is checked in.

davidklaw:
Hmm, I was under the impression that it helped manage large files by maintaining versioned pointers to the source files living off in an LFS store, like Dropbox, Drive, or what have you.

akirmse:
I don't think so: https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage

davidklaw:
[This section](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage) from the same documentation says lfs stores references and not the actual files themselves, which get stored elsewhere. Now that I'm reading a bit more of this, it seems like GitHub offers its own large file storage and this is the "elsewhere" these files are stored.

I'm happy to do whatever. LFS seems like a good option to enable if we ever have significant asset production planned.

davidklaw:
This PR has been updated to include a zip of the mollusk source assets.

jrotunda:
I like the way this looks and feels in-game, the timing matches up well with the sound and attack frames. Thanks for getting this working! The code for mollusk mentions this to be a desire after they added him (he was added with no sounds too, took years to get those) but no on ever got around to it.

PR #905, opened 2024-08-07
Single render pass for translucent overlays

sicoderealm:
We apply the same techniques for overlays as for regular objects (only render each object once and restrict translucent passes)
Logoff ghosts now look consistently translucent (previously overlays -- arms, legs etc were rendered twice and at certain distances were rendered multiple times due to duplication in drawdata) 

image

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -7356,6 +7356,9 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
 
 	anglePitch = PlayerGetHeightOffset();
 
+	// drawdata[] may contain more than one entry with the same id.
+	// We need to keep track of those we've already processed to avoid duplicates.
+	std::unordered_set<int> processedIds;
 	for (curObject = 0; curObject < nitems; curObject++)
 	{
 		list_type	list2;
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender.c
```diff
@@ -7356,6 +7356,9 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
 
 	anglePitch = PlayerGetHeightOffset();
 
+	// drawdata[] may contain more than one entry with the same id.
+	// We need to keep track of those we've already processed to avoid duplicates.
```
akirmse, line 7360: Why?  Explain that this causes problems with transparency.

clientd3d/d3drender.c
```diff
@@ -7383,6 +7390,18 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new *pPool, room_type *room, Draw3DPa
 				continue;
 		}
 
+		// Check for translucent objects
+		bool objTranslucent = (pRNode->obj.flags & TRANSLUCENT_FLAGS) != 0;
+		if ((flags & TRANSLUCENT_FLAGS) != 0) {
```
akirmse, line 7395: Simplify by making this into == and flipping the two branches of the if.

PR #906, opened 2024-08-08
Increase server update rate & incoming packet allowance

paul-bargerstock:
The purpose of these changes is simple, to allow for more 'real time' server/client location updates.  In summary:

1. Decreased the movement interval in which the server is informed of changes
2. Decreased the minimum amount of distance travel distance to where the server should be informed
3. Increased the packet throttle to support the above changes

I chose the values for MOVE_INTERVAL and INCOMING_PACKET_THROTTLE based on values from other open-source forks that have been used 'in the wild' for many years.

**I tested these changes with the following results:**

- I am able to stop on a dime and channel a spell, without a non-moving fizzle
- My attack range (and defense range) is now far more predictable. I no longer need to 'predict' where a target is heading in order to swing in melee combat; nor do monsters hit me as I'm already well out of their range from the client's perspective.
- Entering and exiting through doors is now snappy, and doesn't require delay for consistency

As a disclaimer, I am by no means an expert.  I made these changes through research and testing - any feedback or suggestions are welcome.

### DIFF
--- a/clientd3d/move.c
+++ b/clientd3d/move.c
@@ -54,8 +54,8 @@
 
 #define MAX_STEP_HEIGHT (HeightKodToClient(24)) // Max vertical step size (FINENESS units)
 
-#define MOVE_INTERVAL  1000            // Inform server at most once per this many milliseconds
-#define MOVE_THRESHOLD (FINENESS / 4)  // Inform server only of moves at least this large
+#define MOVE_INTERVAL  250            // Inform server at most once per this many milliseconds
+#define MOVE_THRESHOLD (FINENESS / 8)  // Inform server only of moves at least this large
 #define MOVE_THRESHOLD2 (MOVE_THRESHOLD * MOVE_THRESHOLD)
 #define TIME_THRESHOLD 1000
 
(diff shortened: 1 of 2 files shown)


paul-bargerstock:
@CardiovascularMike Brought to my attention that these changes, as fabulous as they feel to play on, trigger the existing speedhack detections in user.kod > UserMove.

Due to that fact this PR is not ready for merge - does anyone have suggestions on how we can have our cake and eat it too?

paul-bargerstock:
I just made an adjustment (after testing) to the iDelta calculation to account for the movement updates being increased four fold, along with updates to the surrounding comments explaining as much.  Speedhack detection is no longer triggered.

To my (limited) knowledge, this is ready for review.

akirmse:
I haven't gotten a chance to look at this yet.  It is a far-reaching proposal with many consequences that will take time to analyze and test.  Probably it will need a runtime switch to turn it on or off, which you can get started on while you wait.

This will increase peak CPU usage by between 4x and 16x, depending on how much players cluster in the same rooms.  What is your plan for funding that?

paul-bargerstock:
Before I ask details on how you’d like to see that runtime switch implemented, or the large amount of time to analyze and test is spent; I want to ensure I’m understanding the last question clearly.

Are you asking what type of fundraising I’m planning on doing in order to cover the increased CPU use?

Meridian59:
The money will have to come from somewhere.. curious as to your ideas.

paul-bargerstock:
I do have several I think could work, but none of them come without questions of their own.  In the interest of keeping this PR related to the code/change itself, I'll take that conversation offline and reach out.

PR #909, opened 2024-08-15
Fix heat scrolls by updating of target retrieval logic

### LINKED ISSUE #728: Heat scrolls don't work at all
I vaguely recall these being problematic in the past, but at the moment, trying to use a heat scroll results in failure, with the following system message sent to the player:

"There are no proper targets for heat."

To contrast, casting the heat spell in the exact same room works as expected.

davidklaw:
In the game currently are two harmful scrolls: heat and earthquake. Earthquake is disabled and heat is enabled, but fails to cast. The issue as I understand it, is that as a `SpellItem`, scrolls expect a target. This seems to work fine for most scrolls, but _harmful_ scrolls/spells like heat and EQ run through additional logic that fails the spell because the target is the player themselves.

This PR adds a small bit of logic to spell item casts that says, ignore any "apply_on" target if this is a scroll. It does makes a bold assumption that no future scroll will be a targeted spell, which makes me wonder if I should be checking for a room enchantment spell rather than a scroll. I'm curious to hear thoughts on that as I'm leaning towards enchantment.

### The issue
When used, the scroll fails and says:
> There are no proper targets for heat.

https://github.com/user-attachments/assets/bc73ceea-be27-4024-ae6a-25bdce37d1db

### The fix in action

https://github.com/user-attachments/assets/41bf2a6e-c6bd-41c5-85b6-dee700ad2281

Closes #728


kod/object/item/passitem/spelitem.kod
```diff
@@ -216,8 +216,16 @@ messages:
       oSpell = Send(SYS,@FindSpellByNum,#num=viSpellEffect);
 
       % Get proper targets according to spell, assume that apply_on is our basic target.
-      lTargets = Send(oSpell,@GetTargets,#who=what,#lTargets=[apply_on]);
 
+      if IsClass(self,&Scroll)
```
akirmse, line 220: Rather than a gross IsClass, this looks like a prime candidate for a new little function the Scroll can override.  GetTargetsForSpell or something like that.

davidklaw:
This PR's commit history went off the rails. A new PR -- #932 -- has been created to replace it.

PR #911, opened 2024-08-15
Stretch classic software renderer with classic original scale and fov

sicoderealm:
Full screen mode of the software renderer does not currently scale using the classic resolution and FOV. The FOV is calculated by the height and width defined within the draw room routine which we are setting with the height and width available. This is wrong as it also increases the view port and FOV significantly, above and beyond the original classic client view (and the hardware renderer). Instead we now improve the software renderer full screen implementation by scaling the classic view to the size of the available view port area -- importantly keeping the original scale and FOV.

Current full screen view:

image

With this new classic stretched implementation (this is large full screen max):

image

Scaled from original smaller ~classic view port size (this is a small view):

image

### DIFF
--- a/clientd3d/draw3d.c
+++ b/clientd3d/draw3d.c
@@ -86,10 +86,10 @@ BYTE light_rows[MAXY/2+1];      // Strength of light as function of screen row
 
 PDIB background;                      /* Pointer to background bitmap */
 
-static void Stretch(BYTE* src, BYTE* dest, int width, int height, int new_width, int new_height, int max_width);
-
 /* local function prototypes */
 static void StretchImage(void);
+static void Stretch(BYTE* src, BYTE* dest, int width, int height,
+    int new_width, int new_height, int max_width);
 /************************************************************************/
 /* return TRUE iff successful */
 Bool InitializeGraphics3D(void)
@@ -536,14 +536,20 @@ void StretchImage(void)
   Stretch(gBits, gBufferBits, area.cx, area.cy, main_viewport_width, main_viewport_height, MAXX);
 }
 /************************************************************************/
-void Stretch(BYTE* src, BYTE* dest, int width, int height, int new_width, int new_height, int max_width)
+/************************************************************************/
+/*
+ * Stretch: Function to stretch a source image to a new size while maintaining the aspect ratio.
+ * - src: Pointer to the source image data
+ * - dest: Pointer to the destination image buffer where the stretched image will be stored.
+ * - width: Width of the source image.
+ * - height: Height of the source image.
+ * - new_width: Desired width of the stretched image.
+ * - new_height: Desired height of the stretched image.
+ * - max_width: Maximum width of the destination buffer (used for row size calculations).
+ */
+void Stretch(BYTE* src, BYTE* dest, int width, int height, 
+    int new_width, int new_height, int max_width)
 {
-    // If the new size is exactly the same as the original, perform a direct copy
-    if (new_width == width && new_height == height) {
-        memcpy(dest, src, width * height * sizeof(BYTE));
-        return;
-    }
-
     // Calculate the scaling ratios
     int half_new_width = new_width / 2;
     int x_ratio = (width << 16) / half_new_width;
--- a/clientd3d/draw3d.c
+++ b/clientd3d/draw3d.c
@@ -86,7 +86,7 @@ BYTE light_rows[MAXY/2+1];      // Strength of light as function of screen row
 
 PDIB background;                      /* Pointer to background bitmap */
 
-static void StretchC1To2(BYTE *src,BYTE *dest,int width,int height);
+static void Stretch(BYTE* src, BYTE* dest, int width, int height, int new_width, int new_height, int max_width);
 
 /* local function prototypes */
 static void StretchImage(void);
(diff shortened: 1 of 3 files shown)


clientd3d/draw3d.c
```diff
@@ -86,7 +86,7 @@ BYTE light_rows[MAXY/2+1];      // Strength of light as function of screen row
 
 PDIB background;                      /* Pointer to background bitmap */
 
-static void StretchC1To2(BYTE *src,BYTE *dest,int width,int height);
+static void Stretch(BYTE* src, BYTE* dest, int width, int height, int new_width, int new_height, int max_width);
```
akirmse, line 89: split long line

clientd3d/draw3d.c
```diff
@@ -533,30 +533,48 @@ void SetLightingInfo(int sun_x, int sun_y, BYTE intensity)
...
-  register BYTE *s,*d,*d2;
-  
-  /* 1->2 stretch */
-  for(i = 0;i < height;i++)
-  {
-    s = src + i * MAXX;
-    d  = dest + 2 * i * MAXX * 2;
-    d2 = d + MAXX * 2;                // One row offset from dest
-    for(j = 0; j < width; j++)
+    // If the new size is exactly the same as the original, perform a direct copy
```
akirmse, line 541: It seems like this is extremely unlikely; is it worth bothering with this special case?

clientd3d/graphics.c
```diff
@@ -111,14 +110,15 @@ void ResizeAll(void)
...
+    float scale_y = static_cast<float>(main_viewport_height) / CLASSIC_HEIGHT;
 
-   *x = (client_x - view.x) / 2;
-   *y = (client_y - view.y) / 2;
+    // Translate client coordinates to room coordinates, considering scaling
+    *x = static_cast<int>((client_x - view.x) / scale_x);
+    *y = static_cast<int>((client_y - view.y) / scale_y);
 
-   return True;
+    return true;
```
akirmse, line 121: true is not a Bool

clientd3d/graphics.c
```diff
@@ -111,14 +110,15 @@ void ResizeAll(void)
 Bool TranslateToRoom(int client_x, int client_y, int *x, int *y)
 {
 
-   if (client_x < view.x || client_x > view.x + view.cx ||
-       client_y < view.y || client_y > view.y + view.cy)
-      return False;
+    // Calculate the scaling factor
+    float scale_x = static_cast<float>(main_viewport_width) / CLASSIC_WIDTH;
+    float scale_y = static_cast<float>(main_viewport_height) / CLASSIC_HEIGHT;
```
akirmse, line 115: This function no longer ever returns False, which seems like it must break something somewhere, as the comment for the function describes the conditions under which it should return False.

clientd3d/draw3d.c
```diff
@@ -533,30 +533,48 @@ void SetLightingInfo(int sun_x, int sun_y, BYTE intensity)
  */
 void StretchImage(void)
 {
-  StretchC1To2(gBits,gBufferBits,area.cx,area.cy);
+  Stretch(gBits, gBufferBits, area.cx, area.cy, main_viewport_width, main_viewport_height, MAXX);
 }
 /************************************************************************/
-void StretchC1To2(BYTE *src,BYTE *dest,int width,int height)
+void Stretch(BYTE* src, BYTE* dest, int width, int height, int new_width, int new_height, int max_width)
```
akirmse, line 539: I'm not sure what this function is doing, and there's no comment on that, or what the parameters mean.  It's always doing a 2:1 stretch, in addition to some other stretching to fill the viewport? 

We want to make sure that source pixels are always drawn as squares, without anisotropy.  i.e. it would be better to reduce the vertical FOV (by restricting to the middle of the source bitmap) than to distort the image by stretching it in a non-uniform way.

clientd3d/draw3d.c
```diff
@@ -533,30 +533,54 @@ void SetLightingInfo(int sun_x, int sun_y, BYTE intensity)
...
-    s = src + i * MAXX;
-    d  = dest + 2 * i * MAXX * 2;
-    d2 = d + MAXX * 2;                // One row offset from dest
-    for(j = 0; j < width; j++)
+    // Calculate the scaling ratios
+    int half_new_width = new_width / 2;
+    int x_ratio = (width << 16) / half_new_width;
+    int y_ratio = (height << 16) / new_height;
+
+    int i, j, x, y;
```
akirmse, line 558: Declare variables at point of first use

clientd3d/draw3d.c
```diff
@@ -533,30 +533,54 @@ void SetLightingInfo(int sun_x, int sun_y, BYTE intensity)
...
+/************************************************************************/
+/*
+ * Stretch: Function to stretch a source image to a new size while maintaining the aspect ratio.
+ * - src: Pointer to the source image data
+ * - dest: Pointer to the destination image buffer where the stretched image will be stored.
+ * - width: Width of the source image.
+ * - height: Height of the source image.
+ * - new_width: Desired width of the stretched image.
+ * - new_height: Desired height of the stretched image.
+ * - max_width: Maximum width of the destination buffer (used for row size calculations).
```
akirmse, line 548: This is usually called "row stride" (it's not really the maximum of any width).

clientd3d/draw3d.c
```diff
@@ -533,30 +533,54 @@ void SetLightingInfo(int sun_x, int sun_y, BYTE intensity)
...
 void StretchImage(void)
 {
-  StretchC1To2(gBits,gBufferBits,area.cx,area.cy);
+  Stretch(gBits, gBufferBits, area.cx, area.cy, main_viewport_width, main_viewport_height, MAXX);
 }
 /************************************************************************/
-void StretchC1To2(BYTE *src,BYTE *dest,int width,int height)
+/************************************************************************/
+/*
+ * Stretch: Function to stretch a source image to a new size while maintaining the aspect ratio.
```
akirmse, line 541: Clarify the units of the heights and widths: pixels or bytes?

PR #912, opened 2024-08-17
Reorganization and README for dialog folder (`/dlg`) 

davidklaw:
This PR consolidates currently in-use dialog and archives others for reference.

### Consolidation
`konima.tab`, `ren.tab`, and `npc.tab` were all dialog already in `npc1.tab` (renamed to `npc.tab`) and therefore unnecessary unless desirable for insight into iterations of dialog. In this PR, I have deleted these files as I felt they were no longer needed.

### Archival
`councilor.tab`, `murder.tab`, `stage1.tab`, and `stage2.tab` all contain dialog revolving around the Duke's murder and the introduction of Miriana. These contain dialog that was seemingly temporary and event-based. I kept them around in the case that they might be re-used or inspire new dialog.

### README
I've added a README file with some simple instructions.

### Additional:
- I cleaned up a couple of duplicates, but there are others related to "games" like the coin flip that still need investigation.
- Re-reading the same dialog will create duplicates in the speech library. Dialog updates seem infrequent enough that my proposal is for folks to create new tab and txt files for new dialog, independent on the existing dialog (noted in a README file I've added).


dlg/README
```diff
@@ -0,0 +1,26 @@
+# NPC Dialog Management
+
+These files manage random and triggered NPC dialog, primarily managed through `.tab` files and converted into admin commands using `dialogconvert.py`.
+
+## Setting Up Dialog
+
+After recreating the world (e.g., `send o 0 recreate all`) or during initial startup:
+
+1. Move `dialog.txt` to the server run directory
+1. Execute the admin command `read dialog.txt` to load NPC dialog into the speech library.
```
akirmse, line 10: All numbered 1?

dlg/README
```diff
@@ -0,0 +1,26 @@
...
+
+## Adding New Dialog
+
+To avoid duplicates, new dialog should be adding using iterative `.tab` files:
+
+1. Create a new `.tab` file with the desired dialog.
+1. Convert it to admin commands with `python dialogconvert.py newfile.tab newdialog.txt`
+1. Loading the dialog using `read newdialog.txt` in Blakserv.
```
akirmse, line 18: Here too.

dlg/README
davidklaw: I did this with GitHub Markdown rendering in mind, but this isn't a Markdown file! I'll update.

dlg/README
davidklaw: Fixed.

dlg/README
davidklaw: Fixed.

PR #914, opened 2024-08-18
Modifications to Izzio's messaging

### LINKED ISSUE #509: Izzio's purchase success message triggers incorrectly
When buying items from Izzio that you cannot carry, he will correctly state that you can't carry them but also incorrectly act as if you had a successful purchase.

From the chat log:
> Izzio says, "I'm unable to give you the mushrooms.  Perhaps you carry too much?"
> Izzio says, "Thanks!  May that bring you good health and wealth."

The expected behavior is that he tells you you're unable to carry them when you can't, and thanks you when you can (and makes the transaction of course).

davidklaw:
This PR makes it so that Izzio no longer reacts to items bought using `ItemBoughtHook` (Issue #509). Izzio was unique in being the only NPC to do this and it becomes problematic when player try to purchase single item they cannot carry. Izzio responds by telling them that they can't carry it and thanks them for their purchase at the same time.

Rather than modify the workings of `monster.kod` to handle this type of situation, I removed Izzio's unique use of the hook to bring him inline with other NPCs. Furthermore, I added 10 new random quotes to help make his presence known and add to his personality.

#### Here you can see the new quotes in action (`send o <Library> DebugRandomLib mob o <IzzioId>`)

Screenshot 2024-08-18 080215

#### This is the behavior now when the player is unable to carry items:

Screenshot 2024-08-18 080314

### Dialog files
The new random quotes are in new `.tab` and `.txt` files that if merged and applied, should perhaps then be incorporated into `npc.tab` and `dialog.txt`.

Closes #509


akirmse (commented):
Since ItemBoughtHook is no longer used, you should remove it.

davidklaw:
> Since ItemBoughtHook is no longer used, you should remove it.

Can do! I refrained thinking it might over some future value.

PR #917, opened 2024-08-19
Make objects clickable in the hardware renderer

sicoderealm:
During https://github.com/Meridian59/Meridian59/pull/911 we assume all `TranslateToRoom`'s are being stretched from classic width/height. This is not the case for the hardware renderer, so we correct it for both cases.

### DIFF
--- a/clientd3d/graphics.c
+++ b/clientd3d/graphics.c
@@ -109,13 +109,21 @@ void ResizeAll(void)
  */
 Bool TranslateToRoom(int client_x, int client_y, int *x, int *y)
 {
-    // Calculate the scaling factor
-    float scale_x = static_cast<float>(main_viewport_width) / CLASSIC_WIDTH;
-    float scale_y = static_cast<float>(main_viewport_height) / CLASSIC_HEIGHT;
+    if(D3DRenderIsEnabled())
+    { 
+        *x = (client_x - view.x) / 2;
+        *y = (client_y - view.y) / 2;
+    }
+    else
+    {
+        // Calculate the scaling factor
+        float scale_x = static_cast<float>(main_viewport_width) / CLASSIC_WIDTH;
+        float scale_y = static_cast<float>(main_viewport_height) / CLASSIC_HEIGHT;
 
-    // Translate client coordinates to room coordinates, considering scaling
-    *x = static_cast<int>((client_x - view.x) / scale_x);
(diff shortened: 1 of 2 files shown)


clientd3d/graphics.c
```diff
@@ -109,13 +109,21 @@ void ResizeAll(void)
...
 Bool TranslateToRoom(int client_x, int client_y, int *x, int *y)
 {
-    // Calculate the scaling factor
-    float scale_x = static_cast<float>(main_viewport_width) / CLASSIC_WIDTH;
-    float scale_y = static_cast<float>(main_viewport_height) / CLASSIC_HEIGHT;
-
-    // Translate client coordinates to room coordinates, considering scaling
-    *x = static_cast<int>((client_x - view.x) / scale_x);
-    *y = static_cast<int>((client_y - view.y) / scale_y);
+    if(D3DRenderIsEnabled())
```
akirmse, line 112: Looks like this can be simplified a little.  Set the scale factors to 1, and adjust them for the 3D renderer.

akirmse, line 112: Space before (

clientd3d/graphics.c
```diff
@@ -109,13 +109,21 @@ void ResizeAll(void)
...
+    { 
+        *x = (client_x - view.x) / 2;
+        *y = (client_y - view.y) / 2;
+    }
+    else
+    {
+        // Calculate the scaling factor
```
akirmse, line 119: This comment doesn't help the reader.  *Why* is there a scaling factor for hardware only?

clientd3d/graphics.c
```diff
@@ -109,14 +109,26 @@ void ResizeAll(void)
  */
 Bool TranslateToRoom(int client_x, int client_y, int *x, int *y)
 {
-    // Calculate the scaling factor
-    float scale_x = static_cast<float>(main_viewport_width) / CLASSIC_WIDTH;
-    float scale_y = static_cast<float>(main_viewport_height) / CLASSIC_HEIGHT;
-
-    // Translate client coordinates to room coordinates, considering scaling
+    // Default scale factors assume no scaling required.
```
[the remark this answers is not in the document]
akirmse: > This is just based on returning things to the status quo: https://github.com/Meridian59/Meridian59/pull/911/files

Readers of the code probably aren't going to go through the change history.  Can you give them a hand by explaining where the 2 comes from?

PR #918, opened 2024-08-20
Expires faction service quests when resigned from faction

### LINKED ISSUE #465: Gameplay: QuestNodes not updating when Starting Liege loyalty then joining a different liege
If the player starts a loyalty quest for their liege, then leaves or gets booted from their faction in any way, such as by attacking a soldier NPC - the loyalty quest will stay active on their character and they will receive a game mail from that NPC later.

If they join a new faction, or even try to rejoin that faction and start another liege quest - when the time for the original quest ends, the player will receive the game mail for the failed loyalty quest from the most recent NPC to give them a quest.

This can result in either a false confusing game mail, outlaw status, or being removed from their current faction.

Expectation: Starting a quest then losing the faction that gave you the quest should immediately remove that quest from the players history; the player to receive the mail from the correct NPCs for appropriate quests.

To reproduce:

* In game using a faction level character (dm boost stats), go to any of the three lieges (Duke, Princess, Jonas...Necro - j/k ;) and say join.

* Go to another liege and say join 

* Do the quest for the 2nd liege 

* Allow the timer to expire for the first quest *

* Punishment - note the name of the NPC who sends you mail 

* Occurs for both Join and Loyalty Quest

(Note: you can Go into settings to manually lower the timer for the quest to speed this up and trigger the mail)

davidklaw:
This PR addresses an issue brought up in #465. When players are on a service quest, but lose faction, the service quest remains active. At best, the player will receive a late gmail about being booted from the faction. At worst, they'll be booted from their new faction, if they joined one in the meantime. This happens because the quest penalty calls `ResignFaction`. This PR seeks to solve this issue by immediately expiring the related service quest when a player is booted from a faction.

## Testing
Due to the complexity of the quest system, this can be challenging to test. I dialed down the quest engines `piQuestTimerDelay`, `piQuestDeadlineTimerDelay`, and `piQuestHistoryRecentTime` to speed up triggering the loyalty quest in quick succession, as I was jumping from one faction to another to ensure each factions quest stats were accurate.

### General testing steps
1. Join a faction
2. Trigger a loyalty quest
3. Immediately get yourself kicked out (e.g. attack a faction solder)
4. Confirm you've lost faction

When testing, it's most important to verify that the quest is properly expired. Faction quest slots are limited to the number of faction members, so it's important to check that the maximum number of available quests in the quest template adjust with membership (+1 when joining, -1 when resigning). Also, expiring this quest in this new way shouldn't simply delete it from history. This PR expires it quietly, allowing it to do all the things it would normally do without triggering the gmail. I've added a `report` flag to manage this.

I've extracted some of this functionality into new methods in `QuestEngine` and `QuestNode` thinking they might be helpful in the future. I looked up and down the codebase and could not find anything that handled this currently.

### Things to consider
- The faction service quest template ids are 8 (Princess), 9 (Duke), and 69 (Rebel).

### Thoughts
- Ideally, _all_ related faction quests would be failed if booted from a faction, but that felt like opening up a can of worms. As of today, only the 3 service quests have a faction loss penalty and cause this issue.
- The same issue was dealt with very differently in OpenMeridian [here](https://github.com/OpenMeridian/Meridian59/pull/479).

This closes #465


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -12226,6 +12235,30 @@ messages:
       return;
    }
 
+   ExpireQuestByTemplate(questTemplate = $)
```
davidklaw, line 12238: I'm thinking I'll refactor this into QuestEngine as `ExpireQuest(index=0, who=$)`

davidklaw:
Not all faction loyalty quests are structured the same. Jonas simply wants an item while the Duke might want you to go on a multi-step journey to collect taxes. And upon further inspection, the contents of the failed quest gmail that's sent in each case doesn't quite align with the situation.

I attack a faction soldier and Jonas gets on me about my "lack of action." If I'm on the first leg of the Duke's tax collection quest, he tells me "I could not wait any longer...". These responses only make sense in the context of someone who hasn't been kicked out. I think that quietly expiring the quest while still logging it to the player's quest history is the way to go and am going to pursue that.

davidklaw:
This is ready for review but I'm going to keep it in draft form while I continue to look for any bugs that might be lurking in the shadows of the quest system.

kod/object/passive/quest.kod
```diff
@@ -605,6 +605,11 @@ SetQuestNode( node_object = $ )
 	return;
 }
 
+GetQuestNode()
```
akirmse, line 608: I think the main question is this one---whether the quest should expose its quest node to the outside world.  My intuition is not to do this, as it makes even more complexity possible, as now anything outside the quest engine can start doing arbitrary things with quests, without the engine knowing.

The alternative would be to take whatever you're currently using this for (DeadlineExpired?) and move it into this file instead, as "ExpireQuest" or something like that.

davidklaw: Makes sense and confirms a suspicion of mine, thanks 👍🏻. I'll refactor.

davidklaw: I've addressed this in the latest commit.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -11213,6 +11213,10 @@ messages:
          {
             Debug("ResignFaction miscounted number of faction members!");
          }
+
+         % Prematurely expire the player's active PRINCESS service quest, if present, to avoid faction loss when it organically expires.
```
akirmse, line 11217: Is the faction loss the core issue, or is it just that it makes sense to cancel the quest if you leave the faction?

Also, the three code blocks for the various factions seem awfully duplicative.  Is it possible to combine them with a loop over the factions, with a few different constants?

kod/util/questengine.kod
```diff
@@ -7327,6 +7343,38 @@ CheckQuestDeadlines()
       return;
    }
 
+   ExpireQuests(index = 0, who = $, report = TRUE)
+   "Given a template, Locate and prematurely expire a player's active quests"
```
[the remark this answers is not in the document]
davidklaw: If you agree that `index` is the appropriate name, I'm thinking "Given a _template index_, locate..."

kod/util/questengine.kod
```diff
@@ -7168,6 +7168,22 @@ CheckQuestDeadlines()
       return;
    }
 
+   GetActiveQuests(index = 0)
```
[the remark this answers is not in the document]
akirmse: Ah, OK, that's fine.

kod/object/active/holder/nomoveon/battler/player.kod
davidklaw: I think that it's primarily the loss of faction. The penalty "prize" for these quests is faction loss regardless of the faction you started it in and the one you might currently be in. Are you thinking a comment modification is needed?

I'll look into consolidating the code here.

akirmse: Yeah, to me it's more like it doesn't make sense to have a quest from the faction when you've been booted.  Faction loss is a very annoying symptom but not the underlying explanation as to why the quest is being expired.

kod/util/questengine.kod
```diff
@@ -7327,6 +7343,38 @@ CheckQuestDeadlines()
...
+         return;
+      }
+
+      for oQuest in Send(self,@GetActiveQuests,#index=index)
+      {
+         for oQuester in Send(oQuest,@GetQuesters)
+         {
+            if oQuester = who
+            {
+               Post(oQuest,@ExpireQuest);
```
davidklaw, line 7371: ⚠️ This should be passing `report` through. Fixing in a follow-up commit.

davidklaw:
> I'm converting this to draft as I'm seeing this service timer fail and want to understand why.
> 
> > `[player.bof (10960)] DeleteTimer can't find timer ###`

It turns out this is pre-existing and I will submit a separate PR to address it.

davidklaw:
I've merged master and #944 into this and tested again to verify it's still ready for review.

davidklaw:
Just bringing awareness to this again as all feedback has been addressed.

PR #922, opened 2024-08-25
Sort objects before rendering for consistent world object ordering

sicoderealm:
As we receive objects in different orders from the BSP walk, when we come to render they may also be sorted by z-when at an identical location. This causes visual issue with items being sorted differently depending on the players position in a room - we now ensure objects are sorted by their ids to create a consistent view.

### DIFF
--- a/clientd3d/d3drender_objects.c
+++ b/clientd3d/d3drender_objects.c
@@ -1238,11 +1238,11 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
 		// If both items are of the DrawObjectType, compare their draw.id
 		if (a.type == DrawObjectType && b.type == DrawObjectType) {
 			// Compare draw.id from the object union member
+			// note: drawdata[] may contain more than one entry with the same id so this is not a stable sort.
 			return a.u.object.object->draw.id < b.u.object.object->draw.id;
 		}
 
 		// If the types are the same and not DrawObjectType, keep the original order.
-		// note: drawdata[] may contain more than one entry with the same id so this is not a stable sort.
 		return false;
 	});
 
--- a/clientd3d/d3drender_objects.c
+++ b/clientd3d/d3drender_objects.c
@@ -6,9 +6,11 @@
 //
 // Meridian is a registered trademark.
 #include "client.h"
+#include <algorithm>
+#include <chrono>
 #include <unordered_map>
 #include <unordered_set>
-#include <chrono>
+#include <vector>
 
 // Variables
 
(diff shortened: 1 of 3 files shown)


clientd3d/d3drender_objects.c
```diff
@@ -1225,7 +1227,27 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
 	// drawdata[] may contain more than one entry with the same id.
 	// We need to keep track of those we've already processed to avoid duplicates.
 	std::unordered_set<int> processedIds;
```
akirmse, line 1229: Move this down to be closer to the code that uses it.

clientd3d/d3drender_objects.c
```diff
@@ -1225,7 +1227,27 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
...
+			return false;
+		}
+		if (a.type == DrawObjectType && b.type != DrawObjectType) {
+			return true;
+		}
+
+		// If both items are of the DrawObjectType, compare their draw.id
+		if (a.type == DrawObjectType && b.type == DrawObjectType) {
+			// Compare draw.id from the object union member
+			return a.u.object.object->draw.id < b.u.object.object->draw.id;
```
akirmse, line 1244: Heads up about this comment just above:
drawdata[] may contain more than one entry with the same id.

So I think that means the sort is not stable when there are two entries with the same ID.

clientd3d/d3drender_objects.c
```diff
@@ -1273,6 +1295,9 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
 
 		angle = (pRNode->angle - intATan2(-dy, -dx)) & NUMDEGREES_MASK;
 
+		if (pRNode->obj.animate == NULL)
```
akirmse, line 1298: This doesn't seem related to this PR; remove?

clientd3d/drawbsp.h
```diff
@@ -99,4 +99,6 @@ void DrawMiniMap(room_type *room, Draw3DParams *params); /* in Draw3d.h */
 void EnterNewRoom3D(room_type *room);
 Bool GetRoomHeightRad(BSPnode *tree, long *ceiling, long *floor, int *flags, int x, int y, long r);
 
+#define MAX_ITEMS 2400L
```
akirmse, line 102: Not sure why this was moved; revert?

clientd3d/d3drender_objects.c
```diff
@@ -1225,7 +1227,27 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
 	// drawdata[] may contain more than one entry with the same id.
 	// We need to keep track of those we've already processed to avoid duplicates.
 	std::unordered_set<int> processedIds;
-	// base objects
+
+	// Sorting drawdata based on the draw.id field, with a condition to place non-DrawObjectType items last
+	std::sort(drawdata, drawdata + MAX_ITEMS, [](const DrawItem& a, const DrawItem& b) {
+		// Compare the type field directly, which is a BYTE
```
akirmse, line 1233: Is it relevant that it's a BYTE?

clientd3d/d3drender_objects.c
```diff
@@ -1225,7 +1227,27 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
 	// drawdata[] may contain more than one entry with the same id.
 	// We need to keep track of those we've already processed to avoid duplicates.
 	std::unordered_set<int> processedIds;
-	// base objects
+
+	// Sorting drawdata based on the draw.id field, with a condition to place non-DrawObjectType items last
```
akirmse, line 1231: A comment would be good here explaining *why* you're sorting.

clientd3d/d3drender_objects.c
```diff
@@ -1222,10 +1224,32 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
...
+		}
+
+		// If both items are of the DrawObjectType, compare their draw.id
+		if (a.type == DrawObjectType && b.type == DrawObjectType) {
+			// Compare draw.id from the object union member
+			return a.u.object.object->draw.id < b.u.object.object->draw.id;
+		}
+
+		// If the types are the same and not DrawObjectType, keep the original order.
+		// note: drawdata[] may contain more than one entry with the same id so this is not a stable sort.
```
akirmse, line 1245: This comment is in the wrong place.  The sort isn't stable because of the comparison on line 1241.  If there are two objects with the same draw ID, they are ordered nondeterministically.

Issue #428, opened 2022-12-08
Monster: Ice Creature With Shield and Spear are missing


jrotunda:
There are three versions of the Ice Creature, similar to the Thrasher (weapon heads), the Ice Creature is to spawn randomly have a Shield, Sword or Spear.

The art for these variants is in the bgf for the Ice Person and referenced in iceperson.kod

The Spear also has a throwing animation.

Currently only the Ice Creature with the Sword is spawning.

jrotunda:
Needs:

Add unused graphic resource to Iceperson.kod

davidklaw:
Looking into this but wanted to note that while I see the spear in hand frames and what I think are spear projectile frames, I see no shield or reference to a shield.

Issue #427, opened 2022-12-08
Monster: Ice Person has no treasure list


jrotunda:
The Ice Creature [ice person in the code] has no treasure list.

Currently when killed, it will vanish and leave nothing behind.

jrotunda:
I'll be working on this to address the needs for other unique loot arrangements and for firesand to drop on the Mainland for Faren mages working spells there. It will also give some need for players to still farm from lower level mobs and give economic power to new players.

jrotunda:
Addressed by #923

PR #923, opened 2024-08-25
Expanded functionality for the ice creature

davidklaw:
This PR address #428 by building out additional functionality ice creature based on existing assets and clues in the code. This PR does NOT make the ice creature game ready in the sense that it's ready to drop into a room. It has no treasure list, I haven't considered karma or difficulty, it's description is outdated, etc.

image

As it stands today, the ice creature has a single variant carrying a sword. This update leverages existing assets for a spear weapon and projectile to create a spear variant capable of melee and ranged attacks. It also includes a new death (`X`) asset that didn't previously exist.

**I'm posting this as a draft for both review and discussion about the ice creature's default settings.**
- What should it's difficulty be?
- What karma should it have?
- What should it's loot table be? (#427)
- It was setup with `ATCK_SPELL_COLD`, but how should it use it? Random attack chance?
- What should the spear range be?

## Asset work
The original assets were _tiny_, so I extracted, upscaled, and recompiled after making a number of bbg updates using `bbgun`. The death asset was created by simply pulling in a few shards of ice from other ice creature assets.

## Notes
- The attack animation and the angles of the weapons don't feel right, but they're seemingly the animations provided.

## Footage

https://github.com/user-attachments/assets/54f38218-cd8f-48ed-832a-0d00d4524a02

https://github.com/user-attachments/assets/5888ed77-33f1-4008-b01e-4f49562bb466

https://github.com/user-attachments/assets/607df3a9-2b2d-4a1d-bf32-c0bba1078e7d

https://github.com/user-attachments/assets/7b8a6978-c7b7-4369-965e-857f54b73e98


davidklaw:
Marking as ready for review. This isn't intended to necessarily be a game-ready iteration of the ice creature, but I think it is. There are other questions about it's treasure list and perhaps an ice attack of some sort that I don't address in this PR.

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
```diff
@@ -28,9 +62,11 @@ resources:
    iceperson_rightarm = iceperR.bgf
 
    iceperson_spear = icepersp.bgf
-   iceperson_sword = icperswd.bgf
+   iceperson_sword = iceperswd.bgf
```
akirmse, line 65: You're not limited to 8.3 filenames.  Want to rename at least the new files so that they're easier to understand?

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
```diff
@@ -64,19 +100,40 @@ properties:
 
    piAnimation = ANIM_NONE
    piDrawFX = DRAWFX_TRANSLUCENT_50
-   
+
+   prWeaponOverlay = iceperson_sword % sword as default weapon
```
akirmse, line 104: Add at least one more space between the code and comment to make it easier to read.

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
```diff
@@ -64,19 +100,40 @@ properties:
...
+   "Random chance to spawn a spear variant"
+   {
+      local iRandom;
+
+      iRandom = random(0,1);
+      if iRandom = 1
+      {
+         prWeaponOverlay = iceperson_spear;
+         piBehavior = AI_MOVE_OPTIMAL_RANGE;
+      }
```
akirmse, line 128: Should probably also have an else clause the sets the default, otherwise if you call SetWeapon more than once, the variant is no longer random (much more likely to be spear).

davidklaw: Oops. Thanks for catching this!

davidklaw:
Feedback has been addressed along with one uncaught bug:
- When viewed from the left, the weapon was not visible. This is because an incorrect hotspot value was used in the right arm bbg. This is fixed now.

Aesica:
Thinking about your bullet point list a bit:

- Karma:  Originally they had negative karma, making them yet another thing for positive karma players to hunt.  What the game could really use though is something with positive karma around that of skeletons/tusked skeletons.  So negative karma players have something to fight.  Would this infringe on any lore that exists for them?  I know their description says they're evil, but since they never truly existed in the game outside DM invasions before, does this count as established M59 lore?
- Difficulty:  The negative karma monster (positive player) track is pretty well filled-out, so any difficulty seems like it'd be fine if they're to have negative karma.  But for positive karma, around skeleton/tusked skeleton strength would fill a nice gap that currently exists.
- Drops:  Waterskin [100%]. :D  Okay that one leads to the reason I've put off revising certain problematic treasure tables on existing enemies--there's not enough stackable drops besides the basic reagents and food, so it might be advisable to create a few items unique to them, just as it would for other enemies.  I always thought it'd be cool if they had a low chance to drop their swords--longswords or shortswords that slowly melt away, automatically losing durability and being completely broken after about an hour or two.  But that's all for another PR I think.  When in doubt, waterskin makes a lot of sense, along with, I dunno, some gems or mushrooms like so many other things.
- ATCK_SPELL_COLD:  That's just a damage type for basic attacks, meaning it does cold damage with each swing.  One thing to be wary of though is any enemy with spell damage types in its basic attack can be hard-countered by the appropriate resist spell.  So Resist Cold will trivialize melee attacks from them as long as ATCK_SPELL_COLD is set.
- Spear range:  Avar warriors also use spears and have a slightly increased range.  I'd use them as an example.  I think their range is 3 but don't quote me on that.

davidklaw:
@sicoderealm Tagging you in the hopes that you can shed some light on the known issues I captured in the PR description.

davidklaw:
With the exception of the documented known issues, this PR is ready for review. I recognize that it's a hefty one and added some clips of what to expect when engaging ice creatures.

davidklaw:
I’ve updated the PR description to remove the issue related to client crashes, which we’ve been discussing on Discord. I was the only one able to reproduce a crash seemingly caused by having more than one ice creature, dark angel, or ghost present when firing. Interestingly, the crash disappeared when I upgraded from VSC22 17.4.4 to 17.12.8. Regardless, it seems unrelated to this PR and the PR updated as a result.

davidklaw:
I'm going to convert this to a draft while I improve the monster corpse artwork, but this is otherwise ready for review.

davidklaw:
Finally ready for a full review. 

This proved to be a tricky one, particularly around the need to throw a spear, appear empty-handed, then generate a new spear again later. I managed to do this with an extra timer. Hoping I handled it well🤞🏻. I also linked to a quick demo video on Google Drive that I'll be deleting if/when this PR is merged.

kod/include/blakston.khd
```diff
@@ -1591,6 +1591,7 @@
    TID_DARK_ANGEL    = 51
    TID_MUMMY         = 52
    TID_THRASHER      = 53
+   TID_ICE_PERSON    = 54
```
sicoderealm: I'm curious - why do you use ice person when its an ice creature?

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
```diff
@@ -8,29 +8,91 @@
...
 
    include blakston.khd
 
-resources:
+   ANIM_THROW = 3
 
-   iceperson_koc_name_rsc = "atzmoch"
-   iceperson_name_rsc = "ice creature"
-   iceperson_icon_rsc = iceperT.bgf
-   iceperson_desc_rsc = "This is a creature made completely from frozen water!"
```
sicoderealm, line 22: Can we keep this text somehow?

davidklaw:
Merge conflicts resolved.

kod/include/blakston.khd
sicoderealm: Same comments through - e.g. `IcePersonTreasure`

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
```diff
@@ -8,29 +8,91 @@
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+% Ice Creature Overview
```
sicoderealm, line 12: Just to loop back on the call out in this PR to say I did spend quite some time solving the local build issues you had with the crash - which is now solved - but I also spent a decent amount of time looking into the graphical related issues with transparency mention in this PR, I couldn't solve it and won't be investigating further in the near future. Personally, I don't think it looks too bad even with the invisible leg but perhaps worth fixing before merging.

kod/include/blakston.khd
davidklaw: IIRC, all of the original code references use “ice person” (for example `Iceperson`, `iceperson_rightarm`, etc.). The only place “creature” was used was in the UI. From a development standpoint, it felt best to keep the existing terminology intact behind the scenes.

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
davidklaw: Absolutely, I can get behind that. Let me revise.

kod/object/active/holder/nomoveon/battler/monster/iceper.kod
davidklaw: Thanks for your time on this 👏🏻

davidklaw: For the record, this transparency issue may have nothing to do with this PR. It's seen in the presence of slimes as well, at least from certain sides. A clue for later, perhaps.

Slime not rendering:
image

Slime rendering:
image

davidklaw:
@Meridian59 This is ready for a final look. The idea is to use this on live for play-testing, with admins being the only ones generating these for now.

akirmse:
Let's remove the .bgfs from the PR, since we don't want the artwork to be open source.  Send them to me separately and I'll make sure they get into the next update.  If you have the original source artwork, send me that too.

PR #924, opened 2024-08-25
Null pointer check for animation

sicoderealm:
null ptr observed during this code - adding a null pointer check


akirmse (commented):
Are you sure this is necessary?  I don't see any crashes here in bugsplat.

akirmse:
Doesn't seem necessary?  Reopen if needed.

PR #926, opened 2024-08-26
Make logoff ghost overlays visible

sicoderealm:
Logoff ghosts currently do not display overlays (such as masks or helms) - we add that functionality.

image

image

image

image

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -11878,7 +11878,7 @@ messages:
    % Facial overlays such as face masks and helmets are displayed by statues 
    % and logoff ghosts. Such overlays (helms/mummy masks) may may also require a 
    % players hair, nose and/or mouth to be hidden to look right.
-   GetFacialOverlayRsc()
+   GetFacialOverlay()
    {
       local overlay;
       
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -8488,6 +8488,11 @@ messages:
       return;
    }
 
+   GetOverlays()
+   {
+      return plOverlays;
+   }
+
    RemoveOverlayObject(what = $)
    "Removes <what> from plOverlays if it's in there."
    {
@@ -8504,6 +8509,11 @@ messages:
       }
 
       return;
+   }
+
+   GetHairRemoved()
+   {
+      return poHair_remove;
    }
 
    RemoveHair(what = $)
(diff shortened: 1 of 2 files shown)


kod/object/active/logghost.kod
```diff
@@ -569,6 +574,16 @@ messages:
          piOverlays = piOverlays + 1;
       }
 
+      for overlay in prOverlays
+      {
+         piOverlays = piOverlays + 1;
```
akirmse, line 579: There is a Length() function.

kod/object/active/logghost.kod
```diff
@@ -569,6 +574,16 @@ messages:
          piOverlays = piOverlays + 1;
       }
 
+      for overlay in prOverlays
```
akirmse, line 577: The prefix "pr" generally refers to a resource name, not a list.  I don't see where prOverlays is declared, kind of confused why this compiles.

kod/object/active/logghost.kod
```diff
@@ -620,6 +632,9 @@ messages:
       piShield_translation = Send(poGhostedPlayer,@GetShieldTranslation);
       oWeapon = Send(poGhostedPlayer,@LookUpPlayerWeapon);
 
+      plOverlays = Send(poGhostedPlayer,@GetOverlays);
```
akirmse, line 635: I'm on the fence about whether GetOverlays() is a good idea.  It exposes the internal overlays to anyone, breaking encapsulation and potentially leading to some code chaos if someone later comes along and starts modifying the list outside of Player.  I wonder if there are other options.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -8488,6 +8488,16 @@ messages:
       return;
    }
 
+   IsHairVisible()
+   {
+      if poHair_remove = $
```
akirmse, line 8493: return poHairRemove = $;

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -11870,6 +11880,21 @@ messages:
       return piAction;
    }
 
+   GetMaskOrHelmRsc()
```
akirmse, line 11883: Let's make this a little more generic, so if something like scarves are added in the future, someone won't have to go around renaming tons of stuff.

Also, add a comment as to the spirit of this function.  I'm not quite sure why the head overlay is special; the comment will enlighten me.

kod/object/active/logghost.kod
```diff
@@ -103,6 +103,8 @@ properties:
    prRight_arm = LogoffGhost_default_rarm
    prLeft_arm = LogoffGhost_default_larm
    prLegs = LogoffGhost_default_legs
+   pbIsHairVisible = $
+   prMaskOrHelm = $
```
akirmse, line 107: rename

PR #929, opened 2024-08-26
Seduce: Verify conditions before continuing enchantment

### LINKED ISSUE #461: Spell: Seduce mana drain doesn't end after the caster loses focus and takes mana to 0.
Seduce turns any mob within its parameters into a temporary minion of the caster. 

It doesn't require focus, but has a mana drain/time tick, the same as most Jala Spells and another Qor spell, Silence.

The caster is able to cast the spell on multiple targets at once - however when the spell wears off, the target hatred goes up and it aggros on the caster.

Currently, the spell has a bug where, even after the caster loses power over a target mob, the mana tick continues as if you are still actively seducing the mob, and will take your mana down to 0 unless you log off to stop the spell.

This can cause a debilitating and large stack of active mana drains at once if the user uses this on more than one mob, making it worthless in combat in its current state.

Expectation:

When the caster loses control of any 1 seduced mob, the mana drain ends for that mob.

Possible solution:

Remove the mana drain entirely, tweaking other aspects of the spell to make it fun to use before its official introduction.

Note: The "wand of charming", which casts seduce, has no mana drain/caster attached, and the spell functions as one would expect using the wand.

davidklaw:
This PR fixes #461 by verifying that the target monster is still alive during each iteration of Seduce's mana drain timer. Currently, if an affected mob is killed, `EndEnchantmentTimer` doesn't recognize this and the mana drain continues until the caster is out of mana. Additionally, it performs new checks to ensure the master (player) is present.

**About Seduce**: When cast on a monster, seduce will cause the monster to fight on your behalf for as long as your mana allows. If your mana runs out, the monster turns on you. Seduce can be cast on multiple monsters. If a monster is killed, the mana drain (from that monster) should stop.

### DIFF
--- a/kod/object/passive/spell/seduce.kod
+++ b/kod/object/passive/spell/seduce.kod
@@ -218,15 +218,24 @@ messages:
    
    EndEnchantment(who = $, report = TRUE, state = $)
    {
-      if send(state,@GetMana) >= (send(who,@getlevel) / 10)
+      % Verify the monster is still alive
+      if Send(who,@GetHealth) > 0
       {
-         Send(state,@LoseMana,#amount=send(who,@getlevel) / 10);
-         Post(who,@StartEnchantment,#what=self,#time=viDrainTime,#state=state);
-         return;
+         if Send(state,@GetMana) >= (Send(who,@getlevel) / 10)
+         {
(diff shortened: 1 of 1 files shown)


davidklaw:
Converting to a draft so that I can consider other scenarios like:
- The player moving to another room
- The player logging off
- The player dying

PR #931, opened 2024-08-27
Adds folly broadcast to Riija bridge deaths

### LINKED ISSUE #625: System: User: Unique ### Player Death Message from Invisible Bridge fall not displayed
When a player falls from the Invisible Bridge at the Temple of Riija, either by going out of the bounds of the bridge area, or by not using/holding their personalized Riija Sword, they'll audibly scream, fall, leave a random item, then wake up in the Ko'catan Inn. 

Along with the fall, there's a unique death message that is supposed to go out that currently isn't being sent to all.

Found in: 

system.kod

<html>
<body>

system_user_killed_bridge_faith = \
--
56 | "### %q just hurled %sself from the cliffs of Riija to find a watery "
57 | "grave."

</body>
</html>

Missing from:
[kod/object/active/holder/room/temprii.kod](https://github.com/Meridian59/Meridian59/blob/819e3ada61f978f738dd949d36b6767315bfc179/kod/object/active/holder/room/temprii.kod#L11)

Expectation:

When a player falls from the Riija Bridge, users should see this message to add appropriate shame, tension, etc.

davidklaw:
In response to #625, this PR adds a broadcast to all currently logged in users when a player unsuccessfully crosses the Riija bridge. It uses `GetTrueName`, so there's no hiding from the shaming.

> `### PlayerName just hurled hisself from the cliffs of Riija to find a watery grave.`

Closes #625

### DIFF
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -52,9 +52,6 @@ resources:
    system_killed_self = "### %q was just slain by %s own folly."
    system_user_killed_room = \
       "### %s have laid claim to yet another poor soul, %q."
-   system_user_killed_bridge_faith = \
-      "### %q just hurled %sself from the cliffs of Riija to find a watery "
-      "grave."
    system_user_killed_token = "### %q lost a token to %s%q."
    system_user_killed_shield = "### %q was bested by faction rival %s%q."
    system_user_killed_guild = \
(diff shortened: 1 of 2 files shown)


kod/util/system.kod
```diff
@@ -1507,16 +1504,8 @@ messages:
 
       if IsClass(killer,&Room)
       {
-         if IsClass(killer,&TempleRiija)
```
akirmse, line 1510: Do you know why this didn't work before?  Is there a danger that a temple death will now send two messages?

It would be slightly better to have all the death messages in one place.

davidklaw: The player isn't killed, so this code is never reached. The player is sent to the Ko'catan Inn and hit with the ALLRED effect to simulate death.

akirmse: ah, ok

Issue #491, opened 2023-05-19
Disciple Quests need better feedback and faster timing


adrienlaws:
### **The Problem**

- Several players have reported problems in getting a disciple quest completed. 
- Either they already had the spell and didn't know or the quest is not firing.  
- The issue is there's no real way to know the state of the discipline quest.  If they inquire about a spell it gives them the congratulations you may learn it, even if they already bought it, and the user may become confused.

In short, the notifications on what to do next are unclear and the length of time players have to wait is long.

**Anecdotes from Discord**
>she tells me I have to become her disciple if I say enfeeble..  and then she says Priestess Zuxana says, "I see promise in you, young lad, I am not ready to take you on as a disciple yet."

>What really sucks is that for the last 3 days ive had real life time to work these spells, and the game basically told me NO..   discouraging and frustrating at the same time.

>What is this junk?
Priestess Zuxana says, "I see promise in you, young lad, I am not ready to take you on as a disciple yet."
I already have lvl3 stuff from Maleval..

>Maleval says, "You are currently qualified to learn vampiric drain.  Congratulations."
Maleval says, "I'm afraid I have nothing to sell right now."
I've never seen this happen but Karma is well below requirements and we all agree I'm qualified but I still can't buy vampiric drain
Any help would be appreciated

>Yes - I already had it in my spell inventory it looks like
so embarrassing
I have no recollection of buying it

**Possible Solutions**
* A brief message corresponding to why you can't buy the spell (already have it, aren't qualified, haven't done the quest) could be of help in troubleshooting or user notification
* Somehow change the dynamics of disciple quests so they don't get "jammed"
* Have quests fire quicker or perhaps remove the limit on the number of players performing a disciple quest at once

CardiovascularMike:
Another odd thing that can happen - for example for the quest to buy Profane Resonance:   The priestess tells you to go say something and invoke qor to some NPC.  The NPC has absolutely no reaction at all, no matter what you do.  It all seems to strange.

You go back to the Qor priestess and say 'Done' and she congratulates you and lets you buy the spell.  In my opinion the NPC target should always react.

PR #937, opened 2024-08-31
Decreases time between quest retries

davidklaw:
This PR is another small change aimed at improving the questing experience by decreasing the default quest retry time from 4 hours to 1 hour. Note: It's currently 20 minutes on 101.

This aims to ease issues raised in #491


jrotunda:
Just to back up what was stated here already regarding the quest timing on 101: Yes, for the past few years the quest timer had been decreased so as not to make failures ruin momentum, but still give them enough weight to be unpleasant (the player already spent the time). It was mainly an issue for newer players, who have a hard time navigating, so these timers mainly hurt them. In case of a recreate, it would be good to get this hard-coded in to keep the quest timing consistent.

davidklaw:
Any concerns or possible controversy with this one, or maybe something else I'm not considering?

PR #939, opened 2024-09-01
Adds purge audio to purge weapon hits

### LINKED ISSUE #820: Sound: Weapon: Purge weapons missing audio / sound effect when casting the spell.
The special enchanted weapons that have spells attached to them and named after the developers, Zjiria, Q - will cast their spells and also make the sound effect when the weapon 'goes off' or 'procs' (as players say).

Zjiria casts hold, Q casts blind - the weapons also have a text message explaining the effect of the weapon, visible to target and wielder.

For the purger weapon, the "Psychochild" enchanted weapons don't make the expected spell sound effect for the Shal'ille level 5 spell Purge.

_for testing:
dm create itematt purger
dm create itematt blinder
dm create itematt paralyzer_

davidklaw:
This PR plays the purge spell audio cue on successful hits. It does so by going through `CastSpell`, as `waparal.kod` does.

### What the attacker sees:
```
Your weapon tears at the magic surrounding Test3.
Your axe slashes Test3.
```

### What the target sees:
```
Test1's weapon tears at the magic surrounding you!
The spirit of battle has left you.
Test1 slashes you with his axe.
```

In my testing, purge often failed to remove a single PE because of the randomness of it. However, I only wanted to address the audio and not get into the probability of its effectiveness.

Fixes #820

### DIFF
--- a/kod/object/passive/itematt/weapatt/wapurge.kod
+++ b/kod/object/passive/itematt/weapatt/wapurge.kod
@@ -49,7 +49,7 @@ properties:
   
    piValue_modifier = 200   %% modify the object's price by 100%
    piValue_power_modifier = 10
-   piEffect_percent = 15
+   piEffect_percent = 50
 
 messages:
 
@@ -71,28 +71,32 @@ messages:
    %%% Effect Functions
       
    ModifyDamage(damage = 0, target = $, wielder = $, lData= $)
-   "Wielder has a 1% chance of blinding target with any hit."
+   "Wielder has a percent chance of purging target with any hit."
    {
-      local oSpell;
+      local oSpell, oRoom;
 
       if random(1,100) < piEffect_percent
       {
-         oSpell = send(SYS,@FindSpellByNum,#num=SID_PURGE);
+         oSpell = Send(SYS,@FindSpellByNum,#num=SID_PURGE);
          
          %% Gotta have something to try to remove....
          if Send(target,@IsEnchanted)
          {
-            % Try to purge off random(25,75)% of the enchantments.
-            if send(oSpell,@DoPurge,#who=target,#iChance=random(25,75))
+            oRoom = Send(wielder,@GetOwner);
+
+            % Try to purge off random number of the enchantments
+            % Only tell our victim if they lost enchantments
+            if Send(oSpell,@DoPurge,#who=target,#iChance=random(25,75))
             {
-               % Only tell our victim if they lost enchantments.
-               send(target,@MsgSendUser,#message_rsc=purger_worked_target,
-                    #parm1=send(wielder,@GetCapDef),#parm2=send(wielder,@GetName));
+               Send(oSpell,@PlaySpellSound,#room_obj=oRoom,#what=wielder);
+
+               Send(target,@MsgSendUser,#message_rsc=purger_worked_target,
+                    #parm1=Send(wielder,@GetCapDef),#parm2=Send(wielder,@GetName));
             }
 
-            % Tell the wielder every time we fired so they think it's cool.
-            send(wielder,@MsgSendUser,#message_rsc=purger_worked_user,
-                 #parm1=send(target,@GetCapDef),#parm2=send(target,@GetName));
+            % Tell the wielder every time we fired so they think it's cool
+            Send(wielder,@MsgSendUser,#message_rsc=purger_worked_user,
+                 #parm1=Send(target,@GetCapDef),#parm2=Send(target,@GetName));
          }
       }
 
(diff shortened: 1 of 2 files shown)


kod/object/passive/itematt/weapatt/wapurge.kod
```diff
@@ -71,28 +71,32 @@ messages:
...
+            oRoom = Send(wielder,@GetOwner);
+
+            % Try to purge off random number of the enchantments
+            % Only tell our victim if they lost enchantments
+            if Send(oSpell,@DoPurge,#who=target,#iChance=random(25,75))
             {
-               % Only tell our victim if they lost enchantments.
-               send(target,@MsgSendUser,#message_rsc=purger_worked_target,
-                    #parm1=send(wielder,@GetCapDef),#parm2=send(wielder,@GetName));
+               Send(oSpell,@PlaySpellSound,#room_obj=oRoom,#what=wielder);
```
akirmse, line 91: Just checking that this isn't going to stack with existing sounds (hit, death, whatever) and play two at once.  ModifyDamage doesn't seem like the safest place to do this.

davidklaw: I was thinking about refactoring to use `CastSpell` like some of the other spell driven weapon attributes. This would tap into the existing audio cues but would also have it fall into quite a bit of other logic as well. I wonder if this is an issue for these other weapon attributes?

davidklaw:
This is ready for another look.

kod/object/passive/itematt/weapatt/wapurge.kod
```diff
@@ -71,29 +71,28 @@ messages:
...
-                 #parm1=send(target,@GetCapDef),#parm2=send(target,@GetName));
+            Send(target,@MsgSendUser,#message_rsc=purger_worked_target,
+               #parm1=Send(wielder,@GetCapDef),#parm2=Send(wielder,@GetName));
+
+            % Try to purge off random number of the enchantments
+            % Only tell our victim if they lost enchantments
+            Send(oSpell,@CastSpell,#who=self,#lTargets=[target],#iSpellPower=random(25,75),#bItemCast=TRUE);
          }
+
+         % Tell the wielder every time we fired so they think it's cool
```
akirmse, line 93: This looks like a change in behavior---the wielder now gets this message even if the target wasn't enchanted?

kod/object/passive/itematt/weapatt/wapurge.kod
```diff
@@ -71,29 +71,28 @@ messages:
...
-            {
-               % Only tell our victim if they lost enchantments.
-               send(target,@MsgSendUser,#message_rsc=purger_worked_target,
-                    #parm1=send(wielder,@GetCapDef),#parm2=send(wielder,@GetName));
-            }
-
-            % Tell the wielder every time we fired so they think it's cool.
-            send(wielder,@MsgSendUser,#message_rsc=purger_worked_user,
-                 #parm1=send(target,@GetCapDef),#parm2=send(target,@GetName));
+            Send(target,@MsgSendUser,#message_rsc=purger_worked_target,
```
akirmse, line 85: I don't know about this.  The target is now getting an additional, duplicative message, as you noted:

Test1's weapon tears at the magic surrounding you!
A bright, holy light is trying to burn away your beneficial enchantments.

kod/object/passive/itematt/weapatt/wapurge.kod
```diff
@@ -71,29 +71,28 @@ messages:
...
+               #parm1=Send(wielder,@GetCapDef),#parm2=Send(wielder,@GetName));
+
+            % Try to purge off random number of the enchantments
+            % Only tell our victim if they lost enchantments
+            Send(oSpell,@CastSpell,#who=self,#lTargets=[target],#iSpellPower=random(25,75),#bItemCast=TRUE);
```
akirmse, line 90: In addition to getting the audio cue, this also triggers a large amount of code in Spell::CastSpell() that wasn't run before.  Just checking that you've looked through it all, and it all looks safe.

kod/object/passive/itematt/weapatt/wapurge.kod
davidklaw: I felt like this could make or break this PR. If we aren't okay with the standard purge message coming through, this would appear to become a more complicated matter. Though with my recent discovery of bItemCast, perhaps a bit of logic in `Purge` around the display of this message wouldn't be all that intrusive?

kod/object/passive/itematt/weapatt/wapurge.kod
davidklaw: I feel confident that it's safe. Much of that code will be ignored because the `who` is the weapon attribute and  `bItemCast` is TRUE. The only one that stands out to me is the final call to record the cast of purge against the system statistics. I figured casts from a weapon attribute would be useful data.

kod/object/passive/itematt/weapatt/wapurge.kod
davidklaw: Resolved via new logic in `purge.kod`.

kod/object/passive/itematt/weapatt/wapurge.kod
davidklaw: Reverted.

PR #944, opened 2024-09-06
Fixes erroring faction service timer deletion

davidklaw:
When a player joins a faction, `FactionServiceTimer()` creates a timer callback to itself to determine faction status and if good, create another timer to do the same in the future. The problem seen in the logs, is that when control returns to this method, the reference to the timer remains but the timer has since been deleted. The result is a `DeleteTimer()` error.

I'm unable to pinpoint a point in the code where an inadvertent deletion of this timer exists and am kind of assuming that when the timer expires and the call to `FactionServiceTimer()` is made, the timer has been cleaned up and the reference via `ptFactionTimer` remains.

To fix this, I've removed any timer deletion code from `FactionServiceTimer()` and have given that responsibility to `FactionServiceRoutine()`.

### Testing
Testing this is made much easier by tuning `FACTION_RESIGN_TIME` and `FACTION_UPDATE_TIME` way down to create rapid faction service cycles.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -10972,6 +10972,8 @@ messages:
    {
       local iHistory;
 
+      ptFactionTimer = $;
+
       if piFaction = FACTION_NEUTRAL
       {
          return;
@@ -11175,7 +11177,6 @@ messages:
          factliege = Send(Send(SYS,@GetParliament),@GetLiege,#faction=piFaction);
       }
                           
-      ptFactionTimer = $;
       Send(Self,@FactionServiceRoutine);
       Send(Send(SYS,@GetParliament),@MoveToFaction,#who=self,#faction=piFaction);
       Send(Send(SYS,@GetParliament),@UpdateEffects,#who=self);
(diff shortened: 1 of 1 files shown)


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -11182,7 +11176,7 @@ messages:
       }
                           
       ptFactionTimer = $;
```
akirmse, line 11178: This line looks the potential source of problems.  It doesn't check whether there's currently a faction timer going, so if there is, it becomes orphaned, and then there are two faction timers (or more) running at the same time, due to the next line, which winds up creating a new timer.  Probably the right thing to do here is to DeleteTimer if appropriate here first.

Not sure how/if that affects this PR.

davidklaw: I think this is okay, because either the player is joining a faction for the first time or joining a faction after being resigned from another, which would perform a timer check and delete.

I did try adding the same check here as you suggested, but it generates a DeleteTimer() error.

davidklaw:
I've made a small tweak and tested a variety of scenarios to verify timers get cleaned up and the log is freed of errors. I tested this by joining factions and remaining in them until I'm kicked, joining factions and getting kicked out by attacking soldiers, joining factions and moving to other factions via a token, logged off and back on, and just generally hammering a way at this. Additionally, I set up a debug statement upon faction timer creation and when each new timer is created have been able to confirm that the previous timer(s) no longer exist. I'm also confident that the system is still operating as expected with just one less error showing up in the log.

PR #945, opened 2024-09-09
Adds `bItemCast=TRUE` flag to Hold weapatt spell cast

davidklaw:
This further avoids traditional spell cast logic by flagging this weapatt's spell cast as an item cast via `bItemCast`. Doing so avoids a few bits of traditional spell casting code in Spell's `CastSpell()`. 

In Discord, I had suggested that without this flag the player might be improving off a weapon attribute. However, an instance of `self` (the weapon attribute) is being passed to `CastSpell()`, standing in as the caster, and causes `ImproveAbility()` to return immediately.

### DIFF
--- a/kod/object/passive/itematt/weapatt/waparal.kod
+++ b/kod/object/passive/itematt/weapatt/waparal.kod
@@ -61,14 +61,14 @@ messages:
 
       if random(1,100) <= piEffect_percent
       {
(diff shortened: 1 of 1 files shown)


akirmse (commented):
In the future please try to avoid mixing capitalization / syntax changes with behavior changes, as it makes it a lot harder to see what the behavior change is.  I'm happy to approve pure "cleanup" syntax changes.

PR #946, opened 2024-09-09
Revises map marker sizing and color

davidklaw:
This PR is meant to address some feedback heard on #809. Currently, other player's markers (blue dots) can sometimes be tiny in comparison to the player's marker and difficult to spot. This PR increases the size of them by introducing new pen so that _the_ player and other players can be styled differently. To help identity yourself against others, this PR also updates the player's marker to cyan.

I've also heard some concerns about situations where many players are in close proximity to one another. While there are situations and zoom levels where players dot will mash together, it would seem uncommon for those dots to remain static.

I created some representative examples of the current state vs what's proposed in this PR by screenshotting, copying, and pasting player markers. If there are more situations folks would like to see just drop a reply.

image

### DIFF
--- a/clientd3d/map.c
+++ b/clientd3d/map.c
@@ -20,12 +20,14 @@
 #include "client.h"
 
 #define MAP_WALL_THICKNESS 1
-#define MAP_PLAYER_THICKNESS 2
+#define MAP_SELF_THICKNESS 2
+#define MAP_PLAYER_THICKNESS 5
 #define MAP_OBJECT_THICKNESS 3
 
 #define MAP_PLAYER_MARKER_SIZE 3
 
 #define MAP_WALL_COLOR          PALETTERGB(0, 0, 0)
+#define MAP_SELF_COLOR          PALETTERGB(0, 255, 255)
 #define MAP_PLAYER_COLOR        PALETTERGB(0, 0, 255)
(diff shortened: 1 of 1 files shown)


akirmse:
I like the size change but I imagine the color change might be controversial for old-timers.

davidklaw:
I've reverted the player color back to the original blue, so the "Updated state" screenshot above is out of sync in that respect. If this is merged and it's desirable, I'll follow this up with a PR to allow these colors to be changed via the Colors menu:

- Self
- Players
- Guildmates
- Guild Allies
- Guild Enemies

akirmse:
I don't think these colors necessarily need to be settable, unless there's an outcry all of a sudden.

Aesica:
From a player's perspective, it'd be very handy to have separate colors for these things, even if guild wars are extremely rare these days. Especially guildmates and guild enemies.

davidklaw:
> From a player's perspective, it'd be very handy to have separate colors for these things, even if guild wars are extremely rare these days. Especially guildmates and guild enemies.

I've got a branch handy with it mostly complete. Happy to finish it up sometime in the future if it's worthwhile.

Meridian59:
Sounds fine, didn't realize it would be so handy.  The translations will be
a hassle.

PR #947, opened 2024-09-13
Fixes territory game recreate by skipping post-processing on flag claims

### LINKED ISSUE #29: Make RecreateAll preserve factions, shrines

davidklaw:
This PR modifies the `SetFaction` function to only change the flag's faction without performing post-processing steps. Specifically, the flag claim process and other related actions will no longer occur when only the faction is updated. This is needed to properly recreate the territory game and specifically, the full version of the territory game.

## Testing
* I manually converted flags to another faction when testing. Using `send o <TerritoryGame> GetFlagCounts` to verify the counts were the expected values.
* Tested existing functionality to ensure no side effects or unintended behaviors were introduced when recreating the territory game and when performing a system `RecreateAll`.

## Notes:
* This has been attempted and discussed before in #450 
* This attempts to address the issue with the least change possible, so it doesn't address some [desires in the original issue](https://github.com/Meridian59/Meridian59/issues/29#issuecomment-1478850125). I can address those in this or we can leave that for a follow-up.

Fixes #29 (at least what I understand to be the remaining issue)

### DIFF
--- a/kod/object/active/flag.kod
+++ b/kod/object/active/flag.kod
@@ -695,8 +695,8 @@ messages:
       return poClaimer;
    }
 
-   SetFaction(faction=FACTION_NEUTRAL,processClaim=TRUE)
-   "Sets the flag's faction and process the claim through the territory game."
+   SetFaction(faction=FACTION_NEUTRAL,processClaim=FALSE)
+   "Updates the flag's faction. By default, this updates the faction only, which is needed when the territory game is recreated. Pass `processClaim=TRUE` to notify the territory game of the change."
    {
       piFaction = faction;
       if poOwner <> $
--- a/kod/object/active/flag.kod
+++ b/kod/object/active/flag.kod
@@ -695,8 +695,8 @@ messages:
       return poClaimer;
    }
 
-   SetFaction(faction=FACTION_NEUTRAL,processClaim=FALSE)
-   "Updates the flag's faction. If just updating the flag's ownership and nothing else--such as during a recreate--leave the `processClaim` flag as `FALSE`. However, if changing the flag as part of an action in an ongoing territory game, set `processClaim=TRUE` so that the territory game can recount flags and persist state changes."
+   SetFaction(faction=FACTION_NEUTRAL)
+   "Admins use `AdminSetFaction` when forcing a faction change to ensure the territory game updates flag counts and persists state changes."
    {
       piFaction = faction;
       if poOwner <> $
@@ -707,11 +707,6 @@ messages:
          {
             ptGenerateTroops = createtimer(self,@GenerateTroops,piGen_time);
          }
-      }
-
-      if (send(SYS,@GetTerritoryGame) <> $) AND processClaim
-      {
-         send(send(SYS,@GetTerritoryGame),@FlagClaimed,#flag=self);
       }
 
       return;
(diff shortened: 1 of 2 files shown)


kod/object/active/flag.kod
```diff
@@ -695,8 +695,8 @@ messages:
       return poClaimer;
    }
 
-   SetFaction(faction=FACTION_NEUTRAL)
-   "for admin use"
+   SetFaction(faction=FACTION_NEUTRAL,processClaim=TRUE)
+   "Sets the flag's faction and process the claim through the territory game."
```
akirmse, line 699: Explain why the processClaim parameter is there.

kod/util/factgame/territry.kod
```diff
@@ -564,7 +564,7 @@ messages:
                oFlagpole = send(oRoom,@FindHoldingActive,#class=&FlagPole);
                if send(oFlagPole,@GetFaction) <> (iFaction-1)
                {
-                  send(oFlagPole,@SetFaction,#faction=(iFaction-1));
+                  send(oFlagPole,@SetFaction,#faction=(iFaction-1),#processClaim=TRUE);
```
akirmse, line 567: Deserves a comment since it's the one place where processClaim is set to TRUE.

davidklaw:
Updated per feedback. I'm struggled a bit with my comments and how they might express what processing a claim means today and into the future. I kept it general.

kod/object/active/flag.kod
```diff
@@ -695,8 +695,8 @@ messages:
       return poClaimer;
    }
 
-   SetFaction(faction=FACTION_NEUTRAL)
-   "for admin use"
+   SetFaction(faction=FACTION_NEUTRAL,processClaim=FALSE)
+   "Updates the flag's faction. By default, this updates the faction only, which is needed when the territory game is recreated. Pass `processClaim=TRUE` to notify the territory game of the change."
```
akirmse, line 699: I mean, I literally don't know how the territory game works, so try to explain it to me like I'm 5.  If I am writing new code and I want to set a flag's faction, all of a sudden I see this mysterious extra parameter and a reference to recreating the territory game, which I thought I didn't care about.  I am instantly confused as to what value I should pass.  A perfect comment would make it obvious to me what I should do.

Words like "process" are like "manage" --- they convey almost no information.

akirmse:
Now that I see and understand the comment, what about this option: remove the new parameter entirely, and move the logic that calls FlagClaimed to TerritoryGame itself (where SetFaction is called with processClaim=TRUE)?  That would be even less of a change, and I think it's unlikely that anyone else would ever use this parameter otherwise.

davidklaw:
> Now that I see and understand the comment, what about this option: remove the new parameter entirely, and move the logic that calls FlagClaimed to TerritoryGame itself (where SetFaction is called with processClaim=TRUE)? That would be even less of a change, and I think it's unlikely that anyone else would ever use this parameter otherwise.

I like it. I'll do that.

davidklaw:
> > Now that I see and understand the comment, what about this option: remove the new parameter entirely, and move the logic that calls FlagClaimed to TerritoryGame itself (where SetFaction is called with processClaim=TRUE)? That would be even less of a change, and I think it's unlikely that anyone else would ever use this parameter otherwise.
> 
> I like it. I'll do that.

Something that is lost by doing this is the ability for an admin to change the faction of flag. They can, but it won't persist unless they follow up `SetFaction` with the now moved `FlagClaimed`. I don't anticipate an admin needing to do this, but wanted to point that out.

Meridian59:
If that's a concern, you could make a new AdminSetFaction, and put a
comment why admins shouldn't call SetFaction directly.

kod/util/factgame/territry.kod
```diff
@@ -564,7 +564,13 @@ messages:
                oFlagpole = send(oRoom,@FindHoldingActive,#class=&FlagPole);
                if send(oFlagPole,@GetFaction) <> (iFaction-1)
                {
-                  send(oFlagPole,@SetFaction,#faction=(iFaction-1));
+                  Send(oFlagPole,@SetFaction,#faction=(iFaction-1));
```
akirmse, line 567: Call AdminSetFaction instead of duplicating the code?

davidklaw: Calling a function labeled for Admin use within the standard context of the territory game code seemed dangerous to me. I had a tiny Andrew sitting on my shoulder whispering, "What if someone updates AdminSetFaction to perform additional admin code?"

PR #948, opened 2024-09-13
Deconstruction small cleanup

sicoderealm:
Small clean/tidy up of particles, removing unused/multiple definitions


clientd3d/d3drender.c
```diff
@@ -185,6 +183,8 @@ void					D3DPostOverlayEffects(d3d_render_pool_new *pPool);
 
 void					*D3DRenderMalloc(unsigned int bytes);
 
+void					D3DRenderParticles();
```
akirmse, line 186: Can be static?

PR #949, opened 2024-09-13
Update to innkeeper room cost command

Aesica:
This PR adds the following improvements to the "room cost" command and its functionality:

- Adds an underline to the various room rental keywords to make them pop better.
- Adds text for the newer commands: "secure" keyword for copying keys, as well as the "change lock" command.
- Added an alternative command to reach the "room cost" dialog:  "room help" which seems like something a player is more likely to say when they want to know the commands and such.

image

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqinnk.kod
@@ -35,9 +35,11 @@ resources:
       "our rooms to be most well-appointed.  Enjoyest thine stay!"
 
    BarloqueInnkeeper_cost = \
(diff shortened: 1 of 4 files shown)


akirmse:
Is this underlining done everywhere an NPC is listing keywords?  It would look out of place if it only happened here.

Aesica:
I'm not sure, actually.  Italic alone seems insufficient though, since I've seen italics also used by NPCs to put pronunciation emphasis on certain words (See Skivlat and his infamous _guh_).  If it's a no-go, I'll remove it, as the main purpose was to add the new keywords, make the dialog easier to stumble upon (room help|cost), and add at least some emphasis to keywords which previously had none, like "copy key."

Meridian59:
Yes, if we can't make it consistent, let's remove the underlining.

Aesica:
Underlines removed.  I definitely see the point now about how one part of the game doing it while others don't would appear messy.

PR #953, opened 2024-09-14
Adds looping water sound to stream in Off the beaten path (`f7`)

### LINKED ISSUE #621: Room - Sound - Off the Beaten Path [OutdoorsF7] Missing Audio for running water
As you go around Meridian, water that is animating has a 'source' location for sound that makes the water sound as if it's running against land surfaces at various speeds. 

In Off the Beaten Path, the water for the river in the norther part of the room doesn't have any audio.

Expectation:
The running water for this river should be heard as well as seen, with a rushing water sound effect similar to other rivers and creeks.

M59 - Issue - Off Beaten Path - RID F7 - Texture - Animation - River Water - Map View

davidklaw:
This PR adds a looping sound to the room with a source location of the northern stream.

**Note:** I'm unable to test this locally as the fallback WaveMix audio driver doesn't support radius and volume.

Fixes #621

### DIFF
--- a/kod/object/active/holder/room/monsroom/f7.kod
+++ b/kod/object/active/holder/room/monsroom/f7.kod
@@ -19,6 +19,7 @@ resources:
    room_OutdoorsF7 = f7.roo
    room_name_OutdoorsF7 ="Off the beaten path"
    OutdoorsF7_music = walk1.mid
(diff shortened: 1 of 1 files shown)


akirmse:
Would require recreating the room, of course.

PR #954, opened 2024-09-16
Remove hidden admin imp audio cue

### LINKED ISSUE #819: Sound: Improvement sound effect plays for Hidden Administrator or DM.
The "You have improved in the art of..." message now occurs with a soft audio cue that can be heard by the user and any other player in the room (from this repo).

When a DM or Administrator is hidden, the spells sound effects were also removed from being heard a few years ago (also in this repo).

When a hidden DM or Admin casts a spell that gains them an improvement, the audio cue for the improvement message can be heard as if they weren't hidden. 

It's expected that this sound effect be muted for a DM or Admin when Hidden.

davidklaw:
This PR adds logic to skip skill and spell improvement room audio cues (added some time ago) when the improving player is a hidden admin.

This is in response to and fixes #819

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -7162,8 +7162,12 @@ messages:
       {
          oRoom = Send(self, @GetOwner);
          Post(self,@MsgSendUser,#message_rsc=player_improved,#parm1=skillName);
-         Send(oRoom, @SomethingWaveRoom, #what=self, 
-              #wave_rsc=player_improved_wav_rsc);
+
+         if NOT (isClass(self,&Admin) AND Send(self,@IsHidden))
+         {
+            Send(oRoom, @SomethingWaveRoom, #what=self, 
+               #wave_rsc=player_improved_wav_rsc);
+         }
       }
       
       if iChange > 0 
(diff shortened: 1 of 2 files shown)


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -6588,9 +6588,13 @@ messages:
       if report AND iChange > 0
       {
          oRoom = Send(self, @GetOwner);
-         Post(self,@MsgSendUser,#message_rsc=player_improved,#parm1=spellName);
-         Send(oRoom, @SomethingWaveRoom, #what=self, 
-              #wave_rsc=player_improved_wav_rsc);
+         Post(self,@MsgSendUser,#message_rsc=player_improved,#parm1=spellName);   
+         
+         if NOT (isClass(self,&Admin) AND Send(self,@IsHidden))
```
akirmse, line 6593: The bug says "DM or Admin", so change this to &DM instead?

davidklaw: This was written to match the related logic, which returns from `Spell`:`CastSpell` when Admin and hidden. Updating that as well as a part of this.

kod/object/passive/spell.kod
```diff
@@ -1409,8 +1409,7 @@ messages:
 
       if oRoom <> $
       {
-         if isClass(who,&Admin)
-            AND Send(who, @IsHidden)
+         if isClass(who,&DM) AND Send(who,@IsHidden)
          {
              if oTarget = $
              {
```
akirmse, line 1415: Should update these debug statements now, I suppose.

PR #956, opened 2024-09-20
Remove ranged weapons' impact on berserker ring durability 

davidklaw:
This PR is an alternative to #658. Currently, when ranged weapons are used in tandem with a berserker ring, the ring's durability will decreases even though it's damage buff does not apply to ranged weapons. This will remove the invisible effect of ranged weapons on the ring's durability. It _also_ replaces some magic numbers. I can revert if preferred.

### DIFF
--- a/kod/object/item/passitem/ring/besring.kod
+++ b/kod/object/item/passitem/ring/besring.kod
@@ -110,8 +110,6 @@ messages:
             Send(poOwner,@MsgSendUser,#message_rsc=berserkerring_break_rsc); 
             Send(poOwner,@TryUnuseItem,#what=self);
             Send(self,@Delete);
-
-            return hit_roll;
          }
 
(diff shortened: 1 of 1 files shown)


kod/object/item/passitem/ring/besring.kod
```diff
@@ -90,17 +97,27 @@ messages:
...
-         Send(self,@Delete);
+         piHits = piHits - 1;
+
+         if piHits <= 0
+         {
+            Send(poOwner,@MsgSendUser,#message_rsc=berserkerring_break_rsc); 
+            Send(poOwner,@TryUnuseItem,#what=self);
+            Send(self,@Delete);
+
+            return hit_roll;
```
akirmse, line 114: This (existing) line looks like a bug.  There was one charge left on the ring, so it should have an effect.

PR #958, opened 2024-09-22
Adjust alpha test threshold to prevent unwanted blending

sicoderealm:
We have some unwanted blending taking place between static world geometry and the skybox. This issue is most obvious at certain times of the day when the sky is bright on screens such as the outskirts of Barloque or the Flatlands.

Before:
image
image

After:
image
image

### DIFF
--- a/clientd3d/d3drender_world.c
+++ b/clientd3d/d3drender_world.c
@@ -40,11 +40,10 @@ extern int gNumObjects;
 extern Vector3D sun_vect;
 extern D3DMATRIX view;
 
-// The alpha_test_threshold value of 200 was chosen based on tuning through experimentation.
-// This value provided the best visual results for our specific textures and scenes, helping to 
-// prevent unwanted alpha blending between static geometry and the skybox, as well as eliminating 
-// lines that appeared between doorways. It balances visual quality and performance while minimizing
-// transparency issues.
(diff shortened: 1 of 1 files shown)


akirmse:
Is this new after the recent renderer changes since the last update, or has it always been this way?  I'm concerned about regressions that we haven't found.

clientd3d/d3drender_world.c
```diff
@@ -40,7 +40,7 @@ extern int gNumObjects;
 extern Vector3D sun_vect;
 extern D3DMATRIX view;
 
-static const auto alpha_test_threshold = 128; // Threshold for alpha testing to determine transparency
```
akirmse, line 43: Can you add a comment here explaining why 200 is the right number?  It seems super arbitrary.

clientd3d/d3drender_world.c
```diff
@@ -40,10 +40,15 @@ extern int gNumObjects;
...
-extern Bool gWireframe;
+// The alpha_test_threshold value of 200 was chosen based on tuning through experimentation.
```
akirmse, line 43: I'm still confused by this.  An alpha threshold doesn't seem like a tuning parameter---I would think that one value is correct, and all other values are wrong.  What does this value actually do?

PR #959, opened 2024-09-23
Adds additional Identify functionality around hidden attributes

### LINKED ISSUE #702: Identify spell inconsistently reveals true item descriptions
# What

Identify will reveal the true name of an item but not always it's true description.  
- Attributes seem to be revealed (like a blind scimitar) 
- But wands in particular seem to not show their true description when identified.  
- Identify should show **both** the true name and the true description in the pop-up object description dialog box.

# Examples 
Items where this is reproduceable
- ring of lethargy
- wands

## Screenshots
Wand of forgetfulness identified
image

Wand of forgetfulness revealed
image

davidklaw:
This PR addresses issue #702, where Identify wasn’t revealing item name and description as expected. To address this, I added code to Identify that will temporarily reveal an items hidden attributes, then hide them once again. I took this a step further by also revealing and hidden an items hidden color, if it has one. Examples of this include many of the magical rings.

With this fix, Identify now works as intended _and_ will (temporarily) reveal an item's hidden color.

1. Screenshot of an item with hidden attributes, including color:
image

2. The same item when Identify is cast upon it, revealing it's true name, true description, and hidden color, but only temporarily in the dialog:
image

3. If the user inspects the item again, any attributes that were hidden will be hidden once again:
image

4. And if a user casts Reveal, these changes become permanent:
image

This fixes #702

### DIFF
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -1255,9 +1255,19 @@ messages:
 
    HideHiddenAttributes()
    {
-      % TODO: Reverse RevealHiddenAttributes()
+      local i, bDone;
 
-      return;
+      bDone = FALSE;
+      for i in plItem_Attributes
+      {
+         if NOT Send(self,@GetIDStatusFromCompound,#compound=First(i))
+         {
+            setNth(i,1,(First(i)-1));
+            bDone = TRUE;
+         }
+      }
+      
+      return bDone;
    }
 
    AddAttributeSpecifics(litemAtt=$)
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -1253,6 +1253,13 @@ messages:
       return bDone;
    }
 
+   HideHiddenAttributes()
+   {
+      % TODO: Reverse RevealHiddenAttributes()
+
+      return;
+   }
+
    AddAttributeSpecifics(litemAtt=$)
    "Don't call this directly!  Call this from itematt.kod!"
    {
@@ -1792,6 +1799,19 @@ messages:
          AND Send(self,@GetPaletteTranslation) = (viUnrevealedColor & ITEM_PALETTE_MASK)
       {
          Send(self,@SetPaletteTranslation,#translation=viRevealedColor);
+         return TRUE;
+      }
+      return FALSE;
+   }
+
+   HideHiddenColor()
+   "Hides the item's true color."
+   {
+      if viUnrevealedColor <> 0
+         AND viRevealedColor <> 0
+         AND Send(self,@GetPaletteTranslation) = (viRevealedColor & ITEM_PALETTE_MASK)
+      {
+         Send(self,@SetPaletteTranslation,#translation=viUnrevealedColor);
          return TRUE;
       }
       return FALSE;
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -1237,6 +1237,7 @@ messages:
    }
 
    RevealHiddenAttributes()
+   "Reveals the item's attributes, returning TRUE if any were hidden."
    {
       local i, bDone;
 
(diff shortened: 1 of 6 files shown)


davidklaw:
My proposed implementation is complete and this is ready for review.

davidklaw:
@adrienlaws Tagging you since you're familiar with this area of the code.

jrotunda:
Must admit, this issue is so minor it was hard to see the problem even when broken down, but do understand it now and it's a good catch for consistency.
For anyone else like me: Reveal permanently removes a shroud, Identify only temporarily - but when showing the temporary item, it was still missing some information, including the color of the revealed ring. 

With the fix, Identify is working as it should. This should also now work with all other items using the Shroud/Hidden Item Atttribute.

davidklaw:
I’ve updated the description to better explain what was changed and how. Let me know if anything needs more clarity or if we should scale this back to just fix the original issue. I added the color reveal since it felt like a related improvement, but I’m open to dialing that back if needed.

davidklaw:
@Meridian59 Any concerns with this one? If needed, I can split this into two parts:

1. Fixing #702
2. Enhancing with the addition of a temporary color reveal

Let me know how you’d prefer to proceed.

kod/object/item.kod
```diff
@@ -1253,6 +1253,23 @@ messages:
...
+      for i in plItem_Attributes
+      {
+         if NOT Send(self,@GetIDStatusFromCompound,#compound=First(i))
+         {
+            setNth(i,1,(First(i)-1));
+            bDone = TRUE;
+         }
+      }
+      
+      return bDone;
```
akirmse, line 1270: This return value is not documented (neither is the function), and appears to have no effect.

kod/object/item.kod
```diff
@@ -1797,5 +1814,18 @@ messages:
...
+   "Hides the item's true color."
+   {
+      if viUnrevealedColor <> 0
+         AND viRevealedColor <> 0
+         AND Send(self,@GetPaletteTranslation) = (viRevealedColor & ITEM_PALETTE_MASK)
+      {
+         Send(self,@SetPaletteTranslation,#translation=viUnrevealedColor);
+         return TRUE;
+      }
+      return FALSE;
```
akirmse, line 1827: Return value has no effect?

kod/object/item.kod
davidklaw: Not that this is valid documentation (and I’ll remove it), but this function serves the opposite purpose of the `RevealHiddenAttributes` function just above it. I hadn’t considered that the return value might not be necessary here. Thinking it through, the return value seems unnecessary for this function’s purpose and appears to only be passed by the reveal function to handle a few unrelated tasks.

davidklaw:
I've addressed the two comments, but also refactored `Identify::CastSpell` for greater clarity.

kod/object/passive/spell/reveal.kod
```diff
@@ -119,10 +119,9 @@ messages:
       %% on something with no hope of revealing anything.
       piAdvance = send(oTarget,@RevealHiddenAttributes);
       
-      % Show newly revealed color
+      % Show newly revealed attributes
       if piAdvance
       {      
-         Send(oTarget,@RevealHiddenColor);
```
davidklaw, line 125: Note: This is removed because my PR calls `RevealHiddenColor()` from `RevealHiddenAttributes()`.

kod/object/item/passitem/ring/lethring.kod
```diff
@@ -118,7 +118,6 @@ messages:
       % at this point, it's pretty obvious it's not helping them.
       if Send(self, @RevealHiddenAttributes)
       {
-         Send(self, @RevealHiddenColor);
```
davidklaw, line 121: Note: This is removed because my PR calls RevealHiddenColor() from RevealHiddenAttributes().

PR #963, opened 2024-09-25
Allows for Defile to be cast upon the corpses of good monsters

davidklaw:
This PR makes it possible to Defile the corpses of good monsters, in addition to good player corpses. This is in response to discussion in Discord arguing that it's a very challenging spell to improve without the use of mules.

It borrows a player-based flag in `body` to make this possible and sets this `pbWas_good_player` flag to true when generating the corpse in `monster`. On submission, I realized that corpses are completely ephemeral and we're unlikely to have any problems if I rename `pbWas_good_player` to `pbWas_good_battler` along with this change. I can make that change if my thinking is valid.

As far as gameplay balance is concerned, should this come at a higher reagent cost if we're increasing the target pool?

### DIFF
--- a/kod/object/passive/body.kod
+++ b/kod/object/passive/body.kod
@@ -73,7 +73,7 @@ messages:
       pbMob = mob;
       piDrawfx = drawfx;
       prPlayer_name = playername;
-      pbWas_good_player = good; 
+      pbWas_good_player = good;           % Used to track monster karma as well
       pbAssassinated = assassinated;
       prPlayerBodyOverlay = PlayerBodyOverlay;
       piBodyTrans = BodyTranslation;
@@ -184,7 +184,7 @@ messages:
       return;
    }
 
-   WasGoodPlayer()
+   WasGood()
    {
       return pbWas_good_player;
    }
@@ -196,11 +196,6 @@ messages:
 
    GetObjectFlags()
    {
-      if pbMob
-      {
-         return LOOK_NO | piDrawfx;
-      }
-      
       return viObject_flags | piDrawfx;
    }
    
--- a/kod/object/passive/spell/defile.kod
+++ b/kod/object/passive/spell/defile.kod
@@ -105,7 +105,7 @@ messages:
 	      return FALSE;
       }
 
-      if NOT Send(target,@WasGoodPlayer)
+      if NOT Send(target,@WasGood)
       {
 	      Send(who,@MsgSendUser,#message_rsc=defile_not_good);           
 
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -3062,7 +3062,8 @@ messages:
       {
          oBody = Create(&DeadBody,#icon=vrDead_icon,#name=vrDead_name,
                         #mob=TRUE,#playername=Send(killer,@GetTrueName),
-                        #monstername=vrName,#drawfx=viDead_drawfx);       
+                        #good=(Send(self,@GetKarma) > 0),#monstername=vrName,
+                        #drawfx=viDead_drawfx);       
       }
 
       return oBody;                 


kod/object/passive/body.kod
```diff
@@ -73,7 +73,7 @@ messages:
       pbMob = mob;
       piDrawfx = drawfx;
       prPlayer_name = playername;
-      pbWas_good_player = good; 
+      pbWas_good_player = good;           % Used to track monster karma as well
```
akirmse, line 76: maybe "pbHad_good_karma" would be clearer

kod/object/passive/body.kod
```diff
@@ -196,11 +196,6 @@ messages:
 
    GetObjectFlags()
    {
-      if pbMob
```
akirmse, line 199: Was this removed intentionally?  It seems important and its removal could have a lot of side effects.

davidklaw: It was, as with it the monster corpse is not targetable.

And yes, there are side effects I didn't catch. Notably, animate can now be cast upon these monster corpses and that's outside of the scope of this change. That I can address, but removing this flag also means that corpses (possibly many) will show up in the Look dialog list. That seems undesirable to me.

Going to drop this into draft mode while I consider next steps. Open to suggestions.

davidklaw:
> As far as gameplay balance is concerned, should this come at a higher reagent cost if we're increasing the target pool?
> 
> Yah working this is a pain and requires mules. Then again, being able to cast on any corpse is too powerful. This adjustment might make this spell change from one of the hardest to easiest spells to work. If it doesn’t already do so - perhaps if we make the mana gain based on the karma of the monster it would make more sense - so the lower the karma the higher the mana gain. I think keeping original reagents is good. I’d like to see it so casting on a good monster can give an improve but won’t give you a mana boost like a goodly monster will. It would also be good to see the stats on that too.

Note that with this change it would only castable upon _good_ monsters' corpses, though that's still a fair number of new targets. It seems like you might also be suggesting that it be castable on evil/neutral monsters, just without the mana boost?

I like the idea of scaling based on the monsters karma, though that'll require something like storing their karma on the body object. Currently, the body only knows if it was good or not.

davidklaw:
This is now ready for further discussion and review. The PR description has been updated to reflect the new state.

jrotunda:
Having just seen a player roll up a character just to murder themselves on another to have defile targets - this change can't come quicker.

kod/object/passive/spell/defile.kod
```diff
@@ -49,14 +49,30 @@ classvars:
 
    viSpell_level = 3
    viMana = 0
-   viMana_bonus = 10
 
    viCast_time = 10000
    viChance_to_Increase = 100
```
jrotunda, line 54: This should be much lower...this is a guaranteed improve (I could've sworn chance to increase was 50?).

davidklaw: Good catch. Updated to 50.

kod/object/passive/body.kod
```diff
@@ -44,7 +47,7 @@ properties:
    prMonster_name = $
 
    prPlayer_name = $
-   pbWas_good_player = FALSE
```
akirmse, line 47: Removing a property will cause the value to be lost for all existing objects in the save game (the server will print an error trying to load each DeadBody, since it will be unable to restore the saved pbWas_good_player value).  Maybe that's OK, just pointing out the consequences of removing a property.

kod/object/passive/spell/defile.kod
```diff
@@ -49,14 +49,30 @@ classvars:
...
-   viChance_to_Increase = 100
+   viChance_to_Increase = 50
 
    vrSucceed_wav = defile_sound
    vbCastable_in_HappyLand = FALSE
 
+   % Mana Effect Variables
+
+   % Changing this value will adjust the base mana. 
+   % Increasing it raises the minimum mana across the board.
```
akirmse: Were these comments copied from somewhere?  They seem out of place, as they aren't at all specific to Defile.  No need to duplicate comments.

davidklaw:
I found an opportunity to clean up some of the documentation and revised the description to better clarify the changes in this PR.

This has been discussed both here and in a feedback thread on Discord. This approach seems non-controversial and should help address concerns about defile’s reliance on muling.

akirmse:
The changes to defile are fine (the worst that can happen is it breaks defile), but I'm unsure about the new "selectable" flag.  It's documented as meaning that the object is selectable, but that doesn't appear to be accurate.  It's only set for "body".  So it means "wasn't selectable before 2024, but is now".  I'm wondering if there's a cleaner way to do this.  I haven't gone in-depth on all the flags and how the client interprets them.

davidklaw:
The latest commit fixes a missed accessibility issue, enabling OF_SELECTABLE objects (e.g., the corpse in this case) to be included in next/previous target cycling. Previously, casting defile on these corpses was only possible via point-and-click.

davidklaw:
I haven’t updated the PR description yet, but after circling back, I’ve made a couple of key changes:
1. I’ve renamed SELECTABLE to UTILITY_TARGET, a flag for objects that can be targeted for non-combat purposes. In this case, I’m not considering defile casts to be attacks or combat. This feels like a better name than SELECTABLE and could potentially be used for objects like levers in the future, if they aren’t already targetable.
2. I updated the object targeting function to use this flag, which I had previously overlooked. I think corpses should be targetable using the next and previous target keys, and they currently aren’t. By adding this flag to all corpses, they can now be targeted for casting.
3. I’m allowing monster corpses to appear in the look dialog. Due to the complexity of the flag system, it feels too convoluted to come up with a simple, clean, and risk-free solution here. I’m not sure if having monster corpses in the look dialog will result in a negative player experience, but it’s something to keep an eye on.

davidklaw:
@Meridian59 This one has been hanging for a while now and could use a fresh look. There was a notable flag change since you last reviewed.

kod/object/passive/spell/defile.kod
akirmse: Ignore my comment here.. it's old and I now see that these comments do make sense.

kod/object/passive/body.kod
davidklaw: Maybe I leave it in and flag it as deprecated, to be removed in a future update? I assumed the impact would be minimal, but happy to do this if desired.

davidklaw:
> Will players be annoyed that all corpses will be clickable now? e.g. might it make it harder to deal with loot?

Hard to say. It impacts `[`/`]` targeting and mouse clicking, but not the `get` loot dialog. If we believe most players use the latter to loot, then it's the Qor casters using defile amongst piles of corpses/loot that may struggle. Maybe that's a fair tradeoff?

kod/object/passive/body.kod
akirmse: Go ahead and remove it---it won't be any different in the future.

sicoderealm:
Can you make a video showing these changes in action and how targeting is impacted? It would be a shame if look, spell click casting and/or target are polluted for all players (even those without defile) by this change. Hopefully it isn't the case - but if you need to skip over corpses to target an enemy or scroll over monster corpses in the look window, then that would be a downgrade.

davidklaw:
@sicoderealm Dropped a low quality, but I hope serviceable, overview in the PR description.

An alternative approach to defiling monster corpses (and I guess player corpses as well) I considered was it being an area spell that cast upon some amount monster corpses within range.

sicoderealm:
> @sicoderealm Dropped a low quality, but I hope serviceable, overview in the PR description.
> 
> An alternative approach to defiling monster corpses (and I guess player corpses as well) I considered was it being an area spell that cast upon some amount monster corpses within range.

For me - the look window is probably ok, but the targeting and mouse click additions now showing the corpses isn't good (especially for those who don't have defile). Maybe we can discuss some ideas offline to improve it - AOE sounds interesting but it would be nice to keep the original mechanic.

PR #964, opened 2024-09-26
Remove skybox render externs

sicoderealm:
Removing externs from skybox using a `SkyboxRenderParams` structure.


clientd3d/d3drender_skybox.c
```diff
@@ -282,11 +277,11 @@ void D3DRenderSkyboxDraw(d3d_render_pool_new* pPool, int angleHeading, int angle
 /**
 * Set the current background texture for the skybox by background ID.
 */
-void D3DRenderBackgroundSet(ID background)
+bool D3DRenderBackgroundSet(ID background)
```
akirmse, line 280: What does the return value mean?  Add a comment.

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,30 @@
 #ifndef _D3DRENDERSKYBOX_H
 #define _D3DRENDERSKYBOX_H
 
-void D3DRenderSkyBoxBegin();
-void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode, 
-	int& angleHeading, int& anglePitch);
+struct skybox_render_object {
+    LPDIRECT3DVERTEXDECLARATION9& decl1dc;
```
akirmse, line 18: Reference to a pointer?  I think most/all of these should not be references.

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,30 @@
...
+          current_room(current_room),
+          gD3DDriverProfile(driver_profile),
+          gWorldPool(pool),
+          gWorldCacheSystem(cache_system)
+    {}
+};
+
+bool D3DRenderSkyBoxBegin(skybox_render_object* skybox_render_obj_param);
+void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode,
+    int& angleHeading, int& anglePitch, D3DMATRIX& view);
```
akirmse, line 40: Confused about the references here.

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,30 @@
 #ifndef _D3DRENDERSKYBOX_H
 #define _D3DRENDERSKYBOX_H
 
-void D3DRenderSkyBoxBegin();
-void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode, 
-	int& angleHeading, int& anglePitch);
+struct skybox_render_object {
```
akirmse, line 17: Maybe change the name, since "object" doesn't mean anything.  We have Draw3DParams, so perhaps SkyboxRenderParams?

clientd3d/d3drender_skybox.c
```diff
@@ -137,15 +128,17 @@ static unsigned char gSkyboxBGRA[] =
...
+static bool D3DRenderBackgroundSet(ID background);
 
 // Implementations
 
 /**
 * Initialize the skybox rendering process.
+* Return true the background has changed and false otherwise.
 */
-void D3DRenderSkyBoxBegin()
+bool D3DRenderSkyBoxBegin(skybox_render_object* skybox_render_obj_param)
```
akirmse, line 139: This should be a const reference, not a pointer.

clientd3d/d3drender.c
```diff
@@ -455,7 +455,18 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 	Bool draw_particles = can_see;
 	Bool draw_background_overlays = can_see;
 
-	D3DRenderSkyBoxBegin();
+	skybox_render_object skybox_render_object_param(
```
akirmse, line 458: Indentation is weird on the parameters here.

clientd3d/d3drender_skybox.c
```diff
@@ -137,15 +128,18 @@ static unsigned char gSkyboxBGRA[] =
...
 /**
-* Initialize the skybox rendering process.
+* Initialize the skybox with some rendering params.
+* Return true the background has changed and false otherwise.
+* If the background has changed, the caller should trigger a redraw all.
 */
-void D3DRenderSkyBoxBegin()
+bool D3DRenderSkyBoxBegin(SkyboxRenderParams& params)
 {
+	skyboxRenderParams = &params;
```
akirmse, line 142: Super dangerous / will crash: taking the address of a stack variable and storing it as a global variable.  Would be much, much better to pass this to functions that need it and remove the global variable.

clientd3d/d3drender_skybox.c
```diff
@@ -137,15 +128,18 @@ static unsigned char gSkyboxBGRA[] =
...
 // Implementations
 
 /**
-* Initialize the skybox rendering process.
+* Initialize the skybox with some rendering params.
+* Return true the background has changed and false otherwise.
+* If the background has changed, the caller should trigger a redraw all.
 */
-void D3DRenderSkyBoxBegin()
+bool D3DRenderSkyBoxBegin(SkyboxRenderParams& params)
```
akirmse, line 140: params should be const.

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,30 @@
 #ifndef _D3DRENDERSKYBOX_H
 #define _D3DRENDERSKYBOX_H
 
-void D3DRenderSkyBoxBegin();
-void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode, 
-	int& angleHeading, int& anglePitch);
+struct SkyboxRenderParams {
+    LPDIRECT3DVERTEXDECLARATION9 decl1dc;
+    d3d_driver_profile gD3DDriverProfile;
```
akirmse, line 19: "g" is a prefix meaning global variable, which this no longer is.  Should rename these.

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,30 @@
 #ifndef _D3DRENDERSKYBOX_H
 #define _D3DRENDERSKYBOX_H
 
-void D3DRenderSkyBoxBegin();
-void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode, 
-	int& angleHeading, int& anglePitch);
+struct SkyboxRenderParams {
+    LPDIRECT3DVERTEXDECLARATION9 decl1dc;
```
akirmse, line 18: Better name?

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,30 @@
...
+        LPDIRECT3DVERTEXDECLARATION9 decl1dc,
+        d3d_driver_profile driver_profile,
+        d3d_render_pool_new pool,
+        d3d_render_cache_system cache_system,
+        room_type& current_room)
+        : decl1dc(decl1dc),
+          gD3DDriverProfile(driver_profile),
+          gWorldPool(pool),
+          gWorldCacheSystem(cache_system),
+          current_room(current_room)
```
akirmse, line 34: room_type is a big structure, so this winds up doing a memcpy of the whole thing, which is almost surely not what you want.  How about making current_room a pointer?  Also, always dangerous to name two active variables the same thing.  Rename the parameter.

clientd3d/d3drender_skybox.c
```diff
@@ -137,14 +126,16 @@ static unsigned char gSkyboxBGRA[] =
...
 // Implementations
 
 /**
-* Initialize the skybox rendering process.
+* Initialize the skybox with the current room.
+* Return true the background has changed and false otherwise.
+* If the background has changed, the caller should trigger a redraw all.
 */
-void D3DRenderSkyBoxBegin()
+bool D3DRenderSkyBoxBegin(room_type* current_room)
```
akirmse, line 138: current_room is only used to get bkgnd, so how about just passing in bkgnd instead?

clientd3d/d3drender.c
```diff
@@ -534,7 +537,8 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	if (draw_sky) // Render the skybox first
 	{
-		D3DRenderSkyBox(room, params, pRNode, angleHeading, anglePitch);
+		SkyboxRenderParams skyboxRenderParams(decl1dc, gD3DDriverProfile, gWorldPool, gWorldCacheSystem);
+		D3DRenderSkyBox(room, params, pRNode, angleHeading, anglePitch, view, skyboxRenderParams);
```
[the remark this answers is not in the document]
akirmse: "which is actually p"  I don't understand.

clientd3d/d3drender_skybox.h
```diff
@@ -14,9 +14,27 @@
 #ifndef _D3DRENDERSKYBOX_H
 #define _D3DRENDERSKYBOX_H
 
-void D3DRenderSkyBoxBegin();
-void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode, 
-	int& angleHeading, int& anglePitch);
+struct SkyboxRenderParams {
+    LPDIRECT3DVERTEXDECLARATION9 vertexDeclaration;
+    d3d_driver_profile driverProfile;
+    mutable d3d_render_pool_new renderPool;
```
akirmse, line 20: Ideally these would not be mutable.  const can be kind of like a virus, but can you do a quick check to see if the parameters to D3DRenderPoolReset, D3DRenderSkyboxDraw, D3DCacheFill, D3DCacheFlush can be const without further changes?  If so, you should be able to remove the mutable tag.

clientd3d/d3drender_skybox.c
```diff
@@ -154,17 +145,19 @@ void D3DRenderSkyBoxBegin()
...
+		return D3DRenderBackgroundSet(tempBkgnd);
 	}
+	return false;
 }
 
 /**
 * Function to render the skybox with the current background.
 */
-void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode, int& angleHeading, int& anglePitch)
+void D3DRenderSkyBox(room_type* room, Draw3DParams* params, room_contents_node* pRNode,
```
akirmse, line 159: "room" is unused in this function?

clientd3d/d3drender.c
```diff
@@ -455,7 +455,7 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 	Bool draw_particles = can_see;
 	Bool draw_background_overlays = can_see;
 
-	if (D3DRenderSkyBoxBegin(&current_room))
+	if (D3DRenderUpdateSkyBox(current_room.bkgnd))
```
akirmse, line 458: This uses the global extern "current_room", but there is also a parameter "room" to this function.  Can we get rid of the extern?

akirmse:
Nice!

PR #965, opened 2024-09-28
Puts a cap on map object and annotation sizes

davidklaw:
This PR introduces improvements to how map objects and annotations are scaled during zoom operations. The changes are meant to ensure that both objects and annotations are constrained in a way that offers proper visibility and prevents scaling issues at various zoom levels and in the various room sizes.

### Demos

https://github.com/user-attachments/assets/d48e6432-91b7-4861-a56c-d57237c719ec

https://github.com/user-attachments/assets/3d9f08ee-ac6d-4611-aeab-52cc52a7ebc9

https://github.com/user-attachments/assets/ac849105-be9a-4077-a2e8-492d72df7ffd

### DIFF
--- a/clientd3d/annotate.c
+++ b/clientd3d/annotate.c
@@ -98,11 +98,11 @@ void MapMoveAnnotations( MapAnnotation *annotations, int x, int y, float scale, 
        continue;
 
      // Set up tooltip for annotation
-     ti.rect.left = view.x + x + (int) ((annotations[i].x - MAP_ANNOTATION_RADIUS) * scale);
-     ti.rect.top  = view.y + y + (int) ((annotations[i].y - MAP_ANNOTATION_RADIUS) * scale);
+     ti.rect.left = view.x + x + (int) ((annotations[i].x - MAP_ANNOTATION_SIZE) * scale);
+     ti.rect.top  = view.y + y + (int) ((annotations[i].y - MAP_ANNOTATION_SIZE) * scale);
 
-     ti.rect.right  = ti.rect.left + (int) (MAP_ANNOTATION_RADIUS * 2 * scale);
-     ti.rect.bottom = ti.rect.top  + (int) (MAP_ANNOTATION_RADIUS * 2 * scale);
+     ti.rect.right  = ti.rect.left + (int) (MAP_ANNOTATION_SIZE * 2 * scale);
+     ti.rect.bottom = ti.rect.top  + (int) (MAP_ANNOTATION_SIZE * 2 * scale);
 
      // Clip tooltip rectangle to graphics view window
      ti.rect.left = max(ti.rect.left, view.x);
@@ -144,8 +144,8 @@ void MapAnnotationClick(int x, int y)
      if (a->text[0] == 0)
        continue;
 
-     if (a->x <= x + MAP_ANNOTATION_RADIUS && a->x >= x - MAP_ANNOTATION_RADIUS &&
-	 a->y <= y + MAP_ANNOTATION_RADIUS && a->y >= y - MAP_ANNOTATION_RADIUS)
+     if (a->x <= x + MAP_ANNOTATION_SIZE && a->x >= x - MAP_ANNOTATION_SIZE &&
+	 a->y <= y + MAP_ANNOTATION_SIZE && a->y >= y - MAP_ANNOTATION_SIZE)
        {
 	 existed = True;
 	 index = i;
--- a/clientd3d/annotate.c
+++ b/clientd3d/annotate.c
@@ -74,8 +74,16 @@ void MapAnnotationGetText(TOOLTIPTEXT *ttt)
 /*****************************************************************************/
 /*
  * MapMoveAnnotations:  Recompute tooltip rectangles for map annotations.
+ *
+ * Parameters:
+ *  annotations - An array of MapAnnotation objects containing annotation details
+ *  x           - X-offset to adjust the annotation position relative to the view
+ *  y           - Y-offset to adjust the annotation position relative to the view
+ *  scale       - Map scaling factor applied to annotation positions for zooming
+ *  bMiniMap    - True if minimap, false otherwise
+ *  size        - The full size of the square annotation, already adjusted by the zoom factor and clipped
  */
-void MapMoveAnnotations( MapAnnotation *annotations, int x, int y, float scale, Bool bMiniMap, int radius )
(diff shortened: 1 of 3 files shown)


clientd3d/map.c
```diff
@@ -526,7 +528,9 @@ void MapDrawAnnotations( HDC hdc, MapAnnotation *annotations, int x, int y, floa
 		new_x = x + (int) (annotations[i].x * scaleToUse);
 		new_y =	y + (int) (annotations[i].y * scaleToUse);
 
-		radius = max(MAP_ANNOTATION_MIN_SIZE / 2, (int) (MAP_ANNOTATION_SIZE * scaleToUse / 2));
+      // Scale annotation, capping it between the minimum and maximum limits
+      capped_scaled_size = min(MAP_ANNOTATION_SIZE * scaleToUse, MAP_ANNOTATION_MAX_SIZE);
+      radius = max(MAP_ANNOTATION_MIN_SIZE, capped_scaled_size) / 2;
```
akirmse, line 533: Seems like there might be an inconsistency here.  ANNOTATION_MAP_SIZE refers to the whole size (diameter) of the object, but MAP_OBJECT_MAX_SIZE refers to the half size (radius) in line 387 above?

davidklaw: Yes, it does seem like an inconsistency. I tried to keep things largely as they were and it would seem like renaming `MAP_OBJECT_MAX_SIZE` to `MAP_OBJECT_MAX_RADIUS` could clear this up. Would you like me to take a look at reworking annotations to be based on radius or vice versa? (Assuming this is what you're commenting on)

akirmse: Yes, I think we should.  Either one is fine as long as they're all consistent.

davidklaw: 👍🏻

I felt it made the most sense to switch annotations over to radius seeing as that the position code is all based around having the radius.

clientd3d/annotate.h
```diff
@@ -12,9 +12,9 @@
 #ifndef _ANNOTATE_H
 #define _ANNOTATE_H
 
-#define MAP_ANNOTATION_SIZE     (2 * FINENESS)    // Size of annotation indicator in FINENESS units
-#define MAP_ANNOTATION_MIN_SIZE 14                // Min size of drawn annotation in pixels
-
+#define MAP_ANNOTATION_RADIUS     FINENESS          // Radius of annotation indicator in FINENESS units
```
akirmse, line 15: This particular name is confusing (the others are OK).  The annotation is a square bitmap, so referring to a "radius" is super confusing, as it's the "radius of the inscribed circle", which no one cares about.  Mind putting just this one back?

Also, the comment could be clearer about this being the size of the source artwork (versus MAP_ANNOTATION_*_RADIUS being the size on screen, which is described in a comment).

clientd3d/annotate.c
```diff
@@ -98,11 +98,11 @@ void MapMoveAnnotations( MapAnnotation *annotations, int x, int y, float scale,
        continue;
 
      // Set up tooltip for annotation
-     ti.rect.left = view.x + x + (int) ((annotations[i].x - MAP_ANNOTATION_SIZE / 2) * scale);
```
davidklaw, line 101: I realized just now that this does not take into account hitting the maximum annotation size, so the tooltip hotspot zone is larger than the capped size of the annotation graphic. I'll follow up once I figure this out.

davidklaw: I'm stumped by this one even though the solution seems obvious. When `MapMoveAnnotations()` draws the annotation tooltip control, it needs to do so using the scaled radius now used to draw the annotation. I figured this was just a matter of passing this calculated radius to `MapMoveAnnotations()` and using that, but after various attempts I've come up empty-handed. I'll keep at this of course, but will gladly accept advice.

davidklaw: Managed to figure it out in the latest commits. I've updated the function signature to support passing in the scaled radius rather than calculate it again within `MapMoveAnnotations()`.

davidklaw:
This has been updated and the annotation tooltip handling is ready for review. Prior to these latest commits, zooming in past the point of the newly defined maximum annotation size would cause the annotation (which stops scaling) to inconsistent with the tooltip area (which continued scaling). This has been fixed and test at different minimap zoom levels and minimap dimensions.

clientd3d/annotate.h
```diff
@@ -12,9 +12,9 @@
 #ifndef _ANNOTATE_H
 #define _ANNOTATE_H
 
-#define MAP_ANNOTATION_SIZE     (2 * FINENESS)    // Size of annotation indicator in FINENESS units
-#define MAP_ANNOTATION_MIN_SIZE 14                // Min size of drawn annotation in pixels
-
+#define MAP_ANNOTATION_SIZE       FINENESS          // Size of annotation graphic in FINENESS units
```
akirmse, line 15: Confused here--what is "size of annotation graphic"?

clientd3d/annotate.c
```diff
@@ -75,7 +75,7 @@ void MapAnnotationGetText(TOOLTIPTEXT *ttt)
 /*
  * MapMoveAnnotations:  Recompute tooltip rectangles for map annotations.
  */
-void MapMoveAnnotations( MapAnnotation *annotations, int x, int y, float scale, Bool bMiniMap )
+void MapMoveAnnotations( MapAnnotation *annotations, int x, int y, float scale, Bool bMiniMap, int radius )
```
akirmse, line 78: A comment would be good here now that there are a zillion parameters.

clientd3d/annotate.h
davidklaw: > the comment could be clearer about this being the size of the source artwork

Maybe "graphic" is the wrong word here? Artwork? Asset? Something else entirely?

akirmse: Oh weird, this is the size of the bitmap on disk?  As in, if that artwork were resized, it wouldn't draw correctly?

davidklaw:
I ended up reverting annotations back to using size rather than radius. There was a previous ask to align objects and annotations around one or the other. But as I was making changes this morning, dealing with annotations in terms of radius simply felt wrong and confusing. The way I've handled it here in this latest commit feels most natural (to me).

I'll see what I can do on the object side of things.

davidklaw:
Ready for another review. Constants for objects are now in size rather than radius. While doing this I also made object and (other) player ellipses the same size and thickness, and slightly scaled up the standard object size.

PR #970, opened 2024-10-01
Remove render objects externs

sicoderealm:
Removing externs from render objects using several new structures.

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -608,7 +608,7 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 		PlayerViewParams playerViewParams(&player, gScreenWidth, gScreenHeight, main_viewport_width, main_viewport_height, gD3DRect);
 
-		timeObjects = D3DRenderObjects(pRNode, objectsRenderParams, gameObjectDataParams, lightAndTextureParams, fontTextureParams, playerViewParams);
+		timeObjects = D3DRenderObjects(objectsRenderParams, gameObjectDataParams, lightAndTextureParams, fontTextureParams, playerViewParams);
 	}
 
 	D3DRENDER_SET_COLOR_STAGE(gpD3DDevice, 1, D3DTOP_DISABLE, D3DTA_CURRENT, D3DTA_TEXTURE);
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -597,12 +597,12 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	if (draw_objects)
 	{
-		ObjectsRenderParams objectsRenderParams(decl1dc, decl2dc, gD3DDriverProfile, gObjectPool, gObjectCacheSystem, view, proj, room, params);
+		ObjectsRenderParams objectsRenderParams(decl1dc, decl2dc, gD3DDriverProfile, &gObjectPool, &gObjectCacheSystem, view, proj, room, params);
 
 		GameObjectDataParams gameObjectDataParams(nitems, num_visible_objects, gNumObjects, drawdata, visible_objects, 
 			gpBackBufferTexFull, gpBackBufferTex);
 
-		LightAndTextureParams lightAndTextureParams(gDLightCache, gDLightCacheDynamic, gSmallTextureSize, sector_depths);
+		LightAndTextureParams lightAndTextureParams(&gDLightCache, &gDLightCacheDynamic, gSmallTextureSize, sector_depths);
 
 		FontTextureParams fontTextureParams(&gFont, base_palette, gSmallTextureSize, sector_depths);
 
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -600,7 +600,7 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 	{
 		ObjectsRenderParams objectsRenderParams(decl1dc, decl2dc, gD3DDriverProfile, &gObjectPool, &gObjectCacheSystem, view, proj, room, params);
 
-		GameObjectDataParams gameObjectDataParams(nitems, num_visible_objects, gNumObjects, drawdata, visible_objects, 
+		GameObjectDataParams gameObjectDataParams(nitems, &num_visible_objects, &gNumObjects, drawdata, visible_objects, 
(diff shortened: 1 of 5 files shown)


clientd3d/d3drender_objects.c
```diff
@@ -14,58 +14,58 @@
 
 // Variables
```
akirmse, line 15: Obsolete comment; there are no longer any variables here.

clientd3d/d3drender_objects.c
```diff
@@ -14,58 +14,58 @@
...
-extern ObjectRange visible_objects[];    /* Where objects are on screen */
-extern int num_visible_objects;
-extern int gNumObjects;
-
-extern d_light_cache gDLightCache;
-extern d_light_cache gDLightCacheDynamic;
-
-extern long timeObjects;
-
 static auto TRANSLUCENT_FLAGS = OF_TRANSLUCENT25 | OF_TRANSLUCENT50 | OF_TRANSLUCENT75 | OF_DITHERTRANS;
```
akirmse, line 17: should be const

clientd3d/d3drender_objects.c
```diff
@@ -14,58 +14,58 @@
...
+	DWORD flags, 
+	bool fogEnabled,
+	const player_info* player,
+	const LightAndTextureParams& lightAndTextureParams);
+
 static int getKerningAmount(font_3d* pFont, char* str, char* ptr);
-static bool D3DComputePlayerOverlayArea(PDIB pdib, char hotspot, AREA* obj_area);
+static bool D3DComputePlayerOverlayArea(PDIB pdib, char hotspot, AREA* obj_area, const PlayerViewParams& playerViewParams);
 
 extern int FindHotspotPdib(PDIB pdib, char hotspot, POINT* point);
```
akirmse, line 70: These should be in a header somewhere rather then externed.

clientd3d/d3drender_objects.h
```diff
@@ -13,6 +13,153 @@
...
+          d3dRect(d3dRectParam)
+    {}
+};
+
+struct FontTextureParams
+{
+    font_3d* font;
+    Color* basePalette;
+    int smallTextureSize;
+    int* sectorDepths;
```
akirmse, line 53: sector depths definitely doesn't belong here; it has nothing to do with the font texture.

clientd3d/d3drender_objects.c
akirmse: TRANSLUCENT_FLAGS is a constant, not a variable.

clientd3d/d3drender_objects.c
```diff
@@ -1055,23 +1073,23 @@ void D3DRenderOverlaysDraw(d3d_render_pool_new* pPool, room_type* room, Draw3DPa
...
+								range = &gameObjectDataParams.visibleObjects[gameObjectDataParams.numVisibleObjects];
 								range->id = pRNode->obj.id;
 								range->distance = distance;
 								range->left_col = tempLeft;
 								range->right_col = tempRight;
 								range->top_row = tempTop;
 								range->bottom_row = tempBottom;
 
-								num_visible_objects = ++num_visible_objects > MAXOBJECTS ? MAXOBJECTS : num_visible_objects;
+								gameObjectDataParams.numVisibleObjects = ++gameObjectDataParams.numVisibleObjects > MAXOBJECTS ? MAXOBJECTS : gameObjectDataParams.numVisibleObjects;
```
akirmse, line 1092: This (existing) code is super gross, that's what min() is for.
```
gameObjectDataParams.numVisibleObjects = std::min(gameObjectDataParams.numVisibleObjects + 1, MAXOBJECTS);
```

clientd3d/d3drender_objects.c
```diff
@@ -1589,23 +1616,23 @@ void D3DRenderObjectsDraw(d3d_render_pool_new* pPool, room_type* room,
...
+					range = &gameObjectDataParams.visibleObjects[gameObjectDataParams.numVisibleObjects];
 					range->id = pRNode->obj.id;
 					range->distance = distance;
 					range->left_col = tempLeft;
 					range->right_col = tempRight;
 					range->top_row = tempTop;
 					range->bottom_row = tempBottom;
 
-					num_visible_objects = ++num_visible_objects > MAXOBJECTS ? MAXOBJECTS : num_visible_objects;
+					gameObjectDataParams.numVisibleObjects = ++gameObjectDataParams.numVisibleObjects > MAXOBJECTS ? MAXOBJECTS : gameObjectDataParams.numVisibleObjects;
```
akirmse, line 1635: Replace with min()

clientd3d/d3drender_objects.h
```diff
@@ -13,6 +13,150 @@
...
+    DrawItem* drawData;
+    ObjectRange* visibleObjects;    // Where objects are on screen
+
+    // Textures for the back buffer
+    PDIRECT3DTEXTURE9 backBufferTexFull;
+    LPDIRECT3DTEXTURE9 (&backBufferTex)[16];
+
+    GameObjectDataParams(
+        long numItemsParam,
+        int& numVisibleObjectsParam,
```
akirmse, line 101: I'm not too excited about these as references, because it hides the fact that these are pointers rather than just ints.    If there's no way to avoid having them here, I guess convert to pointers?

clientd3d/d3drender_objects.c
```diff
@@ -12,64 +12,60 @@
...
+	const LightAndTextureParams& lightAndTextureParams);
+
+static void D3DRenderPlayerOverlayOverlaysDraw(
+	const ObjectsRenderParams& objectsRenderParams,
+	const PlayerViewParams& playerViewParams,
+	const LightAndTextureParams& lightAndTextureParams,
+	list_type overlays,
+	PDIB pDib, 
+	AREA* objArea, 
+	BOOL underlays);
```
akirmse, line 56: Use bool instead

clientd3d/d3drender.c
```diff
@@ -159,6 +159,8 @@ extern BYTE				light_rows[MAXY/2+1];      // Strength of light as function of sc
 extern ViewElement		ViewElements[];
 extern HDC				gBitsDC;
 
+D3DMATRIX view, mat, rot, trans, proj;
```
akirmse, line 162: View and proj are included in one of the structs; do we also need them as global variables, or can they be static now?

PR #971, opened 2024-10-04
Updated treasure tables of certain "messy" monsters

Aesica:
This PR addresses an issue where certain monsters drop too many unstackable items.  As these items build up on the ground, graphical glitches such as flickering or vanishing/invisible objects occur after only a few minutes due to the number of unique objects on the screen.  The only current remedy is to leave the room and reset it every few minutes.

This addresses that issue by making the unstackable items much rarer, or removing them in some cases, then padding the table with stackable items that are typically of lesser value.

The drop rates of each monster's "valuable" stackable items is preserved, so I don't expect much of an economic impact.

Monsters affected:

Avar
- Added:  uncut seraphym, turkey leg
- Removed:  purify potion, cure poison potion, cure disease potion (all sold by lady aftyn)

Battered Skeleton**
- Added:   mushroom, red mushroom, edible mushroom (snack), emerald, sapphire, identify wand
- Removed:  scimitar (dropped by orcs, cave orcs), metal shield (sold by various blacksmiths)

Tusked Skeleton** (drops themed more around combat)
- Added:  blue mushroom, edible mushroom, emerald, sapphire, diamond, gauntlets, simple helm
- Removed:  short sword, hammer (buyable from several blacksmiths), scimitar (dropped by orcs/cave orcs), knight shield (dropped by various monsters), nodeburst potion (dropped by black spiders, certain quests), enfeeble wand, forget wand (moved to daemon skeleton), light scroll (dropped by several monsters - TID_MED_TOUGH)

Daemon Skeleton** (drops now themed more around magic)
- Added:  blue mushrooms, emeralds, sapphires, diamonds, enfeeble wand, forget wand
- Removed:  scimitar (dropped by orcs, cave orcs), gauntlet, simple helm (dropped by other monsters, including tusked skeletons now), karahol's curse potion, wind scroll, summon poison fog scroll, (not available anymore*), killing fields scroll (dropped by avar chieftains), light scroll (dropped by several other monsters)

\* Items no longer available as drops are all very low value and will be made available again by another PR at a later date.

\*\* Items added to the skeleton variants are based on the original skeleton's loot table.  It's also worth noting that all skeletons will still drop a varied-durability long sword 100% of the time, which is created independently of the drop tables.  This can be explored in another PR, as it seems outside this one's scope.

Edit:  As an added note, this will increase the rarity of the scimitar ever so slightly, which in my opinion is a good thing.

### DIFF
--- a/kod/object/passive/trestype/avart.kod
+++ b/kod/object/passive/trestype/avart.kod
@@ -27,19 +27,18 @@ messages:
    
    constructed()
    {
-      plTreasure = [ [ &RainbowFern, 20],
+      plTreasure = [ [ &RainbowFern, 30],
                      [ &Bread, 15],
                      [ &KriipaClaw, 15],
                      [ &Money, 14],
                      [ &WebMoss, 8],
-                     [ &NeruditeSword, 7],
-                     [ &NeruditeArmor, 7],
-                     [ &IvyCirclet, 5],
-                     [ &CurePoisonPotion, 2],
-                     [ &CureDiseasePotion, 2],
-                     [ &RemoveCursePotion, 2],
-                     [ &PurifyPotion, 2],
-                     [ &Yrxlsap, 1]
+                     [ &UncutSeraphym, 8],
+                     [ &TurkeyLeg, 5],
+                     [ &RemoveCursePotion, 1],
+                     [ &Yrxlsap, 1],
+                     [ &NeruditeSword, 1],
+                     [ &NeruditeArmor, 1],
+                     [ &IvyCirclet, 1]
 		             ];
 
       propagate;
--- a/kod/object/passive/trestype/skel2t.kod
+++ b/kod/object/passive/trestype/skel2t.kod
@@ -28,13 +28,17 @@ messages:
    
    Constructed()
    {
-      plTreasure = [ [ &MetalShield, 50 ],
-                     [ &Knightshield, 15 ],
+      plTreasure = [ [ &Mushroom, 25 ],
+                     [ &RedMushroom, 20 ],
+                     [ &Snack, 19 ],
                      [ &Money, 15 ],
-                     [ &Scimitar, 10 ],
                      [ &NeruditeArrow, 5 ],
+                     [ &Emerald, 5 ],
+                     [ &Sapphire, 4 ],
                      [ &InkyCap, 4 ],
-                     [ &DiscordScroll, 1 ]
+                     [ &KnightShield, 1 ],
+                     [ &DiscordScroll, 1 ],
+                     [ &IdentifyWand, 1 ]
                    ];
 
       propagate;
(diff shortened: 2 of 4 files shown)


Aesica:
My thinking here was that the game sees 1 rainbow fern and 500 rainbow ferns as the same thing when they're on the ground, so there's no added objects on the map in that regard.  I suppose that extra "10" could be given to some new reagent they've never dropped before, like firesand, if the extra weight on ferns is seen as an issue.

davidklaw:
It's me just being picky. When I've built in avar-filled rooms I prioritize all the other reagents because of their value and fill any remaining space with ferns, inevitably leaving behind 100+ fern. If that's useful for others looking for regs and not shillings, there's no change needed, but I was looking at that as a sign that there are already too many dropping.

Aesica:
That's a fair point, as I usually leave them behind.  Even if players don't pick them up, having multiple sets of ferns drop and consolidate into a fern stack of 100 seems better than having a pile of several dozen nerudite swords, armors, and potions (each of which is +1 objects on the screen).

Now I'm really thinking about actually lowering the fern rate a bit, then bumping up uncut a bit in addition to adding firesand.  Both are fairly inexpensive in terms of island reagents, and might be useful to have dropping in the wild like that.

davidklaw:
I don't know the price of uncut or firesand offhand, but replacing fern with more of these could create an even more lucrative farming scenario in rooms like Outside the Guard Tower. Maybe best to keep the remainder on fern -- I think I know realize that was the reason they were bumped up.

Aesica:
Late reply, as I was hoping a few more might chime in, but I think this has generated all the discussion it can, so it might be ready for a full review and either a pass or fail.

Now for farming purposes:  Both firesand and uncut seraphym are the two next cheapest island reagents after ferns.  As of this post, Zala will buy uncut seraphim or firesand for 18 per, only a pinch higher than rainbow fern for 12 per.  In contrast, the other avar reagent drops are web moss for 42 per and kriipa claws for 36 per.  Uncut won't add too much value to farming outside the guard tower, but firesand might because it's a reagent for illusionary firewall along with ferns.  So after thinking about it, firesand probably shouldn't be added so that illusionary firewall users are forced to take a trip back to town after they've exhausted their supply.  Indefinite campouts aren't ideal.

jrotunda:
This is good, you covered a lot of bases. I like the reagent changes. Will make it slightly less painful to be in the Marion Crypt.

Nixing shields and armor from being massive drops will help in restoring balance lost - it makes mend more important. 

One note - Colhorr in Marion buys those cheap light round shields dropped by the battered skeletons in level 2 of the crypt - an intentional design. I don't think this is major however - and farming round shields has never happened despite this intention.

We'll have to carefully monitor how this affects building and player habits when merged.

Aesica:
From a player's perspective, it's more lucrative to sell the long swords that are in good condition as opposed to the shields.  Shields (and especially armor) are seldom worth farming, given their weight-to-value ratio.

PR #974, opened 2024-10-06
Updates advertisements

davidklaw:
This PR updates the advertisements display on login to handle bitmaps instead of AVIs. It includes two ads for the official Discord and website. That this might be silly and nostalgic while hopefully leading players to helpful resources.

### Ad files
[ads.zip](https://github.com/user-attachments/files/17308231/ads.zip)

### Screenshot
image


akirmse:
Are the new ad bitmaps missing?

I'm not sure how the update system works with ad files; I think there's something weird about them.  They might have to go in a separate update package.

davidklaw:
Oh, right. Ads! I'll attach them to this PR as `run/` is gitignored.

As for updating, here's what you wrote 20 years ago 😏. If we want to move forward with this.

## Advertisements
Andrew Kirmse/ 25 January 1996

The Meridian client can display up to 2 advertisements in its character selection dialog. Clicking on an advertisement opens the user's Web browser on a URL associated with the advertisement, and logs the advertisement and user names in the server log. An ad is an AVI file with no sound.

### Client side

On the client side, advertisements are stored in the "ads" subdirectory of the client installation. A message from the server identifies which two advertisements to display in the character selection dialog. [Here](http://127.0.0.1:3000/doc/makingadverts.doc) is a description of how to make advertisement AVI files (Word file).

If a client update consists solely of advertisement files, the normal download dialog is not displayed. Instead, the client shows a small dialog with the words "Performing routine maintenance" while the advertisements are downloaded.

### Server side

The advertisement filenames and associated URLs are set in the server's configuration file (blakserv.cfg). In the [Advertise] section, the keys File1 and File2 point to the two AVI filenames. The keys Url1 and Url2 point to the associated URLs. The default setup is

```
[Advertise]
File1=ad1.avi
Url1=http://www.3do.com
File2=ad2.avi
Url2=http://www.microsoft.com
```
These can be changed dynamically with the "set config" administrator command. Any changes should be saved by also issuing the "save config" command.

### Updating advertisements

Advertisements are updated via the normal client update mechanism. First, build a package containing the new AVI files, using the arq program. By convention, packages are named pkg####$.avi, where #### gives the month and day, and $ is a sequence letter for each day, starting at "a".

Example (typed at the command line):

`arq -a pkg0125a newad1.avi`

Next, add a line to the end of the server's packages.txt file, containing the package name, a new sequence number, and the package type number 24. (If you want guests to also download the new advertisement, use the package type number 56. However, this is not recommended, since it forces guests to do a download when they log in.)

Example (new line in packages.txt file):

`pkg0125a.arq 50 24`

where the sequence number 50 is one more than the number appearing in the previous line of packages.txt.

Finally, type the command "reload packages" in administrator mode. If the server is not running when the packages.txt file is modified, this step is not necessary; the file will be reread when the server starts.

davidklaw:
I had thought these would make for some fun graphical flair that leads players to two very helpful resources, but perhaps the move here is removing advertisements entirely and linking to these resources in the MOTD?

akirmse:
Sorry, I thought you were going to check in the ad files.  I didn't notice that they were attached as a zip file.

akirmse:
I'm OK adding this but there's a merge conflict that needs to be fixed first.

davidklaw:
Merge conflict has been resolved ✅

PR #976, opened 2024-10-11
feature: add cow sounds

adrienlaws:
# What 🐄 
Adds new cow and guard cow sounds `cow_awr.wav`,`cow_atk.wav`, and `cow_dth.wav`. 

# Why
Adds character to two longtime, beloved Meridian creatures.
> There's nothing like sitting back and talking to your cows. - Russell Crowe

# Lore 🔉 
Long known for their stoic presence, the cows of the land now fill the air with a distinct bovine experience.


davidklaw:
Finally!

adrienlaws:
It seemed to be okay when I tested it in game with `dm get monster cow` and `dm get monster guard cow`.

I can post some clips later for review if that's wanted.

The guard cow is very fast movement wise, but they both attacked at the same speed from what I remember.  The sound didn't seem off but I can check again.

adrienlaws:
https://github.com/user-attachments/assets/5cac8b45-eff9-4418-82aa-ac020243b093

Here is a video recording showing what the attack sounds are like for both `cow` and `guard cow`.  Cows seem to "attack" once every 2 seconds.

PR #978, opened 2024-10-14
Adds vscode settings file with encoding settings

davidklaw:
This PR adds a Visual Studio Code settings file to the project that enforces `windows1252` encoding for `rc` files.

Currently, contributors must manually reopen `rc` files with the correct encoding before saving to avoid introducing unwanted characters, especially in files containing translations. This settings file automates the process, ensuring `rc` files always use the proper encoding, preventing encoding errors and preserving file integrity.

Windows-1252 supports ISO-8859-1 plus full support for the following ligature forms (for English):

* German
* Portuguese
* Spanish
* French
* and [more](https://en.wikipedia.org/wiki/Windows-1252)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,8 @@
 {
-  "files.encoding": "windows1252",
   "[rc]": {
     "files.encoding": "windows1252"
+  },
+  "[kod]": {
+    "files.encoding": "utf8"
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,4 @@
   "[rc]": {
     "files.encoding": "windows1252"
   },
-  "[kod]": {
-    "files.encoding": "utf8"
-  }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,8 @@
 {
-  "[rc]": {
-    "files.encoding": "windows1252"
+  "files.associations": {
+    "*.rc": "resource_files"
   },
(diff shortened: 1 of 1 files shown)


akirmse:
I'm not sure about this.  Do foreign language translations in rc files look OK with Windows-1252?  It looks like maybe they're actually UTF-16?  https://stackoverflow.com/questions/72143553/visual-studio-resource-editor-corrupts-rc-files-with-utf-8-encoding

As for kod files, the only thing that reads them is the Blakod compiler, and it definitely doesn't understand UTF-8, nor do the String functions in the server.  Maybe UTF-8 strings happen to work sometimes because they're usually passed through unchanged to the client, where they're displayed in a text box that maybe (??) can display UTF-8?

davidklaw:
> I'm not sure about this. Do foreign language translations in rc files look OK with Windows-1252? It looks like maybe they're actually UTF-16?

I've been opening and saving them as Windows-1252 but I'll be honest, I don't know what they should actually be.

davidklaw:
Windows-1252 supports ISO-8859-1 plus full support for the following ligature forms (for English):
- German
- Portuguese
- French
- Finnish
- Irish
- Italian
- Norweigian
- Spanish
- Swedish
- Icelandic
- and [more](https://en.wikipedia.org/wiki/Windows-1252)

davidklaw:
Updated to be limited to rc files. Updated description.

davidklaw:
Ready for review again.

This configuration will open `.rc` files in Windows-1252 encoding.

PR #980, opened 2024-10-23
Denies Shadow Rift in `NO_COMBAT` rooms

davidklaw:
This commit adds `NO_COMBAT` rooms to the list of rooms excluded from shadow rift portals. Previously, only the `HISTORIC` rooms were excluded. This commit also moves the logic checking this to `CanPayCosts()`, so that the user gets immediate feedback that it cannot be cast.

`NO_COMBAT` rooms include:
- Inns
- Taverns
- Adventurer Halls
- Raza
- Many of the NPC/town rooms (blacksmiths, the vault, banks, etc)
- The Clearing of the Trading Post, Outside the Mad Scientist's Hut, The Remote Trading Post of Wulfgang zax'Ak, and a few others
- Rented Rooms

### DIFF
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -3869,8 +3869,11 @@ messages:
    CanHavePlayerPortal()
    "Return whether it's permissible to have a player-created portal in this room. "
    {
-      % No portals into "historic" (i.e. inaccessible, dead) areas
-      if Send(self, @GetRegion) = RID_HISTORIC
+      % No portals in 
+      % - "historic" (i.e. inaccessible, dead) areas
+      % - "No combat" rooms like inns, adv halls, rented rooms, and others
+      if Send(self,@GetRegion) = RID_HISTORIC
+         OR Send(self,@CheckRoomFlag,#flag=ROOM_NO_COMBAT) 
       {
          return FALSE;
       }
--- a/kod/object/passive/spell/multicst/shadrift.kod
+++ b/kod/object/passive/spell/multicst/shadrift.kod
@@ -81,6 +81,15 @@ messages:
 
    CanPayCosts(who = $, lTargets = $, iSpellPower = 0)
    {
+      % Ask room if we're allowed to cast here.  For example, portals into 
+      % guild halls are a bad idea.
+      if (NOT Send(Send(who, @GetOwner), @CanHavePlayerPortal))
+      {
+         Send(who,@MsgSendUser,#message_rsc=spell_bad_location,#parm1=vrName);
+         
+         return FALSE;
+      }
+
       propagate;
    }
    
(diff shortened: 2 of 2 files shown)


Gar103:
Helpful intention here - this doesn't really solve the problem as you guys seem to perceive it. I don't agree with your perception, but I can see what you're going for. I'm pretty sure the subsequent sequence of events will be:

* Veterans who use shadow rift will shift to using portals outside in combat areas.
* Per a concern expressed before by you guys: newbies will encounter these portals and walk into them.
* The administration, will, I'm guessing, not like that either.
* Another nerf or rule change will be made without having a conversation with the veterans who use these spells.

Using portals in rooms isn't abuse, it's respect and staying out of the way. This is something really important I hope I can convey: the administration seems to perceive veterans as problems, rather than as something valuable. We are consistently being invisibly observed, and the game is being _live edited around us_, with detectable anger. Then, private conversations are had without players involved, and nerfs are made that miss the mark. We veterans are just assumed to be bad actors.

Countless games strive to create fun mechanics. Most fail. Prism spells like shadow rift somehow, amazingly, turned out to be very fun. Why isn't design work being focused on figuring out why it's fun, and preserving that fun? Past the anger that this spell seems to provoke, I've seen no quantification or analysis of exactly what the problem is. Just vague gut allusions that it's somehow 'abusive' or 'too advantageous.' Can you actually quantify that advantage? How many shillings per hour while farming? Does it help in PvP? Does it alter win rates between players or guild wars?

Shouldn't there be cool systems for high level players to have fun with? Can the advantages or abuse you feel this spell entails be trimmed without gutting the thing? Because I'm fairly certain no one in the administration has actually played with it as a player. They don't feel it or experience the effort that goes into it and the fun that comes out of it. They're only seeing the admin side: invisibly teleporting in, seeing something they don't like on a gut instinct level, and getting angry. Anger has no place in design decisions, and I myself have witnessed it repeatedly at even the most basic and obvious uses of existing game mechanics.

There are ways to fix shadow rift in the way you guys want, while keeping the fun part. An open conversation would be fantastic. Please start with asking people who use spells like these why they do it, if they're having fun, and how much advantage they really get out of it. I think you'd be surprised. Elusion is by far the more versatile and powerful spell, and there's no issue there, so maybe it's just a communication and understanding issue?

Gar103:
I actually happen to agree with you guys that shadow rift is a very strong spell. Here's what I turned it into in my own space:

https://github.com/GarAdminTools/Meridian59/blob/Patch31/kod/object/passive/spell/teleportspell/deathrift.kod

Death Rift was a pretty beloved hit. It basically turns Shadow Rift into a personal Qor 6 spell instead, that only works for the caster once per cast. You can only use it in an inn or ad hall, and it lets you walk through the Underworld while still alive. Flavor win, right? The town portals work, so you can go to any mainland inn - but the 'random' portal has a destination of your choice, which you set by personally going somewhere and binding yourself to a spot, just like Shadow Rift did with prisms. There's also a caveat that anyone in the destination area gets warned that someone infernal is approaching through the underworld, so PKs can't just pop up on people out of nowhere. There is also a message to the inn the Qor 6 player casts it in, telling those around them in the inn where their custom destination is, so there's no surprise or element of confusion.

I would probably do it a little differently now, but it's been around for about 12 years now and has been nothing but awesome. It's like a reverse Rescue, letting you go out into the world, but not getting you safe, and it never really presented any major problems or abuses. People couldn't really use it to farm like you might imagine since it costs 4 DAFs every cast, and well, good luck with that.

PR #981, opened 2024-10-29
Make client.rc UTF-16.

akirmse:


akirmse:
I guess this isn't needed after all, since the bug reporter was able to build without converting all the RC files to UTF-16.

PR #983, opened 2024-10-31
Quality of Life: Auto-delete depleted wands and Froz

paul-bargerstock:
This PR serves two purposes - to remove the clutter that inevitably clogs inventories when using wands, or Jewels of Froz; and bringing consistency to the way that consumable combat items are disposed of.

Both wands and Froz remove delete themselves after depletion, like Berserker Rings.  The "shatters" message for wands already makes perfect sense, but the Froz depletion message was updated to be more accurate and lore friendly.

### DIFF
--- a/kod/object/item/passitem/attmod/frozjewl.kod
+++ b/kod/object/item/passitem/attmod/frozjewl.kod
@@ -28,7 +28,7 @@ resources:
 
    frozjewel_desc_broken_rsc = "The jewel seems ordinary and dull."
 
-   frozjewel_broken = "The Jewel of Froz now seems quite ordinary."
+   frozjewel_broken = "Having expended its power, the Jewel of Froz crumbles to crimson dust."
 
 classvars:
 
@@ -62,7 +62,7 @@ messages:
          Send(poOwner,@MsgSendUser,#message_rsc=frozjewel_broken,
 	           #parm1=Send(self,@GetCapDef),#parm2=vrName);
          Send(poOwner,@TryUnuseItem,#what=self);
-         vrDesc = frozjewel_desc_broken_rsc;
+         Send(self,@Delete);
       
          return hit_roll;
       }
--- a/kod/object/item/passitem/attmod/frozjewl.kod
+++ b/kod/object/item/passitem/attmod/frozjewl.kod
@@ -26,8 +26,6 @@ resources:
       "aggression, using raw mana to boost elemental touch spells.  "
       "Your hand tingles when you hold it."
 
-   frozjewel_desc_broken_rsc = "The jewel seems ordinary and dull."
-
    frozjewel_broken = "Having expended its power, the Jewel of Froz crumbles to crimson dust."
 
 classvars:
--- a/kod/object/item/passitem/specwand.kod
+++ b/kod/object/item/passitem/specwand.kod
@@ -70,7 +70,7 @@ messages:
       {
          Send(poOwner,@MsgSendUser,#message_rsc=SpecialWand_broken,
            #parm1=send(self,@getcapdef),#parm2=send(self,@getname));
-         vrDesc = SpecialWand_desc_broken_rsc;
+         Send(self,@Delete);
       }
       
       return;
--- a/kod/object/item/passitem/specwand.kod
+++ b/kod/object/item/passitem/specwand.kod
@@ -38,8 +38,6 @@ classvars:
 
 properties:
 
-   vrDesc = SpecialWand_desc_rsc
-
 messages:
 
    %the subclass should define what the wand can apply itself to


kod/object/item/passitem/attmod/frozjewl.kod
```diff
@@ -28,7 +28,7 @@ resources:
 
    frozjewel_desc_broken_rsc = "The jewel seems ordinary and dull."
```
davidklaw, line 29: Maybe clean this resource up since it's no longer used.

kod/object/item/passitem/specwand.kod
```diff
@@ -70,7 +70,7 @@ messages:
       {
          Send(poOwner,@MsgSendUser,#message_rsc=SpecialWand_broken,
            #parm1=send(self,@getcapdef),#parm2=send(self,@getname));
-         vrDesc = SpecialWand_desc_broken_rsc;
```
davidklaw, line 73: Like the froz, this resource can be removed now.

davidklaw:
We talked about this on Discord, and I like the move. The only observation I have is that broken items can hint someone’s nearby, but overall, it’s a good QOL improvement—players won’t have to check which items are broken.

jrotunda:
Thanks for the work @paul-bargerstock ! I like the idea of this - it helps streamline ones inventory in a more user-friendly manner, and feels intuitive based on how some other magic items work, like scrolls.

Some personal design feedback and notes for posterity:

- Management of broken gear in inventory, during combat - is a sub-construct of the use of items that cast spells or enhance spell.
- This management doesn't extend to rings, which are double-clicked to be worn and imitate a Personal Enchantment, constantly using 'hits' as they are kept on
- This management doesn't extend to scrolls, which cast an Area Effect Spell. Scrolls *currently only appear in single-use( format - but scrolls *can* be made to have more than one use.
- Spell wands (with colored tops indicating their spell), fire wands, heal wands, gnarled staves and vampiric wands, which use a 'hit' every time they are clicked, do not currently vanish.
- This change specifically makes it so that spell wands vanish from the inventory - not heal wands, staves, or other specific wands.
- The Jewel of Froz is similar to the Berserker Ring - it only uses a 'hit' when the person attacks while it's in use. The Ring uses the 1 finger slot currently allotted to players.
-  The Jewel of Froz uses the hand slot. Using the Jewel of Froz requires a lot of clicking to use and unuse. It does not swap to another Jewel of Froz, or a weapon - the player gets told "Your hands are too full to use that."
- The JoF automatically unequips when it breaks, just like a weapon or a pair of gauntlets (another damage enchancing item)

Design question:

Are players now, after 30 years, having trouble throwing out a used Jewel of Froz or even a used wand? Are players running around with that many Jewels at one time that it's causing them to be confused? Are players who use the Jewel of Froz and aware of its capabilities not also aware of how to throw them out? Basically -Is this an actual problem and is the solution here actually making the people already skilled at using these things - more skilled? 

**This change does not include heal wands but want to mention I do feel strongly that heal wands in particular should never use this mechanism to vanish when depleted due to the speed in which one can use them - the fact they don't automatically vanish helps balance out that they have no post-cast cooldown timer. Making it so someone can just hit tab-space and plow through a bunch of wands that automatically select the next one down the line would be bad.**

Along those lines - We DO have a hospice wand - this wand can be used like a heal wand, but there is a post-cast timer on the 'spellwand' class and spamming to use them won't work and will make the wand break faster (it uses a hit whether or not the spell casts when you use it)

Personal thoughts:

I think changes like this can have an opposite effect, and make it even easier for the ones who are already good with these items, but time will tell. I think for the spell wands, it's ok, as long as the heal/fire/vamp and staves are left out of this.

I'm conflicted on the Jewel of Froz as it would be a shame to lose the "lifeless, ordinary and dull"  - but I do like the crumbling to dust message and mechanism. I just want to be sure this is something players would want.

If the idea is to make it more obvious that the Jewel of Froz is broken, or make them easier to use perhaps:

- Add a dynamic light to the Jewel of Froz, not just when it's on its platform in the Underground Lake.
- Remove dynamic light if the Jewel of Froz is depleted, retain "ordinary and dull message"
- Additional: dynamic light set to the hand slot when Jewel of Froz is 'held' by caster
- Make Jewel of Froz brighter/dimmer in Inventory View if alive/dead.
- Make it so that you can **swap the use on a Jewel of Froz** just like you would any other weapon _**(a change that was recently made to gauntlets)**_ <--- This should be done regardless.

paul-bargerstock:
Hey @jrotunda - a lot to unpack here, I'll do my best.

As to your notes, they appear to be statement of fact so I don't have anything to add.  To your questions:

> Are players now, after 30 years, having trouble throwing out a used Jewel of Froz or even a used wand? Are players running around with that many Jewels at one time that it's causing them to be confused? Are players who use the Jewel of Froz and aware of its capabilities not also aware of how to throw them out? Basically -Is this an actual problem and is the solution here actually making the people already skilled at using these things - more skilled?

Having trouble dragging and dropping?  I don't think so.  This a quality of life PR.  You use up wands, you use up Froz, and now you have useless objects clogging your inventory with the exact same art as the useful ones.  You've just created a chore for the user which requires both right-click to read, and then drag and drop.  Being more or less efficient or acclimated to chores doesn't equate to skill/non-skill.  As one of those 30 year players who recently finished off a pure mage on thrashers burning through JoF's I can tell you that yes, it is indeed terrible.

> **This change does not include heal wands but want to mention I do feel strongly that heal wands in particular should never use this mechanism to vanish when depleted due to the speed in which one can use them - the fact they don't automatically vanish helps balance out that they have no post-cast cooldown timer. Making it so someone can just hit tab-space and plow through a bunch of wands that automatically select the next one down the line would be bad.**

For the same reasons above (unintuitive chore / no feedback loop), I whole-heartedly disagree here.  If the player base feels heal wands are unbalanced (do they..?) we should balance them!  A poor user experience is not a good tool for balancing.

> I think changes like this can have an opposite effect, and make it even easier for the ones who are already good with these items, but time will tell. I think for the spell wands, it's ok, as long as the heal/fire/vamp and staves are left out of this.

I don't agree that 'far less painful' equates to easier.  It is "hard" to hit a target at any distance with the wonky wand targeting system.  It is "painful" to not know which of the wands has charges left, and need to go whack-a-moling to find out.

> I'm conflicted on the Jewel of Froz as it would be a shame to lose the "lifeless, ordinary and dull" - but I do like the crumbling to dust message and mechanism. I just want to be sure this is something players would want.

This brings me to my final point that I think can keep all parties happy!  Having discussed further with @davidklaw, we're going to revert this PR and pivot.  He's graciously offered to make assets for broken wands, and 'dull' Froz which we can then use when charges or usefulness have run out.

Although I believe that dragging and dropping useless items in a game where inventory space is always at a premium, and the interface to do so is rather wonky - I think this will make for a good compromise.  Since the majority of the chore is needing to go spelunking for which items are actually still useful, I'm all for an 80% solution if everyone is happy.

PR #988, opened 2024-11-03
Adds window capture of mouse wheel to zoom map

davidklaw:
This PR introduces mouse wheel support for zooming in and out of the map, making navigation a bit more intuitive and user-friendly. Users can now use their mouse to zoom both the map view and minimap.

https://github.com/user-attachments/assets/da80de59-45ea-4aee-961a-cf898f2738e5

### DIFF
--- a/clientd3d/client.c
+++ b/clientd3d/client.c
@@ -105,6 +105,18 @@ LRESULT CALLBACK WndProc (HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 
 	switch (message)
 	{
+		case WM_MOUSEWHEEL:
+		{
+				if (state == STATE_GAME)
+				{
+						int zDelta = GET_WHEEL_DELTA_WPARAM(wParam);
+						int direction = (zDelta > 0) ? 1 : -1; // Scroll up is positive, scroll down is negative
+						MapZoom(direction);
+						return 0;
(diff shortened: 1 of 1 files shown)


clientd3d/client.c
```diff
@@ -105,6 +105,18 @@ LRESULT CALLBACK WndProc (HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 
 	switch (message)
 	{
+		case WM_MOUSEWHEEL:
+		{
+				if (state == STATE_GAME)
+				{
+						int zDelta = GET_WHEEL_DELTA_WPARAM(wParam);
+						int direction = (zDelta > 0) ? 1 : -1; // Scroll up is positive, scroll down is negative
```
akirmse, line 113: Do you want to make the amount of zoom proportional to how far the wheel has moved?

davidklaw: **Thinking out loud:**

I think my solution might end up a bit more complex than what I have here now. My mouse (notch scrolling) emits fixed increments of 120. To speed up the zoom when the wheel is scrolled quickly, I suppose I'd need to record consecutive scroll events and adjust zoom speed accordingly.

Then there's smooth scrolling mouse wheel support and I'd need to track down one of these mice to test.

Issue #990, opened 2024-11-06
Sun / Moon disappear  after view display reset.


jrotunda:
The sun and moon will disappear from the sky when the display is reset between height transitions.

When the sun or moon shift again, they'll render correctly.

This will occur if you login as a DM and do "dm reset", or send object invalidatedata, or if you are online during a system save as a user or DM.

PR #992, opened 2024-11-06
Add background objects following a system save 

sicoderealm:
Currently, background objects are only sent to players on login. When a system save takes place they disappear from the room overlays and are therefore not rendered (in both hardware/software renderer). This could have never worked or been supported by the game until now.

drawing

This change adds them again following a system save as well as on login.

fixes https://github.com/Meridian59/Meridian59/issues/990

### DIFF
--- a/kod/object/passive/moon.kod
+++ b/kod/object/passive/moon.kod
@@ -114,7 +114,7 @@ messages:
    }
 
    AddBackgroundObject(who = $)
-   "Called by user upon logon and after system save"
+   "Sends the background overlay to the client specified by the 'who' parameter."
    {
       Send(who,@ToCliAddBackgroundOverlay,#what=self);
       return;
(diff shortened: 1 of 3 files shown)


kod/object/passive/moon.kod
```diff
@@ -114,7 +114,7 @@ messages:
    }
 
    AddBackgroundObject(who = $)
-   "Called by user upon logon"
+   "Called by user upon logon and after system save"
```
akirmse, line 117: Change the comment to say what the function does, not when it is called.  This is a perfect illustration of why not to document when it's called.

kod/object/passive/sun.kod
```diff
@@ -72,7 +72,7 @@ messages:
    }
 
    AddBackgroundObject(who = $)
-   "Called by user upon logon"
+   "Called by user upon logon and after system save"
```
akirmse, line 75: here too

PR #916, opened 2024-08-19
Deconstruction of d3drender for world objects

sicoderealm:
Deconstruction of draw world objects from d3d render.

PR #999, opened 2024-11-07
Correction to bHotspot default value

sicoderealm:
Incorrectly set `bHotspot` value during deconstruction - needs to default to false not true.
This was causing incorrect rendering issues when another player had some equipment and used magic.
Also remove a useless comment that was added previously

https://github.com/Meridian59/Meridian59/pull/916/files#diff-006128dfe5afe1d5b5b42cb04abcb98e0b6c96c74c3e4c6ab70fd3177cb9021cL7447

### DIFF
--- a/clientd3d/d3drender_objects.c
+++ b/clientd3d/d3drender_objects.c
@@ -643,7 +643,7 @@ void D3DRenderOverlaysDraw(
 
 			for (list2 = *pRNode->obj.overlays; list2 != NULL; list2 = list2->next)
 			{
-				bHotspot = true;
+				bHotspot = false;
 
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender_objects.c
```diff
@@ -1204,7 +1204,7 @@ void D3DRenderOverlaysDraw(
 					}
 
 					*gameObjectDataParams.numObjects++;
-				TEMP_END2: // ewww we have a goto.. todo: lets remove
```
akirmse, line 1207: You can certainly reword this if you want, but having a TODO to remove a goto seems eminently reasonable.

PR #1002, opened 2024-11-08
Adjust alpha enablement to prevent unwanted blending world edges

sicoderealm:
We have some unwanted blending taking place between static world geometry and the skybox. This issue is most obvious at certain times of the day when the sky is bright on screens such as the outskirts of Barloque or the Flatlands.

To solve this we correct when and where alpha blending is enabled/disabled.

Before:
image
image

After:
image
image


clientd3d/d3drender_world.c
```diff
@@ -147,6 +141,9 @@ long D3DRenderWorld(
 	IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_CULLMODE, D3DCULL_NONE);
 	SetZBias(gpD3DDevice, 1);
 
+	// Disable alpha testing and blending
```
akirmse, line 144: Is this comment correct?  It looks like it only disables alpha testing, not blending.

PR #1003, opened 2024-11-08
Prevent flicking of textures

sicoderealm:
When we solved visual issues when a player recovers from blind https://github.com/Meridian59/Meridian59/pull/907 we introduced a new issue than can cause flicking textures on some screens.

We revert that change and install the alternative solution that was provided in the original PR to solve the recovery from blind issues and solve the new flickering texture issue.


akirmse:
What is the issue?  Why are textures flickering?

akirmse:
I can see that.  But do you know why the previous change was causing the textures to flicker?

akirmse:
I'll merge this one, but in general I'm very wary of including code that is not understood, as it tends to lead to more problems later.

PR #1005, opened 2024-11-09
Removal of render materials externs

sicoderealm:
Removing externs from render materials.

### DIFF
--- a/clientd3d/d3drender_materials.c
+++ b/clientd3d/d3drender_materials.c
@@ -423,6 +423,8 @@ Bool D3DMaterialObjectInvisiblePool(d3d_render_pool_new *pPool)
 	IDirect3DDevice9_SetSamplerState(gpD3DDevice, 1,
                                          D3DSAMP_MIPFILTER, D3DTEXF_NONE);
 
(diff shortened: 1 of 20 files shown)


clientd3d/d3drender_materials.c
```diff
@@ -427,9 +423,6 @@ Bool D3DMaterialObjectInvisiblePool(d3d_render_pool_new *pPool)
 	IDirect3DDevice9_SetSamplerState(gpD3DDevice, 1,
                                          D3DSAMP_MIPFILTER, D3DTEXF_NONE);
 
-	IDirect3DDevice9_SetTexture(
```
akirmse, line 430: This looks like a behavior change.

clientd3d/game.h
```diff
@@ -115,4 +115,6 @@ M59EXPORT player_info *GetPlayerInfo(void);
 M59EXPORT void SetActiveStatGroup(int stat_group);
 M59EXPORT int GetActiveStatGroup(void);
 
+M59EXPORT const room_type& getCurrentRoom();
```
akirmse, line 118: Not sure this should be M59EXPORT.  This means you are exporting it to modules and using it there, but I don't see any usage.

PR #1006, opened 2024-11-09
Correctly return from SengBackgroundObjects

sicoderealm:
Toko detected an error in the logs following https://github.com/Meridian59/Meridian59/pull/992/

```
[user.bof (2241)] SendBlakodMessage can't propagate MESSAGE SendBackgroundObjects (14377) in CLASS User (1)
```

We correct this by returning rather than propagating in SengBackgroundObjects.


akirmse:
Be sure took look for server errors anytime you make a Blakod change.

PR #1011, opened 2024-11-12
fix: Re-save roo file to clean up skybox creep in Ko'catan (`settle1`)

### LINKED ISSUE #1009: Ko'catan, the island settlement (walk through wall and hairlines)
You are able to walk inside this wall section (but not the opposite side of the door)
image
There are also seems to be plenty of hairlines between buildings throughout this screen, one example being:
image

davidklaw:
This PR addresses the skybox creep or gaps seen in #1009 by simply opening re-saving `settle1.roo`. These issues are hard to spot and not visible at all window sizes.

This PR originally took this a step further and fixed unclosed sector warnings by adjusting door geometry, but no longer does so.


davidklaw:
> Will this fix all buildings in the city with this issue?

It "fixes" the issue throughout the room--the gap remains, but is far less visible.

davidklaw:
I'm dropping this to a draft while I try to re-investigate what's going on here.

This screenshots illustrates the sector issue with the NW "door". When 1 point deep, this issue is flagged. When the door is simply made 2 points, it goes away. I cannot find issues with these doors that would lead to this. Most doors in this room are 2-4 points deep, but there are 1-point deep doors along the north edge of the map. However, they are not diagonal 🤷🏻‍♂️.

It would be nice to resolve these warnings without resorting to changing the door geometry, though if we must, I don't think anyone will notice.

image

davidklaw:
This PR addresses the issue without attempting to resolve the unclosed sector warning. Instead, I simply opened and re-saved the file in roomedit, which appears to have cleaned up some edges that were allowing the skybox to show through. A better fix will likely require addressing the sector warning directly.

davidklaw:
@Meridian59 This change is quite minor now. It’s just a simple open-and-save operation that doesn’t alter the geometry as the original change did (e.g., the doors). Saving the file appears to clean up many rough edges, especially where slopes meet level surfaces.

akirmse:
That makes sense, since it converts fixed point to floating point, which has more precision.

PR #1014, opened 2024-11-12
Adds 24-hour duration to DM-added weapon spell attributes

### LINKED ISSUE #1007: Error: `DM create itematt` generates errors
## What

- Creating an item with `dm create itematt holy` generates 2 server errors
>  `[weapon.bof (679)] InterpretBinaryAssign can't and 2 vars 1,0 and 0,0`
>  `[weapon.bof (667)] InterpretBinaryAssign can't or 2 vars 1,0 and 0,0`
- Viewing the item after adding the attribute also generates a server error
>  `[waspellt.bof (122)] C_AppendTempString can't set from NIL`

### Notes
- occurs occurs with the available `spelltypes`
    - fire
    - cold
    - shock
    - holy
    - unholy
    - acid
- only seems to occur the first time you do it to a weapon as a dm
  - speculating the other attributes are not added in subsequent commands

### Example
https://github.com/user-attachments/assets/b5c80c07-be24-4bc8-b7d8-6818b0e19668

### Expected Result
- using `dm create itematt <spelltype>` should not generate any errors
- looking at an item created with `dm create item att <spelltype> should not generate errors

davidklaw:
This PR addresses an issue with DM-added weapon attributes of the spell attack type. Currently, when these attributes are added via `dm create itematt holy`, for example, a timer is not provided. It's needed by a later check and its absence causes the errors seen in #1007.

These weapon attributes are randomly generated with a duration of 1-24 hours. For the DM version, I chose 24-hours as anything created by a god ought to last a while.

I also snuck in a small change to the spell type descriptions that get appended for consistency with all other text formatting.

### Testing
* `dm get item mace`
* Equip the mace
* `dm create itematt holy`
* Examine the mace to verify it's enchanted with a holy attack
* Verify the error logs are clear of issues

This fixes #1007

### DIFF
--- a/kod/object/passive/itematt/weapatt/waspellt.kod
+++ b/kod/object/passive/itematt/weapatt/waspellt.kod
@@ -20,7 +20,7 @@ WeapAttSpellType is WeaponAttribute
 %
 %      [WA_ATTACKSPELLTYPE, timer, spell_attack_type ]
 %   
-%       timer::the length of time, from 1 to 25 realtime hours,
+%       timer::the length of time, from 1 to 24 realtime hours,
 %         that the enchantment lasts (depending on spell factors).
 %       spell_attack_type : as defined in blakston.khd.
 %          examples: ATCK_SPELL_FIRE, ATCK_SPELL_UNHOLY, ATCK_SPELL_ACID
@@ -35,6 +35,9 @@ WeapAttSpellType is WeaponAttribute
 constants:
 
    include blakston.khd
+
+   ONE_HOUR_IN_MILLISECONDS = 3600000
+   ONE_DAY_IN_MILLISECONDS = 86400000
 
 resources:
    
@@ -193,7 +196,7 @@ messages:
       if send(self,@ReqAddToItem,#state1=iSpellType,#oItem=oWeapon)
       {
          send(self,@AddToItem,#oItem=oWeapon,#state1=iSpellType,
-              #timer_duration=86400000,#random_gen=TRUE);
+              #timer_duration=ONE_DAY_IN_MILLISECONDS,#random_gen=TRUE);
               
          return TRUE;
       }
@@ -265,7 +268,8 @@ messages:
       if send(self,@ReqAddToItem,#state1=iSpellType,#oItem=oItem)
       {
          send(self,@AddToItem,#oItem=oItem,#state1=iSpellType,
-              #timer_duration=random(3600000,86400000),#random_gen=TRUE);
+              #timer_duration=random(ONE_HOUR_IN_MILLISECONDS,ONE_DAY_IN_MILLISECONDS),
+              #random_gen=TRUE);
       }
       
       return;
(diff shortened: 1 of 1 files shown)


kod/object/passive/itematt/weapatt/waspellt.kod
```diff
@@ -191,8 +191,8 @@ messages:
       
       if send(self,@ReqAddToItem,#state1=iSpellType,#oItem=oWeapon)
       {
-         send(self,@AddToItem,#oItem=oWeapon, #state1=iSpellType,
-              #random_gen=TRUE);
+         send(self,@AddToItem,#oItem=oWeapon,#state1=iSpellType,
+              #timer_duration=86400000,#random_gen=TRUE);
```
akirmse, line 195: For big numbers like this it's generally better to do something like "60 * 60 * 24 * 1000" (or whatever) to make it clearer what's going on, or better yet, add a constant like "ONE_DAY_IN_MILLISECONDS" with that value.

kod/object/passive/itematt/weapatt/waspellt.kod
```diff
@@ -40,12 +40,12 @@ resources:
...
-   iaSpelltype_fire = "The weapon is hot to the touch."
-   iaSpelltype_cold = "It is icy cold to the touch."
-   iaSpelltype_shock = "Nearly invisible currents run the length of the blade."
-   iaSpelltype_acid = "A thin coating of acid covers the length of the blade."
+   iaSpelltype_holy = "  The weapon glows with a pure, white light."
+   iaSpelltype_unholy = "  The weapon seems to reek of malevolence."
+   iaSpelltype_fire = "  The weapon is hot to the touch."
+   iaSpelltype_cold = "  It is icy cold to the touch."
+   iaSpelltype_shock = "  Nearly invisible currents run the length of the blade."
+   iaSpelltype_acid = "  A thin coating of acid covers the length of the blade."
```
akirmse, line 48: What's the deal with these spaces?

davidklaw: Dumb move on my part. This is to create space between the base description and these modifier strings, but I need to move this down to where they're appended.

davidklaw: This has been streamlined in the latest commit. There's a bit of logic that follows to add the maker's string to the description that DOES NOT need this, as the spaces for that string are in `spell.kod:134`.

kod/object/passive/itematt/weapatt/waspellt.kod
davidklaw: Makes sense. Did the same for another `AddToItem` call in this class, from which I borrowed this parameter.

kod/object/passive/itematt/weapatt/waspellt.kod
```diff
@@ -36,6 +36,9 @@ constants:
 
    include blakston.khd
 
+   ONE_HOUR_IN_MILLISECONDS = 3600000
+   ONE_DAY_IN_MILLISECONDS = 86400000
```
akirmse, line 40: Same feedback here... 60 * 60 * 24 * 1000 is clearer.  No counting zeros to see if it's correct.  (And no performance penalty, as the compiler will simplify the expression.)

davidklaw: 👍🏻

PR #1015, opened 2024-11-13
Increment draw objects per leaf in draw bsp to allow more objects to be visible

sicoderealm:
When using the new renderer changes Mayhem noticed that objects are disappearing from view.
Checking the busier rooms you will see many `too many objects per leaf: throwing some away` debug messages.
We increase the number of leaf nodes to allow for even more objects to be visible in game.
We've been optimizing the graphics and rendering for the last few years but this value hasn't changed for at least 12 years.

With this change the error debug messages no longer appear and everything is visible.


clientd3d/draw3d.h
```diff
@@ -28,10 +28,10 @@
 
 #define LIGHT_NEUTRAL 192       // Light level of sector drawn at ambient light level
 
-// XXX Temporary maximum on # of objects in room 
+// Memporary maximum on # of objects in room
```
akirmse, line 31: Spelling.  And is this actually temporary?

clientd3d/drawbsp.c
```diff
@@ -23,7 +23,7 @@
 
 #include "client.h"
 
-#define MAX_OBJS_PER_LEAF 100
+#define MAX_OBJS_PER_LEAF 1000
```
akirmse, line 26: This is absurdly high; there is no reason to draw 1000 objects in one BSP leaf.  It could cause performance problems, though.

clientd3d/drawbsp.c
```diff
@@ -23,7 +23,7 @@
 
 #include "client.h"
 
-#define MAX_OBJS_PER_LEAF 100
```
akirmse, line 26: Increasing the max object limit is perhaps OK, but the O(N^2) bubble sort on line 2271 needs to be replaced if so.  Let's put in std::sort instead.

PR #1018, opened 2024-11-16
handle errors in saving in the server better

ckirmse:
not tested, unfortunately


akirmse:
Seems to work locally!

PR #1019, opened 2024-11-16
Various Portuguese client translation corrections, 'GPU Efficiency' option fix [keybind]

davidklaw:
This PR addresses various issues related to the Portuguese language translations. For the most part these issues are incorrect sizes that lead to cutoff text, but in a case or two it was the language itself that was updated with the help of Coloral, one of our Brazilian players.

One particular change of note is that this PR converts what I assume had to be a non-functioning _button_ for the GPU Efficiency configuration option into a proper checkbox.

For cataloging purposes, here are the updated UIs:

### 'Help' Option in Menu
image

### Character Creation
image
image
image

### Configuration
image
image
image
image
image

### Command Aliases
image


davidklaw:
Converting to a draft to allow time for any additional corrections.

PR #1020, opened 2024-11-17
Make the Chalice of the rain function like rescue and elusion

sicoderealm:
The Chalice was adjusted back in 2013 to allow all players (not just non-angled) to partake. However, it was changed to prevent all outlaw/red characters from using it instead. As per original follow up discussions in 2013: we correct the use of the Chalice to make it fair and consistent with rescue/elusion: https://github.com/Meridian59/Meridian59/pull/37#issuecomment-13698080

This brings a level playing field to the use of the chalice for all.

testing:

Confirmed working as expected when used on local server. Unable to chalice away and see new message until timer condition is met.


davidklaw:
+1

PR #927, opened 2024-08-26
Make Statue overlays visible

sicoderealm:
Statues currently do not display overlays (such as masks or helms) - we add that functionality.

Currently:
image

With these changes:
image
image
image

Linked to: https://github.com/Meridian59/Meridian59/pull/926

PR #1022, opened 2024-11-18
Upgrade existing statues

sicoderealm:
After we added the ability to show overlays on statues: https://github.com/Meridian59/Meridian59/pull/927

We found out that any existing statues that had overlays that were created before this upgrade would have a mismatch between the number of calculated overlays and the overlays sent to the client. We create a one time upgrade to solve this, fixing any existing statues that need to be recalibrated.

Testing:

Created a server pre-upgrade and reset the statues and saved the game state. Upgraded post https://github.com/Meridian59/Meridian59/pull/927 - got the bug and then ran this fix, saw the one time upgrade applied to all statues and the bug was gone.

### DIFF
--- a/kod/object/passive/statue.kod
+++ b/kod/object/passive/statue.kod
@@ -521,7 +521,6 @@ messages:
       local iSkin_xlat;
 
       % Send overlay bitmap info to user.  
-
       % Player has 8 standard overlays: right arm, left arm, legs, head, eyes, 
       % mouth, nose, hair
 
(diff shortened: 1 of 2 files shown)


akirmse:
I don't think this is the right place to put a fix.  It will run every time anyone enters a room with a statue, once per statue, forever.  Instead this should be a one-off conversion function (maybe in System) that we run once, then delete.

kod/object/passive/statue.kod
```diff
@@ -505,6 +505,17 @@ messages:
       return;
    }
 
+   UpgradeStatue()
```
akirmse, line 508: Add a comment that this function should be removed after it has been run.

PR #1023, opened 2024-11-21
Fix for Portuguese client account creation

davidklaw:
This addresses a critical issue with the Portuguese translations and tackles some housekeeping:

1. Adds Portuguese translations for the signup dialog, which do not currently exist. This includes incorrect entries for account creation endpoints that are currently causing account creation to render a false positive.
2. Replaces Portuguese string table numerical ids with their constant equivalent.
3. Replaces references to non-existent help pages with the Guides section of the website.

### DIFF
--- a/clientd3d/client.rc
+++ b/clientd3d/client.rc
@@ -523,6 +523,32 @@ BEGIN
     PUSHBUTTON      "Oferta...",IDC_BTN_DEMO,138,167,93,14
 END
 
+IDD_SIGNUP DIALOGEX 0, 0, 288, 253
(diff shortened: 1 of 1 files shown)


clientd3d/client.rc
```diff
@@ -566,195 +592,203 @@ END
...
 STRINGTABLE
 BEGIN
-    192                     "Nome de usu�rio � inv�lido ou j� est� em uso."
-    193                     "Endere�o de email inv�lido."
-    194                     "Endere�o de email j� est� em uso."
-    195                     "Senhas s�o inv�lidas ou n�o s�o iguais."
-    196                     "Servidor desconhecido selecionado."
-    197                     "O servidor de cadastro est� offline."
-    198                     "Cadastro gerou um erro desconhecido, por favor, tente novamente."
-    199                     "/api/signup/"
```
davidklaw, line 752: These couple of entries are the cause of the false positives.

PR #1024, opened 2024-11-22
Sets English as the neutral language and fallback in `.rc` files

davidklaw:
This PR sets English as the neutral and fallback language for all resource scripts.

This change resolves an issue where certain entries, such as API endpoints, were language-specific and are inhibiting registration for users of Portuguese translations. By setting English as the fallback, these critical entries are now universally available, ensuring consistent functionality across all languages.

Additionally, this adjustment applies to other entries, like rarity grades, which currently exist only in English. As a result, modifiers like "(rare)" will now appear in English for all clients until translations for Portuguese and German are added in a future PR.


akirmse:
Ooh, nice.  Did you double-check that the fallback works in keybind?  I have this memory that MFC does things differently.

And we should delete all of the URLs and other non-translatable stuff from the Portuguese and German translations, since they provide no benefit and can only get out of sync.

davidklaw:
This won't work with keybind and has been reverted. From my research, it seems like it would be much more involved. I'm not sure that the juice is worth the squeeze as all of the keybind strings _should_ be translated. As it stands, if an entry is missing, keybind will error.

davidklaw:
- Entries needed for API calls or that are universal, like most URLs, are now dependent on the English entries.
- keybind excluded from this change.

akirmse:
A few more:

IDS_HTTPKEY
IDS_HELPFILE
IDS_TROUBLESHOOT

Also please remove all of these from German (even if they don't have the matching symbolic name for the string; a lot of those just have the integer values instead of the IDS constants).

davidklaw:
Huh, I thought I caught those! Latest commit cleans these up:
| ID | English Value |
|---|---|
| IDS_CONFIGMENUEXE | "m59bind.exe" |
| IDS_HTTPKEY | "Software\\Classes\\http\\shell\\open\\command" |
| IDS_HELPFILE | "https://www.meridian59.com/guides/" |
| IDS_INIFILE | "meridian.ini" |
| IDS_TROUBLESHOOT | "https://meridian59.com/troubleshoot/" |
| IDS_HOMEPAGEURL | "https://www.meridian59.com/" |
| IDS_SIGNUPAPI | "create" |
| IDS_WEBAPIDOMAIN | "accounts.meridian59.com" |
| IDS_VERIFIEDAPI | "check_verified" |
| IDS_RESENDAPI | "resend" |

PR #1026, opened 2024-11-25
Overhauls resource files to to convert numerical IDs, and add and improve some translations

davidklaw:
This PR is significant in that it touches all of the `.rc` files again.

* It converts German's use of numerical IDs to named constants. It does the same for Portuguese in a few instances.
* It specifies _Brazilian_ Portuguese as the sublang.
* It corrects German encoding issues I came across.

It would be helpful if others could test this. Simply "Add a language" in Windows Language Settings.

image

When you toggle to another language you can switch your input method to English.
image


davidklaw:
In hindsight, I maybe shouldn't have reordered dialogs and stringtables. I did this to match English, but it's making comparison a nightmare.

clientd3d/client.rc
```diff
@@ -21,9 +21,11 @@
 /////////////////////////////////////////////////////////////////////////////
 // Brazilian resources
 
-#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_DEU)
-LANGUAGE LANG_PORTUGUESE, SUBLANG_NEUTRAL
+#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_PTB)
+#ifdef _WIN32
+LANGUAGE LANG_PORTUGUESE, SUBLANG_PORTUGUESE_BRAZILIAN
```
akirmse, line 26: Just double-checking on this.  I guess this would not match Portuguese from Portugal (pt-pt)?  Is that what we want?

davidklaw: I gathered a bit of feedback from a Brazilian player who stated that while there are differences, Portuguese (Brazilian) can be largely understood by Portuguese (Portugal) speaking players. Should there ever be an influx of players from Portugal, neutralizing this directive would provide them with something familiar without the need for patch cycles to introduce it. For that reason, I'm planning to revert these and display Portuguese (Brazilian) for both.

An interesting bit of feedback shared with me is that there might be a preference among some non-English speaking players to use the English settings. They've either become accustomed to them, and I imagine align best with documentation, guides, and content they're finding about the game online. We should consider adding a language picker in-game somehow.

I'll update the PR to make these neutral.

PR #1029, opened 2024-12-05
Fixes mail subject handling tied to recipient count

davidklaw:
This fixes a critical issue where mail subject handling was incorrectly tied to the number of recipients.

Previously, the subject line could be written in the wrong position. If there were more than three recipients, the client would crash.

To resolve this, I adjusted the logic by tracking the subject ID separately.

This supports blank subjects, which appeared to be a problem when this bug surfaced.

### DIFF
--- a/module/mailnews/mailfile.c
+++ b/module/mailnews/mailfile.c
@@ -133,7 +133,6 @@ void MailNewMessage(int server_index, char *sender, int num_recipients,
    auto subject_strs = std::vector<std::string>(num_subjects);
    char *subject = "";
    char *ptr;
-   int i, num;
    int subject_id_index = 0;
    char filename[FILENAME_MAX + MAX_PATH];
    char date[MAXDATE];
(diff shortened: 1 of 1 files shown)


module/mailnews/mailfile.c
```diff
@@ -134,6 +134,7 @@ void MailNewMessage(int server_index, char *sender, int num_recipients,
    char *subject = "";
    char *ptr;
    int i, num;
+   int subject_id_index = 1;
```
akirmse, line 137: Are you sure this should be 1 and not 0?  I think that means that the default in error cases will be German.

module/mailnews/mailfile.c
```diff
@@ -134,6 +134,7 @@ void MailNewMessage(int server_index, char *sender, int num_recipients,
    char *subject = "";
    char *ptr;
    int i, num;
```
akirmse, line 136: While we're at it, let's declare i inside the for loop on line 158 (and remove it here) so that we can't get into this buggy situation again.

PR #1031, opened 2024-12-10
Event-friendly version of the lich queen

Aesica:
This adds a monster to the game that is a copy of Queen Venya'cyr, except without any location restrictions or ties to the hunter/necromancer scenario.

It has the following minor changes compared to Queen Venya'cyr:

- Attack type is now physical (magical weapon + monster claw) instead of unholy.  This is to prevent cheesing her with resistance from a very high level holy resolve spell, which trivializes any damage she deals.  Her attacks can instead be mitigated by resist magic, but that can't go nearly as high as holy resolve.  (Xeos also use magical physical attacks and are quite tough)
- She no longer counter-blinds immortal DMs when they try to cast spells.
- Removed all code tied to the hunter/necromancer scenario and the necromancer guild hall.
- Fixed a bug where certain text wasn't being shown to the victor of a solo lich kill.
- She is no longer forced to remain in her room in Brax.
- She leaves an ordinary, decomposable corpse upon defeat.
- Her name has been changed to the more generic "ancient lich" and her description has also been updated to match.
- She cannot be affected by dispel illusion and is now always in her dispelled appearance.  This way, the actual Venya'cyr remains unique.
- Moved her spell resistance value from constants to properties, so DMs can adjust it as they see fit.
- Added AI_FIGHT_NEWBIESAFE to her behavior flags.
- Removed some unused resources.
- Fixed a typo which prevented her from having a proper death sound.
- Updated her definite/indefinite articles, as she is a generic monster instead of a named one.

image

Some things to note that go beyond the scope of this PR, but are worth addressing in another PR later:

- Many of the little fixes exist only in this version, as the original Venya'cyr is untouched.
- Despite being a lich, her only spell is blind and she is a melee attacker.  I feel like this should be expanded upon for both versions of her.  A few more spells so she can do something when out of melee range, etc.
- Her loot (like many bosses) is a bit weird and mixed.  There's some useless stuff in there, like scimitars and some things that make no sense for her to have, like ivy circlets.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/ancientlich.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/ancientlich.kod
@@ -1,0 +1,463 @@
+% Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
+% the LICENSE file that accompanies it.
+%
+% Meridian is a registered trademark.
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+AncientLich is Monster
+
+constants:
+
+   include blakston.khd
+   
+resources:
+
+   ancient_lich_name_rsc = "ancient lich"
+   ancient_lich_icon_rsc = lichb.bgf
+   ancient_lich_desc_rsc = \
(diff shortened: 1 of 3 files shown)


akirmse:
Overall there is a lot of duplicate code with Lich, so I wonder if you looked at inheriting, rather than copying the whole class and then modifying.

Aesica:
It occurred to me, but I wasn't sure how having Venya'cyr suddenly inherit from this version (the logical choice, given her greater complexity) would affect the game, if at all.  If it's a harmless thing to do, I can do it either in this PR or in a follow-up PR.

PR #1033, opened 2024-12-18
feature: re-add Tos and BQ hometown chance

adrienlaws:
# What

- re-added Tos and Barloque (BQ) as possible hometowns when a player leaves Raza through the portal
- kept old towns in and changed chance to be equal among all 5 towns (20%)

# Why

- a long time ago (maybe 10+ years?), Tos and BQ were removed as possible hometowns
- it appears that this was done to reduce pking of newbies outside of Tos
- that doesn't seem to apply any longer

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -6939,22 +6939,35 @@ messages:
    "This sets one of the many hometowns to be the initial hometown."
    {
       local rand;
+      rand = random(1,8);
       
-      rand = random(1,10);    
-      
-      if rand < 4
+      if rand = 1
       {
          piHomeroom = RID_MAR_INN;
-      }   
+      }
       else
       {
-         if rand > 7
+         if rand = 2
          {
-            piHomeroom = RID_JAS_INN;
+            piHomeroom = RID_COR_INN;
          }
          else
          {
-            piHomeroom = RID_COR_INN;
+            if rand = 3
+            {
+               piHomeroom = RID_JAS_INN;
+            }
+            else
+            {
+               if rand < 6
+               {
+                  piHomeroom = RID_BAR_INN;
+               }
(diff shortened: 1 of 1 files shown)


akirmse:
Why were Tos and Barloque removed as starting towns?  Are those reasons
still relevant?

adrienlaws:
I'm not sure.  Perhaps @davidklaw or @jrotunda can comment.  I believe it's been this way for at least 12 years.

Aesica:
Since Tos and Barloque are the most commonly visited towns, it might actually make more sense for them to be the default hometown choices--especially Tos.  Most people seem to change their hometown to Tos anyway.  It would allow new players to land in an area populated by other players, rather than being cast off into the often-empty towns of Jasper or Marion to find their way around alone.

adrienlaws:
I change this PR to make all 5 hometowns equally likely to be assigned (20% chance).

It might be that eventually we offer an option to enter the world by walking out of Raza, with the portal as a fun "random" option if people want that.

If this PR doesn't make any sense anymore (now that I have more context) I can close it.

davidklaw:
> I'm not sure. Perhaps @davidklaw or @jrotunda can comment. I believe it's been this way for at least 12 years.

Just noting that I don't know. This would be before my time.

adrienlaws:
Closing this as the intent has kind of changed.

The description of the `SetRandomHomeroom()` function says
> "This sets one of the many hometowns to be the initial hometown."

So I originally felt that this PR with 20% chance for each of the 5 towns makes sense, instead of excluding Tos and Barloque.

However, it's become clear we're talking about new player progression after Raza instead of just adjusting the function.  The Raza portal is what calls this function to determine where to drop the player when they leave.  I think some of the suggestions of having a map connection to Raza like on other servers could make sense.

That is a bit out of scope of my original intent of just removing missing towns, so I think we can close this for now and revisit this PR if needed.

PR #1038, opened 2024-12-24
feature: add Codespell kod folder scanning GitHub action

adrienlaws:
# What

- adds [Codespell with GitHub Actions](https://github.com/codespell-project/actions-codespell) to the build pipeline
  - uses a parameter to only run in the `/kod/` directory and subdirectories
- adds a `.codespellignore` file to avoid false positives
  - these all flagged as positive (count of 232 instances)

# Why

- Codespell is a well regarded tool and helped with #1037 
- it can catch things in the build pipeline before they're implemented
- limited scope for now but could be expanded if it is useful
- very easy to modify by editing the ignore file or just disabling the GitHub action
- very inexpensive to run (seems to be a few seconds)

# Notes

- Codespell is not a spell checker for everything
- Codespell has a built-in dictionary of common misspellings and will flag on them (it will not flag on gibberish like "asjdklfasld")
- some words in the ignore file may seem strange, but they are true false positives that do flag without the ignore file


adrienlaws:
It seems that the action ran in this PR.  No issues detected other than encoding on certain `.kod` files.

image

.codespellignore
```diff
@@ -0,0 +1,32 @@
+asssign
+commend
+crate
+crystalize
```
akirmse, line 4: Can you add some comments (lines starting with #) in this file to explain why some misspellings have to be allowed (e.g. "CrystalizeMana", "Villify" spell, etc.)?

.codespellignore
```diff
@@ -1,3 +1,9 @@
+# The following words are intentionally added to the ignore list because they are valid Meridian 59 codebase words.
```
akirmse, line 1: I was hoping for more specific comments on individual words.  For example, next to "Crystalize Mana", put a comment that this is a spell whose name is misspelled in the game.  For "asssign", say that this is a class for a sign for the assassin's game (I think?).  You don't need a general comment saying that Meridian 59 is a game that includes unique terms---you can assume the reader knows this.  Put yourself in the place of someone new to the codebase.  They come across "asssign" and they are confused as to why this is not a misspelling.

You don't necessarily need to comment on every single word, just the ones that are confusing.

adrienlaws: Understood, I refactored it based on your feedback.

I had to do comments on the following line, because when I included "#" after a word, it would not filter that word anymore.  Putting the comment on the following line seems to work.

akirmse:
Sorry for all the back and forth, but it's usual to put comments before the line they apply to.

Issue #1039, opened 2024-12-24
Bug: Unable to click sun or moon


sicoderealm:
Unable to click sun and moon since orientation was corrected

PR #1040, opened 2024-12-24
Mirror background textures on the x-axis

sicoderealm:
Adjust the texture coordinates (s and t) of the vertices of a background overlay to mirror the texture on the x-axis.
Since [rotating the backgrounds](https://github.com/Meridian59/Meridian59/pull/1016) to match the origin image orientations, the background objects were no longer clickable. To prevent this, we revert the change and instead we swap the s coordinates of the vertices. The background objects are once again clickable.

before:
image

after:
image

Fixes: #1039

### DIFF
--- a/clientd3d/d3drender_bgoverlays.c
+++ b/clientd3d/d3drender_bgoverlays.c
@@ -138,6 +138,8 @@ void D3DRenderBackgroundOverlays(const BackgroundOverlaysRenderStateParams& bgoR
 		static const float u[4] = { epsilon, 1.0f - epsilon, 1.0f - epsilon, epsilon };
 		static const float v[4] = { epsilon, epsilon, 1.0f - epsilon, 1.0f - epsilon };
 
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender_bgoverlays.c
```diff
@@ -142,13 +138,13 @@ void D3DRenderBackgroundOverlays(const BackgroundOverlaysRenderStateParams& bgoR
 		static const float u[4] = { epsilon, 1.0f - epsilon, 1.0f - epsilon, epsilon };
 		static const float v[4] = { epsilon, epsilon, 1.0f - epsilon, 1.0f - epsilon };
 
-		pChunk->st0[0].s = u[0];
+		pChunk->st0[0].s = u[1];
```
akirmse, line 141: Can you add a comment here explaining why the texture coordinates aren't in the expected order?

PR #1043, opened 2024-12-27
Remove erroneous Graveyard of Tos wall texture

sicoderealm:
While the front view looking towards the exit in the Graveyard looks fine:
image

However, there is an erroneous wall texture as you enter, we remove it to prevent this inconsistency.

Before:
image

After:
image


davidklaw:
Nice!

PR #1046, opened 2025-01-02
Enable gpu efficiency mode by default

sicoderealm:
We cannot guarantee or know upfront the performance of players machines - to ensure maximum compatibility and performance for new players we enable gpu efficiency by default.

To opt in for high quality graphics, players will now need to manually go to the configuration settings and untick the gpu efficiency option (requires some patch notes).

Testing:
- GPU efficiency mode one time enablement flip takes place successfully, after which manual changes are successful.
- GPU efficiency now enabled by default when config.ini doesn't contain an entry for `gpuefficiency`.
- GPU efficiency can be deselected and will appear correctly in config.ini (correctly disabling gpu efficiency mode).
- GPU efficiency can be toggled on and off correctly and is reflected in config.ini (reflecting the mode correctly).
- Tested successfully against my steam installation setup

### DIFF
--- a/clientd3d/config.c
+++ b/clientd3d/config.c
@@ -109,7 +109,7 @@ static char INITechnical[]    = "Technical";
 
 /* config.ini file entries (preferences) */
 static char config_section[] = "config";  /* Section for configuration stuff */
-static char INIGpuEfficiency [] = "gpuefficiency";
+static char INIGpuEfficiency[] = "gpuefficiency";
 static char INIGpuEfficiencyOneTimeFlip[] = "gpuefficiencyonetimeflip";
 
 static char INITextAreaSize[] = "TextAreaSize";
@@ -184,7 +184,6 @@ void ConfigInit(void)
    ini_file = ini_filename;
 
    snprintf(config_ini, sizeof(config_ini), "%s%s", dir, GetString(hInst, IDS_INIFILECONFIG));
-
 }
 /****************************************************************************/
 /*
@@ -326,7 +325,7 @@ void ConfigLoad(void)
       // This is to ensure players start from a clean slate with maximum performance.
       // After this one-time flip, future changes to gpu efficiency preferences will be respected.
       config.gpuEfficiency = true;
-      WriteConfigInt(config_section, INIGpuEfficiencyOneTimeFlip, config.gpuEfficiency, config_ini);
+      WriteConfigInt(config_section, INIGpuEfficiencyOneTimeFlip, True, config_ini);
       WritePrivateProfileString(config_section, INIGpuEfficiency, "true", config_ini);
    }
 


akirmse:
Do we also need to do a one-time flip of all current users to enable GPU efficiency?  There might be a lot of broken players out there.

clientd3d/config.c
```diff
@@ -102,6 +102,11 @@ static char INIDebug[]        = "Debug";
 static char INISecurity[]     = "Security";
 static char INITechnical[]    = "Technical";
 
+static char config_ini_file[] = "./config.ini";
```
akirmse, line 105: You definitely do not want to hard-code and redefine the location of this file, which is already available as ini_file.

clientd3d/config.c
```diff
@@ -294,9 +299,23 @@ void ConfigLoad(void)
    config.active_stat_group = GetConfigInt(misc_section, INIActiveStatGroup, 5, ini_file);
 
    // Determine if we should be using gpu efficiency mode or not.
-   char config_value[255];
-   GetPrivateProfileString("config", "gpuefficiency", "error", config_value, 255, "./config.ini");
-   config.gpuEfficiency = (0 == strcmp(config_value, "true"));
+   auto one_time_gpu_efficiency_enablement = GetConfigInt(config_section, 
+	   "gpuefficiencyonetimeflip", False, config_ini_file);
+   if (!one_time_gpu_efficiency_enablement)
```
akirmse, line 304: Get rid of the ! and exchange the then and else clauses.

clientd3d/config.c
```diff
@@ -294,9 +299,23 @@ void ConfigLoad(void)
...
+      // This is to ensure players start from a clean slate with maximum performance.
+      // After this one-time flip, future changes to gpu efficiency preferences will be respected.
+      config.gpuEfficiency = true;
+      WriteConfigInt(config_section, INIGpuEfficiencyOneTimeFlip, config.gpuEfficiency, config_ini_file);
+      WritePrivateProfileString(config_section, INIGpuEfficiency, "true", config_ini_file);
+   }
+   else
+   {
+      char config_value[255];
+      GetPrivateProfileString(config_section, INIGpuEfficiency, "true", config_value, 255, config_ini_file);
```
akirmse, line 316: sizeof(config_value) rather than 255.  Also 255 seems rather large.

clientd3d/config.c
```diff
@@ -294,9 +299,23 @@ void ConfigLoad(void)
...
 
    // Determine if we should be using gpu efficiency mode or not.
-   char config_value[255];
-   GetPrivateProfileString("config", "gpuefficiency", "error", config_value, 255, "./config.ini");
-   config.gpuEfficiency = (0 == strcmp(config_value, "true"));
+   auto one_time_gpu_efficiency_enablement = GetConfigInt(config_section, 
+	   "gpuefficiencyonetimeflip", False, config_ini_file);
+   if (!one_time_gpu_efficiency_enablement)
+   {
+      // Perform one-time flip to enable GPU efficiency mode regardless of current preference.
```
akirmse, line 306: Add a TODO with the date you added it, mentioning that this should be removed after the next update (since essentially all existing users will presumably run it after that).

clientd3d/config.c
```diff
@@ -102,6 +102,11 @@ static char INIDebug[]        = "Debug";
 static char INISecurity[]     = "Security";
 static char INITechnical[]    = "Technical";
 
+static char config_ini_file[] = "./config.ini";
+static char config_section[] = "config";
```
akirmse, line 106: Add a comment specifying what this section is for (see the other section definitions).

clientd3d/config.c
akirmse: I'd like you to research this some more and test it in both Steam and Web installer configurations (they run in different directories).  Notice how the main config file has its directory set explicitly:

https://github.com/Meridian59/Meridian59/blob/master/clientd3d/config.c#L172

So I'm not convinced that hard-coding the path to the file will work in all cases.

Also, given that this situation with multiple config files is confusing, you need to document it.

clientd3d/config.c
```diff
@@ -21,18 +21,22 @@
 Config config;
 char inihost[MAXHOST];
 
-// Full pathname of INI file
+// Full pathname of meridian.ini (maintained in-game via the Game>Preferences menu)
```
akirmse, line 24: Remove reference to particular menu item.  If the name of that menu item ever changes, no one is going to remember to update this comment.

clientd3d/config.c
```diff
@@ -21,18 +21,22 @@
...
 static char ini_filename[MAX_PATH + FILENAME_MAX];
 char *ini_file;  // Pointer to ini_filename
 
+// Full pathname of config.ini (currently maintained by m59bind)
+static char config_ini_filename[MAX_PATH + FILENAME_MAX];
+char *config_ini_file;  // Pointer to config_ini_filename
```
akirmse, line 30: Why do we need two variables pointing at the same buffer?

clientd3d/config.c
```diff
@@ -83,25 +87,30 @@ static char INIShow[]        = "Show";
...
 
-static char users_section[]  = "Users";  /* Section for dealing with other users */
+static char users_section[]  = "Users";  /* Section for dealing with other users in meridian.ini */
 static char INIDrawNames[]   = "DrawNames";
 static char INIIgnoreAll[]   = "IgnoreAll";
 static char ININoBroadcast[] = "NoBroadcast";
 static char INIIgnoreList[]  = "IgnoreList";
 
-static char special_section[] = "Special";  /* Section for hidden stuff in INI file */
+static char special_section[] = "Special";  /* Section for hidden stuff in in meridian.ini */
```
akirmse, line 104: You don't need to put the filename in every comment here; it's redundant.  Instead beef up the comment at the top, making it clear that there are two different ini files.

clientd3d/config.c
```diff
@@ -294,9 +308,25 @@ void ConfigLoad(void)
...
+      GetPrivateProfileString(config_section, INIGpuEfficiency, "true", config_value, 
+		  sizeof(config_value), config_ini_file);
+      config.gpuEfficiency = (0 == strcmp(config_value, "true"));
+   }
+   else
+   {
+      // TODO: Added January 2025 - remove after the next update when everyone will have been reset.
+      // Perform one-time flip to enable GPU efficiency mode regardless of current preference.
+      // This is to ensure players start from a clean slate with maximum performance.
+      // After this one-time flip, future changes to gpu efficiency preferences will be respected.
```
akirmse, line 325: Should one_time_gpu_efficiency_enablement also be removed at that time?  Make it easy on the future maintainer (who could be you).

clientd3d/config.c
```diff
@@ -102,6 +107,11 @@ static char INIDebug[]        = "Debug";
 static char INISecurity[]     = "Security";
 static char INITechnical[]    = "Technical";
 
+/* config.ini file entries (preferences) */
+static char config_section[] = "config";  /* Section for configuration stuff */
+static char INIGpuEfficiency [] = "gpuefficiency";
```
akirmse, line 112: remove extra space

clientd3d/config.c
```diff
@@ -160,17 +170,21 @@ void SaveSettings(void)
...
 
-   //memset(&config,0,sizeof(config));
    GetGamePath( dir );
+
+   // Retrieve the filenames for meridian.ini (preferences) and config.ini (configuration)
    snprintf(ini_filename, sizeof(ini_filename), "%s%s", dir, GetString(hInst, IDS_INIFILE));
    ini_file = ini_filename;
+
+   snprintf(config_ini, sizeof(config_ini), "%s%s", dir, GetString(hInst, IDS_INIFILECONFIG));
+
```
akirmse, line 187: Remove extra blank line.

clientd3d/config.c
```diff
@@ -294,9 +308,27 @@ void ConfigLoad(void)
...
+   else
+   {
+      // TODO: Added February 2025 - remove after the next update when everyone will have been reset.
+      // Also remove the `one_time_gpu_efficiency_enablement` variable and associated check above.
+
+      // Perform one-time flip to enable GPU efficiency mode regardless of current preference.
+      // This is to ensure players start from a clean slate with maximum performance.
+      // After this one-time flip, future changes to gpu efficiency preferences will be respected.
+      config.gpuEfficiency = true;
+      WriteConfigInt(config_section, INIGpuEfficiencyOneTimeFlip, config.gpuEfficiency, config_ini);
```
akirmse, line 329: Weird to use config.gpuEfficiency here.  Shouldn't this be simply "True"?

PR #1049, opened 2025-01-02
Add fire arrow visual fx

sicoderealm:
We make fire arrows have a dull red glow, similar to fireball spells.
This provides some additional pleasure to using fire arrows in dark areas.

Video updated following recent flickering lighting updates:
https://github.com/user-attachments/assets/d9a58b06-ef99-41e8-b705-b5783f4a7eac


kod/object/item/passitem/numbitem/ammo.kod
```diff
@@ -81,8 +86,17 @@ messages:
...
-      return LIGHT_FLAG_NONE;
+      return viProjectileLightFlags;
+   }
+
+   GetProjectileLightIntensity()
+   {
+      return viProjectileLightIntensity;
+   }
+
+   GetProjectileLightColor()
```
akirmse, line 97: I'm not sure these extra functions are needed.  Looks like your new code calls only GetProjectileLightFlags?

kod/object/item/passitem/numbitem/ammo.kod
```diff
@@ -35,7 +35,12 @@ classvars:
    viAmmo_type = AMMO_ARROW     
 
    viGround_group = 2
-   viInventory_group = 1   
+   viInventory_group = 1
+
+   viProjectileFlag = PROJ_FLAG_NONE
```
akirmse, line 40: Was LIGHT_FLAG_NONE.. which is correct?

kod/object/item/passitem/numbitem/ammo/firearow.kod
```diff
@@ -46,6 +46,15 @@ properties:
    piNumber = 10
    piAttack_spell = ATCK_SPELL_FIRE
 
+   viProjectileFlag = PROJ_FLAG_LIGHT_SOURCE
+
+   % Projectile lighting information.
+   viProjectileLightFlags = LIGHT_FLAG_ON | LIGHT_FLAG_DYNAMIC
+   % Medium light range, out of 255 (not as bright as a fireball)
+   viProjectileLightIntensity = 50
```
akirmse, line 54: A little bright for my taste from the video.. any other opinions?

akirmse (commented):
Tested on software renderer?

akirmse:
I am super confused.  I asked about a Blakod function, and now there is a new C lighting scaling function filled with magic numbers.  Where did this come from and why was it necessary?

jrotunda:
This looks great. And yes, the red here looks more like a heated element than the more bright orange 'fire' color. I should add that Fire arrows are a newer ammo type that are not yet entered into the code to appear in-game. Should other elemental projectiles be added, this code could serve as a functional foundation.

> I've`` updated the arrow to use LIGHT_DRED as the colour (see video) - 

> call.

> > there is a new C lighting scaling function filled with magic numbers. Where did this come from and why was it necessary?
> 
> I added some noticeable scaling for the dynamic lights based on the object intensity. I've reverted that - perhaps we can bring that back in another PR. 
> 

There was an issue present with the 2003 engine updated that added dynamic lighting in regards to intensity when the lights are close to textures. It has become more noticeable with further upgrades to the graphics. 

I'm going to add some new issues separately with some examples.

Other than that, this one looks ready for a review...?

sicoderealm:
I going to do some more testing on this one this weekend

PR #1051, opened 2025-01-03
Enable Invisible to Fading spell flow

sicoderealm:
It turns out that you cannot cast fade while invisible. If there are no negative consequences of doing so - this would be neat to enable as will open up some interesting gameplay mechanics for these existing spells.

https://discord.com/channels/650824489885630475/1099485818797703178/threads/1324500688608165939

### DIFF
--- a/kod/object/passive/spell/fade.kod
+++ b/kod/object/passive/spell/fade.kod
@@ -25,8 +25,6 @@ resources:
       "you invisible.  The spell lasts as long as your trance "
       "is maintained.  "
       "Requires dark angel feather."
-   
-   fade_already_enchanted = "You are already invisible."
 
    fade_on = "Your physical form fades from existence leaving only a shimmering apparition."
    fade_off = "Your physical form is visible once more."


akirmse:
I'm not so sure about this---there's explicit code that was added to disallow this.  Surely it's there for a reason?

Meridian59:
What happens in the opposite case of casting invisibility when faded?

kod/object/passive/spell/fade.kod
```diff
@@ -69,12 +69,12 @@ messages:
...
    {
-      % check for enchantment already applied
+      local oIvisibilitySpell;
+      % Remove existing ivisibility spell if it exists
       if Send(who,@CheckPlayerFlag,#flag=PFLAG_INVISIBLE)
       {
-         Send(who,@MsgSendUser,#message_rsc=fade_already_enchanted);
-         
-         return FALSE;
+         oIvisibilitySpell = Send(SYS,@FindSpellByNum,#num=SID_INVISIBILITY);
```
akirmse, line 76: It looks like PFLAG_INVISIBLE does not imply the invisibility spell.  For example, Fade itself sets this player flag.  Is it a good idea to attempt to remove the enchantment if it isn't applied?  With more research, it looks like it might be safe, but I'm not 100% sure.

kod/object/passive/spell/fade.kod
```diff
@@ -69,12 +69,12 @@ messages:
 
    CanPayCosts(who=$,lTargets=$)
    {
-      % check for enchantment already applied
-      if Send(who,@CheckPlayerFlag,#flag=PFLAG_INVISIBLE)
+      local oIvisibilitySpell;
+      % Remove existing ivisibility spell if it exists
```
akirmse, line 73: spelling.. c'mon man

davidklaw:
While I’m not necessarily opposed to this change, there’s something particularly underhanded about entering a room invisible and then transitioning to a “persistent” invisibility state. Perhaps players should need to strategically position themselves and activate fade _or_ enter rooms invisible and act quickly before their invisibility wears off. It’s not a major concern for me, but perhaps that’s why it was designed this way originally.

PR #1052, opened 2025-01-03
Reclaim arrows

sicoderealm:
Working fire and archery is time consuming and expensive.
This is perhaps one of the most grindy experiences in the game, even when using a practice bow.
You will likely need additional built characters or mules to help you finish these skills.

To mitigate this we add a chance for arrows, especially when fired from a practice bow, to be reclaimable. These reclaimable arrows will be dropped when a monster is killed, in addition arrows can also be reclaimable from other players!

To start with the following simple reclaimable calculations are used:

- At 99% a practice bow will make an arrow reclaimable 50% of the time (based on the players Fire ability) but the damage and range of the practice bow has been significantly reduced.
- When using a non-practice bow (e.g. long or battle bow) there is still a chance to reclaim arrows (based on the players Archery ability) at most 20% of the time.

A similar idea was also suggested in 2023 in the discord feedback and ideas channel (I only noticed this on 11th January 2025):
`Archery Improvement: Arrow Retrieval` https://discord.com/channels/650824489885630475/1099485818797703178/threads/1105207838390882336

https://github.com/user-attachments/assets/72194a22-042f-4878-a4dd-32bac8cbe776


akirmse:
I'm closing this on principle because it contains undocumented changes to audio and the strength of practice bows.  Those kinds of surprises are dangerous---if I accidentally let one of these in and it causes a problem later, it will be very difficult to debug because there's no description of it anywhere.

Please split up changes sensibly and document them completely.

PR #1056, opened 2025-01-05
Replace typographer’s apostrophe with standard apostrophe in NPC dialog

davidklaw:
This PR updates a handful of apostrophes in the NPC dialog tab and txt files by replacing typographer’s apostrophes (’) with the more commonly used standard apostrophe (').


davidklaw:
> Nice, thanks. I wonder if we can have the codespell pre-commit hook run on these files (or at least the .tab). Bonus points if we can do it in the existing action, without having to add a new one to hand the different directory.

It seems this sort of thing is outside of the scope of codespell, which only appears to be geared towards misspellings. We could introduce a prettier or prettier-like pre-commit check for this and more, if interested.

Meridian59:
Yeah, I was actually thinking of using it to catch misspellings in NPC
dialog.

davidklaw:
Ah, I don’t think so. Codespell doesn’t natively support regex for narrowing checks to quoted text, and I haven’t found a way to ignore color flags like `~r`. For example, text like `~rin`—where the `~r` indicates a Riija monk speaking in color—is treated as “rin” instead of “in”. @adrienlaws: I don't know if your experience with codespells suggests something different (we're talking about running it against `npc.tab`.

adrienlaws:
I just submitted PR #1061 as an attempt.  There is a `codespell` command line argument you can provide but I haven't tested it yet.  That probably makes the most sense for ignoring the first character after a tilda.

PR #1057, opened 2025-01-05
Add conditional seller attribute to Hanla zax'Ta

sicoderealm:
Hanla zax'Ta has a conditional sale of Kriipa claws - however it is currently broken.

We add the conditional seller attribute `MOB_COND_SELLER` that will enable this functionality. The `MOB_COND_SELLER ` attribute is required for the seller to have their `InitCondSale` called, enabling any conditional sales (via monster Constructor or library ReInitialize)

Will require a `send o 0 RecreateAll` from server side to enable from an existing setup.


akirmse:
We're not going to run a recreateall just for this, but I don't mind adding the flag.

PR #1059, opened 2025-01-05
Hanla zax'Ta conditional arrow sales

sicoderealm:
A typical statement from Hanla zax'Ta : `Hanla says, "I make different types of arrows."`
https://github.com/Meridian59/Meridian59/blob/master/dlg/dialog.txt#L5072

Rather than just selling standard arrows only - Hanla zax'Ta will now conditionally sell each arrow type, with a nice description of each in her artisan manner.


Aesica:
Are you sure we should be making special arrow types commonly available like this?

akirmse:
I don't know whether this is a good idea or not.  In the future I'd suggest coming to consensus first, rather than dropping a balance change like this out of the blue.

PR #1060, opened 2025-01-06
feature: add npc.tab to codespell checks

adrienlaws:
# What

- add `npc.tab` as part of the `codespell` GitHub Actions workflow

# Why

- extending spellchecking beyond `.kod` files to include additional NPC chatter

# Notes

- I tested it locally and found some things, but the true test is once the pipeline runs (hence why I need to submit this PR)


adrienlaws:
I messed up this PR, I used my old branch by accident instead of my newly created one.  Please disregard, sorry.

PR #1061, opened 2025-01-06
feature: add npc.tab and regex filtering to codespell spellchecker

adrienlaws:
# What

- add `npc.tab` as part of the `codespell` GitHub Actions workflow
- changed codespell to install on github runners and use command line arguments
- added `--ignore-regex "[~`][a-zA-Z]+"` to ignore Meridian 59 specific text formatting ("~I", "~b", etc)

# Why

- extending spellchecking beyond `.kod` files to include additional NPC chatter
- the codespell GitHub action doesn't have all the available configuration options that the command line tool does
- adding in regex means we don't have to ignore as many words

# Notes

- I tested it locally and found some things, but the true test is once the pipeline runs (hence why I need to submit this PR)
- renamed a few things in the workflow now that codespell checks more than `.kod` files


adrienlaws:
I'm tryingto find a way to incorporate regex to ignore M59 specific text formatting (~I, ~r, etc)

In `codespell` you can do this with regex, but not with the `codespell` GitHub action.  It looks like there is a workaround mentioned [here](https://github.com/codespell-project/actions-codespell/issues/42) that requires you to create a local codespell configuration file in the repository.  That configuration file can contain regex.  I will try this.

akirmse:
Nice!  Can you fix the misspellings in npc.tab while you're at it, so that the autobuild will be green for future PRs?

I'm also curious why the file isn't utf-8.  We should fix that.

davidklaw:
Dang! Way to go.

adrienlaws:
This is ready for review.
- added `npc.tab` to the codespell checking path
- added regex to ignore M59 specific text formatting (so we don't have to add words to the ignore list like "~iits")
- fixed all the typos in `npc.tab` and added valid words to the `.codespellignore` file
- reworked `codespell` to install via pip and use command line arguments instead of the `codespell` github action since the command line tool can do better arguments and configurations 

Maybe we could do the UTF-8 fix for `npc.tab` in another PR since this one has multiple fixes in it already.

akirmse:
This is great, thanks for plowing through this!
 
For one or more follow-up PRs:
- Can we move .codespellignore out of the top level?  It's kind of litter there.
- We need to regenerate dlg\dialog.txt by running a Python script (see dlg\README.txt)

adrienlaws:
Sounds good, I will submit follow-on PRs to move the ignore file out of the top-level directory and to regenerate `dialog.txt`.

Is this PR okay to merge now, or is there something else I should address?

akirmse:
I sent some grammatical comments last time that haven't been addressed.

PR #1065, opened 2025-01-10
fix: MOTD `CenterWindow` function in `blakserv` win interface

adrienlaws:
# What

- makes MOTD pop-up dialog box appear centered on the `blakserv` application window on Windows
- modifies `CenterWindow` function in`blakserv` Windows application
- adds support for multiple monitors

# Why

- #608 fixed issues with the Meridian 59 client, this adds similar fixes to the Meridian 59 `blakserv` application
- Fixes the MOTD dialog box appearing on the wrong screen 
    - for example, screen 1 has the application but the MOTD dialog box will open on screen 2

# Notes

- The "Are you sure you want to exit?" dialog box will appear on the correct monitor, but centered on the desktop not the application
  - I looked into fixing this, but that message box is a system-managed dialog box which is difficult to modify
    - I tried a few workarounds but since it does open on the correct monitor, it seems ok
- I also looked into trying to have the application "remember" where the application was placed on the desktop, so it would re-open in the same spot
  - I looked into how to save the position in the registry but it also seemed like I was adding too much complexity since the server should really be run from Linux in production

# Examples

### Before
- MOTD appears on first monitor, always, even when the application is on the second monitor
https://github.com/user-attachments/assets/a0885076-6a85-4151-8703-9dad9a25b5c1

### After
- MOTD opens centered on the application
https://github.com/user-attachments/assets/c96e9f79-8d92-40a0-97ec-df24d0aa290e

### DIFF
--- a/blakserv/interface.c
+++ b/blakserv/interface.c
@@ -586,7 +586,6 @@ void StartupComplete()
 void InterfaceCreate(HWND hwnd,WPARAM wParam,LPARAM lParam)
 {
 	SetWindowText(hwnd,BlakServNameString());
-	CenterWindow(hwnd, NULL);
 }
 
 void InterfaceCreateTabControl(HWND hwnd) 
--- a/blakserv/interface.c
+++ b/blakserv/interface.c
@@ -22,7 +22,6 @@
 	
 */
 
-#define NOMINMAX
 #include "blakserv.h"
 #include <mmsystem.h>
 #include <windowsx.h>
--- a/blakserv/interface.c
+++ b/blakserv/interface.c
@@ -1046,12 +1046,9 @@ void CenterWindow(HWND hwnd, HWND hwndParent)
     RECT rcDlg, rcParent, rcScreen;
     int x, y;
 
-    #define min(a, b) ((a) < (b) ? (a) : (b))
-    #define max(a, b) ((a) > (b) ? (a) : (b))
-
     if (hwndParent == NULL)
         hwndParent = GetDesktopWindow();
-  
+
     GetWindowRect(hwndParent, &rcParent);
     GetWindowRect(hwnd, &rcDlg);
 
@@ -1071,8 +1068,8 @@ void CenterWindow(HWND hwnd, HWND hwndParent)
         rcScreen = mi.rcWork;  // Use working area excluding taskbar
 
         // Adjust position to ensure the child window is fully visible
-        x = max(rcScreen.left, min(x, rcScreen.right - rcDlg.right));
-        y = max(rcScreen.top, min(y, rcScreen.bottom - rcDlg.bottom));
+        x = std::max(rcScreen.left, std::min(x, rcScreen.right - rcDlg.right));
+		y = std::max(rcScreen.top, std::min(y, rcScreen.bottom - rcDlg.bottom));
     }
     else
     {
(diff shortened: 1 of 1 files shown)


blakserv/interface.c
```diff
@@ -1027,50 +1028,78 @@ void InterfaceReloadSystem()
...
-	screen_height = GetSystemMetrics(SM_CYSCREEN);
-	
-	x = std::max(0, std::min(x, (int) (screen_width  - rcDlg.right)));
-   y = std::max(0, std::min(y, (int) (screen_height - rcDlg.bottom)));
-	
-	SetWindowPos(hwnd, NULL, x, y, 0, 0, SWP_NOSIZE | SWP_NOACTIVATE);
+    RECT rcDlg, rcParent, rcScreen;
+    int x, y;
+
+    #define min(a, b) ((a) < (b) ? (a) : (b))
```
akirmse, line 1050: Ugh... use std::min and std::max instead of macros.

blakserv/interface.c
```diff
@@ -1027,50 +1028,78 @@ void InterfaceReloadSystem()
...
+        int screen_height = GetSystemMetrics(SM_CYSCREEN);
+
+        x = (screen_width - rcDlg.right) / 2;
+        y = (screen_height - rcDlg.bottom) / 2;
+    }
+
+    SetWindowPos(hwnd, NULL, x, y, 0, 0, SWP_NOSIZE | SWP_NOACTIVATE);
+
+    #undef min
+    #undef max
```
akirmse, line 1091: Remove these.

blakserv/interface.c
```diff
@@ -1027,50 +1028,78 @@ void InterfaceReloadSystem()
...
 INT_PTR CALLBACK InterfaceDialogMotd(HWND hwnd,UINT message,WPARAM wParam,LPARAM lParam)
 {
 	char s[2000];
+	HWND hwndParent = GetParent(hwnd);
 	
 	switch (message)
 	{
 	case WM_INITDIALOG :
-		CenterWindow(hwnd,NULL);
+        CenterWindow(hwnd, hwndParent);
```
akirmse, line 1102: simplify to 
```
CenterWindow(hwnd, GetParent(hwnd));
```

since there's no other use for the parent window.

blakserv/interface.c
```diff
@@ -586,6 +586,7 @@ void StartupComplete()
 void InterfaceCreate(HWND hwnd,WPARAM wParam,LPARAM lParam)
 {
 	SetWindowText(hwnd,BlakServNameString());
+	CenterWindow(hwnd, NULL);
```
akirmse, line 589: I'm not sure this is a good idea.  Does this always center the server window on the monitor?  I certainly don't want that; I want it to come up where it currently does.

akirmse:
Nice fix, though I don't want the app centered on my monitor.  (For one thing, it will be hidden by the client if I do that, which will make local testing harder).  I believe there are other shell-related ways to control initial location.

blakserv/interface.c
adrienlaws: Removing as requested.

blakserv/interface.c
adrienlaws: Removed

blakserv/interface.c
adrienlaws: Simplified as requested.

adrienlaws:
This is ready for another review, when possible.

I have addressed all the comments and removed the automatic centering of the application on the desktop.  

The `blakserv` application starts in it's original position now, but the MOTD dialog box will appear centered on the `blakserv` application.

blakserv/interface.c
```diff
@@ -22,6 +22,7 @@
 	
 */
 
+#define NOMINMAX
```
akirmse, line 25: Are you sure this is needed?  It looks like the original code used std::max and didn't have this.

Precompiled headers in C++ is a complicated topic, but extra #defines like this can slow down compilation.

adrienlaws: Removed, you are right.

PR #1066, opened 2025-01-13
feature: Add server-configurable client movement interval for real-time updates

davidklaw:
This PR introduces a server-configurable method for adjusting the interval at which client movement updates are sent to the server. The goal is to enable more 'real-time' location updates between the server and client, improving the overall responsiveness of movement in the game.

🚧 **Drafted for discussion.**


blakserv/synched.c
```diff
@@ -662,6 +662,17 @@ void SynchedGotoGame(session_node *s,int last_download_time)
...
    // All set to go to game mode.
 
+   move_interval = ConfigInt(UPDATE_MOVE_INTERVAL);
+   // Sanity check to ensure move_interval isn't 0 or negative
+   if (move_interval <= 0)
+   {
+      move_interval = DEFAULT_MOVE_INTERVAL; // Use a default value if invalid
+   }
+
+   AddByteToPacket(AP_MOVE_INTERVAL);
```
akirmse, line 672: Are you sure you need to introduce a new message type for this?  Could you just add the move interval value to the existing AP_GAME message instead (the very next lines)?

davidklaw:
Would we like to see these values and their defaults added to `run/server/blakserv.cfg`?

akirmse:
We generally do not add default values to blakserv.cfg, because then if the in-code default ever changes, it will be overridden by the old default in the cfg file, resulting in painful debugging for the server operator.

blakcomp/function.c
```diff
@@ -85,7 +85,8 @@ function_type Functions[] = {
 {"RecycleUser",         RECYCLEUSER,     AEXPRESSION,   ANONE},
 {"MinigameNumberToString", MINIGAMENUMBERTOSTRING,     AEXPRESSION,   AEXPRESSION,   ANONE},
 {"MinigameStringToNumber", MINIGAMESTRINGTONUMBER,     AEXPRESSION,   ANONE},
-   };
+{"GetServerConfigValue", GETSERVERCONFIGVALUE,  C_STRING,   ANONE},
```
davidklaw, line 88: Should I be using `AEXPRESSION` here rather than a `C_STRING`? Or `C_NUMBER`?

blakserv/ccode.c
```diff
@@ -2344,3 +2345,45 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+	if (config_id.v.tag != TAG_INT)
+	{
+		bprintf("C_GetServerConfigValue can't set from non-int %i,%i\n",
+			config_id.v.tag,config_id.v.data);
+		return NIL;
+	}
+
+	config_node *cnode = GetConfigByID(config_id.v.data);
+
+	if (cnode == NULL)
```
davidklaw, line 2367: If I update the server config to include the config entry ID without a value, the server crashes. It seems like this might be accepted by `config.c::GetConfigByID`, and perhaps there’s an assumption that a config entry will always have both an ID and a value. If we want to safeguard against this scenario, what would you recommend?

Example:
```
[Update]
MoveInterval	500
MoveCountThreshold 
```

Meridian59:
Having no value is meaningless and should crash.

davidklaw:
Ready for review.

blakcomp/function.c
akirmse: Yes, AEXPRESSION.

run/server/blakserv.cfg
```diff
@@ -39,7 +39,7 @@ LogDisk              Yes
 
 [Email]              
 
-[Update]             
+[Update]
```
akirmse, line 42: Revert change to this file.

blakserv/ccode.c
```diff
@@ -2344,3 +2345,45 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+		normal_parm_array[0].value);
+  
+	if (config_id.v.tag != TAG_INT)
+	{
+		bprintf("C_GetServerConfigValue can't set from non-int %i,%i\n",
+			config_id.v.tag,config_id.v.data);
+		return NIL;
+	}
+
+	config_node *cnode = GetConfigByID(config_id.v.data);
```
akirmse, line 2365: This seems way too permissive, as it allows Blakod to read any config value that exists now or in the future.  Let's lock it down to the one(s) we actually want to allow.

blakserv/synched.h
```diff
@@ -13,6 +13,8 @@
 #ifndef _SYNCHED_H
 #define _SYNCHED_H
 
+#define DEFAULT_MOVE_INTERVAL 1000
```
akirmse, line 16: Unless this needs to be in a header visible to all source files, move it to the C file that needs it, and make it a constant (static const int DEFAULT_MOVE_INTERVAL = 1000).  Macros are gross, but in 1994 they were the accepted method.

clientd3d/server.c
```diff
@@ -1650,8 +1651,24 @@ Bool HandleLoginFailed(char *ptr, long len)
...
-   if (len != 0)
+   int server_move_interval;
+
+   if (len < 4)
+      return False;
+
+   Extract(&ptr, &server_move_interval, 4);   
+
+   // Sanity check to ensure move interval isn't negative or zero
+   if (server_move_interval <= 0)
```
akirmse, line 1662: Maybe a more aggressive check?  If the move interval is, say, 1, a single client could flood the server.

clientd3d/server.c
```diff
@@ -1650,8 +1651,24 @@ Bool HandleLoginFailed(char *ptr, long len)
...
+   {
       return False;
+   }
+
+   debug(("Client move interval set by server: %d\n", server_move_interval));
```
akirmse, line 1667: Remove log spam

blakserv/ccode.c
```diff
@@ -17,6 +17,7 @@
 */
 
 #include "blakserv.h"
+#include "config.h"
```
akirmse, line 20: Needed?  blakserv.h already includes config.h.

blakserv/synched.c
```diff
@@ -577,6 +577,7 @@ void SynchedSendMenuChoice(session_node *s)
 void SynchedGotoGame(session_node *s,int last_download_time)
 {
    int num_new_files;
+   int move_interval = DEFAULT_MOVE_INTERVAL;
```
akirmse, line 580: This default value is never used, so remove it.  Then declare the variable at the first point of use instead.

clientd3d/server.c
davidklaw: Yes, of course! I'm thinking 250, or 4 moves per second, at most.

blakserv/synched.c
davidklaw: Are you suggesting I just do away with the `DEFAULT_MOVE_INTERVAL` constant entirely?

```
AddByteToPacket(AP_GAME);

int move_interval = ConfigInt(UPDATE_MOVE_INTERVAL);
if (move_interval <= 0)
{
  AddIntToPacket(move_interval);
}

SendPacket(s->session_id);
```

I'm thinking yeah...

davidklaw: If I did this, I'd need to update the client's packet handling of this. But now I'm doubting my thinking and wondering if always having the server send the move interval on game entry is the way to do this.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -2944,7 +2969,7 @@ messages:
 
       % Check for cheating by sending more move packets than allowed.
       % Don't log too many cheat detections, because it'll fill up our logs.
-      if piMovesCounter > MOVEMENT_COUNT_THRESHOLD
+      if piMovesCounter > piMoveCountThreshold
```
davidklaw, line 2972: There's a speedhack detection component of this effort that I overlooked in my work here: `iDelta` a few lines above. 

I believe this needs to scale in relation to the new server values to properly apply the decay, like so:
```
% Number of moves “consumed” due to the passage of time
decayFactor = (1000/MOVE_INTERVAL));

piMovesCounter = (piMovesCounter + 1) - (iDelta * decayFactor);
```

To do this, I'll also need to grab move interval from the config and that has me wondering: should we store these in `System` and then retrieve and store them on `User` during logon?

davidklaw:
I've updated the PR to use the configured move interval in speedhack detection, and added a good bit of plumbing to support that.

blakcomp/function.c
akirmse, line 88: I still see C_STRING here?

blakserv/ccode.c
```diff
@@ -20,6 +20,8 @@
 
 #define iswhite(c) ((c)==' ' || (c)=='\t' || (c)=='\n' || (c)=='\r')
 
+static const int server_config_whitelist[] = { UPDATE_MOVE_INTERVAL, UPDATE_MOVE_COUNT_THRESHOLD };
```
akirmse, line 23: Should be documented.

blakserv/ccode.c
```diff
@@ -2344,3 +2346,61 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
 	
 	return ret_val.int_val;
 }
+
+bool IsAllowedConfigID(int config_id) {
```
akirmse, line 2350: If you make server_config_whitelist a std::vector instead, this function could be one line (using the find() function).

akirmse, line 2350: Needs a comment

akirmse, line 2350: Function can be static?

clientd3d/move.c
```diff
@@ -773,21 +774,30 @@ void MoveUpdatePosition(void)
 {
    int x, y;
    BYTE speed;
+   DWORD current_time;
+   time_t rawtime;
+   struct tm * timeinfo;
+   char buffer[9]; // Buffer to hold HH:MM:SS
```
akirmse, line 780: Declare variables at point of first use instead of at the beginning of the function, where possible.

kod/util/system.kod
```diff
@@ -474,6 +483,48 @@ messages:
       return;
    }
 
+   RetrieveServerConfigValues()
+   "Retrieves server config values needed for operation."
```
akirmse, line 487: Not a very illuminating comment.

kod/util/system.kod
```diff
@@ -474,6 +483,48 @@ messages:
...
+      return;
+   }
+
+   RetrieveServerConfigValue(config_id=$)
+   {
+      local config_value;
+
+      config_value = GetServerConfigValue(config_id);
+
+      if config_value = $
```
akirmse, line 510: Not sure I see the purpose of this function.  If the config_value is ```$```, it returns ```$```.  What does this function add over calling GetServerConfigValue directly?

kod/util/system.kod
```diff
@@ -41,6 +41,9 @@ constants:
    RECREATE_PHASE_FOUR = 4
    RECREATE_PHASE_FIVE = 5
 
+   MOVE_INTERVAL = 1000
```
akirmse, line 44: Document these

davidklaw:
I'm wrapping up the feedback changes and wanted to check in on a question from my PR description:

Currently, server config values are only saved to System when the server is first created. If they change, admins must manually call a new function to update them. Would it make more sense to save them on server startup or reload instead? What do you prefer?

akirmse:
You mean reloading the game?  That feels weird that reloading the game would also apply newly-changed config values.  I think for this first version we could do it just at startup.  But in a follow-up PR, if there's a function that admins can call to apply the new values, I'd think we could have the server do that automatically whenever these particular config values are changed (with some extra work).

davidklaw:
Alright, all feedback has been addressed.

I moved the call to pull in the server config values to `System::LoadedFromDisk`, which I believe to be the appropriate place to pull these on server startup.

davidklaw:
@Meridian59 This is ready for your review after dealing with some merge conflicts.

PR #1067, opened 2025-01-13
No eats under second wind

CardiovascularMike:
Currently players are able to consume foods with an active second wind enchantment.  This fills up their stomach capacity and removes the food item from their inventory but does nothing to help the player.

This update would prevent players from eating any foods - notifying them that they are still under the surge of adrenaline from the second wind.

To keep gameplay impacts as small as possible an exception is made for chocolate mints - these will be consumed and will increase the player's mana but will not increase the player's vigor.  In addition, potions and flasks are eaten according to KOD but they are not food items and will be consumed and have the normal effect.

This is originally submitted as a draft PR in the hopes of finding some confirmation that the IsItemType() message is used correctly.  This message lives in Item.kod already and returns:

      return (viItem_type & type);
      
 Where each item can have multiple types so the bitwise and is a bit strange here to me.  In testing this current set up works.  I've also considered adding a new IsFood() message in a few spots (in item it returns false, in Food.kod we return true, etc).


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -5563,7 +5565,29 @@ messages:
          
          return FALSE;
       }
-      
+
+      % Prevent eating if player has active second wind enchantment
```
akirmse, line 5569: When adding the first thing of a new type, it's often useful to ask yourself what would happen if you added a lot more.  i.e. if there were 20 enchantment types that all cared about what the player was eating, would you have 20 checks here in Player, which is already 12000 lines long?

Hopefully this convinces you that this code belongs in SecondWind.  You could have a loop here in Player that asks each active enchantment whether it's OK to eat the food.  If any returns false, you prevent the food from being eaten.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -5563,7 +5565,29 @@ messages:
...
+      if what <> $
+      AND Send(self,@IsEnchanted,#byClass=&SecondWind)
+      {
+         if Send(what,@IsItemType,#type=ITEMTYPE_FOOD)
+         {
+            if IsClass(what,&Mint)
+            {
+               return TRUE;
+            }
+            else
```
akirmse, line 5584: You don't need "else", since the "then" clause always returns.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -321,6 +321,7 @@ resources:
    player_desc_enchanted_none = "%q"
    player_desc_enchanted = "%s has %s%s.\n\n%s"
    
+   player_not_hungry_sw = "Still feeling a surge of vitality, you are uninterested in food or drink right now."
```
akirmse, line 324: This will move to SecondWind (see below).  This will have the advantage of not requiring an awkward "_sw" suffix, which is a sign that this doesn't really belong here.

kod/object/active/holder/nomoveon/battler/player.kod
CardiovascularMike: I understand.  I have updated this and now we loop over all enchantments on the player and make sure that none are preventing the player from eating.

kod/object/active/holder/nomoveon/battler/player.kod
CardiovascularMike: Thanks - fixed

CardiovascularMike:
We are using IsItemType() which lives in Item.kod.  Its not a commonly used function.  Here it is:

   IsItemType(type = ITEMTYPE_MISC)
   {
      % Check for misc items (type 0)
      if type = viItem_type            
      {
         return TRUE;
      }               

      return (viItem_type & type);
   }
   
  I'm posting this here to make sure we review the use of this function for this specific set of requirements.  My testing hasn't shown any issues but I'm not thrilled with returning using the bitwise and.  For example, some items are reagents, food, and sundry.  I will test all foods before this is merged.

kod/object/passive/skill/secwind.kod
```diff
@@ -222,5 +223,29 @@ messages:
...
+      if what <> $
+      {
+         % Prevent consumption of food - except mints for the mana effect
+         if Send(what,@IsItemType,#type=ITEMTYPE_FOOD)
+         {
+            if IsClass(what,&Mint)
+            {
+               return FALSE;
+            }
+            if who <> $
```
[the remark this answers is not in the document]
CardiovascularMike: Updated this a little - will add the error log aspect for what=$.  I was thinking that the orc cave waters was what=$ but actually you are drinking the orc caves room object!

kod/object/passive/trance.kod
```diff
@@ -133,6 +133,13 @@ messages:
       return FALSE;
    }
 
+   PreventsEating(what=$, who=$)
```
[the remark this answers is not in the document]
CardiovascularMike: I did a search for StartEnchantment to find these.  Another possible culprit is sickness / disease, palsy, and dement.

kod/object/passive/skill/secwind.kod
```diff
@@ -222,5 +223,29 @@ messages:
       return FALSE;
    }
 
+   PreventsEating(what=$, who=$)
```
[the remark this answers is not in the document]
CardiovascularMike: Ok - this is done.

kod/object/passive/trance.kod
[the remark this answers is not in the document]
CardiovascularMike: I tested and dement from a zombie as well as palsy does send the ReqEatSomething message to sickness and would cause an error without a message there.

PR #1071, opened 2025-01-16
fix: Allow guild allies to exit allied guild halls

### LINKED ISSUE #703: Guild Hall: Functionality: Guild hall doors don't open from the inside for players who don't own that hall
There's a new issue that was introduced in the last patch when Guild Hall alliances were reverted back to their original design, which allowed Allied Guilds to share one another's guild halls, use the doors and go in and out on their as long as they were above the lowest rank (Initiate).

If a player who is not in the guild is in the hall, past the entrance door (examples below), this player will not be able to open the front door to get out - effectively locking them inside.

This has resulted in situations where players have snuck into a hall and gotten locked in, or allied players getting locked in to their allies guild hall. This bug affects all Guild Halls.

This did not occur before - and the following behavior is expected:

Expectation:

- A player who is inside a guild hall can open the front door from the inside
- Allied players who go inside a guild hall belonging to their allies (Rank 2 and above) should be able to open the front doors ( alliance once featured being able to share halls without this issue) so they can get out

How to Reproduce:

Guildmasters of Mature Guilds can purchase a Guild Hall from Frular the Guild Master (GMHall) in South Barloque.

Once they own a Guild hall, all entrances and all inside doors including the guild chest, guild news, and other rooms within the hall will open for Guildmaster and all members of the guild from Rank 2 (2, in this use - being the level just past Initiate, under the level of 'Member', which adds the Guild invite / initiation functionality.) There is no issue with these users being able to use their hall or navigate it.

Create a guild. Make the guild mature using the admin commands or by putting 3 members in the guild.

Create an alliance with another guild.

Purchase any Guild Hall with the 1st guild.

Place the allied guild members inside the hall, and place a non allied player above 30 hit points in the hall.

Try to open the door with the outside of guild players.

davidklaw:
This PR fixes #703, addressing an issue where guild allies can enter an allied guild hall but are unable exit. The root cause lies in the inconsistent logic between entry and exit handling. Most guild halls use `GuildHall::SomethingTryGo` for entry, but rely on their own overridden `SomethingTryGo` for exiting. The exit logic, which uses `ReqLegalEntry`, did not account for allies.

To resolve this, `ReqLegalEntry` has been updated to ensure that anyone permitted to use the door to enter from the outside can also use it to exit from the inside.

~**Updated:** For consideration, the latest commit refactors this setup for clarity. It limits `ReqLegalEntry` to focus solely on legal entry per the hotplate system, and updates exit handling to use `CanEnter`.~

### DIFF
--- a/kod/object/active/holder/room/ghall.kod
+++ b/kod/object/active/holder/room/ghall.kod
@@ -1037,8 +1037,9 @@ messages:
    }
 
    CanEnter(who=$)
-   "Return TRUE if the player is allowed to use the entrance door and log "
-   "off safely inside the hall, through membership or alliance, and rank."
+   "Only guild members and their allies can enter the guild, but not their "
+   "apprentices.  Also determines who can log off safely and remain inside "
+   "the guild hall."
    {
       local oGuild;
 
@@ -1354,17 +1355,14 @@ messages:
    }  
 
    ReqLegalEntry(who=$)
-   "Returns TRUE if the player is in the guild hall legally; otherwise."
    {
       local i;
-      
-      % Unowned guild halls are always open
-      if poGuild_Owner = $
+
+      if Send(self,@CanEnter,#who=who)
       {
          return TRUE;
       }
 
-      % Check the legal entry list
       for i in plLegal_entry
       {
          if i = who
@@ -1373,8 +1371,8 @@ messages:
          }
       }
 
-      % This player did not enter the guild hall legally, through the hotplate system.
-      Debug("Player",Send(who,@GetTrueName),"entered a guild hall illegally",
+      % Okay, we're pretty sure this is a bastard PKK wuss.
+      Debug("Player",Send(who,@GetTrueName),"broke into guild hall",
             Send(self,@GetName),"illegally!");
 
       return FALSE;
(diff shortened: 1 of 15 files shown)


akirmse:
Very nice, that looks right to me.

kod/object/active/holder/room/ghall.kod
```diff
@@ -1355,20 +1354,17 @@ messages:
...
       local i;
-
+      
+      % Unowned guild halls are always open
       if poGuild_Owner = $
       {
          return TRUE;
       }
 
-      if poGuild_owner = Send(who,@GetGuild)
```
davidklaw, line 1366: I don't think this belongs here. It isn't needed and just because someone is in the guild doesn't mean they entered legally.

akirmse: Doesn't this make it impossible to take anything from the guild chest?  I think we should go back to a more minimal change that just fixes the allied exit bug.

davidklaw: Yes, since it will now rely entirely on hotplates. I can reintroduce this logic here to allowlist guild members, even if they’re not on the legal entry list, or simply roll this back to reduce the complexity. I'll assume you're in favor of the latter and plan to roll it back.

davidklaw:
Simplified and ready for action.

PR #1074, opened 2025-01-18
Welcome message time interval adjustments

sicoderealm:
The mechanism used to override the welcome message interval for specific NPC doesn't work correctly due to use of constants over properties. Currently all NPCs will have the default of 30 seconds - the original PR in https://github.com/Meridian59/Meridian59/pull/775/ intended to make the most popular bar in town (Fams) less active with these messages.

We now switch to using properties against each NPC that can be adjusted as required.
There is also a more specific way to target NPCS, over monsters, by using `Towns` over `Monster` for the implementation.

Testing: reviewed properties of NPCs in game and timed against expected intervals successfully + also successfully set the class values as an admin.


kod/object/active/holder/nomoveon/battler/monster.kod
```diff
@@ -5827,17 +5819,9 @@ messages:
       return;
    }
 
-   % True if the monster is ready to say a welcoming message to players
-   %  who have entered their room.
    ReadyForWelcome()
```
akirmse, line 5822: a) This function doesn't no longer returns TRUE or FALSE, which looks like a bug.
b) Is this function still needed?  It looks like you now only call it on Towns.

kod/object/active/holder/nomoveon/battler/monster.kod
```diff
@@ -347,10 +343,6 @@ properties:
    piMinDamage = $
    piMaxDamage = $
 
-   
-   % Tracking the last time this monster welcomed a player
-   piLastWelcomeTime = 0
```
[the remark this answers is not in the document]
akirmse: When a saved game is loaded, you should get one server error message for every existing monster, stating that it can't find this property from the saved game.  But after that, things should work OK.  Please verify this before/after behavior.

kod/object/active/holder/room.kod
```diff
@@ -1779,9 +1779,9 @@ messages:
             for i in plActive
             {
                each_obj = First(i);
-               if IsClass(each_obj,&Monster)
+               if IsClass(each_obj,&Towns)
```
akirmse, line 1782: Having to make this change is a good example of the crappiness of IsClass.  What do you think about this alternative design:

- ActiveObject gets a UserEntered() which by default does nothing.
- Instead of checking IsClass here, you always call UserEntered() on each active object.
- In Towns, UserEntered() does the logic that is now in ReadyForWelcome(), and calls a new function X if the user hasn't entered too recently.
- Rename the existing UserEntered to X.  This is tedious, but I think an improvement, as currently the name is misleading.  It is not called whenever a user enters.

kod/object/active/holder/nomoveon/battler/monster/towns.kod
```diff
@@ -32,6 +32,13 @@ properties:
    piFace_translation = PT_GRAY_TO_SKIN1
    piHair_translation = PT_HAIR_DKBROWN
 
+   % Tracking the last time this NPC welcomed a player
+   piLastWelcomeTime = 0
+
+   % Minimum interval (in seconds) between NPC welcome messages to prevent excessive 
+   % chatter when a player enters their room.
+   piWelcomeTimeInterval = 120
```
akirmse, line 40: Not sure if this should be a class variable or a property, as there's typically only one of each NPC type.  Your thoughts?

kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzapoth.kod
```diff
@@ -59,7 +59,7 @@ messages:
       return False;
    }
 
-   UserEntered(who = $)
+   UserEnter(who = $)
```
akirmse, line 62: I am confused as to why some UserEntered became WelcomeUser, and some became UserEnter.

kod/object.kod
```diff
@@ -685,6 +685,11 @@ messages:
    {
       return;
    }
+   
+   UserEntered(who = $)
```
akirmse, line 689: Since this is the highest place in the hierarchy this appears, there should be documentation here.

Also, I guess this could be on ActiveObject rather than Object, since Room iterates over its active objects list to make the call.  I'm not sure why all of these similar functions wound up on Object instead.

kod/object/active/holder/nomoveon/battler/monster/towns.kod
```diff
@@ -26,12 +26,19 @@ classvars:
 
    viSpeed = SPEED_AVERAGE
 
+   % Minimum interval (in seconds) between NPC welcome messages to prevent excessive 
+   % chatter when a player enters their room.
+   viWelcomeTimeInterval = 120
```
akirmse, line 31: 2 * 60

PR #1078, opened 2025-01-22
feature: Revised waterskin `piNumber` to modify availability

davidklaw:
This PR adjusts the `piNumber` value for waterskins to align better with other food items, making them more suitable for monster treasure drops.

It has been adjusted from 20 to 2, consistent with similar food items.

### Impact
* **Zombies:** Now drop 2 waterskins instead of 20.
* **Zhieu B'hob:** Now starts with a quantity of 4 waterskins instead of 20 for sale.

### Deployment Notes
* Send `SetForSale` to Zhieu B'hob to update his item list

### DIFF
--- a/kod/object/item/passitem/numbitem/food/watrskin.kod
+++ b/kod/object/item/passitem/numbitem/food/watrskin.kod
@@ -43,8 +43,8 @@ classvars:
 
 properties:
 
-   viNutrition = 10
-   viFilling = 20
+   viNutrition = 3
+   viFilling = 6
    piNumber = 2
 
 messages:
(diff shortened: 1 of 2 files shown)


kod/object/item/passitem/numbitem/food/watrskin.kod
```diff
@@ -45,7 +45,7 @@ properties:
 
    viFilling = 6
    viNutrition = 3
-   piNumber = 20
```
akirmse, line 48: I don't know what the implications are of changing the default number of an object.  What have you found?

davidklaw: It tracks the quantity of the item instance (e.g., 20 mushrooms) and acts as the default value for any instance created without a specified quantity. I’m assuming that weight and bulk are defined per single item, even for items like arrows, which are generated in multiples. If this assumption is correct, updating `piNumber` should only affect the default quantity generated when no value is specified, without causing any miscalculations.

PR #1079, opened 2025-01-28
feat: Add female door locked audio and refactor door locked wave logic

### LINKED ISSUE #1041: Sounds: male sound when door is locked
When trying to open a locked door, regardless of character gender, a male sound is played.
Propose we add a female variant of `doorlck.wav`.

davidklaw:
This update adds a new audio file for the female gender when attempting to enter a locked door. Since there's also a guild hall that handles an additional case independently for a unique balcony door, the logic has been refactored into a new `Room` message to centralize the behavior.

Sound byte courtesy of @sicoderealm 🎶

### Female

https://github.com/user-attachments/assets/be4bb061-70b8-4a61-8745-a0f41159fc47

### Male (existing)

https://github.com/user-attachments/assets/9584f7f9-cb51-4b7c-89c2-733f0ca24922

Closes #1041


kod/object/active/holder/room.kod
```diff
@@ -3957,5 +3958,19 @@ messages:
       return;
    }
 
+   DoorLockedSound(what=$)
```
akirmse, line 3961: Name for this function could be better, as could its documentation.  "DoorLockedSound" makes it sound like it returns the appropriate sound resource, whereas it actually assumes that "what" is a player (rename?), and sends the sound to the client.

davidklaw: Yes, absolutely. A slip-up on my part. I've made the updates, and it's ready for another review.

akirmse:
Just one question: what is a "sound byte"? :-)

PR #1081, opened 2025-01-28
fix: Properly move `IDC_APPLY` control when description dialog is resized

### LINKED ISSUE #1028: Wands: USE button for the Karma-based spell-casting wands misplaced.
For all spellwands that are karma-based, the USE button on the window overlay interface (OK, DROP, USE) is out of place and showing in the bottom of the second block of text related to those specific wands.

-- Karma-wands all have an extra sentence mentioning the karma requirement.

The wand works, but the player will only see the very bottom of the USE button. One user reported not being able to use them as they couldn't see the button (they _were_ told what to do with the wand...)

The player can still click the bottom of the use button, or double-click the wand to activate/use it.

This occurs on the Steam Client.
Note: I set my text to Heidelbrand.

https://github.com/user-attachments/assets/b5376750-30d0-44ad-84fa-0bbf2dc33250

davidklaw:
This PR adds `IDC_APPLY` to the set of controls that are repositioned when the object description dialog is resized, such as when using non-default fonts. Previously, `IDC_APPLY` was not included and remained stationary, which appears to have been an oversight rather than an intentional choice.

Here is an example of what happens today:
image

And once updated:
image
image

Closes #1028

### DIFF
--- a/clientd3d/dialog.c
+++ b/clientd3d/dialog.c
@@ -31,6 +31,7 @@ static ChildPlacement desc_controls[] = {
 	{ IDC_USE,         RDI_BOTTOM },
 	{ IDC_UNUSE,       RDI_BOTTOM },
 	{ IDC_INSIDE,      RDI_BOTTOM },
+	{ IDC_APPLY,       RDI_BOTTOM },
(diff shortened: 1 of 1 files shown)


akirmse:
Good find!

PR #1085, opened 2025-01-30
Fix rescue bug in kocatan rentable rooms

todasj:
There is a bug with the new rentable rooms in kocatan, if you cast rescue when in a rented room it sends you to your hometown on mainland instead of The Aerie Guest House, this is because this room is not treated as a RID_KOCATAN.

The problem is that rentable rooms are not using piRoom_num (they do not have a room number) and you cannot distinguish a mainland room from a kocatan room.

### DIFF
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -870,8 +870,6 @@ messages:
 
       % this is ugly to do it here, but was lowest risk to add as an emergency patch
       % change this later by overriding GetRegion() please!
-	  
-	  local LocRentedRoom;
 
       if (piRoom_num >= RID_GUEST_BASE) and (piRoom_num < RID_NEWB_BASE)
       {
@@ -887,15 +885,8 @@ messages:
       {
          return RID_KOCATAN;
       }
-	  
-      % Rentable rooms are not using piRoom_num like other rooms.
-      % To distinguish a rented room on the kocatan island from one on mainland
-      % we are using GetLocation to retrieve which Inn the rented room belongs to,
-      % if it belongs to RID_KOC_INN we return RID_KOCATAN  
-	  
-      LocRentedRoom = Send(self,@GetLocation);
-	  
-      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13) or (LocRentedRoom = RID_KOC_INN)
+
+      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13)
       {
          return RID_KOCATAN;
       }
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -886,7 +886,7 @@ messages:
          return RID_KOCATAN;
       }
 
-      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13) or (prRoom = KocatanRentableRoom_roo)
+      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13) or (vrName = KocatanRentableRoom_name)
       {
          return RID_KOCATAN;
       }
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -886,7 +886,7 @@ messages:
          return RID_KOCATAN;
       }
 
-      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13)
+      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13) or (prRoom = KocatanRentableRoom_roo)
       {
          return RID_KOCATAN;
       }
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -871,8 +871,7 @@ messages:
       % this is ugly to do it here, but was lowest risk to add as an emergency patch
       % change this later by overriding GetRegion() please!
 	  
-	  local lRoom_data;
-	  lRoom_data = Send(self,@GetLocation);
+	  local LocRentedRoom;
 
       if (piRoom_num >= RID_GUEST_BASE) and (piRoom_num < RID_NEWB_BASE)
       {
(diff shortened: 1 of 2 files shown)


todasj:
Realised it does not pass all the checks, looking into that

kod/object/active/holder/room.kod
```diff
@@ -886,7 +886,7 @@ messages:
          return RID_KOCATAN;
       }
 
-      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13)
+      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13) or (prRoom = KocatanRentableRoom_roo)
```
davidklaw, line 889: Is the problem actually here? Because the `if` statement above should cover `RID_KOC_INN` (2001), which is between `RID_KOCATAN` (2000) and `RID_KOCATAN_END` (2499), and return `RID_KOCATAN` before this is even reached.

todasj: I am 100 % sure, rented room does not have room numbers like normal numbers, on my test server ALL rented rooms starts from 10.000 and then ticks up one per rented room I created
image

davidklaw: Ah, right. Sorry, I was looking at the inn itself and not the rented room.

todasj: This is for the rented rooms, not The Aerie Guest House, which is as you say RID_KOC_INN (2001)

todasj:
Ugh, why does this keep failing...

davidklaw:
What is your build showing locally? You can see what's happening in the GitHub build by looking at the details:
> room.kod(889): error: Unknown identifier KocatanRentableRoom_name

todasj:
> What is your build showing locally? You can see what's happening in the GitHub build by looking at the details:
> 
> > room.kod(889): error: Unknown identifier KocatanRentableRoom_name

It does not show any errors at all now, I did get one before but it stopped when I changed to vrName

`C:\Users\micro\Documents\GitHub\Meridian59>nmake

Microsoft (R) Program Maintenance Utility Version 14.39.33523.0
Copyright (C) Microsoft Corporation.  All rights reserved.

Making in .\blakserv
Making in .\util
Making in .\external\zlib
Making in .\external\libpng
Making in .\external\libarchive
Making in .\external\wavemix
Making in .\clientd3d
Making in .\module
Making in .\blakcomp
Making in .\kod
Building util
Building factgame
Building object
Building active
Building holder
Building room
Building barlqrm
Building crnthrm
Building jasperrm
Building kocarm
Building marnrm
Building tosrm
Building ghall
Building monsroom
Building objroom
Building feyforst
Building guest6
Building bossroom
Building kcforest
Building guest1
Building guest2
Building guest3
Building guest4
Building guest5
Building guest7
Building guest8
Building rentroom
Building nomoveon
Building battler
Building monster
Building mummy
Building skel
Building temples
Building towns
Building barlqtwn
Building crnthtwn
Building jasprtwn
Building marntwn
Building tostwn
Building hazartwn
Building kocatwn
Building wanderer
Building council
Building factions
Building grdworm
Building troop
Building xeochctl
Building player
Building user
Building dm
Building admin
Building storebox
Building portal
Building wallelem
Building passive
Building spell
Building atakspel
Building boltspel
Building dmspell
Building boon
Building jala
Building hinder
Building persench
Building resist
Building touchatk
Building roomench
Building multicst
Building walspell
Building news
Building trestype
Building sickness
Building disease
Building skill
Building profic
Building stroke
Building unarmed
Building mananode
Building guild
Building guildcmd
Building brain
Building itematt
Building weapatt
Building flikerer
Building item
Building actitem
Building passitem
Building attmod
Building defmod
Building armor
Building robe
Building shield
Building soldshld
Building helmet
Building mask
Building pants
Building shirt
Building healer
Building numbitem
Building food
Building ammo
Building weapon
Building ranged
Building bow
Building necklace
Building ring
Building resring
Building pforget
Building token
Building minigame
Building instrum
Building lute
Building totem
Building offering
Building swshard
Building inscript
Building factflag
Building roomkeyc
Building hstone
Building spelitem
Building potion
Building scroll
Building wand
Building specwand
Making in .\blakdeco
Making in .\club
Making in .\bbgun
Making in .\makebgf
Making in .\resource
Building graphics
Building textures
Making in .\roomedit

C:\Users\micro\Documents\GitHub\Meridian59>`

akirmse:
This is failing because it's using a name from lower down in the class hierarchy.  This might happen to work for you locally depending on the order you built things (which is a deficiency/bug in the compiler), but if you build clean in the kod directory and rebuild, you'll likely see the error.

Can you think of a way to implement this doesn't depend on these constants?  Constants that are globally visible (e.g. in blakston.khd) are OK.

todasj:
@davidklaw Thanks for the idea, it works now and passes all checks

akirmse:
> The problem is that rentable rooms are not using piRoom_num (they do not have a room number) and you cannot distinguish a mainland room from a kocatan room, so I used prRoom = KocatanRentableRoom_roo instead to detect if it is a rented kocatan room.

This is outdated; please change.  (This is why it isn't a good idea to duplicate information--change it one place, and the other is wrong.)

kod/object/active/holder/room.kod
```diff
@@ -886,7 +889,7 @@ messages:
          return RID_KOCATAN;
       }
 
-      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13)
+      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13) or (lRoom_data = RID_KOC_INN)
```
akirmse, line 892: You don't need lRoom_data until here, so don't call GetLocation until this point.  Also, can you think of a more meaningful name?

Also, please add a comment explaining why this check is here.

kod/object/active/holder/room.kod
```diff
@@ -870,6 +870,8 @@ messages:
 
       % this is ugly to do it here, but was lowest risk to add as an emergency patch
       % change this later by overriding GetRegion() please!
+
```
akirmse, line 873: Something felt fishy about this implementation, and this comment explains it:

```
 % this is ugly to do it here, but was lowest risk to add as an emergency patch
 % change this later by overriding GetRegion() please!
```

The right fix is to add a function GetRegion() to the class KocatanRentableRoom, and have it return RID_KOCATAN.  This way you don't need to modify the Room class at all.

And I guess it wasn't clear the first time---you shouldn't put code details in the description of the PR.  It's going to be outdated again after this, and you'll have to change it again  :-)

todasj: Added GetRegion to the KocRentedRoom and removed code details from the description. I did not know you could make a overriding function like this. Thank you for the guiding!

kod/object/active/holder/room/rentroom/kocrent.kod
```diff
@@ -123,6 +123,17 @@ messages:
 
       return;
    }
+   
+   GetRegion()
+   {
+      % Rentable rooms are not using piRoom_num like other rooms.
```
akirmse, line 129: This comment is no longer necessary, so you can remove it.  Thanks to good naming, it's self-evident that KocatanRentableRoom should be in the KOCATAN region.

todasj: Gotcha, done!

PR #1086, opened 2025-02-01
refactor: Improve ugliness of `GetRegion` logic

davidklaw:
Tidied up `GetRegion` by moving logic into the relevant subclasses, making things a bit cleaner. No functional changes—just easier to read and maintain.


kod/object/active/holder/room.kod
```diff
@@ -886,11 +886,6 @@ messages:
          return RID_KOCATAN;
       }
 
-      if (piRoom_num = RID_GUILDH12) or (piRoom_num = RID_GUILDH13)
```
akirmse, line 889: Did you forget these?

davidklaw: These two rooms already override `GetRegion` and return `RID_KOCATAN`.

PR #1107, opened 2025-02-15
Fix: Mana focus to give mana regen

todasj:
Many years ago a changed was added to the spell mana focus, it was supposed to make mana focus increase mana regen, but it has never worked because the spellpower was never recieved.

The code in player.kod:
**iTime = Send(oSpell,@AdjustManaTime,#who=self,#time=iTime);**

the _old_ code in focus.kod:
**AdjustManaTime(time=0, iSpellPower=0)**

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5499,7 +5499,7 @@ messages:
    CalculateManaTime()
    "Calculate # of milliseconds until user will gain a mana point"
    {
-      local iTime, lJalaInfo, oJalaSpell, iJalaPower, oSpell;
+      local iTime, lJalaInfo, oJalaSpell, iJalaPower, oSpell, iSpellPower;
 
       if piMana > piMax_mana
       {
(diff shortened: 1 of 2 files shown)


akirmse:
This doesn't look right.  See https://github.com/search?q=repo%3AMeridian59%2FMeridian59%20adjustmanatime&type=code.  I would expect all of these calls to be consistent.

Also, can someone comment on the balance implications of making this spell stronger?

todasj:
This is the old PR that you accepted https://github.com/Meridian59/Meridian59/pull/53 that did not work.

Seems like he just did a copy of the old function but did not edit what was sent in it.

I have not looked to deep into it but there is a difference from getting spellpower from a spell and from a jala spell (room enchantment in the rejuv case). But till you are certain if you want this fixed I will just let it rest

davidklaw:
> Also, can someone comment on the balance implications of making this spell stronger?

This would increase regeneration speed from no reduction to potentially 50%, putting it on par with Jala’s Rejuvenate (just before the mana focus check in `player.kod`), so it’s not unprecedented. The player remains immobile, can cast elemental spells, and benefit from fast mana regen—but any movement will break the effect. No alarms going off for me.

PR #1131, opened 2025-03-03
Removal of keybind

sicoderealm:
Keybind removal from Meridian 59.

Testing: able to build and no mentions of keybind remain in the codebase.


akirmse:
It also needs to be removed from the installer.

PR #1132, opened 2025-03-08
Update disarm.kod

UprisingEngineer:
Fix `SuccessChance()` so that low agility users can have Disarm succeed sometimes.  Without this fix, users with agility between 1 and 8 will not be able to have Disarm trigger successfully.

### DIFF
--- a/kod/object/passive/skill/disarm.kod
+++ b/kod/object/passive/skill/disarm.kod
@@ -83,8 +83,17 @@ messages:
        iAbility=send(who,@GetSkillAbility,#skill_num=viSkill_num);
        num = ((100-reqstat)*iAbility/100) + reqstat;
        num = (num + 1 + modifier) / DISARM_FACTOR;
-       
-       if random(1,100) < num
+
+       if(num < 1)
+       {
+         num = 1;
+       }
+       else
+       {
+         num = num * 10;
+       }
+
+       if random(1,1000) < num
        {
          return TRUE;
        }


kod/object/passive/skill/disarm.kod
```diff
@@ -83,8 +83,17 @@ messages:
        iAbility=send(who,@GetSkillAbility,#skill_num=viSkill_num);
        num = ((100-reqstat)*iAbility/100) + reqstat;
        num = (num + 1 + modifier) / DISARM_FACTOR;
-       
-       if random(1,100) < num
+
```
davidklaw, line 86: Regardless of gameplay changes, you could simplify this by bounding `num` before the random check to ensure it stays within the expected range:

```
% Guarantee at least a 1% chance
num = bound(num, 1, 100);
```

UprisingEngineer: True, very true.  But that would mean an improvement of chances for low-agility players, whereas my fix was only to make _some_ chance without  affecting other users, in the spirit of the original code.  I'm fine with either!  I'd like to use this skill eventually!  I purchased it at 100hp and haven't had it fire since I got it 130hp.

davidklaw: If I'm not mistaken, bounding it in that way would give anyone with 1-8 agility a 1% chance of success. Everyone else's chances should remain the same.

UprisingEngineer:
>If I'm not mistaken, bounding it in that way would give anyone with 1-8 agility a 1% chance of success. Everyone else's chances should remain the same.

TRUE!  Except in PvP! ;)

davidklaw:
> > If I'm not mistaken, bounding it in that way would give anyone with 1-8 agility a 1% chance of success. Everyone else's chances should remain the same.
> 
> TRUE! Except in PvP! ;)

Caught my comment before I deleted it :) I saw the flaw in what I was suggesting vs. what you're trying to do. I think you're down the right path if the idea is to make it challenging at the lowest of levels.

davidklaw:
Oh, and definitely update the PR title to something meaningful. Maybe:
> fix: Allow low-agility players a small chance to succeed at disarm

akirmse:
It looks like you accidentally put a second unrelated change in here.. that should go in its own PR (in a separate branch).

UprisingEngineer:
> It looks like you accidentally put a second unrelated change in here.. that should go in its own PR (in a separate branch).

Hey, I'm sorry, I tried to make a separate branch, but GitHub won't let me. =\

This is my second account, my first account with 10k+ stars is `holdoffhunger`.  I have other contributions to Meridian59 under that name.  But that account was hacked by M$ and is completely within their hands and control, so I can't make contributions or new separate branches.

So, to be clear, I'm only allowed to make one branch. Making another branch doesn't work. I TRIED to make another branch, and put changes in there, but Github just put them into THIS branch.  It sucks a lot.  I'm sorry, I wish we were all on SourceForge or some other website that properly implements Git.

UprisingEngineer:
> > It looks like you accidentally put a second unrelated change in here.. that should go in its own PR (in a separate branch).
> 
> Hey, I'm sorry, I tried to make a separate branch, but GitHub won't let me. =\
> 
> This is my second account, my first account with 10k+ stars is `holdoffhunger`. I have other contributions to Meridian59 under that name. But that account was hacked by M$ and is completely within their hands and control, so I can't make contributions or new separate branches.
> 
> So, to be clear, I'm only allowed to make one branch. Making another branch doesn't work. I TRIED to make another branch, and put changes in there, but Github just put them into THIS branch. It sucks a lot. I'm sorry, I wish we were all on SourceForge or some other website that properly implements Git.

To be clear, I really don't care about clout here.  I just want a working game.  It sucks that Github is what it is.  I can't do anything about it.

davidklaw:
@UprisingEngineer I think you might be confusing forks with branches (judging from your fork of the codebase being named `Meridian59-DisarmFix`. You can have 1 fork of the project, but branches are unlimited. At the very least, you'll need to revert the latest commit on the branch you PR'd and then push that up.

UprisingEngineer:
No, actually, it's 1 fork per different project.  That way, the different forks can't interfere with each other.  Whereas branches aren't supposed to interfere with each other to begin with.

Look at my second change -- It has nothing to do with this, and since it wasn't committed as part of this, it shouldn't even be here.  It's for a different issue, so it should be a different fork.

Github Docs:

>>A fork is a new repository that shares code and visibility settings with the original “upstream” repository. Forks are often used to iterate on ideas or changes before they are proposed back to the upstream repository, such as in open source projects or when a user does not have write access to the upstream repository.

Since these are different problems, the shield max and the fix to the disarm skill, these should be different forks.  It's possible one fix could be accepted, but not the other.  Why should building the photovoltaic cells in solar plants be tied to the amount of nickel-cadmium found in riverbeds of perch and salmon?  It shouldn't.  That's why you fork it.

davidklaw:
Respectfully, that's not accurate. A fork is a place to iterate on ideas and changes that can be proposed back to the upstream repository in the form of branches. If it's a hassle, I don't mind submitting an updated PR on your behalf.

PR #1134, opened 2025-03-16
Allow robes to be worn with pants 

sicoderealm:
Support wearing robes with clothes (pants and shirts).

We use the same approach as soldier shields to allow leg positions to support robes with pants and update relevant code to allow clothes and robes combinations.

https://github.com/user-attachments/assets/aa6872a1-d23d-42ba-b411-5b32d2dc1261

https://github.com/user-attachments/assets/92d36d15-e8c7-4cbe-8d78-4b3058f6e911

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -3275,8 +3275,6 @@ messages:
             {
                if Send(i,@GetItemUseType) & type
                {
-                  % Found conflicting item - check if they can be worn together (bidirectional)
-                  % Check use type to ensure we're dealing with wearable items (not weapons)
                   if (Send(what,@CanBeWornInSameSlotWith,#what=i) 
                           OR Send(i,@CanBeWornInSameSlotWith,#what=what))
                   {
--- a/kod/object/item/passitem/defmod/armor/robe.kod
+++ b/kod/object/item/passitem/defmod/armor/robe.kod
@@ -148,7 +148,7 @@ messages:
          
          return FALSE;
       }
-      
+
       propagate;
    }
 
@@ -200,9 +200,6 @@ messages:
 
    CanBeWornInSameSlotWith(what = $)
    "Robes can be worn in the same slot as pants and shirts despite overlapping use types."
-   "Robes use ITEM_USE_BODY | ITEM_USE_LEGS, but can still be worn with:"
-   "- Pants (which also use ITEM_USE_LEGS)"
-   "- Shirts (which use ITEM_USE_SHIRT, visual overlap only)"
    {
       % Robes can be worn simultaneously with pants and shirts
       if IsClass(what, &Pants) OR IsClass(what, &Shirt)


Gar103:
This is actually quite important to the look and feel of the game, can we get this one in? The pants/robes conflict fundamentally makes everyone walk around with no pants on. Makes us look a bit silly.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -3234,7 +3234,16 @@ messages:
          }
          else
          {
-            Send(self,@MsgSendUser,#message_rsc=player_use_full);
+            if ((IsClass(what,&Pants) AND (Send(self,@FindUsing,#class=&Robe) <> $))
+               OR ((IsClass(what,&Robe) AND (Send(self,@FindUsing,#class=&Pants) <> $))))
```
akirmse: Doesn't this mean you can *only* wear pants if you have a robe on, and vice versa?

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -8719,6 +8728,13 @@ messages:
    "Sets our legs to <new_legs>, and keeps track of <what> to undo when necessary.  "
    "Tells everyone we changed."
    {
+      if Send(self,@FindUsing,#class=&Robe) <> $ 
+         AND IsClass(what,&Pants) 
+      {
+         % Player is wearing a robe so We don't show pants here as they are underneath.
```
akirmse: Fix capitalization

kod/object/item/passitem/defmod/armor/robe.kod
```diff
@@ -126,8 +126,9 @@ messages:
 
    ReqUseSomething(what = $)
    {
-      if send(what,@GetItemUseType) & ITEM_USE_LEGS
-         AND NOT IsClass(what,&Robe)
+      if (NOT IsClass(what,&Robe)
```
akirmse: There used to be a check for send(what,@GetItemUseType) & ITEM_USE_LEGS here.  Doesn't removing this change more than just robe behavior?

kod/object/item/passitem/defmod/armor/robe.kod
```diff
@@ -126,8 +126,9 @@ messages:
 
    ReqUseSomething(what = $)
    {
-      if send(what,@GetItemUseType) & ITEM_USE_LEGS
-         AND NOT IsClass(what,&Robe)
+      if (NOT IsClass(what,&Robe) 
+         AND NOT IsClass(what,&Pants) 
+         AND NOT IsClass(what,&Shirt))
```
akirmse: I think if you reversed the sense of this if using de Morgan's Law and instead returned early, there would be less indentation and the function would be easier to read.

akirmse: These IsClass checks are ugly and can easily break in the future if other clothes types are added.  Or more realistically, it would make adding new clothes so difficult that no one will try.  Might there be a way of adding a new function that instead indicated whether the given object (Defmod?) is compatible with robes, or something along those lines?  Hopefully a function that you can also use in the corresponding place in Pants.kod too.

sicoderealm: Replaced hardcoded class checks with IsCompatibleWithItem(). Added base method in DefenseModifier with overrides in Robe, Pants, and Shirt classes - agreed, this will make adding new clothing types much easier and maintainable.

sicoderealm: Applied early return pattern to reduce nesting

kod/object/active/holder/nomoveon/battler/player.kod
sicoderealm: switched to check combination compatibility checks from your recommendations

kod/object/item/passitem/defmod/armor/robe.kod
sicoderealm: Also moved to compatibility checks for the logic here

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -8746,22 +8775,40 @@ messages:
    }
 
    ResetPlayerLegs()
-   "Sets our icon to its natural, un-panted state."
+   "Sets our players legs to show their robes, pants (equipped or un-panted state)."
```
akirmse: player's

kod/object/item/passitem/defmod/armor/robe.kod
```diff
@@ -126,20 +126,19 @@ messages:
 
    ReqUseSomething(what = $)
    {
-      if send(what,@GetItemUseType) & ITEM_USE_LEGS
-         AND NOT IsClass(what,&Robe)
+      % Check if this robe is compatible with the item being equipped
+      if Send(self,@IsCompatibleWithItem,#what=what)
```
akirmse: I'm unclear on the semantics of IsCompatibleWithItem.  In Player::CheckPosition, you check both X compatible with Y and Y compatible with X, but here you check only one.  How do we know when to check one vs. both?

Also, I'm surprised by the default implementation of IsCompatibleWith in DefMod.  It returns FALSE, meaning that by default a DefMod isn't compatible with anything.  That seems strange, and is also the opposite of what the comment to that function says.

"Compatible" also isn't super clear to me.  Is this meant to affect only what can be worn with what?  If so, how about "CanBeWornWith"?  Or is it actually is meant to check whether it can be worn *in the same slot* as something else?  In that case, maybe "CanBeWornInSameSlotWith"?  If that's the intention, then I would understand why the default is FALSE.. hopefully you can see the confusion here.  The comment to the function should make this crystal clear.

sicoderealm: Thanks for the feedback, changes made: 
-Bidirectional checking: Now consistent everywhere
-Default FALSE clarified: Documentation explains overlapping slots can't be worn together by default
-Renamed: CanBeWornInSameSlotWith as suggested for clearer semantics
-No IsClass checks: Uses viUse_type for slot conflicts
-Tested in-game - working correctly with no server errors

PR #1143, opened 2025-03-29
Add class exclusion list to `Holder::CountHoldingHowMany` and exclude DMs from customer list

### LINKED ISSUE #1140: NPC: A DM or Admin in a room affects Shopkeeper NPC behavior and can interfere with gameplay.
When a player who is PvP Flagged as a murderer attempts to purchase, sell or trade in any way with a shopkeeper type NPC (smith, healer, apothecary), they'll be blocked from doing so if another player is present - unless the trading player is in the Duke Faction (where this is a bonus for faction members).

This block also occurs if a DM, Admin or Creator is present.

The NPC should ignore all DMs, Admins or Creators, whether they are hidden or unhidden.

davidklaw:
Currently, murderers cannot purchase from lawful NPCs when DMs are present in the room (even hidden DMs), as DMs are counted as fellow customers. This creates unintended interference with normal gameplay.

This PR adds the ability to exclude specific classes (like DMs) when counting objects in a holder. By updating the lawful NPC checks to ignore DMs present in the room, DMs will no longer be counted as fellow customers and murderers can go about their business.

### Testing
- Verify murderers can purchase from lawful NPCs when:
  - No other players are present (only DMs)
  - DMs are hidden in the room
  - DMs are visible in the room
- Verify existing behavior remains unchanged when:
  - Regular players are present
  - No DMs are present

Fixes #1140

### DIFF
--- a/kod/object/active/holder.kod
+++ b/kod/object/active/holder.kod
@@ -158,10 +158,11 @@ messages:
       return FALSE;
    }
 
-   CountHoldingHowMany(class = $)
-   "Does not count quantity in a numbitem (500 apples counts as 1)"
+   CountHoldingHowMany(class = $, lExcludeClasses = $)
+   "Counts objects of the given class, excluding specified classes, "
+   "ignoring quantity in numbitem (500 apples counts as 1)."
    {
-      local n,i,each_obj;
+      local n;
 
       if class = $
       {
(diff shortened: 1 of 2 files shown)


kod/object/active/holder.kod
```diff
@@ -158,10 +158,11 @@ messages:
       return FALSE;
    }
 
-   CountHoldingHowMany(class = $)
-   "Does not count quantity in a numbitem (500 apples counts as 1)"
+   CountHoldingHowMany(class = $, lExcludeClasses = $)
+   "Counts objects of the given class, excluding specified classes, "
```
akirmse, line 162: Given that the "class" is only a single class, is there any utility to having lExcludeClasses be a list instead of a single class?  I can't think of one.  If a class in the list is not in the hierarchy of "class", it will have no effect.  For the others, the topmost class in the list will have an effect, and all others will be ignored.

davidklaw: The only thing that came to mind when I added this was `Food`—randomly excluding `Apples` and `Cheese` for some reason. Odds of anyone actually needing this? Pretty much zero.

davidklaw: Or perhaps `Monster`, but excluding a handful of specific monsters for some reason.

kod/object/active/holder.kod
```diff
@@ -170,25 +171,48 @@ messages:
 
       n = 0;
```
akirmse, line 172: Remove this line; does nothing.

kod/object/active/holder.kod
```diff
@@ -170,25 +171,48 @@ messages:
...
       {
          each_obj = Send(self,@HolderExtractObject,#data=i);
          if isClass(each_obj, class)
          {
-            n = n + 1;
+            % Only increment if no excluded classes or doesn't match any excluded class
+            count = count + 1;
+            
+            % Check excluded classes if any exist
+            if lExcludeClasses <> $
```
akirmse, line 199: You don't need this outer "if", since the "while" check will handle the $ case.

PR #1151, opened 2025-04-06
feat: Add multi-select support for mail deletion

### LINKED ISSUE #1120: Feature Request: Batch delete for mail items
It would be great to add the ability to delete multiple mail items at once in the in-game mail system. Currently, users must delete each mail item individually.

davidklaw:
This PR adds ability to delete multiple mail messages at once by:
- Removing single-select restriction from mail list view
- Adding `DeleteSelectedMessages` function to handle bulk deletions
- Improving selection after deletion: selects the message at the position of the last deleted item, or falls back to the first message if that position no longer exists
- Adding confirmation dialog when deleting multiple messages (with translations in English, German, and Brazilian Portuguese)
- Adding `UpdateButtonStates` helper function to manage button states based on selection:
  - Reply/ReplyAll enabled only with single message selected
  - Delete enabled whenever any messages are selected
- Ensuring proper button states when dialog first loads

The change allows users to select multiple messages using standard Windows controls (Shift/Ctrl+click) and delete them in one operation, with appropriate safety checks and UI feedback.

https://github.com/user-attachments/assets/6e0c605a-5685-4a4b-ad29-d9958c0feb61

Closes #1120


module/mailnews/mailnews.rc
```diff
@@ -333,7 +333,7 @@ CAPTION "Read Mail"
 CLASS "Mail"
 FONT 8, "MS Sans Serif", 0, 0, 0x0
 BEGIN
-    CONTROL         "List1",IDC_MAILLIST,"SysListView32",LVS_REPORT | LVS_SINGLESEL | LVS_SHOWSELALWAYS | WS_BORDER | WS_TABSTOP,4,4,275,131
+    CONTROL         "List1",IDC_MAILLIST,"SysListView32",LVS_REPORT | LVS_SHOWSELALWAYS | WS_BORDER | WS_TABSTOP,4,4,275,131
```
akirmse, line 336: Make sure to modify the dialog for all of the other languages, too.

akirmse (commented):
What happens if there are multiple messages selected and you press Reply or Reply All?

davidklaw:
> What happens if there are multiple messages selected and you press Reply or Reply All?

Oops. I had tunnel vision and didn’t account for that. Right now, it would only open a compose window for the first selected message. But I’ve put a proper fix in place.

* Reply and Delete buttons are now disabled when the dialog opens, and their states update as the list changes or messages are selected.
    * Reply/Reply All is only enabled when a single message is selected.
    * Delete is enabled when one or more messages are selected.
* Added a confirmation dialog for deleting multiple messages (supports all 3 of our supported languages).
* Did some light code refactoring while I was in there.

PR #1152, opened 2025-04-06
gameplay: Only strip Vale boon when karma alignment flips

### LINKED ISSUE #1149: Design: Mana Node: Vale of Sorrows Node is continually lost due to players not having karma in the requisite direction (+ or -)
The Vale of Sorrows Mana Node requires players to have positive (+) or negative (-karma), either +1 or -1 in either direction to be able to successfully use the node when it's raised in either the pos./neg. karma direction. 

To raise the node, 9 rooms, including the one containing the node (vale.kod), must be turned negative or positive by players killing all of the good (fey elhai) or bad (fey dirhai) fairies.

After all of the fairies are killed, the forest will either darken or brighten, and the music for the area will change. There is also a 'neutral' state for the rooms when the number of each living fey in the room is about even.

Once the good or bad node appears, or 'raised' as the players say, players have a brief window of time in which they're able to access the node. Once they leave a room, the monsters begin to reset, and the node goes down. 

Fun Note: if a player doesn't have the 'right' karma to use the node, say, they have +30 karma and the negative karma mana node is raised - using the node will cause the player to be killed by the mana node in a special death. This is the only mana node that does this.

The Node gives double the mana of a regular mana node, functioning as two nodes in one, giving players a large boost to their current mana pool.

The Node can be lost, causing players to lose their gained mana, in a few ways intended by the original design:

- If a player has the negative karma node, and the positive node is raised.
- If a player has the positive karma node, and the negative node is raised.

The above method is more understood by players as a consequence of raising the node. A server will usually agree upon a certain karma node to do in the lifespan as a server as a social agreement during the lifespan of a server. The negative karma node is historically the most popular node to do, as gaining negative karma is a more difficult process.

However, there is also another way to lose the node:

- If the negative karma node is raised, and the player has gotten it already, but doesn't have their karma negative when the node appears again by another group of like-karma players attempting to meld with it.
- If the positive karma node is raised, and the player has gotten it already, but doesn't have their karma positive when the node appears again by another group of like-karma players attempting to meld with it.

The above design causes a contentious gameplay loop, currently observable on the production servers: the Negative Karma Mana Node has been 'raised' at least half a dozen times this year, and every time, players who just got the node wind up losing it because they don't have their karma in the direction of the node. Right now, a Node Run is being planned 1 week after one just occurred, in order for a few players who missed that node to be able to get it. Players who aren't currently playing this week are now going to lose the node they spent an hour or more to get.

If this is the intended overall design, this issue can be closed.  Due to the difficulty of organizing a group of players to activate the Vale of Sorrows Node, based on player feedback, it would be appreciated if the Node was lost only if the opposive karma version of the node was raised.

davidklaw:
This PR updates the Vale node to track its previous karma alignment and only strip mana boons from players when the node flips between good and evil. If the alignment stays within the same polarity, no action is taken.

In practice, this means:
* If a player with negative karma melds with a negative Vale node, then later raises their karma, raising the same negative node again won’t remove their boon.
* But if the server raises the positive node instead, their boon will be stripped — because the node flipped alignment and they no longer match.

This change keeps the intended gameplay behavior but avoids unnecessary disruption when the node is raised without actually changing sides.

Addresses and closes #1149

### DIFF
--- a/kod/object/passive/mananode/feynode.kod
+++ b/kod/object/passive/mananode/feynode.kod
@@ -39,6 +39,7 @@ classvars:
 
 properties:
 
+   piPreviousKarma = FEYNODE_GOOD
    piKarma = 0
 
 messages:
@@ -70,8 +71,15 @@ messages:
    }
 
    StripUsers()
+   "Strip users based on their karma."
    {
       local lUsers, i;
+
+      % If the karma hasn't changed to opposite, don't strip users.
+      if (piPreviousKarma = piKarma)
+      {
+         return;
+      }
 
       lUsers = Send(SYS,@GetUsers);
       if piKarma = FEYNODE_GOOD
@@ -100,6 +108,8 @@ messages:
             }
          }
       }
+
+      piPreviousKarma = piKarma;
 
       return;
    }
--- a/kod/object/passive/mananode/feynode.kod
+++ b/kod/object/passive/mananode/feynode.kod
@@ -13,7 +13,7 @@ FeyNode is ManaNode
 % FeyNode has three properties that make it distinct:
 % 1) It gives twice as much mana
 % 2) Only people of matching karma can meld with it.
-% 3) When created, it strips everyone of opposing mana of their link
+% 3) When created, if polarity flips, it strips everyone of opposing mana of their link
 
 constants:
 
@@ -75,7 +75,7 @@ messages:
    {
       local lUsers, i;
 
-      % If the karma hasn't changed to opposite, don't strip users.
+      % If the karma hasn't changed to the opposite polarity, don't strip users.
       if (piPreviousKarma = piKarma)
       {
          return;


akirmse:
I'm not sure about this.  It makes some sense, but on the other hand the original intent is pretty clear in the comments.

davidklaw:
That’s fair. The original intent appears to be that if you’re evil, you must be evil, or this is one of the consequences. Conversely, if you’re good, be good. You should not be bouncing between the two. @jrotunda: want to weigh in further than #1149?

jrotunda:
This one is mainly for QOL, I don't disagree with the gameplay intentions but wanted to at least raise it here as its come up a lot in player feedback; I defer to akirmse's call on this one. This can remain as-is if its desired to stick with the original design.

akirmse:
I think let's keep it as-is.  The intent is that you can only get the benefit of the node while your karma agrees with it.  In theory this would be checked every time your karma changes (though that could be quite annoying as a player).

PR #1155, opened 2025-04-09
gameplay: Ability to flag items as excluded from logoff penalties

### LINKED ISSUE #269: [General] Flag added for items that can and cannot drop in a penalty.
Adding an additional flag to items so items can be marks as valid/invalid selection for a penalty.

Suggest setting this flag on all created items (created through create weapon etc.), and light item that can be carried in vast quantities (ballots, keys, vials). 

Flag will allow quick response to new low value throw away items being utilized to mitigate/avoid item penalties.

davidklaw:
This PR addresses #269 by updating how item drops are handled during logoff penalties. Previously, players could avoid meaningful penalties by filling their inventory with cheap or disposable items. 

💭 **Submitting as a draft for a sanity check on the overall approach.**
🗒️ **Debug statements in temporarily for testing and validation.**

### Key Changes

* Adds a new `DropsOnLogoffDeath` method to the Item interface.
* Updates `InflictPenalties` item drop logic with a two-phased approach:
    1. Build a list of droppable items and randomly select items from that list to drop
    2. Drop those chosen items in a separate pass
* Applies the new behavior to:
    * Items created by spells (e.g., created weapons, created food).
    * Ballots (set to false for now).

Also fixes an issue in the “drop all” logic where the loop could run unnecessarily. Previously, the loop used a static inventory count even as items were being removed. Now it properly tracks the current inventory size after each drop.

### Some scenarios to consider and test
1. An empty inventory
2. An inventory made up entirely of ballots
3. An inventory with a wedding ring (should not drop on death)
4. An inventory with _created_ weapons and/or created food

### Notes
* **Fun fact:** If a player hits the logoff ghost penalty limit, an Amulet of Shadows can still be selected as a drop—meaning it’s possible for the amulet to kill the player while they’re logged off.

Closes #269

### DIFF
--- a/kod/object/active/logghost.kod
+++ b/kod/object/active/logghost.kod
@@ -247,7 +247,7 @@ messages:
             LossAmount, iNumItemsInventory, numItemsLost, iLoops, iStamina,
             iStart, lSpells, iSpellAbility, iAbilityNum, numSpellPointsLost,
             lSkills, iSkillAbility, numSkillPointsLost, iEquivDeath,
-            oSoldierShield, oItem;
+            oSoldierShield, oItem, iItemsDropped;
 
       % Sanity check: Are they logged on?
       If Send(poGhostedPlayer,@IsLoggedOn)
@@ -295,9 +295,13 @@ messages:
       }
       else
       {
-         debug("LogoffGhost: Starting drop process for",Send(poGhostedPlayer,@GetTrueName),
-               "- Need to drop",LossCounter,"items from",numItemsLost,"total items",
-               "- PenaltyCounter:",penaltyCounter,"of",iEquivDeath);
+         % Log initial state
+         debug("LogoffGhost:",Send(poGhostedPlayer,@GetTrueName),
+               "- Initial inventory:",iNumItemsInventory,"items",
+               "- Need to drop:",LossCounter,"items",
+               "- Penalty:",penaltyCounter,"of",iEquivDeath);
+
+         iItemsDropped = 0;
 
          while LossCounter > 0 AND iNumItemsInventory > 0
          {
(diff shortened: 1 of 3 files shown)


akirmse (commented):
This new loop is complicated and very difficult to reason about.  An alternative implementation might be:
- Make a copy of the inventory list with the undroppable objects filtered out
- N = min(length of new list, number of items to drop)
- Choose N items at random from the new list, and drop them.

Also, does this solve a real problem?

davidklaw:
I considered that approach, but when an item is dropped, all items—droppable or not—shift down, which makes the stored indices invalid. For example, if our inventory looks like this:

1. Droppable
2. Non-droppable
3. Droppable
4. Droppable

Our list of droppable item indices would be [1, 3, 4]. But after dropping item 1, the inventory becomes:

1. Non-droppable
2. Droppable
3. Droppable

Now those stored indices 3 and 4 no longer point to the right items. I thought about adjusting the indices by subtracting the number of dropped items, but since non-droppable items can be anywhere in the list, that gets messy fast.

As for whether this is solving a real problem—hard to say. It does clean up a cheesy edge case in PVP and penalty scenarios, which matters to a vocal group of players right now. But the fix also involves a decent number of class changes just to flag items, and that part feels a little icky.

As always, the code and context live on in this PR, and I’m totally fine closing both the PR and the issue until there's more public support for it. Honestly, I'm just looking to close issues, especially old ones😄

davidklaw:
Well, that's a significant improvement over my "work". Thanks.

davidklaw:
Closing this PR. I’ll likely submit a new one that reuses the droppable item list logic, but avoids introducing any new item-level flag messages like this one.

PR #1159, opened 2025-05-01
gameplay: Adds PvP cooldown to prevent free penalty chance after/during combat

davidklaw:
This PR addresses community feedback about logoff penalties in PVP, notably the ability for some players to rotate through multiple characters for free penalties. This change makes penalties consistent for anyone who has "recently" (arbitrarily set to 12 hours) attacked another player while applying the existing penalty system to everyone else.

### New Property
This change introduces a new Settings property, `piLogoffPenaltyPvPTime`, which can be used to adjust just how "recent" an attack must be to flag a character as not getting a free penalty.

### Testing
- Verify PvP timer blocks free penalties after attacking another player
- Confirm normal penalty system still works for non-PvP scenarios

### DIFF
--- a/kod/object/active/logghost.kod
+++ b/kod/object/active/logghost.kod
@@ -270,11 +270,22 @@ messages:
       numItemsLost = Send(poGhostedPlayer,@GetNumItemsInInventory);
       LossCounter = Send(poGhostedPlayer,@GetNumItemsInInventory)
                     * penaltyCounter / iEquivDeath;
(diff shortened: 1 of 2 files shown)


davidklaw:
Converting to a draft while I reconsider the logic here.

davidklaw:
I'm going to close this in favor of the larger but, I feel, better suited solution I proposed in #1161.

PR #1160, opened 2025-05-01
fix: fixes item dropping logic in logoff ghost

davidklaw:
This PR fixes a loop issue in the logoff ghost's item dropping logic when dropping everything.

**Bug Example**
With 1 droppable item in inventory:
1. `iNumItemsInventory = 1`
2. First iteration: `i = 1`, drop succeeds
3. `i` stays at 1 (we don't increment on successful drops)
4. Second iteration: `i = 1 <= iNumItemsInventory (1)` is true
5. Try to drop item at index 1 again, but inventory is empty

On the server side, an error will be thrown from `Player::FindItemByIndex`, as there is no longer an item at index 1.

### DIFF
--- a/kod/object/active/logghost.kod
+++ b/kod/object/active/logghost.kod
@@ -290,8 +290,8 @@ messages:
             }
             else
             {
-               % Don't need to increment unless the item stayed in the
-               %  inventory.
+               % Only increment the index if the item could not be dropped and
+               % remained in inventory.
                i = i + 1;
             }
          }
--- a/kod/object/active/logghost.kod
+++ b/kod/object/active/logghost.kod
@@ -282,13 +282,17 @@ messages:
       {
          % just drop everything
          i = 1;
-         while i <= iNumItemsInventory AND Send(poGhostedPlayer,@GetNumItemsInInventory) > 0
+         while i <= iNumItemsInventory
          {
             if NOT Send(poGhostedPlayer,@DropItem,#index=i,#targetGhost=self)
             {
                % Don't need to increment unless the item stayed in the
                %  inventory.
                i = i + 1;
+            }
+            else
+            {
+               iNumItemsInventory = iNumItemsInventory - 1;
             }
          }
       }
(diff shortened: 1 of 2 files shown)


kod/object/active/logghost.kod
```diff
@@ -282,7 +282,7 @@ messages:
       {
          % just drop everything
          i = 1;
-         while i <= iNumItemsInventory
+         while i <= iNumItemsInventory AND Send(poGhostedPlayer,@GetNumItemsInInventory) > 0
```
akirmse, line 285: If there are N items in the inventory, this loop will traverse them N times, leading to N^2 operations.  Performance might not matter too much here, but still this is to be avoided if possible.

What about simply decrementing iNumItemsInventory in the other branch of the "if"?  (This assumes that dropping an item doesn't have side effects like causing other items to be dropped, items to be created in the inventory, etc.)

davidklaw: Good point. Your suggestion probably makes more sense given how things work _today_. As far as I know, nothing adds items to inventory on drop. The Amulet of Shadows comes to mind, but doesn’t actually drop. I’m thinking I’ll go ahead and make the change, and document `DropItem()`. It might not always be the function that gets overridden, but it feels like the right junction place this information, if needed.

```
DropItem(droppedItem=$, targetGhost=$, dropRoom=$, dropRow=0, dropCol=0,
            index=$, merge=TRUE)
   % Removes the specified item (by object or index) from the player's inventory and drops it in the specified location.
   % Returns TRUE if the item was successfully dropped, FALSE otherwise.
   %
   % IMPORTANT: This function is assumed NOT to add items to the inventory.
   % If this behavior changes, review all code that loops over inventory and calls DropItem,
   % as it may result in infinite loops or other unintended consequences.
```

kod/object/active/logghost.kod
```diff
@@ -290,6 +290,10 @@ messages:
                %  inventory.
                i = i + 1;
             }
+            else
```
akirmse, line 293: To simplify, remove the NOT in the if and switch the two branches.

PR #1161, opened 2025-05-05
gameplay: Introduce toggleable flat penalty system as alternative to legacy model

davidklaw:
This PR introduces a new `PenaltyManager` system that replaces the old, progressive penalty model with a flat, deterministic one. The new system unifies how item, spell, and skill penalties are applied and includes a server-side toggle for easy testing or rollback.

### Motivation
The legacy system is overly complex—mixing progressive scaling, randomness, and escalation—which makes it hard to tune and frustrating for players. PvP players in particular have raised concerns about both overly harsh penalties and loopholes, like cycling through alts to avoid them. Knowing how contentious this topic is, I’ve made the new system configurable so it can be rolled out gradually.

### Key Changes
- Introduces a centralized `PenaltyManager` class to handle logoff and (eventually) death penalties. Long term, this aims to phase out logoff ghosts as a PvP meta in favor of meaningful deaths.
- Penalties are now deterministic: `base * multiplier`, no randomness or escalation.
- Admins can configure base values and adjust multipliers for murderers and PvP participants (including outlaws, warring guilds, and faction soldiers) while keeping non-PVP penalties at a minimum (including a freebie).
- A server property allows switching between the old and new systems at any time.


davidklaw:
One challenge for me with this is determining whether these new properties belong in `System` or `Settings`.

davidklaw:
Latest commit adds a server-configurable number of penalty-free logoffs for eligible players.

PR #1162, opened 2025-05-13
consistency: aligns msg structure across quests of similar type

bjorndjurner:
This is a consistency update to align the msg structure across quests of similar type


davidklaw:
@bjorndjurner Interesting. Do you have any theories on why the escaped quotes are problematic?

akirmse:
I don't understand this fix.  How does this affect quest triggering?

Aesica:
This is just the line spoken by the priestess giving the quest, whether it gets displayed in quotes or not should have zero impact on the destination NPC's response.  If anything, I suspect the problem (target NPC not always recognizing the cargo/line spoken) has to do with the fact they're listening for something else, triggering some other response or quest, etc.

That's why repeating the line 2-3 times ensures that this step in the quest completes correctly.

bjorndjurner:
> This is just the line spoken by the priestess giving the quest, whether it gets displayed in quotes or not should have zero impact on the destination NPC's response. If anything, I suspect the problem (target NPC not always recognizing the cargo/line spoken) has to do with the fact they're listening for something else, triggering some other response or quest, etc.
> 
> That's why repeating the line 2-3 times ensures that this step in the quest completes correctly.

I'm reshaping this as a consistency update as that's really all that needs to be considered at the moment .. I'm aware of those aspects and I think things like that is why it's flown under the radar for so long .. I've seen the issue myself and it could possibly be a Mac/VM issue or something similar that is system related as that's what I often play on .. the issue's been in the wild for a very long time but affects a small percentage of players.

There's no guarantee this fixes that issue but the reason I didnt hesitate to do a PR is that the quest msg structure wasn't aligned across all quests and it was an easy win to sort that while at the same time seeing if it affects the mentioned issue.

akirmse:
Thank you for your contribution, but I'm going to close this.  You should always test your changes with a local server setup.  If you did, you'd see that this change does not affect game behavior in any way; it's purely cosmetic.

Regarding the quests prompts being inconsistent, I'd say that the quotation marks are actually useful, as they precisely delineate what the player is supposed to say.  If other quests don't have them and you'd like to add them, that might be a good idea.

PR #1164, opened 2025-05-22
Fix for primary bonus for Jala spells

neiltaco:
Jala's primary bonus is unintentionally capped at 30 when it's supposed to exceed that cap based on your instrument.

Normally the primary bonus to most spell schools is capped at a max of 30.  However, comments left about Jala spells state that its primary bonus is based on the wielded instrument, and that the bonus can go up to 40 because instruments with higher bonuses were harder to find.

Unfortunately, the primary bonus for Jala would first be calculated by getting the instrument level, then afterwards the bounds for primary bonuses sets it back to the max of 30.  This meant true lutes gave the same strength as fine lutes.

This PR does the following:
  - The bounds for the primary and secondary bonuses were moved above the if-statement for Jala, which allows it to override the values with its own calculations for Jala spells.
  - A bound is applied again on iSecondaryBonus for Jala, so it can maintain its intended max cap of 10.

I tested this on a 150HP player character with normal lute, fine lute, and true lute.  The true lute can now give the instrument power of 40, and the secondary bonus is capped at 10 as intended.

### DIFF
--- a/kod/object/passive/spell.kod
+++ b/kod/object/passive/spell.kod
@@ -2182,6 +2182,10 @@ messages:
             iItemModifier = iItemModifier/2;
          }
       }
+	  
+      % Bind the bonuses to reasonable values.
+      iPrimaryBonus = bound(iPrimaryBonus,0,30);
+      iSecondaryBonus = bound(iSecondaryBonus,0,10);
       
       if viSchool = SS_JALA
       {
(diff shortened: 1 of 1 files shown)


davidklaw:
Note that there's some previous context in the discussion of #664, which this revives.

I don't know what the 10 point impact will be on the game itself, but I'm all for updating to match the original intention.

akirmse:
Yeah, I guess that's really my only question.  Only bugs like this that are to the player's advantage ever get fixed, so over time the game gets easier.  What's the impact of making the lutes more powerful, and should there be a corresponding change to make something else more difficult?  (I realize that 40 lute power was the original intent, but that was decades ago, so not really relevant to today's balance.)

neiltaco:
An extra 10 spellpower can give some extra buff to helpful songs like Rejuvenate and Restorate.  However I can see how this could potentially affect PvP.   Anti-school songs like Conciliation and Profane Resonance may be more likely to prevent spell casting or at least reduce spellpower.

I think more data can be collected to see whether or not an extra 10 spellpower will tip the balance too far.

Aesica:
Why not lower the lutes' strength instead, so that a lute is 10, a fine lute 20, and a true lute 30.  This way, the grades matter again but you aren't effectively giving players a buff.

neiltaco:
Honestly that's not a bad idea, Aesica!  It would technically be nerfs to the first two lute, but a normal lute is usually a way to get started on Jala, and I think bards usually try to get a fine lute as soon as they can.  That leaves us with the rarer true lute that now has a better bonus.

I might leave this PR up for a little bit longer if anyone else has some more thoughts to add, and I'm curious what others have to say about Aesica's suggestion.

davidklaw:
I'll +1 @Aesica 's suggestion, or even 20, 25, and 30 if that's a meaningful spread.

neiltaco:
Closing this PR.  I like Aesica's suggestion and I think it can be discussed further in an open issue since it also involves a balance change.

Issue #1167, opened 2025-05-28
Map annotation tooltips appear in game window when map annotations are disabled


davidklaw:
When map annotations are disabled in Preferences, users can still create annotations by right-clicking the minimap. These annotations do not appear on the minimap, but their tooltips show up unexpectedly across large areas of the game window.

### Steps to reproduce
1. Disable "Map annotations" in Preferences
2. Right-click the minimap and create an annotation
3. Observe that no marker is shown, but the tooltip appears when hovering over unrelated parts of the UI

### Proposed fix
Prevent annotation creation entirely when map annotations are disabled.

akirmse:
I don't think this solves the root of the problem, which is why a tooltip is being created in the wrong place on the screen.  But I guess it's OK as a workaround.

PR #1169, opened 2025-05-29
fix: disable map annotation creation when annotations are disabled via Preferences

davidklaw:
This PR adds a check to prevent map annotations from being created when they are disabled in Preferences. Without this, users can still create annotations via the minimap, leading to the issue described in #1167 and tooltip text will appear unexpectedly in the game window.


akirmse:
I don't think this solves the root of the problem, which is why a tooltip is being created in the wrong place on the screen. But I guess it's OK as a workaround.

PR #1170, opened 2025-05-30
fix: Improvements to map annotations and tooltips

davidklaw:
This PR introduces a couple of improvements to map annotations behavior:

1. **Annotations and tooltips are now fully disabled if either "Show dynamic minimap" or "Map annotations" is turned off.** 
One could argue there's some benefit to having them display, but I think it's more confusing to have tooltips appear for annotations the player can't see.

2. **Tooltips are now properly cleared when many dialogs are displayed.** 
I added a wrapper around `DialogBoxParam` that calls an updated `TooltipReset()` beforehand. This helps clear lingering tooltips before most dialogs appear. I skipped a few dialogs that likely don’t interact with tooltips.

I recorded this video to show a variety of conditions under these changes:
- Players cannot create or edit map annotations, or see their tooltips, if either the minimap is disabled
- Players cannot create or edit map annotations, or see their tooltips, if either map annotations are disabled
- An annotation's tooltip will no longer remain under the mouse cursor if an annotation is edited, or if the Look dialog is brought up. Not shown are the other scenarios:
  - When the who dialog is opened
  - When a buy dialog is opened
  - When a newsglobe is opened
  - When the mail reader is opened
  - When an offer dialog is opened
  - When the color selector is opened

https://github.com/user-attachments/assets/a43c7be9-7ef2-4b91-9414-c4fd3da63402

### DIFF
--- a/clientd3d/annotate.c
+++ b/clientd3d/annotate.c
@@ -175,9 +175,7 @@ void MapAnnotationClick(int x, int y)
       }
    }
 
-   // Clear any open tooltips
-   HWND hToolTips = TooltipGetControl();
-   SendMessage(hToolTips, TTM_POP, 0, 0);
+   TooltipReset();
 
    if (DialogBoxParam(hInst, MAKEINTRESOURCE(IDD_ANNOTATE), hMain, MapAnnotationDialogProc,
                       (LPARAM) index) == IDOK)
--- a/clientd3d/annotate.c
+++ b/clientd3d/annotate.c
@@ -54,8 +54,8 @@ void MapAnnotationGetText(TOOLTIPTEXT *ttt)
 {
   int index;
 
-//   if (!MapVisible())
-//      return;
+   if (!config.map_annotations || !config.drawmap)
+      return;
 
   index = ttt->hdr.idFrom;
 
@@ -134,7 +134,7 @@ void MapAnnotationClick(int x, int y)
    MapAnnotation *a;
    Bool existed;  // True if editing an existing annotation
 
-   if (!config.map_annotations)
+   if (!config.map_annotations || !config.drawmap)
      return;
 
    MapScreenToRoom(&x, &y, TRUE);
@@ -175,6 +175,10 @@ void MapAnnotationClick(int x, int y)
       }
    }
 
+   // Clear any open tooltips
+   HWND hToolTips = TooltipGetControl();
+   SendMessage(hToolTips, TTM_POP, 0, 0);
+
    if (DialogBoxParam(hInst, MAKEINTRESOURCE(IDD_ANNOTATE), hMain, MapAnnotationDialogProc,
                       (LPARAM) index) == IDOK)
    {
@@ -185,8 +189,6 @@ void MapAnnotationClick(int x, int y)
          current_room.annotations[index].y = y;
       }
    }
-
-   TooltipReset();
 
    RedrawAll();  // In case we added or removed an annotation
 }
--- a/clientd3d/annotate.c
+++ b/clientd3d/annotate.c
@@ -175,9 +175,7 @@ void MapAnnotationClick(int x, int y)
       }
    }
 
-   TooltipReset();
-
-   if (DialogBoxParam(hInst, MAKEINTRESOURCE(IDD_ANNOTATE), hMain, MapAnnotationDialogProc,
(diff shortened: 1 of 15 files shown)


clientd3d/annotate.c
```diff
@@ -175,6 +175,10 @@ void MapAnnotationClick(int x, int y)
       }
    }
 
+   // Clear any open tooltips
```
akirmse: Is this what TooltipReset() should do?  It's still called from DisplayDesrciption.  Do description dialogs have the same tooltip bug?

davidklaw: I _believe_ `TooltipReset` was doing this at some point, maybe in older versions of Windows? And yes, this bug appears when description dialogs are opened while hovering over a map annotation.

I wonder: would it be overkill to update `TooltipReset` to simply do this:
```
// Clear any open tooltips
SendMessage(hToolTips, TTM_POP, 0, 0);
```
...and not bother with the mouse events?

Doing that and moving the `TooltipReset` calls up before the `DialogBoxParam` calls in these two places seems to fix the issue.

akirmse:
Yeah, that's what I was thinking.

davidklaw:
Turns out this issue affects more than just the annotation and description dialogs. Tooltips are also sticking around when opening other dialogs, like the newsglobe or look list (via “L”).

As you know, `TooltipReset()` is only being called in a couple specific cases. I originally thought removing the `TTS_ALWAYSTIP` style might solve it, but that doesn’t seem to have any effect.

I’m considering a passthrough helper that calls `TooltipReset()` and `DialogBoxParam()` together, then updating existing calls to use that.

Any thoughts on this approach?

Meridian59:
That's a big hammer, but if it's happening for all dialogs, that sounds
like a pretty good approach.  You would need to export the helper function
through the client.def file so that the modules can use it, as a lot of
dialogs are created in the modules.

davidklaw:
Big hammer yes, but it is happening for all the scenarios I'm able to test. Some of these need to be testing by setting up a hotkey alias.
1. Opening the Who dialog
2. Opening a Description dialog
3. Opening a Look list dialog
4. Opening a newsglobe dialog
5. Opening the annotation editor dialog
6. Opening the buy dialog
7. Opening the offer dialog

There are others and for completeness, I figured I'd address them all in this.

davidklaw:
Updated. I left a few of these dialog calls untouched because it didn't seem to me like a plausible scenario to cover. E.g., a player looking at a tooltip and then updating the client. Should I update them _all_? Remaining instances are:
1. `club.c`
2. `charpick.c`

clientd3d/dialog.c
```diff
@@ -734,3 +733,12 @@ void AbortGameDialogs(void)
 	AbortPasswordDialog();
 	AbortPreferencesDialog();
 }
+/************************************************************************/
+/*
+* SafeDialogBoxParam:  A wrapper for DialogBoxParam that resets tooltips when a dialog is opened.
```
akirmse, line 738: Add a comment describing why this function was added.

clientd3d/dialog.c
```diff
@@ -734,3 +733,12 @@ void AbortGameDialogs(void)
...
+*/
+INT_PTR __stdcall SafeDialogBoxParam(HINSTANCE hInstance, LPCSTR lpTemplate, HWND hWndParent, DLGPROC lpDialogFunc, LPARAM dwInitParam)
```
akirmse: I don't think we need __stdcall on our own functions.  Win32 functions are declared this way for historical reasons.

akirmse (commented):
Agreed that we don't need to use this on char or update functions, where there can't be any tooltips visible.

clientd3d/dialog.c
davidklaw: It appears to be necessary, otherwise we see a bunch of these when compiling:
```
error LNK2001: unresolved external symbol _SafeDialogBoxParam@20
```

akirmse: Can you try M59EXPORT instead?  See other exported functions.

davidklaw: This results in the same linker errors. Since we’re replacing a Windows API function, I guess we need to match its calling convention.

akirmse: Hm, no, there's nothing special about this new function.  Where do you get the linker error, and does it persist if you build clean (client and modules) first?

davidklaw: My mistake -- I needed to clean first.

PR #1173, opened 2025-06-20
Default no op welcome message for all npcs

### LINKED ISSUE #1172: Server Error: Entering the Raza Inn generates missing handler errors
## What

- When a player enters the Raza Inn, a server error is sometimes generated
  - It does not happen every time, but it is reproduceable

## Reproducing the Error

- Create a new account
- Log in after character creation
- Error generates
- Alternatively walk in and out of the room (and sometimes it will appear)

### Example Screen Capture
Image

sicoderealm:
We add a empty welcome message handler for all NPCS - fixes #1172


akirmse:
Should this go higher up in the hierarchy so that it covers all NPCs?

sicoderealm:
> Should this go higher up in the hierarchy so that it covers all NPCs?
Sure lets try Towns to cover all the NPCs

PR #1174, opened 2025-06-20
feat: make Windows M59 Server channel messages easily copyable

adrienlaws:
## What
- Make log, error, and debug messages copyable in the Windows Meridian 59 Server `blakserv` by replacing the `MessageBox` with a custom dialog containing a read-only edit control
- When double-clicking a channel log file the text is already selected for rapid copying

## Why
- Troubleshooting and support often require sharing detailed error or log messages
- While logs can be retrieved from text files or manually typed out, it is faster and less error-prone to be able to quickly select and copy messages directly from the interface

## Details
- Replaces the standard `MessageBox` shown on double-clicking a log/error/debug message with a custom dialog
- The new dialog uses a read-only, multiline edit control so users can select and copy the message text
- No previous functionality is removed except the old `MessageBox` call

### Before This Change
explorer_UoEgz1mGOL

### After This Change
explorer_1ElsD0U0WO


blakserv/interface.h
```diff
@@ -43,4 +43,7 @@ void StartAsyncPortalSocket(SOCKET sock);
 
 void FatalErrorShow(const char *filename,int line,const char *str);
 
+INT_PTR CALLBACK CopyableMessageDlgProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam);
```
akirmse: I think you can remove this line.

blakserv/blakserv.rc
```diff
@@ -158,6 +158,14 @@ BEGIN
     LISTBOX         IDC_DEBUG_LIST,0,124,325,52,WS_VSCROLL | WS_HSCROLL | WS_GROUP | WS_TABSTOP
 END
 
+IDD_COPYABLE_MESSAGE DIALOGEX 0, 0, 300, 120
+STYLE DS_MODALFRAME | WS_POPUP | WS_CAPTION | WS_SYSMENU
+CAPTION "BlakStone Server"
```
akirmse: BlakSton

akirmse (commented):
Nice change!

davidklaw:
HUGE! 😄

blakserv/blakserv.rc
adrienlaws: Corrected

blakserv/interface.h
adrienlaws: Removed

blakserv/interface.h
```diff
@@ -43,4 +43,8 @@ void StartAsyncPortalSocket(SOCKET sock);
 
 void FatalErrorShow(const char *filename,int line,const char *str);
 
+#ifdef _WIN32
```
akirmse: Use BLAK_PLATFORM_WINDOWS instead.

Also I think it would be slightly better if the #ifdef were inside the body of ShowCopyableMessageDialog in interface.c instead.  That makes the #ifdef less visible than in a header file.

PR #1175, opened 2025-06-21
fix: make morph who list names color correctly

### LINKED ISSUE #1068: Who List: Colored Names revert to white with Morph
If a user casts morph and their name is red, orange, green, blue - their name will turn to WHITE on the WHO LIST. 

The name immediately switches from Red to White in real time, which makes it appear the user is not a pk, or that they were pardoned.

The name will only change to white on the WHO LIST - users can still see the players name is red when the spell morph wears off and the name is above their head again.

If they log off and on, the name will appear red again to all online. The visual vanishes whenever a client viewing the who list is reset in any way.

Expectation:

When casting morph or changing forms, the colored name of the user not changing on the who list and to remain accurate.

Background:

 Who list colors were added by Gar.

https://github.com/Meridian59/Meridian59/pull/200

For the next few years, the names reverting to white with morph was in existence along with another bug, where the names would render as black when casting shadow form, making them invisible on the who list unless you changed your M59 window font colors.

The shadow form part of this was fixed by Toko, here:

https://github.com/Meridian59/Meridian59/pull/609

Personal Observation:

The reasoning for the change was to 'get rid of dependency' on the names above players heads. 

In turn, this has made players dependent on the Who List, and its inadvertently made players not communicate as well about who PKs are and has caused a lot of confusion and aggravation. Who List and knowing who pks are was never a problem, now it is.

I think this is a good example of a well-meaning change to "make the game easier" that made gamers not play as smart and made the experience worse for both players and admins. 

A full reversion would be pointless now.

adrienlaws:
## What
- Updated who list name coloring logic so that morphed players retain their correct color (PK, outlaw, DM, etc.) instead of defaulting to white
- Added a new `MORPHED_YES `flag to the client object flags
- Fixes #1068

## Why
- Previously, when a player became morphed, their name color would revert to white (even if they were an outlaw, PK, DM, etc)
- With the new flag and updated logic, the client can distinguish morphed players and apply the correct name color throughout morph and afterwards

## Details
- Defined `MORPHED_YES `in the client/server shared flags
  - Took the last available flag bit
  - We needed something to distinguish a morphed player since the normal player flags are stripped when morphed
- Adjusted both server and client code so that the correct status color is preserved for morphed players

## Result
- Morphed players now always display the appropriate name color

### DIFF
--- a/clientd3d/color.c
+++ b/clientd3d/color.c
@@ -449,29 +449,28 @@ WORD GetItemListColor(HWND hwnd, int type, item_rarity_grade text_color_value)
 */
 COLORREF GetPlayerNameColor(int flags,const char*name)
 {
-    debug(("GetPlayerNameColor: flags=0x%08X name=%s\n", flags, name));
 	if (GetDrawingEffect(flags) == OF_BLACK)
 		return NAME_COLOR_BLACK_FG;
 	
-    // If morphed, temporarily add OF_PLAYER for name coloring
+	/* If morphed, temporarily add OF_PLAYER for name coloring */
     int color_flags = flags;
-    if (flags & OF_MORPH)
+    if (flags & OF_MORPHED)
         color_flags |= OF_PLAYER;
 
     switch (GetPlayerFlags(color_flags))
     {
 		case PF_DM:
 			return NAME_COLOR_DM_FG;
-		case PF_KILLER:
-			return NAME_COLOR_KILLER_FG;
-		case PF_OUTLAW:
-			return NAME_COLOR_OUTLAW_FG;
 		case PF_CREATOR:
 			return NAME_COLOR_CREATOR_FG;
 		case PF_SUPER:
 			return NAME_COLOR_SUPER_FG;
 		case PF_EVENTCHAR:
 			return NAME_COLOR_EVENT_FG;
+		case PF_KILLER:
+			return NAME_COLOR_KILLER_FG;
+		case PF_OUTLAW:
+			return NAME_COLOR_OUTLAW_FG;
             
     default:
 			return NAME_COLOR_NORMAL_FG;
--- a/clientd3d/color.c
+++ b/clientd3d/color.c
@@ -449,11 +449,17 @@ WORD GetItemListColor(HWND hwnd, int type, item_rarity_grade text_color_value)
 */
 COLORREF GetPlayerNameColor(int flags,const char*name)
 {
+    debug(("GetPlayerNameColor: flags=0x%08X name=%s\n", flags, name));
 	if (GetDrawingEffect(flags) == OF_BLACK)
 		return NAME_COLOR_BLACK_FG;
+	
+    // If morphed, temporarily add OF_PLAYER for name coloring
+    int color_flags = flags;
+    if (flags & OF_MORPH)
+        color_flags |= OF_PLAYER;
 
-	switch (GetPlayerFlags(flags))
-	{
+    switch (GetPlayerFlags(color_flags))
+    {
 		case PF_DM:
 			return NAME_COLOR_DM_FG;
 		case PF_KILLER:
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -2540,6 +2540,7 @@ messages:
          % so the who list dialog box does not color shadowformed names black 
          AddPacket(4,Send(i,@GetObjectFlags) & ~DRAWFX_INVISIBLE);
       }
+
       SendPacket(poSession);
       
       return;
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -2540,7 +2540,7 @@ messages:
          % so the who list dialog box does not color shadowformed names black 
          AddPacket(4,Send(i,@GetObjectFlags) & ~DRAWFX_INVISIBLE);
       }
-
+      
       SendPacket(poSession);
       
       return;
--- a/kod/object/active/holder/nomoveon/battler/player/user.kod
+++ b/kod/object/active/holder/nomoveon/battler/player/user.kod
@@ -2533,22 +2533,11 @@ messages:
       AddPacket(1,BP_PLAYERS,2,Length(lUsers));
       for i in lUsers
       {
-         if SEnd(i,@CheckPlayerFlag,#flag=PFLAG_MORPHED)
-            AND Send(i,@GetIllusionForm) <> $
-         {
-            % If the user is morphed, we send the illusion form's name.
-            rName = Send(Send(i,@GetIllusionForm),@GetTrueName);
-         }
-         else
-         {
-            % Otherwise, we send the user's true name.
          rName = Send(i,@GetTrueName);
-
          AddPacket(4,i, 4,rName, STRING_RESOURCE,rName);
          % Clear the DRAWFX_INVISIBLE flag of which DRAWFX_BLACK is a subset
          % so the who list dialog box does not color shadowformed names black 
          AddPacket(4,Send(i,@GetObjectFlags) & ~DRAWFX_INVISIBLE);
-         }
       }
       SendPacket(poSession);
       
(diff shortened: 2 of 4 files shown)


clientd3d/color.c
```diff
@@ -451,21 +451,26 @@ COLORREF GetPlayerNameColor(int flags,const char*name)
...
+	/* If morphed, temporarily add OF_PLAYER for name coloring */
+    int color_flags = flags;
+    if (flags & OF_MORPHED)
+        color_flags |= OF_PLAYER;
+
+    switch (GetPlayerFlags(color_flags))
+    {
 		case PF_DM:
 			return NAME_COLOR_DM_FG;
-		case PF_KILLER:
```
akirmse: It looks like this was moved unnecessarily; put it back to minimize the diff.

clientd3d/color.c
```diff
@@ -451,21 +451,26 @@ COLORREF GetPlayerNameColor(int flags,const char*name)
...
 	if (GetDrawingEffect(flags) == OF_BLACK)
 		return NAME_COLOR_BLACK_FG;
-
-	switch (GetPlayerFlags(flags))
-	{
+	
+	/* If morphed, temporarily add OF_PLAYER for name coloring */
+    int color_flags = flags;
+    if (flags & OF_MORPHED)
+        color_flags |= OF_PLAYER;
```
akirmse: What would be the implication of *always* turning on this bit when getting the player color?  I'm not super excited about adding OF_MORPHED in the protocol, because it means that cheaters could use it to detect someone who is morphed.  Is that a problem?

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -7307,16 +7307,24 @@ messages:
...
 
-      % If we're morphed, get rid of any "player" flags
+      % If we're morphed, get rid of certain "player" flags
       if Send(self,@CheckPlayerFlag,#flag=PFLAG_MORPHED)
          AND oIllusion <> $
          AND IsClass(oIllusion,&Monster)
       {
-         iFlags = (iFlags & ~USER_YES & ~OFFER_YES & ~PLAYER_PK
-                   & ~PLAYER_OUTLAW);
-         iFlags = (iFlags & ~PLAYER_DM & ~PLAYER_CREATOR & ~PLAYER_SUPER);
```
akirmse, line 7317: Removing this line seems unrelated to this change.  Wouldn't it mean that e.g. when Mayhem morphs himself for an event, he would still be drawn as a DM, giving himself away?

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -7307,16 +7307,24 @@ messages:
...
-      }
+         iFlags = (iFlags & ~USER_YES & ~OFFER_YES);
+         iFlags = iFlags | MORPHED_YES;
 
+         % Recheck and add PK/OUTLAW flags if appropriate
```
akirmse: Is this necessary?  Now that the flags aren't being removed like they were before, do we have to manually add them in?

adrienlaws:
There is an alternative approach we could take for this morph name color problem if we don't want to use an extra flag:
- leave morphed player names in white text, regardless of player status (pk, outlaw, dm, etc)
   - BUT use the "apparent name" instead of "real name" for the who list
     - a morphed "Toko" would show up in the who list as "a fungus beast" instead of "toko" and would be in white text, always
- fix the end enchantment weirdness where the "white" color remains after morph wears off
   - probably need to leverage SomethingChanged() or a variant implementation

#### Upsides
- no extra flag to deal with or that could be used to detect morphed players (as mentioned in comments)
- a bit easier to implement

#### Downside
- some might say this is against the purpose of the who list since it hides users
- trying to "tell" "a fungus beast" could be annoying for users

kod/object/active/holder/nomoveon/battler/player/user.kod
adrienlaws: His floating name would be missing as with all morphed users.  The who list would list him with his DM color.
However the `dm hidden` command can hide DMs from the who list of players (not other DMs/Admins).

Example with test being an admin using morph and `dm hidden` and test2 being a player murderer that is also morphed:
image
- orc = test2 = player (left)
- lupogg = test = admin (right)

The design philosophy I took was to make the who list always show the real name of the user (DM/Player/whoever) + the appropriate color.  That may or may not be the right philosophy.

kod/object/active/holder/nomoveon/battler/player/user.kod
adrienlaws: Removed, not necessary.

clientd3d/color.c
adrienlaws: Changed the order back to the original order as requested.  I was tinkering too much earlier and I was concerned that the order would matter, but that was unfounded.  I checked the bitspace to make sure and tested afterwards.

adrienlaws:
I massively overengineered this change.  

I've reverted everything it didn't need.  In testing [what used to fail here](https://github.com/Meridian59/Meridian59/issues/1068#issuecomment-2993032037) the issue appears fixed.

kod/object/active/holder/nomoveon/battler/player/user.kod
```diff
@@ -7307,14 +7307,12 @@ messages:
...
-      % If we're morphed, get rid of any "player" flags
+      % If we're morphed, get rid of certain "player" flags
       if Send(self,@CheckPlayerFlag,#flag=PFLAG_MORPHED)
          AND oIllusion <> $
          AND IsClass(oIllusion,&Monster)
       {
-         iFlags = (iFlags & ~USER_YES & ~OFFER_YES & ~PLAYER_PK
-                   & ~PLAYER_OUTLAW);
-         iFlags = (iFlags & ~PLAYER_DM & ~PLAYER_CREATOR & ~PLAYER_SUPER);
+         iFlags = (iFlags & ~USER_YES & ~OFFER_YES);
```
akirmse, line 7315: Just so we don't have any nasty surprises a few months from now when this goes live... why were the PLAYER_PK and PLAYER_OUTLAW flags removed in the past, and what are all the implications of no longer removing them?

adrienlaws: I'm not sure.

There are many checks on the server side for `OF_MORPH` that control how a morphed player is treated.

Some of the other flags make sense to remove in this section:
- `USER_YES `= `OF_PLAYER `on the client side which is what determines if the minimap shows the user as a player or monster and that seems correct
- `OFFER_YES `is self explanatory and it's clear that we don't want to offer morphed players

However I can't think of a reason why `PLAYER_PK `and `PLAYER_OUTLAW `were removed. 
- At one point I thought of putting a PR to add "enemy" halos to outlaws and PKs so perhaps that's where they were going with it? To make it so morphed players didn't have that halo since they are supposed to be a "monster"?
- I can try and get some additional feedback from others

akirmse: I'm a little concerned about putting these flags back without understanding why they're here.  Can you look in the client (and modules) and see what these flags do, and thus what will change if they're sent for morphed characters?

adrienlaws: Here's the mapping of server vs client names for the same flag values:
| Meaning        | Server Side Value         | Client Side Value         |
|----------------|--------------------------|--------------------------|
| Player         | USER_YES = 0x00000004    | OF_PLAYER = 0x00000004   |
| Outlaw         | PLAYER_OUTLAW = 0x8000   | PF_OUTLAW = 0x00008000   |
| Player Killer  | PLAYER_PK = 0x4000       | PF_KILLER = 0x00004000   |

On the client side, `PF_OUTLAW` and `PF_KILLER` are only used in the `GetPlayerNameColor` function in [GetPlayerNameColor implementation](https://github.com/Meridian59/Meridian59/blob/614a436cf4fd3fca4c1d243e18d3b7071440d039/clientd3d/color.c#L450).

I can't find any other references in the codebase for `PF_OUTLAW` or `PF_KILLER` except for their declarations in [`proto.c`](https://github.com/Meridian59/Meridian59/blob/614a436cf4fd3fca4c1d243e18d3b7071440d039/include/proto.h#L397)
- [search results for PF_OUTLAW](https://github.com/search?q=repo%3AMeridian59%2FMeridian59%20PF_OUTLAW&type=code)
- [search results for PF_KILLER](https://github.com/search?q=repo%3AMeridian59%2FMeridian59%20PF_KILLER&type=code)

I also searched for `morph` and read through Psychochild's update notes but I didn't find anything relevant.

So I guess in conclusion I'm not sure what else to check.  If the bitflag is only referenced in the character name color function `GetPlayerNameColor` I would assume the risk to reintroduce these flags is low.  I'm not quite sure what else to check.

I was worried about getting the function wrong (since the bitspace is shared and so for example DM and Outlaw share a bit) but the case matching seems to be only for exact matches.

| Name         | Value (Hex)  | Bits Set         | Notes / Shared Bits                |
|--------------|--------------|------------------|------------------------------------|
| PF_KILLER    | 0x00004000   | 14               | Shares bit 14                      |
| PF_OUTLAW    | 0x00008000   | 15               | Shares bit 15                      |
| PF_DM        | 0x0000C000   | 14, 15           | Combines bits 14 + 15              |
| PF_CREATOR   | 0x00010000   | 16               | Shares bit 16                      |
| PF_SUPER     | 0x00014000   | 14, 16           | Combines bits 14 + 16              |
| PF_EVENTCHAR | 0x0001C000   | 14, 15, 16       | Combines bits 14 + 15 + 16         |

akirmse:
OK, if this is only used for drawing player colors, it seems pretty safe.

PR #1176, opened 2025-06-22
feature: add M59 server webhook calling capability

adrienlaws:
🏗️ Draft PR 🚧
## What
- adds `SendDiscordWebhook` Blakod function to enable M59 server webhook functionality
- example use case: automatically post `###` death messages or player broadcasts to a Discord channel

## Why
- Discord is the prevalent platform for M59 player populations
- the webhook functionality is not Discord specific and can be adapted to other services
- it drives player engagement if players can more easily know what is happening in the M59 world

## Example
r7RDfHlO9E


blakserv/ccode.c
```diff
@@ -18,6 +18,13 @@
 
 #include "blakserv.h"
 
+#ifdef BLAK_PLATFORM_WINDOWS
+#include <windows.h>
+#else
```
adrienlaws: I'll clean up these includes since they should be in blakserv.h if they aren't already

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
 	
 	return ret_val.int_val;
 }
+
+// --- BEGIN MULTI-PIPE WEBHOOK SUPPORT ---
+#ifdef BLAK_PLATFORM_WINDOWS
```
akirmse: Move all of the webhook code to a new source file.  It doesn't belong in ccode.c.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+            last_pipe_index = (pipe_index + 1) % NUM_WEBHOOK_PIPES;
+            return;
+        }
+    }
+
+}
+#else
+#include <time.h>
+#define NUM_WEBHOOK_PIPES 1
+void SendWebhookRoundRobin(const char* message, int len) {
```
akirmse: "RoundRobin" doesn't seem like a great name, since there is only 1 pipe here and it's not round robin.  Not that the internal implementation should be part of the name, anyway.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
 	
 	return ret_val.int_val;
 }
+
+// --- BEGIN MULTI-PIPE WEBHOOK SUPPORT ---
+#ifdef BLAK_PLATFORM_WINDOWS
+#include <time.h>
+#define NUM_WEBHOOK_PIPES 10
```
akirmse: Macros are for the 1990s.  Use a static const int.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
 	return ret_val.int_val;
 }
+
+// --- BEGIN MULTI-PIPE WEBHOOK SUPPORT ---
+#ifdef BLAK_PLATFORM_WINDOWS
+#include <time.h>
+#define NUM_WEBHOOK_PIPES 10
+
+void SendWebhookRoundRobin(const char* message, int len) {
+    static int last_pipe_index = 0;
```
akirmse: There is a lot of shared code between the Windows and Linux implementations.  Factor that out.   Also try to put the #ifdef inside the function, so that there's only one function definition, if possible.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+         "{\"timestamp\":%ld,\"message\":\"%.*s\"}", (long)now, len, message);
+    if (msg_len < 0 || msg_len >= (int)sizeof(full_message)) {
+		// If snprintf fails or the message is too long, truncate it
+        msg_len = (int)sizeof(full_message) - 1;
+        full_message[msg_len] = '\0';
+    }
+    for (int i = 0; i < NUM_WEBHOOK_PIPES; ++i) {
+        int pipe_index = (last_pipe_index + i) % NUM_WEBHOOK_PIPES;
+        char pipe_name[64];
+        snprintf(pipe_name, sizeof(pipe_name), "\\\\.\\pipe\\m59apiwebhook%d", pipe_index + 1);
```
akirmse: It seems like weird things will happen if multiple servers are run on the same machine, i.e. that won't be possible.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+    char full_message[1024];
+    time_t now = time(NULL);
+	
+    int msg_len = snprintf(full_message, sizeof(full_message),
+         "{\"timestamp\":%ld,\"message\":\"%.*s\"}", (long)now, len, message);
+    if (msg_len < 0 || msg_len >= (int)sizeof(full_message)) {
+        msg_len = (int)sizeof(full_message) - 1;
+        full_message[msg_len] = '\0';
+    }
+    int fd = open("/tmp/m59apiwebhook1", O_WRONLY | O_NONBLOCK);
```
akirmse: This will silently fail if more than one server is run on the same machine.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+    if (fd == -1) {
+        return;
+    }
+    write(fd, full_message, msg_len);
+    close(fd);
+}
+#endif
+// --- END MULTI-PIPE WEBHOOK SUPPORT ---
```
akirmse: I don't see any pipe usage on Linux.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+            return;
+        }
+    }
+
+}
+#else
+#include <time.h>
+#define NUM_WEBHOOK_PIPES 1
+void SendWebhookRoundRobin(const char* message, int len) {
+    // For Linux use a single pipe
```
akirmse: Where is the pipe?

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+    close(fd);
+}
+#endif
+// --- END MULTI-PIPE WEBHOOK SUPPORT ---
+
+blak_int C_SendWebhook(int object_id, local_var_type *local_vars,
+    int num_normal_parms, parm_node normal_parm_array[],
+    int num_name_parms, parm_node name_parm_array[])
+{
+    val_type msg_val;
```
akirmse: Declare where first used.

blakserv/ccode.h
```diff
@@ -245,9 +245,14 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
 
+blak_int C_SendWebhook(int object_id, local_var_type *local_vars,
+    int num_normal_parms, parm_node normal_parm_array[],
+    int num_name_parms, parm_node name_parm_array[]);
+
 
 void FuzzyCollapseString(char* pTarget, const char* pSource, int len);
 bool FuzzyBufferEqual(const char *s1,int len1,const char *s2,int len2);
 bool FuzzyBufferContain(const char *s1,int len1,const char *s2,int len2);
+void SendWebhookToPipe(const char* message, int len);
```
akirmse: Remove this line?

blakserv/ccode.c
akirmse: Opening a file (or creating a pipe) on every single message is absurdly expensive.  I'm not even sure that just writing to a file that's already open is going to have acceptable performance.

blakserv/ccode.c
```diff
@@ -2344,3 +2351,76 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+    if (msg_len < 0 || msg_len >= (int)sizeof(full_message)) {
+		// If snprintf fails or the message is too long, truncate it
+        msg_len = (int)sizeof(full_message) - 1;
+        full_message[msg_len] = '\0';
+    }
+    for (int i = 0; i < NUM_WEBHOOK_PIPES; ++i) {
+        int pipe_index = (last_pipe_index + i) % NUM_WEBHOOK_PIPES;
+        char pipe_name[64];
+        snprintf(pipe_name, sizeof(pipe_name), "\\\\.\\pipe\\m59apiwebhook%d", pipe_index + 1);
+        HANDLE hPipe = CreateFileA(pipe_name, GENERIC_WRITE, 0, NULL, OPEN_EXISTING, 0, NULL);
```
akirmse: Up to 10 pipes are created every time a single message is sent?  There are no comments describing what's going on here.

blakserv/ccode.c
adrienlaws: Removed after moving webhook functions to webhook.c

blakserv/ccode.c
adrienlaws: Moved to webhook.c and webhook.h

blakserv/ccode.c
adrienlaws: Removed "RoundRobin" from function name

blakserv/ccode.c
adrienlaws: Replaced macros with static const int

blakserv/ccode.c
adrienlaws: Factored out shared logic into common helper functions and moved platform #ifdefs inside `SendWebhookMessage()`

blakserv/ccode.c
adrienlaws: Added configurable pipe prefix support so multiple servers can run simultaneously with unique pipe names (e.g., server1_m59apiwebhook1 vs server2_m59apiwebhook1).

blakserv/ccode.c
adrienlaws: Replaced hard-coded /tmp/m59apiwebhook1 with configurable prefix system so multiple servers use unique pipe paths (e.g., /tmp/server1_m59apiwebhook1).

blakserv/ccode.c
adrienlaws: Messed this up previously, now uses pipes.

blakserv/ccode.c
adrienlaws: Messed this up previously, now uses pipes.

blakserv/ccode.c
adrienlaws: Moved variable declarations to their first use (e.g., val_type msg_val = RetrieveValue(...))

blakserv/ccode.h
adrienlaws: Removed

blakserv/ccode.c
adrienlaws: Implemented persistent pipe connections that stay open between messages

blakserv/ccode.c
adrienlaws: Added comments explaining persistent connection behavior and round-robin logic; pipes are only opened once and reused, not created per message.  Also created a documentation file under docs called webhooks.md as part of this PR since it is a complicated change.

kod/util/system.kod
```diff
@@ -1466,6 +1466,37 @@ messages:
          return;
       }
 
+      % Send webhook notification for player death
```
adrienlaws: This is just an example - I can remove it from the PR.

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+
+bool InitWebhooks(const char* prefix)
+{
+    if (webhook_initialized) {
+        return true;
+    }
+
+    // Check if webhooks are enabled in config
+    if (!ConfigBool(WEBHOOK_ENABLED)) {
+        return false; // Webhooks disabled, return success but don't initialize
```
akirmse: Comment says "return success" but it returns failure?

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+static const int MAX_WEBHOOK_PIPES_WINDOWS = 10;
+static const int MAX_WEBHOOK_PIPES_UNIX = 10;
+
+// Global state for persistent pipe connections
+static bool webhook_initialized = false;
+static char pipe_prefix[64] = "";
+static int last_pipe_index = 0;
+
+#ifdef BLAK_PLATFORM_WINDOWS
+static HANDLE pipe_handles[10];
```
akirmse: Use MAX constants rather than hard-coding 10 again.

Also, you should try hard to reduce the number of #ifdefs.  For example, osd_linux.h already defines HANDLE to int, so you can probably just use the Windows version here on both platforms.

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+#ifdef BLAK_PLATFORM_WINDOWS
+#include <windows.h>
+#else
+#include <fcntl.h>
+#include <unistd.h>
+#endif
+
+// Configuration constants
+static const int MAX_WEBHOOK_PIPES_WINDOWS = 10;
+static const int MAX_WEBHOOK_PIPES_UNIX = 10;
```
akirmse: Do you really need platform-specific constants?  Seems like one would be enough.

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+*
+
+  Cross-platform webhook message delivery via named pipes (Windows) or FIFOs (Linux/macOS).
+  Keeps pipes open for performance instead of opening/closing on every message.
+
+*/
+
+#include "blakserv.h"
+#include "webhook.h"
+#include <time.h>
```
akirmse: Looks like these are both already included in blakserv.h?

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+
+#ifdef BLAK_PLATFORM_WINDOWS
+#include <windows.h>
+#else
+#include <fcntl.h>
+#include <unistd.h>
+#endif
```
akirmse: See if you can get rid of this ifdef.  You certainly don't need to re-include windows.h here.

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+    } else if (config_prefix && *config_prefix) {
+        snprintf(pipe_prefix, sizeof(pipe_prefix), "%s_", config_prefix);
+    } else {
+        pipe_prefix[0] = '\0';
+    }
+
+#ifdef BLAK_PLATFORM_WINDOWS
+    // Initialize Windows pipe handles
+    for (int i = 0; i < num_pipes; i++) {
+        pipe_handles[i] = INVALID_HANDLE_VALUE;
```
akirmse, line 56: If you #define INVALID_HANDLE_VALUE -1 in osd_linux.h, you can get rid of this ifdef.

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+            } else {
+                // Write failed, disconnect
+                CloseHandle(pipe_handles[pipe_index]);
+                pipe_handles[pipe_index] = INVALID_HANDLE_VALUE;
+                pipe_connected[pipe_index] = false;
+            }
+        }
+    }
+#else
+    // Unix/Linux implementation with round-robin like Windows
```
akirmse: There is too much duplicated code in these ifdef branches.  Try to reduce it.

docs/webhooks.md
```diff
@@ -0,0 +1,367 @@
...
+### 4. Start the Services (Order is Important!)
+```bash
+# Terminal 1: Start webhook listener FIRST
+m59api webhook
+
+# Terminal 2: Start your M59 server SECOND
+./blakserv
+```
+
+**?? Important:** The webhook listener must be started before the M59 server, as the server connects to pipes created by the listener.
```
akirmse: What will happen if you get the order wrong?  Could be useful to tell people what to look for.

kod/util/system.kod
akirmse: I wonder if it might be better to send some kind of text-based protocol, rather than the final user-visible messages.  As-is, if you wanted to change a message, you'd have to wait for a game update (months).  But if the final messages were in the Python process, you could update them much more easily.

Also, system.kod is already super enormous--maybe think about a way to segregate the webhook logic in a separate (probably new) Blakod file.

blakserv/main.c
```diff
@@ -99,6 +99,11 @@ void MainServer()
 	InitTable();
 	AddBuiltInDLlist();
 	
+	/* Initialize webhook system - controlled by [Webhook] Enabled config setting
```
akirmse: This comment is out of place here

blakserv/main.c
```diff
@@ -99,6 +99,11 @@ void MainServer()
 	InitTable();
 	AddBuiltInDLlist();
 	
+	/* Initialize webhook system - controlled by [Webhook] Enabled config setting
+	   Multiple servers are automatically supported - each server claims an available pipe
+	   from the pool of 10 pipes created by external webhook listeners */
+	InitWebhooks(NULL);
```
akirmse: Since NULL is always passed for the prefix, what is the purpose of this argument?

blakserv/webhook.c
```diff
@@ -0,0 +1,222 @@
...
+{
+    if (!webhook_initialized) {
+        return;
+    }
+
+#ifdef BLAK_PLATFORM_WINDOWS
+    // Close all Windows pipe handles
+    for (int i = 0; i < num_pipes; i++) {
+        if (pipe_connected[i] && pipe_handles[i] != INVALID_HANDLE_VALUE) {
+            CloseHandle(pipe_handles[i]);
```
akirmse, line 73: If you make your own platform-dependent close function, you should be able to get rid of this ifdef.

blakserv/webhook.h
```diff
@@ -0,0 +1,72 @@
...
+  MULTI-SERVER SUPPORT:
+  - External webhook listeners create multiple pipe servers (webhookN, where N=1-10)
+  - Each M59 server instance automatically claims an available pipe
+  - Up to 10 concurrent servers supported per webhook listener instance
+  - No configuration needed - automatic pipe distribution
+
+  PERFORMANCE:
+  - Pipes are opened once at startup and kept open (not per-message)
+  - Non-blocking operations prevent game server blocking
+  - Only 1 system call per message vs 3 in naive implementation
```
akirmse: I don't understand this last comment--you're documenting an implementation that doesn't exist?

blakserv/ccode.c
```diff
@@ -2299,3 +2299,131 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+                        json_buffer[json_pos++] = '\\';
+                    }
+                    json_buffer[json_pos++] = value_str[j];
+                }
+                json_pos += snprintf(json_buffer + json_pos, sizeof(json_buffer) - json_pos, "\"");
+            } else {
+                json_pos += snprintf(json_buffer + json_pos, sizeof(json_buffer) - json_pos, "null");
+            }
+        } else if (value_val.v.tag == TAG_INT) {
+#ifdef BLAK_PLATFORM_WINDOWS
```
akirmse: You can do this without an #ifdef by either casting the numeric argument:

https://github.com/Meridian59/Meridian59/blob/e62ce3dc1c2893e2c545775fd37674c8b9189824/blakserv/term.c#L122

or using a portable format specifier:

https://github.com/Meridian59/Meridian59/blob/e62ce3dc1c2893e2c545775fd37674c8b9189824/blakserv/loadgame.c#L541

adrienlaws: Thank you, I just switched to using the portable format specifier as suggested.

kod/util/system.kod
adrienlaws: Fixed - now sends structured `JSON` so external services can format messages without game updates. Only one `SendWebhook()` call added to `system.kod` (for death events), so keeping it here rather than a separate file unless you disagree.

blakserv/webhook.c
adrienlaws: Fixed - now returns true when webhooks are disabled since that's not an error condition.

blakserv/webhook.c
adrienlaws: Fixed both - now uses `MAX_WEBHOOK_PIPES `constant throughout (no more hard-coded 10), and unified to single HANDLE array for both platforms.

blakserv/webhook.c
adrienlaws: Fixed - consolidated to single `MAX_WEBHOOK_PIPES = 10` constant.

blakserv/webhook.c
adrienlaws: Fixed - removed the redundant include block.

blakserv/webhook.c
adrienlaws: Fixed - removed redundant includes since both` webhook.h` and` time.h` are already in `blakserv.h`.

blakserv/webhook.c
adrienlaws: Fixed - added `INVALID_HANDLE_VALUE` to `osd_linux.h` and removed this `#ifdef` block. The init code is now platform agnostic.

blakserv/webhook.c
adrienlaws: Moved platform-specific code into helper functions to eliminate duplication.

blakserv/webhook.c
adrienlaws: Fixed - added `CloseHandle` macro to `osd_linux.h` (defined as close).

blakserv/main.c
adrienlaws: Removed comment and updated` webhooks.md` documentation with this information.

blakserv/main.c
adrienlaws: Removed.

blakserv/webhook.h
adrienlaws: Fixed.  Refactored the file to be relevant to the most current implementation.  I probably will need to check right at the end that the comments here, the PR description, and the webhooks.md documentation all match the final state.  I'm trying to update them as I go, but I missed this one.

docs/webhooks.md
adrienlaws: Added additional documentation.

blakserv/ccode.c
```diff
@@ -2299,3 +2299,119 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+        // Get the value
+        val_type value_val = RetrieveValue(object_id, local_vars,
+            normal_parm_array[i + 1].type, normal_parm_array[i + 1].value);
+        
+        const char *value_str;
+        int value_len;
+        
+        // Add comma if not first parameter
+        if (i > 1) {
+            json_pos += snprintf(json_buffer + json_pos, sizeof(json_buffer) - json_pos, ",");
```
akirmse: These snprintfs are pretty awkward and a bit error-prone with the string offsets.  I'd use a std::string for json_buffer instead.

blakserv/webhook.c
```diff
@@ -0,0 +1,181 @@
...
+                pipe_connected[pipe_index] = true;
+            }
+        }
+        
+        // Try to send message
+        if (pipe_connected[pipe_index]) {
+            if (write_webhook_pipe(pipe_handles[pipe_index], message_to_send, json_len)) {
+                last_pipe_index = (pipe_index + 1) % MAX_WEBHOOK_PIPES;
+                return true;
+            } else {
```
akirmse: You don't need an "else" here since the other branch always returns.

blakserv/webhook.c
```diff
@@ -0,0 +1,181 @@
...
+                // Write failed, disconnect
+                CloseHandle(pipe_handles[pipe_index]);
```
akirmse, line 133: One thing to think about is what happens if the pipe gets backed up (e.g. if the receiving end is slow for some reason -- maybe the Python script is doing something heavy, or it get stuck talking to Discord).  In these conditions, the write() call will return -1 and set errno to EWOULDBLOCK, if I remember correctly.  Since you close the pipe in this case, if there are a bunch of webhook writes while the reader is stalled, all of the pipes will be closed.  And then I guess you'll try to reopen them on every subsequent send?  I'm not sure of the performance implications.

This situation is very similar to the way the server does networking over sockets.  If the client is slow to read, the server queues up the pending send data in its own buffers and sends them later.  However, this system is very complicated and uses more memory.  I think it would be overkill to implement something similar for pipes.  But you should think about this case and document what will happen.

blakserv/ccode.c
```diff
@@ -2299,3 +2299,110 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+    if (num_normal_parms == 1) {
+        msg_val = RetrieveValue(object_id, local_vars, normal_parm_array[0].type, normal_parm_array[0].value);
+        
+        if (!LookupString(msg_val, "C_SendWebhook", &content, &content_len)) {
+            bprintf("C_SendWebhook: LookupString failed\n");
+            return NIL;
+        }
+
+        SendWebhookMessage(content, content_len);
+        return 1;
```
akirmse: "1" is not a valid Blakod value, since the tag field hasn't been set.  If you want to return an integer, you have to do it like this:

https://github.com/Meridian59/Meridian59/blob/e62ce3dc1c2893e2c545775fd37674c8b9189824/blakserv/ccode.c#L1888

But what is the meaning of the return value for SendWebhook?  If there isn't one, you should always return NIL.

blakserv/ccode.c
```diff
@@ -2299,3 +2299,110 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+            json += std::to_string(value_val.v.data);
+        }
+    }
+    
+    // Close JSON: }}
+    json += "}}";
+    
+    // Send the webhook message
+    SendWebhookMessage(json.c_str(), (int)json.length());
+    return 1;
```
akirmse: Same issue with the return value here.

blakserv/webhook.c
```diff
@@ -0,0 +1,211 @@
...
+static HANDLE open_webhook_pipe(const char *pipe_name)
+{
+#ifdef BLAK_PLATFORM_WINDOWS
+    return CreateFileA(pipe_name, GENERIC_WRITE, 0, NULL, OPEN_EXISTING, 0, NULL);
+#else
+    return open(pipe_name, O_WRONLY | O_NONBLOCK);
+#endif
+}
+
+static bool write_webhook_pipe(HANDLE handle, const char *data, int len, bool *should_close)
```
akirmse: "should_close" is a confusing name for this parameter.  You are assuming that the caller of this function wants to close the pipe on certain error codes, but you don't actually know that.  The name should indicate what the value means, not what the caller might want to do with it.

Also, this function needs a comment at the start to describe what it does--this parameter in particular.

blakserv/webhook.c
```diff
@@ -0,0 +1,211 @@
...
+    if (bytes_written == len) {
+        return true;
+    }
+    
+    // Check if pipe buffer is full (temporary) vs broken (permanent)
+    if (bytes_written == -1) {
+        // EAGAIN/EWOULDBLOCK means pipe buffer is full but pipe is still valid
+        *should_close = (errno != EAGAIN && errno != EWOULDBLOCK);
+    } else {
+        // Partial write - treat as permanent error
```
akirmse: Non-blocking I/O is tricky.  If the pipe buffer is almost full, but not totally full, it's quite possible for part of the message to be sent, and then received on the other end.  But since you then close the pipe, the rest of the message will be discarded.  So it could happen that the other end of the pipe (the Python script or whatever) will sometimes get truncated messages.  FYI---not sure how much time you want to spend on making this perfect.

blakserv/ccode.c
adrienlaws: Changed to return `NIL `as suggested.

blakserv/ccode.c
adrienlaws: Changed to return `NIL `as suggested.

blakserv/webhook.c
adrienlaws: Changed parameter name and added function documentation.

blakserv/webhook.c
adrienlaws: I changed the function so if there's not enough room in the pipe, then try another pipe.  If all pipes are unavailable, then drop the webhook message attempt.  Updated the webhooks.md file with this case as an example.

blakserv/webhook.c
adrienlaws: Removed.

blakserv/ccode.c
adrienlaws: Changed as suggested.

blakserv/webhook.c
adrienlaws: Refactored to try and address this scenario.

blakserv/webhook.c
```diff
@@ -0,0 +1,217 @@
...
+ * write_webhook_pipe:  Write data to webhook pipe. Distinguishes temporary errors
+ *    (buffer full, partial write) from permanent errors (broken pipe) via 
+ *    is_permanent_error output parameter. 
+ *    Returns true if full message sent, false otherwise.
+ */
+static bool write_webhook_pipe(HANDLE handle, const char *data, int len, bool *is_permanent_error)
+{
+#ifdef BLAK_PLATFORM_WINDOWS
+    DWORD bytes_written;
+    BOOL success = WriteFile(handle, data, (DWORD)len, &bytes_written, NULL);
```
akirmse, line 195: Are these writes on Windows blocking or non-blocking?  I don't see any reference to making this non-blocking.

blakserv/webhook.c
```diff
@@ -0,0 +1,217 @@
...
+                last_pipe_index = (pipe_index + 1) % MAX_WEBHOOK_PIPES;
+                return true;
+            }
+            
+            // Write failed - only close pipe if it's a real error (not just buffer full)
+            if (is_permanent_error) {
+                CloseHandle(pipe_handles[pipe_index]);
+                pipe_handles[pipe_index] = INVALID_HANDLE_VALUE;
+                pipe_connected[pipe_index] = false;
+            }
```
akirmse, line 136: I'm not sure that this new approach does what you want.  Suppose the message being sent is "ABCDEF", and many of the pipe buffers are almost full.  Since you always retry the full message on a different pipe, here are some potential results the reader may see (since the order of pipe events isn't guaranteed):

ABCDEF

or

ABC
ABCDEF

or

AB
ABCDE
A

blakserv/webhook.c
```diff
@@ -0,0 +1,217 @@
...
+    } else {
+        // Legacy format: wrap plain text in JSON with timestamp
+        format_json_message(message, len, json_message, sizeof(json_message));
+        message_to_send = json_message;
+        json_len = (int)strlen(json_message);
+    }
+
+    // Try to send using round-robin across all pipes
+    for (int i = 0; i < MAX_WEBHOOK_PIPES; i++) {
+        int pipe_index = (last_pipe_index + i) % MAX_WEBHOOK_PIPES;
```
akirmse, line 110: Is there a possibility that the loop never exits now?

adrienlaws: I think this is okay unless I'm missing something. We have `MAX_WEBHOOK_PIPES  `set to 10 and either the loop exists from return true or it iterates after returning false. It should loop 10 times and then exit the loop.

blakserv/webhook.c
adrienlaws: It was blocking - changed it to non-blocking.  I missed it.

blakserv/webhook.c
adrienlaws: Thanks this is a good point.  Changed to close the pipe and try the next one if the target pipe's buffer gets full (instead of partially sending then trying the next pipe).

Also updated webhooks.md with this scenario.

blakserv/ccode.c
```diff
@@ -2299,3 +2299,110 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+    event_val = RetrieveValue(object_id, local_vars, 
+        normal_parm_array[0].type, normal_parm_array[0].value);
+    
+    if (!LookupString(event_val, "C_SendWebhook", &event_name, &event_len)) {
+        bprintf("C_SendWebhook: Event name lookup failed\n");
+        return NIL;
+    }
+    
+    // Build JSON using std::string
+    json.reserve(4096);
```
akirmse: You don't need this.  Strings grow to fit their contents.

blakserv/webhook.c
akirmse: I think you'll still have the same potential problem.  If you try to write 10 bytes and 5 bytes are written, then half the message will be delivered.  Pipes are bytes-based, not message-based.

Again, not sure that this is worth fixing.

blakserv/ccode.c
adrienlaws: Removed.

blakserv/webhook.c
adrienlaws: I documented the edge case in webhooks.md.  The current implementation limits corruption a single message since we close and reopen the pipe (vs before where the pipe was left open).  Is that okay or should I change something?

PR #1177, opened 2025-06-25
fix: refresh wholist colors after changes in a player's PK/OUTLAW flags

adrienlaws:
## What
- automatically update the who list for all players when PK-related flags change
- also added logic through `IsServerCrowded()` to only perform this expensive operation if online player count < 100
 
## Why
- previously, the who list did not reflect changes to murderer/outlaw status until the receiving player logged off and on
- this ensures all players see accurate status in the who list immediately after any PK flag change
- there were some comments [in the Discord about this](https://discord.com/channels/650824489885630475/650824624908533791/1385950917697736835)
  - players noticed some outlaws and murderers appearing as white in person, but still having red or orange in the who list
- addresses maintainability concerns by making the update automatic rather than requiring manual calls

## How
- added conditional logic in `SetPlayerFlag` to call `UpdateWhoListForAll` when `PFLAG_MURDERER` or `PFLAG_OUTLAW` flags are modified
- includes server load protection via `IsServerCrowded()` to skip updates during peak usage
- this automatically handles updates for:
  - admin `villify` and `pardon` commands
  - justicar pardons 🧑‍⚖️ ⚖️
  - combat-related PK flag changes
  - any future code that modifies these flags
- on the margin I think this makes sense over manual calls since PK-related flag changes are relatively infrequent
- consolidated with existing room notification logic to avoid code duplication


akirmse:
This seems like it might be OK, but I can easily imagine that in the future, someone might add a mechanism that changes a flag value but forgets to call UpdateWhoListForAll.  Ideally, this could be done inside SetPlayerFlag so that it's automatic and doesn't require code changes every time a flag is updated.  However, naively adding a call to UpdateWhoListForAll inside SetPlayerFlag would be very inefficient, as often multiple flags are changed at once (murderer + outlaw), and sometimes multiple players' flags are changed simultaneously (Pardon, Villify).  Any thoughts on this?

adrienlaws:
I have updated this PR to include `UpdateWhoListForAll` in  `SetPlayerFlag`

It's a dilemma as you mentioned.
- 1/ Either future proof with known inefficiencies, or
- 2/ Know that over time someone will implement a flag change that won't know anything about `UpdateWhoListForAll` and it will be broken again

Having to choose between the two, I choose option 1/ automatic in `SetPlayerFlag`

The mentioned downside of inefficiency is okay I think, on the margin.  The PK flags are updated relatively infrequently and since the call occurs in only one place now, that can be modified if deemed necessary.  The alternative of chasing PK flag implementations and matching them to `UpdateWhoListForAll` seems difficult for a generational codebase with new contributors joining and others leaving.

akirmse (commented):
Yeah, I agree.  But given that this could make a very busy server worse, what about this idea:

- Add a new System function called IsServerCrowded, which checks whether the number of players logged on is over a certain threshold (maybe 100 to start).  Put a comment there to explain that this function should be used to cut back on expensive operations when the server is busy.

- Only call UpdateWhoListForAll if NOT IsServerCrowded.

adrienlaws:
- added `IsServerCrowded` system message as recommended
- added logic to only use `UpdateWhoListForAll` if <100 players online
- I tested it by setting the `CROWDED_SERVER_THRESHOLD` constant to `1` and it worked as expected when pardoning a player (the who list did NOT update in that case, but DID update when it was set to `100`).
  - no errors seen in the logs
  - also tested sending `IsServerCrowded` to object 0 (the server I believe) and didn't see any errors


image

PR #1178, opened 2025-06-26
feature: add item rarity indicators to inventory UI

adrienlaws:
## What
- this change adds a small visual indicator for item rarity on inventory items
- also adds a configurable option in the client preferences to toggle it on or off (default = on)

## Why
- previously, item rarity grades were not visible in the inventory UI
- players no longer need to open other dialogs to find rare items in their inventory
- completes the rarity notification system so rarity is always visible at a glance

## Notes
- uses existing rarity flags from the object data
- importing from `color.c` was not possible due to module linkage issues, so rarity color logic is duplicated (this might be sloppy/bad)
- the code is perhaps excessively commented to make it easier for others to tweak the colors and 3D effect (I'm no artist)

## Attribution
- special thank you to @davidklaw for significantly enhancing the rarity graphics

### Rarity symbol examples
meridian_Pimx9EaV05

### Toggling preferences on and off
meridian_a4MaHzlVeL

### DIFF
--- a/module/merintr/inventry.c
+++ b/module/merintr/inventry.c
@@ -679,7 +679,7 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
          MoveToEx(hdc, left-1, top-1, NULL);
          LineTo(hdc, right-1, top-1);  // Top edge (stop before corner)
          MoveToEx(hdc, left-1, top-1, NULL);
-         LineTo(hdc, left-1, bottom-1); // Left edge (stop before corner)
+         LineTo(hdc, left-1, bottom-1);  // Left edge (stop before corner)
          SelectObject(hdc, oldPen);
          DeleteObject(hPenDark);
 
@@ -687,16 +687,16 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
          HPEN hPenLight = CreatePen(PS_SOLID, 1, RGB(150,150,150));
          SelectObject(hdc, hPenLight);
          MoveToEx(hdc, left-1, bottom, NULL);
-         LineTo(hdc, right-1, bottom); // Bottom edge (stop before corner)
+         LineTo(hdc, right-1, bottom);  // Bottom edge (stop before corner)
          MoveToEx(hdc, right, top-1, NULL);
-         LineTo(hdc, right, bottom-1); // Right edge (stop before corner)
+         LineTo(hdc, right, bottom-1);  // Right edge (stop before corner)
          SelectObject(hdc, oldPen);
          DeleteObject(hPenLight);
 
          COLORREF shadow = RGB(
-            max(GetRValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT , 0),
-            max(GetGValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT , 0),
-            max(GetBValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT , 0)
+            max(GetRValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT, 0),
+            max(GetGValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT, 0),
+            max(GetBValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT, 0)
          );
          HPEN hPenShadow = CreatePen(PS_SOLID, 1, shadow);
          oldPen = SelectObject(hdc, hPenShadow);


akirmse:
Highly-visible UI changes like this always annoy some fraction of players, so let's also add a setting to turn this off.  It can default to on.

What are "module linkage issues"?

module/merintr/inventry.c
```diff
@@ -636,66 +636,69 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
+         // Draw 3D effect: shadow (top/left), highlight (bottom/right)
+         COLORREF shadow = RGB(
+            max(GetRValue(rarityColor) - 80, 0),
+            max(GetGValue(rarityColor) - 80, 0),
+            max(GetBValue(rarityColor) - 80, 0)
+         );
+         COLORREF highlight = RGB(
+            min(GetRValue(rarityColor) + 80, 255),
+            min(GetGValue(rarityColor) + 80, 255),
+            min(GetBValue(rarityColor) + 80, 255)
```
akirmse: 80 is used 6 times.  Make it a named constant.

module/merintr/inventry.c
```diff
@@ -636,66 +636,69 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
-      DeleteObject(hPenShadow);
-      DeleteObject(hPenHighlight);
+   if (cinfo->config->show_inventory_rarity)
+   {
+      // Add rarity color if item is not normal; with 3D/inset effect
+      if (item->obj && item->obj->rarity != ITEM_RARITY_GRADE_NORMAL) {
+         COLORREF rarityColor = GetItemRarityColor(item->obj->rarity);
+
+         // Square size and position
+         int boxSize = 6;
```
akirmse: should be const

clientd3d/config.h
```diff
@@ -58,6 +58,7 @@ typedef struct {
    int  ini_version;             /* INI version number; restore defaults if it doesn't match */
 
    Bool draw_names;              /* Draw names over players? */
+   Bool show_inventory_rarity;   /* Add inventory item rarity symbols? */
```
akirmse: For safety, new elements of the config need to be added to the end of the struct.  That's because the structure is shared with module DLLs, so if something is added in the middle, the offsets of everything after it will change, potentially causing modules to crash.

module/merintr/inventry.c
```diff
@@ -98,6 +98,26 @@ keymap inventory_key_table[] = {
 { 0, 0, 0},   // Must end table this way
 };
 
+#define COLOR_ITEM_UNCOMMON     RGB(141,242,242)
```
akirmse: Use static const ints, not #defines

module/merintr/inventry.c
```diff
@@ -98,6 +98,26 @@ keymap inventory_key_table[] = {
...
+#define COLOR_ITEM_RARE         RGB(0,255,0)
+#define COLOR_ITEM_LEGENDARY    RGB(255,0,255)
+#define COLOR_ITEM_UNIDENTIFIED RGB(252,128,0)
+#define COLOR_ITEM_CURSED       RGB(255,0,0)
+#define COLOR_ITEM_DEFAULT      RGB(255,255,255) // fallback
+
+static COLORREF GetItemRarityColor(item_rarity_grade rarity)
```
akirmse: Functions go after local function prototypes

clientd3d/meridian.ini
```diff
@@ -72,6 +72,7 @@ ServerNumber=0
 DrawNames=1
 IgnoreAll=0
 NoBroadcast=0
+DrawInvRarities=1
```
akirmse: "Inv" seems like a potentially confusing abbreviation.  Inverse?  Spell out "Inventory", it doesn't cost you anything.

clientd3d/client.rc
```diff
@@ -1190,6 +1191,7 @@ BEGIN
     CONTROL         "Schrittschwankungen anzeigen",IDC_BOUNCE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,111,17,132,9
     CONTROL         "Schmerzen bei Treffern anzeigen",IDC_PAIN,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,111,28,132,9
     CONTROL         "Namen der Spieler anzeigen",IDC_DRAWNAMES,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,111,38,132,9
+    CONTROL         "Seltenheit der Inventargegenstande anzeigen",IDC_INV_RARITY,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,111,61,132,9
```
akirmse: Accents are not optional; you need to put them back.  Probably you're opening this file with the wrong encoding?  @davidklaw might be able to assist here.

module/merintr/inventry.c
```diff
@@ -616,6 +636,81 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
+         RECT rc = { left, top, right, bottom };
+         FillRect(hdc, &rc, hBrush);
+         DeleteObject(hBrush);
+
+         // Draw 3D effect: shadow (top/left), highlight (bottom/right)
+         COLORREF highlight = RGB(255,255,255);
+
+         HBRUSH hBrushHighlight = CreateSolidBrush(highlight);
+
+         int highlightSize = 4;
```
akirmse: const

module/merintr/inventry.c
```diff
@@ -616,6 +636,81 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
+         MoveToEx(hdc, left, bottom-1, NULL);
+         LineTo(hdc, right, bottom-1);
+         MoveToEx(hdc, right-1, top, NULL);
+         LineTo(hdc, right-1, bottom);
+         
+         SelectObject(hdc, oldPen);
+         DeleteObject(hPenShadow);
+
+         DeleteObject(hPenDark);
+         DeleteObject(hPenLight);
```
akirmse, line 713: Creating and deleting 3 objects for each little inventory box seems pretty heavyweight, but maybe there's no noticeable performance problem in practice.  Is performance OK even with lots of these being drawn?

davidklaw: I can provide bitmaps if needed.

clientd3d/client.rc
davidklaw: I always edit these `rc` files with Windows 1252 encoding. 

In Visual Studio Code:
1. ctrl+shift+P
2. "Change File Encoding"
3. "Reopen With Encoding"
4. Search "1252" and select "Western (Windows 1252)"

adrienlaws:
re the concern about the highlights not going away immediately, but coming on immediately:
> In the video, it looks like the highlights don't go away immediately when the config option is turned off? If so, that needs to be fixed.

An example preference I notice in prod (official client) that works is `show amounts for inventory items`
- it functions as expected
  - tick or untick box, click ok, and the numbers appear or disappear
- in dev which has a new way of handling preferences, I see it working exactly the same as my change here
  - tick on and it works immediately after clicking ok, but tick off, click ok, and you must refresh the window for it to take effect

### Example
meridian_53UXTAgH5m

So I guess I can try and fix it in this PR or should we try and fix that behavior in a separate PR?  It sounds like unwanted behavior from `preferences.c`

clientd3d/client.rc
adrienlaws: Re-added the accents after switching encoding in my VS Code to Windows 1252, as suggested.  Thank you both.

clientd3d/meridian.ini
adrienlaws: Renamed `DrawInvRarities` to `DrawInventoryRarities` as suggested.

akirmse:
Hm, that is disturbing... we should try to fix this preference change bug now.  Can you add a github issue for it?  Any idea what caused it?

adrienlaws:
Created issue #1200 for tracking.

I can only speculate that the consolidation of the `preferences` and `configuration` sections (there used to be two distinct configuration areas, now there is only one named `preferences`) means that we need to adjust something.  I don't want to speculate too much but that seems like the first place to look.

I also know that `preferences.c` has `#include "client.h"` at the top which does not contain `inventry.h` since it's in a different module.  Otherwise, I was going to call `InventoryRedraw()` for my change and it would have been trivial to fix for this PR.  So it can't easily call a quick refresh of the inventory window.

adrienlaws:
I have confirmed that with the #1201 fix, the highlights now appear and disappear as wanted after updating preferences.  Updated [this PR's description](https://github.com/Meridian59/Meridian59/pull/1178#issue-3180332610) with a relevant GIF showing this functionality.

module/merintr/inventry.c
adrienlaws: It seems okay I don't notice any hiccups at all.  I did update my system with a newer CPU recently but I don't notice any slowdown at all.  I'm holding a very full inventory with a dozen rare items in my tests.

clientd3d/config.h
adrienlaws: Moved the new element to the end of the struct, as recommended.  I appreciated the explanation as to why it was needed.

module/merintr/inventry.c
adrienlaws: Initial comment is on outdated code but changed shadow + 125 value to named constant `INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT`

module/merintr/inventry.c
adrienlaws: Added const

module/merintr/inventry.c
adrienlaws: Changed to static const ints as recommended

module/merintr/inventry.c
adrienlaws: Moved to functions section

module/merintr/inventry.c
adrienlaws: Changed

adrienlaws:
This PR is ready for another review, when possible.

I have made changes based on feedback from the previous review and I also updated the code to use bool instead of Bool so it compiles correctly after the recent other changes to the codebase.

I don't see any errors in the client or on the server in my local dev environment testing.

module/merintr/inventry.c
```diff
@@ -98,6 +98,13 @@ keymap inventory_key_table[] = {
 { 0, 0, 0},   // Must end table this way
 };
 
+static const int COLOR_ITEM_UNCOMMON     = RGB(141,242,242);
```
akirmse: Shouldn't the types of these be COLORREF, since that's what GetItemRarityColor returns?  It's also clearer.

A one-word comment at the end of each line saying what color it is wouldn't hurt.

module/merintr/inventry.c
```diff
@@ -616,6 +639,81 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
+         SelectObject(hdc, hPenLight);
+         MoveToEx(hdc, left-1, bottom, NULL);
+         LineTo(hdc, right-1, bottom); // Bottom edge (stop before corner)
+         MoveToEx(hdc, right, top-1, NULL);
+         LineTo(hdc, right, bottom-1); // Right edge (stop before corner)
+         SelectObject(hdc, oldPen);
+         DeleteObject(hPenLight);
+
+         COLORREF shadow = RGB(
+            max(GetRValue(rarityColor) - INVENTORY_OBJECT_RARITY_SHADOW_AMOUNT , 0),
```
akirmse: Remove extraneous space (and in following lines)

module/merintr/inventry.c
```diff
@@ -616,6 +639,81 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
+         MoveToEx(hdc, left-1, top-1, NULL);
+         LineTo(hdc, left-1, bottom-1); // Left edge (stop before corner)
+         SelectObject(hdc, oldPen);
+         DeleteObject(hPenDark);
+
+         // Draw outline for inset effect - bottom and right edges (light)
+         HPEN hPenLight = CreatePen(PS_SOLID, 1, RGB(150,150,150));
+         SelectObject(hdc, hPenLight);
+         MoveToEx(hdc, left-1, bottom, NULL);
+         LineTo(hdc, right-1, bottom); // Bottom edge (stop before corner)
```
akirmse: Two spaces between ; and // to make it more visible (throughout the file)

module/merintr/inventry.c
```diff
@@ -616,6 +639,81 @@ void InventoryDrawSingleItem(InvItem *item, int row, int col)
...
+      // Add rarity color with 3D/inset effect if item is not normal
+      if (item->obj && item->obj->rarity != ITEM_RARITY_GRADE_NORMAL) {
+         const COLORREF rarityColor = GetItemRarityColor(item->obj->rarity);
+
+         // Square size and position
+         const int boxSize = 10;
+         const int right = area.x + INVENTORY_BOX_WIDTH - 4;
+         const int left  = right - boxSize;
+         const int top   = area.y + 4;
+         const int bottom= top + boxSize;
```
akirmse: space before =

adrienlaws: Added space before =

module/merintr/inventry.c
adrienlaws: Reviewed the new function and ensured there are two spaces before every // occurrence

module/merintr/inventry.c
adrienlaws: removed space before the comma on three lines

module/merintr/inventry.c
adrienlaws: Changed type to COLORREF and added one word color comments that match color.c

PR #1182, opened 2025-07-21
feat: Add NPC responses to guide disciple quest progression

### LINKED ISSUE #711: Quest: Dialogue: NPC: Priestess Xiana is missing a reply to the player after they say the Qor Disciple Quest speech.
For the Qor version of the Disciple quest (needed to learn level 3 - 5 of any spell school), the player gets a quote from Zuxana to repeat to various NPC's. 

When the target NPC is the Shal'ille Priestess Xiana, after the player says the required words  [the player must say the part after the hyphen and inbetween quotes.], she doesn't reply to the player, which can cause confusion to the player if they said the right section of dialogue.

References:
viQuestID=QST_ID_QOR_DISCIPLE

_Note: I didn't debug which other NPC's were quest nodes and if this is a global issue for all Qor Disciple Quest targets at the time of this issue write-up._

davidklaw:
This PR fixes #711 by adding responses to the NPCs involved in the Qor, Faren, and Shal'ille disciple quests. Each of these quests asks the player to say a specific line to another NPC. Right now, those NPCs stay silent, which makes it unclear if the player did the right thing or should return to the quest giver. This adds a simple response to help communicate that.

Each group of NPCs uses a shared line that reflects their alignment and tone. The Qor targets are good-aligned and reject the message outright. The Faren targets are being chastised and respond with dismissal. And in the Shal'ille quest, the targets reject the player’s offer of peace and sends them back with a cold dismissal.

This should help players move through these quests with a bit more clarity.

### Examples of NPC responses for the message delivery step

faren_1
shal_1
qor_1

### DIFF
--- a/kod/util/questengine.kod
+++ b/kod/util/questengine.kod
@@ -176,6 +176,7 @@ resources:
    shalilledisciple_assign = "`bYou wish to learn more of the gentle, healing ways of our good Lady.  You must "
          "brave the temple of %NPC, and proclaim your willingness to forgive their violence and brutality. Heal "
          "them if you can.  Say, \"%CARGO\"  Then, return and tell me that you have done as I asked."
+   shalilledisciple_invoke_success = "Go back and tell your priestess her blessing is neither wanted nor needed."
    shalilledisciple_success = "`bBlessed be this day. Today there is the birth of another soul filling with the "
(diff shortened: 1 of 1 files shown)


davidklaw:
This PR has been updated to address the 3 disciple quests for a lack of NPC response can be confusing: Qor, Shal'ille, and Faren.

PR #1183, opened 2025-07-21
Disallow invalid username and password characters during signup

sicoderealm:
Previously, non-ASCII characters (e.g., ó) were being silently replaced with visually similar ASCII characters (e.g., ó replaced by regular o) by the signup form itself, which could be especially problematic for fields like passwords. This PR introduces validation to detect and reject invalid characters, rather than silently replacing them, ensuring users are explicitly informed of the issue.

Additionally:

- "Select all" functionality has been added to applicable fields.
- Fields left empty will now trigger clear, field-specific error messages instead of the generic "sign-up failed" message previously shown.

image

### DIFF
--- a/clientd3d/signup.c
+++ b/clientd3d/signup.c
@@ -109,24 +109,6 @@ void ShowSignupErrors(HWND hDlg, long errorFlags)
 }
 
 /*
-* EditSubclassProc to support select all on the signup edit fields.
-*/
-LRESULT CALLBACK EditSubclassProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam, UINT_PTR uIdSubclass,
-                                  DWORD_PTR dwRefData)
-{
-   if (msg == WM_KEYDOWN)
-   {
-      if ((GetKeyState(VK_CONTROL) & 0x8000) && wParam == 'A')
-      {
-         SendMessage(hwnd, EM_SETSEL, 0, -1);
-         return 0;  // handled
-      }
-   }
-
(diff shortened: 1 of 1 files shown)


clientd3d/signup.c
```diff
@@ -47,6 +46,86 @@ INT_PTR ApplyErrorStyleToStaticText(UINT wParam, LONG lParam)
     return (INT_PTR)hBrush;
 }
 
+// Returns true if all characters are ASCII alphanumeric (a requirement for account usernames and passwords).
+bool IsValidNonEmptyAscii(const std::wstring &wstr)
```
akirmse: Make static

clientd3d/signup.c
```diff
@@ -47,6 +46,86 @@ INT_PTR ApplyErrorStyleToStaticText(UINT wParam, LONG lParam)
...
+            ss << " - ";
+            ss << errorStr;
+         }
+      }
+   }
+   ClientError(hInst, hDlg, IDS_SIGNUPFAILED, ss.str().c_str());
+}
+
+/*
+* EditSubclassProc to support select all on the signup edit fields.
```
akirmse: This is unexpected -- is it related to disallowing invalid characters?

sicoderealm: Raised this PR https://github.com/Meridian59/Meridian59/pull/1187 having select all is really helpful for anyone using or testing /debugging this signup flow.

akirmse:
It might be a good follow-up to indicate on the dialog that only A-Z 0-9 are allowed, otherwise we're going to have a lot of confused users as to why their passwords are invalid.

PR #1168, opened 2025-05-29
feat: Add clang-format configuration for C formatting support

davidklaw:
This PR adds `.clang-format` and `.clang-format-ignore` to support C formatting for files in the following directories:
- `blakcomp/`
- `blakdeco/`
- `blakdiff/`
- `blakserv/`
- `clientd3d/`
- `club/`
- `makebgf/`
- `module/`

PR #1186, opened 2025-07-24
format: clang formatting of several files, updated clang-format rules

davidklaw:
This applies the new clang formatting rules (#1168) to a handful of files to clean things up before making further changes. Let me know if there's anything about these rules we’d want to tweak based on how this turned out.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -20,7 +20,11 @@ BraceWrapping:
   BeforeCatch: true
 
 # Comments
+AlignTrailingComments: true
 SpacesBeforeTrailingComments: 2
+ReflowComments: true
+SpacesInLineCommentPrefix:
+  Maximum: 1
 
 # Spacing
 SpaceBeforeParens: ControlStatements
@@ -30,3 +34,6 @@ SpaceBeforeAssignmentOperators: true
 # Line breaks
 ColumnLimit: 120
 MaxEmptyLinesToKeep: 1
+
+# Alignment
+AlignConsecutiveMacros: true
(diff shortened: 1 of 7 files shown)


akirmse:
Mostly a big improvement, but:

- There need to be at least two spaces between a semicolon and a trailing comment.  It looks like it's set up this way for C++ comments, but not for C comments.

- Sometimes we have related #define constants manually spaced so that the constant values are in a vertical column.  This is much more readable than the "standard" single space that clang applies.

davidklaw:
**re: two spaces before C comments:**
This is not supported by clang-format. Would you be willing to accept single spaces here or perhaps have these updated to C++ style when we catch them?

**re: #define constant spacing:**
Also not supported, but we can wrap these to ignore them. E.g.:
```
// clang-format off
...
// clang-format on
```

akirmse:
If you want to make those changes in these files, this would be OK.  You don't have to get literally every #define, just the main blocks, and maybe the comments can be converted automatically?

davidklaw:
Found support for aligning macros in later versions of clang-format. Apparently I was looking at documentation that was 19 versions behind the current 😐. But, aligning block comments is still out of the question, so I converted those.

Possibly controversial clang-format update:
```
SpacesInLineCommentPrefix:
  Maximum: 1
```
This will turn `//  Comment...` into `// Comment`

Meridian59:
Sorry, what's the difference there?  It removes three trailing periods?

davidklaw:
It trims the space between the comment syntax and comment itself down to a single space. "Controversial" was a bit tongue in cheek. I can't think of any reason to have more than a single space and just happened to come across a case of this in the code while formatting these.

PR #1188, opened 2025-07-31
feat: Show hints when player fails restriction checks on message quests

### LINKED ISSUE #491: Disciple Quests need better feedback and faster timing

### LINKED ISSUE #741: NPC: Text: Dialogue: Quest: Players don't know HP level to be a faction soldier; Liege gives inappropriate reply
If a player who doesn't have enough HP to become a soldier tries to join a faction as a soldier, the NPC Liege says, "I am not reviewing applicants right now.  Come back later." (This text example is from DukeLiege, Akardius - there is specific text for each Liege).

To reproduce:

- Create a character with 40 HP
- Join any faction
- Say 'soldier' to the Liege

It's expected the player be instructed that the reason they can't become a soldier due to their hit points not being high enough if their HP is under the requisite level necessary. 

There is also no reference in the game as to how many HP the player needs to be a soldier - the player should know this as well as the NPC having an appropriate response.

Note: Saying an exact number would be clunky - the NPC should say something in roleplaying fashion that they need to be "tougher" before they can become a soldier for the liege.

As I wrote this issue up I had no idea myself, and players often ask in broadcast and I've seen varying answers (70, 75, 80, 85) - the HP required is 70 hit points / base max health.

_**Suggestion: this could be added into "player_intrigue_mail"**_

davidklaw:
### Problem
Players often receive no feedback when they fail to meet the requirements for a quest, making it unclear whether they’re on the right track or missing something.

### Solution
This update adds support for NPC hints when a player can’t progress in a quest due to a quest restriction. It applies only to message quests, where the player speaks to an NPC and expects a response. These hints are designed to clarify why the player is ineligible (e.g., not enough hitpoints, missing a spell, etc.), and can be overridden on a per-node basis if needed.

[Video Walkthrough](https://youtu.be/l3WrUIsIgOo)

### Technical Implementation Details

1. Enhanced `CheckPlayerCriteria`: Modified from returning simple TRUE/FALSE to returning:
	- TRUE if player is eligible
	- FALSE if no restrictions apply
	- Restriction ID if player fails a specific requirement
2. New feedback logic: When `CheckCompletionCriteria` determines a quest node is waiting for a player, it now:
	- Handles `CheckPlayerCriteria` returning a restriction ID by looking up the corresponding hint message and sending it to the player.
3. Hint system:
	- Default hints have been added for all quest restriction types
	- Special token replacement for dynamic hints (e.g. hp requirements, spell counts, karma)
	- Override capability in quest node templates for more unique responses

Closes #741 
Closes #491

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -345,6 +345,11 @@ properties:
    piMinDamage = $
    piMaxDamage = $
 
+   % List of pending quest restriction hints queued for players. Each element is [player, hint_message].
+   % Used when multiple quest message nodes share triggers - hints are queued (not sent immediately)
+   % during eligibility checks so other nodes can be tried first. After iterating all nodes, at most
+   % one queued hint is flushed to the player if none succeeded. On success, any queued hint for
+   % that player is cleared.
    plPendingRestrictHints = $
 
 messages:


kod/include/blakston.khd
```diff
@@ -2829,17 +2829,17 @@
...
-   Q_R2_IS_PLAYER                = 2   % [Q_R2_, player]             (player = $ always returns false)
-   Q_R2_HAS_SKILL                = 3   % [Q_R2_, skill, ability]     (default ability = 1)
-   Q_R2_HAS_SPELL                = 4   % [Q_R2_, spell, ability]     (default ability = 1)
-   Q_R2_HAS_SPELL_SCHOOL         = 5   % [Q_R2_, school, numspells]  (default numspells = 1)
-   Q_R2_HAS_ITEM                 = 6   % [Q_R2_, itemclass]
-   Q_R2_USING_ITEM               = 7   % [Q_R2_, itemclass]
-   Q_R2_NUM_AMULETS              = 8   % [Q_R2_, minNumAmulets]
-   Q_R2_HAS_HEALTH_LEVEL         = 9   % [Q_R2_, minHPs]             (calls GetMaxHealth)
-   Q_R2_DOES_NOT_HAVE_ITEM       = 10  % [Q_R2_, itemclass]
+   Q_R2_QN_ASSIGNABLE            = 0xB0   % [Q_R2_, boolean]            (is this quest in an assignable state?)
```
davidklaw, line 2832: The issue leading to these changes is that restriction hints are stored as pairs `[restriction_type, hint_message]`, where `restriction_type` is the lookup key. If `Q_R2_*` constants were 0-10, they would collide with `Q_PLAYER_*` constants.

These updates prevent ID conflicts in the restriction hinting mapping.

davidklaw:
Added a video walkthrough to the PR description.

davidklaw:
During testing, I discovered an edge case with the TRADING_POST ("inventory") quest. This quest requires players to have positive or negative karma (not neutral) to trigger one of two quest nodes (good/evil karma paths), both sharing the same trigger word "inventory". Previously, if eligibility was only checked on the first matched node, a valid second node would never be reached (i.e., if the player is good but the evil check comes first, they won't get the quest). In this PR, the quest engine now evaluates all active message nodes of type `QN_TYPE_MESSAGE`. The first eligible node is assigned, and if none are eligible, the NPC sends exactly one restriction message explaining why, using the first failure's hint.

To improve the player experience using the revisions in this PR, I'll be overriding both of those restriction hints with the same message, telling players to choose a path one way or the other.

davidklaw:
Updated just now to include some documentation on usage and to include a instance of an overridden hint.

The override mentioned is for Wulfgang, whose "inventory" quests require the player to have non-neutral karma. By overriding messages for both good and bad karma restrictions, _all_ players receive a meaningful message back.

kod/include/blakston.khd
akirmse: What about existing quests that are using the old constants in lists?

kod/object/passive/quest.kod
```diff
@@ -75,12 +75,12 @@ Init( template_index = 0, report = FALSE )
 }
 
 %% CheckPlayerCriteria()
-%	Return TRUE if player eligible, otherwise FALSE
+%	Return TRUE if player eligible
+%	Return FALSE OR restriction ID if player not eligible
```
akirmse, line 79: It's dangerous to return multiple types like this.  It means that the restriction IDs have a silent, undocumented constraint that they can't be 0 or 1, or terrible things will happen.  Better to define new restriction IDs that replace TRUE/FALSE in the return values here (and obviously, change the callers to check for these) and that don't conflict with restriction ID values.

kod/util/questengine.kod
```diff
@@ -7459,5 +7562,411 @@ CheckQuestDeadlines()
...
+               return lRestrictionEntry;
+            }
+         }
+      }
+      
+      return $;
+   }
+
+   GetSchoolDescriptor(iSchool=-1)
+   "Returns the school descriptor for a school number, used when building tokenized hints."
```
akirmse, line 7937: - Document what the function does, not how you expect it to be called.
- Readers can figure out that this function likely returns a school descriptor.  But they want to know: what is a school descriptor?

kod/util/questengine.kod
```diff
@@ -7459,5 +7562,411 @@ CheckQuestDeadlines()
       return;
    }
 
+   GetRestrictionHint( quest_template_index = 0, who = $, quest_node_index = 0, restriction_type = $ )
```
akirmse, line 7565: Can you avoid adding another 500 lines to a file that is already 7500 lines long?  As in, maybe make a new class for restriction hint stuff?

akirmse (commented):
I don't really understand this change yet but here are some comments to get you started.

kod/include/blakston.khd
davidklaw: I hadn't thought about it from that angle, but a call to `QuestEngine::Reset()` is necessary given the other changes and will wipe out any existing quests.

PR #1189, opened 2025-08-05
feat: Add point-in-sector utility to check for presence in room sectors

davidklaw:
This PR adds a new `IsPointInSector` function that lets you check if a specific point (by coordinates) is inside one or more sectors in a room. The function takes an array of IDs (sometimes called server IDs), so you can check multiple sectors at once. The main reason for adding this is to help catch players who hang around too long in guild hall door sectors, but it’s flexible enough to be useful in other situations too.

### Sector Geometry Loading on Startup
When the server starts up, it loads sector geometry from the room files. Each sector’s polygons get stored as arrays of vertex coordinates.

### IsPointInSector Coordinate Lookup Process
`IsPointInSector` converts Blakod coordinates to fine coordinates, grabs the room data, and loops through the sectors. It looks for sectors with IDs matching any of the provided user IDs, and for each match, uses ray-casting to see if the point is inside the sector’s polygon.

### Memory Usage Changes
**BEFORE:**
Memory Allocated: 30096561 bytes

**AFTER:**
Memory Allocated: 34324621 bytes

**DIFFERENCE:** 4,228,060 bytes / 4.22806 Megabytes

### Example: Guild Hall 11 (`guildh11`) (removed)
As a demo, I’ve updated `ghall.kod` and `guildh11.kod` to use this new function. To try it out: open the hall’s entrance door and stand in either of the two door sectors. Let the door close, and after a short delay, you’ll be teleported to the foyer. Modify `plDoorSectors` to test other scenarios, like an invalid sector id or a single integer:

```
plDoorSectors = [ MAINDOOR, MAINDOOR2 ]; // original
plDoorSectors = MAINDOOR; // single sector check
plDoorSectors = [ MAINDOOR, MAINDOOR2, "foo" ]; // invalid sector id, should still work
```

### Usage Examples
A single usage example was originally included included but removed in favor of a follow-up PR. The following is an example of how this could be used, though there's a bit more work needed to support it that can be seen in https://github.com/Meridian59/Meridian59/pull/1189/changes/a936b59e0027042d3d2fb2f92a75287cb9fb4809

```
// Check single user ID
if IsPointInSector(prmRoom, 
    Send(who,@GetRow), 
    Send(who,@GetCol), 
    Send(who,@GetFineRow), 
    Send(who,@GetFineCol), 
    DOOR_SECTOR_ID) 
{
    // Point is in sector for this user
}
```

### DIFF
--- a/blakserv/roofile.c
+++ b/blakserv/roofile.c
@@ -63,17 +63,17 @@ static const int SF_SLOPED_CEILING = 0x00000800;
       } \
    } while (0)
 
-static bool readInt(FILE *fd, int *out)  /* Reads 4-byte little-endian signed int */
+static bool readInt(FILE *fd, int *out)  // Reads 4-byte little-endian signed int
 {
    return fread(out, 4, 1, fd) == 1;
 }
 
-static bool readShort(FILE *fd, short *out)  /* Reads 2-byte little-endian signed short */
+static bool readShort(FILE *fd, short *out)  // Reads 2-byte little-endian signed short
 {
    return fread(out, 2, 1, fd) == 1;
 }
 
-static bool readByte(FILE *fd, unsigned char *out)  /* Reads 1-byte unsigned */ 
+static bool readByte(FILE *fd, unsigned char *out)  // Reads 1-byte unsigned
 { 
    return fread(out, 1, 1, fd) == 1; 
 }


blakserv/ccode.c
```diff
@@ -18,6 +18,10 @@
 
 #include "blakserv.h"
 
+#ifndef FINENESS
+#define FINENESS 64
```
akirmse: Make this a static const int, and comment what it is.  Try to avoid the preprocessor entirely.  (It's used a lot in existing code because C didn't have a lot of the features it does today.)

blakserv/ccode.c
```diff
@@ -2344,3 +2348,174 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
 	
 	return ret_val.int_val;
 }
+
+/**
+ * PointInPoly: Returns True if point (fx, fy) lies inside the given polygon.
+ */
+static Bool PointInPoly(int fx, int fy, server_polygon *poly)
+{
+   if (!poly || poly->num_vertices < 3)
```
akirmse: Make poly a const reference (i.e const server_polygon &poly), then it can never be null.

blakserv/ccode.c
```diff
@@ -2344,3 +2348,174 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+
+/**
+ * PointInPoly: Returns True if point (fx, fy) lies inside the given polygon.
+ */
+static Bool PointInPoly(int fx, int fy, server_polygon *poly)
+{
+   if (!poly || poly->num_vertices < 3)
+      return False;
+
+   Bool inside = False;
```
akirmse: Use bool instead of Bool where possible.  (bool is the native C++ type)

blakserv/ccode.c
```diff
@@ -2344,3 +2348,174 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+   room_val = RetrieveValue(object_id, local_vars, normal_parm_array[0].type, normal_parm_array[0].value);
+   row_val = RetrieveValue(object_id, local_vars, normal_parm_array[1].type, normal_parm_array[1].value);
+   col_val = RetrieveValue(object_id, local_vars, normal_parm_array[2].type, normal_parm_array[2].value);
+   fr_val = RetrieveValue(object_id, local_vars, normal_parm_array[3].type, normal_parm_array[3].value);
+   fc_val = RetrieveValue(object_id, local_vars, normal_parm_array[4].type, normal_parm_array[4].value);
+   id_list_val = RetrieveValue(object_id, local_vars, normal_parm_array[5].type, normal_parm_array[5].value);
+
+   if (room_val.v.tag != TAG_ROOM_DATA)
+   {
+      bprintf("PIS early exit: bad room tag\n");
```
akirmse: Don't abbreviate to PIS, that's going to be pretty meaningless in an error log.

blakserv/ccode.c
```diff
@@ -2344,3 +2348,174 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+   else if (id_list_val.v.tag != TAG_LIST)
+   {
+      bprintf("PIS early exit: 6th parameter must be a list of IDs or a single ID\n");
+      return NIL;
+   }
+
+   roomdata_node *rd = GetRoomDataByID(room_val.v.data);
+   if (!rd || !rd->file_info.sectors)
+   {
+      return NIL;
```
akirmse, line 2478: Print error here if rd is nullptr, i.e. bad room ID passed in.

blakserv/ccode.c
```diff
@@ -2344,3 +2348,174 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+   int row0 = (int) row_val.v.data - 1;
+   int col0 = (int) col_val.v.data - 1;
+
+   // Now, we need to convert the kod coordinates to coordinates that match the room data
+   int fine_x = (int) ((long long) (col0 * FINENESS) + fc_val.v.data) * 16;
+   int fine_y = (int) ((long long) (row0 * FINENESS) + fr_val.v.data) * 16;
+
+   val_type current_list = id_list_val;
+
+   while (current_list.v.tag == TAG_LIST)
```
akirmse: This handling of single value vs. list is a little convoluted.  To simplify, how about making a C++ vector of the sector IDs, either by traversing the Blakod list, or making a single-element vector with the one ID.  Then a second loop that simply iterates over the vector.

Also there should be an error message if a list element isn't an integer.

blakserv/roofile.c
```diff
@@ -17,122 +17,300 @@
...
+
 static unsigned char room_magic[] = { 0x52, 0x4F, 0x4F, 0xB1 };
 
+#define SF_SLOPED_FLOOR   0x00000400
+#define SF_SLOPED_CEILING 0x00000800
+
+static int readFloatAsInt(int fd)
+{
+   float f;
+   if (read(fd, &f, 4) != 4)
```
akirmse: read() is unbuffered I/O, while fread() is buffered and is much faster.  I got a huge speedup in the Blakod compiler by switching to buffered I/O a few years ago.  So, not sure you want to do this at the same time, but we should switch this to buffered reads, as it will likely make a significant difference in server startup time.

include/roomtype.h
```diff
@@ -12,6 +12,20 @@
 #ifndef _ROOMTYPE_H
 #define _ROOMTYPE_H
 
+typedef struct {
```
akirmse: This is the C way of declaring a struct.  In C++ you can do simply

struct server_polygon {
...
}

blakserv/roofile.c
davidklaw: Turns out `readFloatAsInt` was a leftover function and I've removed it, but did make sweeping I/O changes throughout to move to buffered. Seeing a very noticeable difference in startup time!

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
 	return ret_val.int_val;
 }
+
+/**
+ * PointInPoly: Returns True if point (fx, fy) lies inside the given polygon.
+ */
+static bool PointInPoly(int fx, int fy, const server_polygon &poly)
+{
+   if (poly.num_vertices < 3)
+      return False;
```
akirmse: false

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+
+   bool inside = False;
```
akirmse: false

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+   for (int i = 0, j = poly.num_vertices - 1; i < poly.num_vertices; j = i++)
+   {
+      int xi = poly.vertices_x[i], yi = poly.vertices_y[i];
+      int xj = poly.vertices_x[j], yj = poly.vertices_y[j];
+
+      int dy = yj - yi;
+      if (dy == 0)
```
akirmse: dy unused... should be if (yj == yi) ?

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+/**
+ * C_PointInSector: Tests whether a fine-grained room coordinate falls inside any polygon of sectors
+ * matching given ID(s).
+ * Expects 6 params: room data, row, col, fine-row (fr), fine-col (fc), and an ID or list of IDs.
+ */
+blak_int C_PointInSector(int object_id, local_var_type *local_vars, int num_normal_parms, parm_node normal_parm_array[],
+                         int num_name_parms, parm_node name_parm_array[])
+{
+   val_type room_val, row_val, col_val, fr_val, fc_val, id_list_val;
+   std::vector<int> sector_ids;
```
akirmse: Declaring a vector can allocate some memory, so it's better to declare it right before it's used.

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+            bprintf("C_PointInSector: Sector list element is not an integer\n");
+         }
+
+         if (list->rest.v.tag == TAG_LIST)
+         {
+            list = GetListNodeByID(list->rest.v.data);
+         }
+         else
+         {
+            break;
```
akirmse, line 2537: This is an error, no?

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+         }
+      }
+   }
+   else
+   {
+      bprintf("C_PointInSector has bad 6th parameter tag, expected list or int\n");
```
akirmse: A hard-coded number like "6th" is very likely to become outdated, making debugging impossible.  Better to tell the reader which parameter you mean.

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+      return NIL;
+   }
+
+   for (int sector_id : sector_ids)
+   {
+      for (int i = 0; i < r->num_sectors; i++)
+      {
+         if (r->sectors[i].id != sector_id)
```
akirmse: This embedded loop structure is odd.  Instead make a little loop that just finds the sector you're looking for and prints or returns an error if it's not found.  (Also, are the sectors in a room sorted by ID?  If so the lookup can be faster.)

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+         {
+            continue;
+         }
+
+         // Found matching sector, check polygons
+         for (int p = 0; p < r->sectors[i].num_polygons; p++)
+         {
+            // First, check if the sector has polygons
+            if (r->sectors[i].polygons == NULL || p >= r->sectors[i].num_polygons)
```
akirmse: This check seems unnecessary.

blakserv/ccode.c
```diff
@@ -2344,3 +2350,199 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+            {
+               continue;
+            }
+
+            server_polygon *poly = &r->sectors[i].polygons[p];
```
akirmse: This is a good place to use a reference:

const server_polygon &poly = r->sectors[i].polygons[p];

blakserv/ccode.h
```diff
@@ -245,7 +245,10 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
 		  int num_normal_parms,parm_node normal_parm_array[],
 		  int num_name_parms,parm_node name_parm_array[]);
 
-
+blak_int C_PointInSector(int object_id,local_var_type *local_vars,
```
akirmse: I wonder if maybe "IsPointInSector" is a better name?  "PointInSector" suggests that perhaps it finds and returns a point in a sector.

blakserv/ccode.c
davidklaw: I'm fairly certain that sectors are not sorted by _server ID_, the relevant ID here. Maybe I should save sectors with server ids--the only ones this lookup cares about--separately from others?

akirmse: Ah yes, that makes sense.  So a linear search is fine, but the loop structure can still be simplified.

blakserv/ccode.c
davidklaw: My intent was to handle the end of the list by simply breaking and moving on, using the sector ids already collected. But yes, it _could_ be an error, or it could simply be the end of the list.

Gar103:
Very keen on seeing this progress. Happy to help as well. Adding the ability to check for sectors and positioning unlocked a massive array of stuff we could do on OS, and here it could do things like server-side eliminate wallhacking & automatically rescue characters from out of bounds,

davidklaw:
> it could do things like server-side eliminate wallhacking & automatically rescue characters from out of bounds,

You’re raising use cases I didn’t account for, which is good. This PR is scoped to sector server IDs, the optional ID typically used for moving sectors, and whether a point exists within those sectors. I probably need to clean up some naming, but I'm wondering... should this PR also cover checks for a point within a sector by its unique ID, and a check for whether a point falls outside all sectors (and thus outside the room boundaries)?

davidklaw:
The code conflicts have been addressed.

I was thinking, I modified one guild hall. If we want to move forward with this, I can either a) update _all_ guild halls to or b) remove the hall change here and drop a subsequent PR that addresses all the halls at once.

blakserv/ccode.c
```diff
@@ -2366,6 +2372,204 @@ blak_int C_BuildString(int object_id, local_var_type *local_vars, int num_normal
...
+      return NIL;
+   }
+
+   if (col_val.v.tag != TAG_INT)
+   {
+      bprintf("C_IsPointInSector has bad col tag\n");
+      return NIL;
+   }
+
+   if (fr_val.v.tag != TAG_INT)
```
akirmse, line 2444: Not checking fc_val?  Also, improve the error message ("fr" is cryptic).

blakserv/ccode.c
```diff
@@ -2366,6 +2372,204 @@ blak_int C_BuildString(int object_id, local_var_type *local_vars, int num_normal
...
+            bprintf("C_IsPointInSector: Sector id list element is not an integer\n");
+         }
+
+         if (list->rest.v.tag == TAG_LIST)
+         {
+            list = GetListNodeByID(list->rest.v.data);
+         }
+         else
+         {
+						break;
```
akirmse: Fix indentation.

include/roomtype.h
```diff
@@ -12,6 +12,20 @@
 #ifndef _ROOMTYPE_H
 #define _ROOMTYPE_H
 
+struct server_polygon {
+   int num_vertices;
+   int *vertices_x;
+   int *vertices_y;
+};
+
+typedef struct
```
akirmse: struct server_sector, not typedef struct.

Also add a comment saying what this structure is for.

include/roomtype.h
```diff
@@ -12,6 +12,20 @@
...
+   int *vertices_x;
+   int *vertices_y;
+};
+
+typedef struct
+{
+   int id; // server id
+   int num_polygons;
+   int polygon_capacity;
+   server_polygon *polygons;
```
akirmse: Maybe should be a std::vector instead?  Then you could get rid of polygon_capacity.

include/roomtype.h
```diff
@@ -29,6 +43,8 @@ typedef struct
 
    int security;          /* Security number, unique to each roo file, to ensure that client
                              loads the correct roo file */
+   server_sector *sectors;
```
akirmse: Should maybe be a vector?

kod/object/active/holder/room/ghall.kod
```diff
@@ -90,6 +90,8 @@ resources:
    guildhall_new_occupants = "~B%s has been rented to new owners."
    guildhall_someone_rented = "Someone has already rented this guild hall."
 
+   guildhall_teleported = "You blink and find yourself in the foyer."
```
akirmse: I think I agree with your comment, let's split the Blakod changes into a separate change so that the server side changes can go in more quickly.

davidklaw:
Feedback has been addressed. Among the changes:
* Removed the guild hall implementation for a later PR.
* Introduced vectors in room_type.
* Added validation for `fseek()`, `fread()`, and `readShort()` in `LoadSectorPolygons` to prevent crashes.
* Removed unreachable code in `C_IsPointInSector` where `TAG_INT` was checked after it had already been converted to `TAG_LIST`.

blakserv/roofile.c
```diff
@@ -17,122 +17,303 @@
 
 #define ROO_VERSION 4
 
+
 static unsigned char room_magic[] = { 0x52, 0x4F, 0x4F, 0xB1 };
 
+#define SF_SLOPED_FLOOR   0x00000400
```
akirmse: static const int

blakserv/roofile.c
```diff
@@ -17,122 +17,303 @@
...
+      fclose(infile);
+      return false;
+   }
 
-	 close(infile);
-	 return false;
+   room->cols = (short) temp;
+
+   room->grid = (unsigned char **) AllocateMemory(MALLOC_ID_ROOM, room->rows * sizeof(char *));
+   room->flags = (unsigned char **) AllocateMemory(MALLOC_ID_ROOM, room->rows * sizeof(char *));
```
akirmse: Every error return after this leaks all of this memory, no?

blakserv/roofile.c
```diff
@@ -17,122 +17,303 @@
...
+   int room_w = readInt(infile);
+   int room_h = readInt(infile);
 
-   // Read size of room
-   if (read(infile, &temp, 4) != 4)
-   { close(infile); return false; }
-   room->rows = static_cast<short>(temp);
+   // Read in subsection offsets
+   int node_off = readInt(infile);
+   readInt(infile); /* cwall off   */
```
akirmse: The old code checked for errors on every read--the new code needs to do that too.

blakserv/roomdata.c
```diff
@@ -81,6 +82,7 @@ blak_int LoadRoomData(int resource_id)
    }
 
    room = (roomdata_node *)AllocateMemory(MALLOC_ID_ROOM,sizeof(roomdata_node));
+   new(&room->file_info) room_type();
```
akirmse: This use of placement new, and the direct call of the destructor a little bit above, is very rarely seen outside of library code and I think it's to be avoided for its strangeness.  Two options I can think of:

- Allocate "room" with new roomdata_node, not AllocateMemory.  This will throw off the server's accounting of room memory by a little bit, since AllocateMemory keeps some nice stats.  Losing tracking counts for just the sector data maybe isn't enough to worry about?  With extra work you could add methods to room_type to adjust the memory system's counts manually.  New methods would need to be added to the memory system to allow manual adjustment of memory counts, which isn't great, but will be needed if we want to use the STL more aggressively.  This is what has kept me from checking in https://github.com/Meridian59/Meridian59/pull/1254, so it will probably be needed eventually anyway.

- Don't make sectors into a std::vector after all.. it's not worth this weirdness.

davidklaw:
Feedback addressed (and then some), and per your suggestion I added manual memory management methods to `memory`.

One additional change of note: Some of the `FreeMemory` calls in `BSPRooFileLoadServer` are using the wrong variable. They're using `i` instead of `j`. This PR corrects that, but I'll flip another PR for that alone in the morning if it's worthwhile.

https://github.com/Meridian59/Meridian59/blob/c555fbdd6cda1ab1be38ff122ba5f8999a62f8d7/blakserv/roofile.c#L89-L91

blakserv/roomdata.c
```diff
@@ -44,6 +44,23 @@ void InitRoomData()
...
 
+static size_t CalculateRoomTypeMemory(const room_type *room)
+{
+   size_t total = sizeof(room_type);
+   
+   // Track vector's internal buffer capacity
+   total += room->sectors.capacity() * sizeof(server_sector);
+   
+   // Track each sector's polygon vector buffer
+   for (size_t i = 0; i < room->sectors.size(); i++)
```
akirmse: You can just do something like 
```
for (auto &element : room->sectors) {
}
```
since you don't need the loop variable

blakserv/roomdata.c
```diff
@@ -53,7 +70,14 @@ void ResetRoomData()
    {
       temp = room->next;
       BSPRoomFreeServer(&(room->file_info));
-      FreeMemory(MALLOC_ID_ROOM,room,sizeof(roomdata_node));
+
+      // Manually subtract memory for room_type structure
+      SubtractMemoryCount(MALLOC_ID_ROOM, CalculateRoomTypeMemory(&(room->file_info)));
+
+      // Manually subtract memory for roomdata_node structure
+      SubtractMemoryCount(MALLOC_ID_ROOM, sizeof(roomdata_node));
```
akirmse: This is pretty fragile--if someone ever adds a field to roomdata_node, all of the places where you do this will silently be wrong.  Better to add a (say) GetSize() method to roomdata_node that returns its size.  The body of CalculateRoomTypeMemory would also move to this function.

blakserv/memory.c
```diff
@@ -328,3 +328,18 @@ void * ResizeMemory(int malloc_id,void *ptr,int old_size,int new_size)
...
+{
+   if (malloc_id < 0 || malloc_id >= MALLOC_ID_NUM)
+      eprintf("AddMemoryCount adding memory of unknown type %i\n", malloc_id);
+   else
+      memory_stat.allocated[malloc_id] += size;
+}
+
+void SubtractMemoryCount(int malloc_id, size_t size)
+{
+   if (malloc_id < 0 || malloc_id >= MALLOC_ID_NUM)
```
akirmse: Just do
```AddMemoryCount(malloc_id, -size)```

but note that you need to make size signed (or delegate both functions to a single function that takes a signed value if you want the interfaces to remain unsigned).

File: blakserv/roofile.c [hunk 32]
```diff
@@ -17,122 +17,535 @@
...
+      return false;
+   }
+   
+   //
+   // Read in and ignore sidedef offset
+   //
+   if (readInt(infile) == -1)
+   {
+      bprintf("BSPRooFileLoadServer: Failed to read sidedef_off in %s\n", fname);
+      fclose(infile);
```
akirmse: I was hoping to save you all of these zillions of extra fclose calls and kind of excessive prints.  The file can be made to auto-close with various tricks (see RAII, or it can be done with a macro), but this change is already big enough as it is.

Also, bprintf is for Blakod-related errors.  These should all be eprintfs.

blakserv/memory.c
```diff
@@ -328,3 +328,18 @@ void * ResizeMemory(int malloc_id,void *ptr,int old_size,int new_size)
 #endif
 }
 
+void AddMemoryCount(int malloc_id, size_t size)
```
akirmse: Could use a comment.

blakserv/roofile.c
davidklaw: I meant to mention that this was still on my list to address. It feels like something we should do and I'm happy to extract and submit the memory changes as a separate PR to reduce this some.

davidklaw: I've tried to deal with the file closes (RAII) and excessive prints (macros). I'm not exactly in love with the solution, but it does greatly reduce the code. I can trim it down even further by updating many of my comments, but wanted to see if I was taking this in the right direction or not.

blakserv/fileutil.h
```diff
@@ -9,6 +9,38 @@
...
 
+/**
+ * RAII wrapper for FILE* that automatically closes the file when it goes out of scope.
+ * 
+ * This class ensures files are properly closed even if an error occurs or an early return
+ * happens, eliminating the need for manual fclose() calls..
+ */
+class FileGuard
+{
+   FILE *file;
```
akirmse: It's a common convention to put the public section first, then to put the data last in a private: section.  The reason is that the public API is more important than the data representation (which external users shouldn't care about).

blakserv/fileutil.h
```diff
@@ -9,6 +9,38 @@
...
+
+public:
+   FileGuard(FILE *f) : file(f) {}
```
akirmse: Single argument constructors should be marked "explicit" to prevent automatic type coercion (weird C++ feature).

blakserv/roomdata.h
```diff
@@ -19,6 +19,21 @@ typedef struct roomdata_struct
    blak_int roomdata_id;
    room_type file_info;
    struct roomdata_struct *next;
+   size_t GetSize(void) const
+   {
+      size_t total = sizeof(*this);
```
akirmse: Interesting fact: sizeof is an operator, not a function, so parentheses aren't needed (but are OK if you want them).

blakserv/roomdata.h
```diff
@@ -19,6 +19,21 @@ typedef struct roomdata_struct
    blak_int roomdata_id;
    room_type file_info;
    struct roomdata_struct *next;
+   size_t GetSize(void) const
```
akirmse: I might define the return type as int, which is more convenient for callers to use.  Then the annoying casts are limited to the internals of this function.  Up to you.

blakserv/roofile.c
```diff
@@ -19,120 +19,364 @@
...
+{
+   short v;
+   if (fread(&v, sizeof(short), 1, fd) != 1)
+      return -1;
+   return v;
+}
+
+static int readByte(FILE *fd)             /* 1-byte unsigned, returns int for error checking */
+{
+   unsigned char v = 0;
```
akirmse: If you make this "signed char", can you then avoid the later cast to int?

blakserv/roomdata.c
```diff
@@ -44,6 +44,22 @@ void InitRoomData()
    num_roomdata = 0;
 }
 
+static size_t CalculateRoomTypeMemory(const room_type *room)
```
akirmse: Is this still used?

blakserv/roofile.c
davidklaw: I don't believe so. If we read a `0xFF` it would become -1 and be indistinguishable from our (-1) error code.

blakserv/fileutil.h
davidklaw: Got it. Given how `roomdata_node` is used today, nothing can be made `private`. Separate PR to to refactor?

akirmse: I should have been clearer--the default for classes is private, so the "file" data member is currently private.  You should add a private: section at the bottom and move "file" there.

blakserv/roomdata.h
```diff
@@ -13,12 +13,29 @@
 #ifndef _ROOMDATA_H
 #define _ROOMDATA_H
 
-
 typedef struct roomdata_struct
 {
-   blak_int roomdata_id;
-   room_type file_info;
+ public:
```
akirmse: The default for structs is public, so you don't need this, and you should remove the empty private: section.

blakserv/roofile.c
akirmse: Oh yikes---these read functions are all busted.  e.g. if readInt() ever reads a legitimate -1 value, it will be interpreted as an error.  This is why fread takes a pointer to a buffer rather than returning the read value.  So these functions will need to be redesigned to operate in the same way.

Do they work now?  It might be pure luck that there are no "error" values being read?

davidklaw:
The latest commits redesign `readInt`, `readShort`, `readByte`, and `readCoord` helper functions to return `bool` and use output pointers instead of sentinel values, avoiding collisions with valid data. They also refactor the file macros to work with these changes. Other refactor work was to minimize comment real estate.

blakserv/roofile.c
davidklaw: They did work, so pure luck 🎰  Thanks for catching this.

blakserv/roofile.c
```diff
@@ -19,119 +19,365 @@
...
+}
+
+static bool readShort(FILE *fd, short *out)          /* 2-byte little-endian signed short */
+{
+   return fread(out, sizeof(short), 1, fd) == 1;
+}
+
+static bool readByte(FILE *fd, unsigned char *out)             /* 1-byte unsigned, returns int for error checking */
+{
+   // Use unsigned char so byte value 0xFF doesn't collide with -1 error code
```
akirmse: This comment isn't right.

blakserv/roofile.c
```diff
@@ -19,119 +19,365 @@
...
+   char buf[4];
+   if (fread(buf, 4, 1, fd) != 1)
+      return false;
+
+   if (as_float)
+      *out = (int)(*((float *)buf));
+   else
+      *out = *((int *)buf);
+
+   // version ≤12 – bytes contain a plain little-endian int
```
akirmse: This comment is in the wrong place.

blakserv/roofile.c
```diff
@@ -19,119 +19,365 @@
...
+{
+   return fread(out, sizeof(int), 1, fd) == 1;
+}
+
+static bool readShort(FILE *fd, short *out)          /* 2-byte little-endian signed short */
+{
+   return fread(out, sizeof(short), 1, fd) == 1;
+}
+
+static bool readByte(FILE *fd, unsigned char *out)             /* 1-byte unsigned, returns int for error checking */
```
akirmse: Comment isn't right.

blakserv/roofile.c
```diff
@@ -19,119 +19,365 @@
...
+      if (!readShort(fd, out)) { \
+         eprintf("Failed to read short %s in %s\n", desc, fname); \
+         return false; \
+      } \
+   } while (0)
+
+#define CHECK_READ_BYTE(fd, out, fname, desc) \
+   do { \
+      if (!readByte(fd, out)) { \
+         eprintf("Failed to read byte%s in %s\n", desc, fname); \
```
akirmse: Missing space

davidklaw:
re: memory usage before and after: Here are the numbers after server startup:

**BEFORE:**
Memory Allocated: 30096561 bytes

**AFTER:**
Memory Allocated: 34324621 bytes

**DIFFERENCE:** 4,228,060 bytes / 4.22806 Megabytes

davidklaw:
Feedback addressed. I want to call out that I modified `readInt` and `readShort` to use fixed sizes rather than `sizeof(int)` and `sizeof(short)`, in case that could be problematic across platforms.

blakserv/roofile.c
```diff
@@ -63,20 +63,19 @@ static const int SF_SLOPED_CEILING = 0x00000800;
       } \
    } while (0)
 
-static bool readInt(FILE *fd, int *out)              /* 4-byte little-endian signed int   */
+static bool readInt(FILE *fd, int *out) /* Reads 4-byte little-endian signed int */
```
akirmse: So close to done!  But then you removed this extra whitespace, and now the comment is too close to the code :-/  Also a few other times below.  Also might as well make them C++ style comments.

davidklaw: Hah. Damnit! I had two spaces, then used "format selection" to tidy up. Thought my `clang-format` settings had this rule.

Face palm

Issue #701, opened 2024-05-04
Suggestion: Add Soldier News for Territory Game & Faction War


jrotunda:
Guild Warfare and Soldier Shields (TerritoryGame) were added as additional PvP mini-game scenarios for players to have fun competing over. 

These systems come complete with their own unique death messages that always draw attention when they appear, letting all online know there's an end-game PvP scenario, something a little more interesting than the traditional PK/Hunter conflict.

A few years ago [Phil "Oz"] added a tome to Frular (gmhall.kod) that tracks Guild War - Peace Treaties, War Declarations -to the results of the war, ending in peace or disbandment, and all kills of the members for any active warring Guilds. Players love seeing their name in print, and having a public record of these kills adds to the drama and fun.

For the Territory Game, a new News class that does something similar, tracking kills, victories, and activity - would likely also prove popular. A recent in-game player poll returned favorable comments in this regard.

- Show every Player Character (PC) on Player Character Faction Soldier Kill
- Show major land claims (not minor, spammy) including which PC and allies caused a major shift (Tos is now in the name of Jonas thanks to King Blah and Queen Bleh!)
- Show Land Claim Faction Soldier Kill (BlackPanther has been killed by GoofusDeluxorz during a land takeover)
- Keep Track of The Most Kills, Most Land Taken, Most Retreats and post a monthly "Leaderboard" and summary
- Give no other incentives, have this be purely cosmetic and bragging rights only for this group of players to connect to their gameplay

Placement Ideas (_less important than creating the class itself_)

Somewhere political but neutral, using 1 of these:

- Cor Noth Inn (political owner but not a home of Princess, Duke or Jonas), in sight of many players - but mainly newbies and can plant seeds of interest in PvP and that aspect to them - another way to show the Meridian gameplay being simple but having many layers and things to possibly do; also has nothing really going on in that room as you can't rent a private room, owner sells wedding rings
- Cor Noth town square, near the Clocktower or Cibilo Creek
- Brahma & Falcon (Barloque Bar)

One in all 3 home locations:
- Duke's Chambers
- Princess Chambers
- Pietro's 
OR
- Jasper, Tos or Barloque Adv. Halls

PR #1190, opened 2025-08-10
feat: Add territory and faction war newsglobes

davidklaw:
This PR adds territory and faction war ledgers that record and displays war updates. Currently, these updates are only visible to players who are online. These new ledgers are located in the Duke and Princess' throne rooms, and the Jasper Tavern. The ledgers record player soldier vs. player soldier kills and major land claims (i.e. claiming a town). These posts are authored by Baron Gerah Springer, who I opted to create a class for in the hopes of future expansion -- happy to dial back if that's too much.

### Notes
* This PR removes resources in `jastavnr.kod`. The resources were for a Bard's newsglobe that isn't present on 101. There is at least one other instance of it on 101 though, in the Marion adventurers hall.

### Screenshots
image
image
image

Related: #701

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/crnthtwn/cnambassador.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/crnthtwn/cnambassador.kod
@@ -1,0 +1,26 @@
+% Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
+% the LICENSE file that accompanies it.
+%
+% Meridian is a registered trademark.
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+CorNothAmbassador is CorNothTown
+
+constants:
+
+   include blakston.khd
+
+resources:
+
+   cornothambassador_name_rsc = "Baron Gerah Springer"
+
+classvars:
+
+   vrName = cornothambassador_name_rsc
+
+end
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/kod/object/active/holder/room/duke3.kod
+++ b/kod/object/active/holder/room/duke3.kod
@@ -88,6 +88,7 @@ messages:
 
    CreateStandardObjects()
    {
+      local oNews;
 
       Send(self,@NewHold,#what=Create(&Brazier),#new_row=9,#new_col=18,
            #fine_row=48,#fine_col=48);
@@ -100,6 +101,15 @@ messages:
 
       Send(self,@NewHold,#what=Create(&Throne),
            #new_row=10,#new_col=20,#fine_row=32,#fine_col=32,#new_angle=ANGLE_SOUTH);
+
+      oNews = Send(SYS,@FindNewsByNum,#num=NID_TERRITORY_NEWS);
+      if oNews = $
+      {
+         oNews = Create(&TerritoryNews,#nid=NID_TERRITORY_NEWS);
+      }
+      Send(self,@NewHold,#what=oNews,#new_row=15,#new_col=16,
+         #fine_row=24,#fine_col=24,#new_angle=ANGLE_SOUTH_EAST);
+
       propagate;
    }
 
--- a/kod/object/passive/news/territorynews.kod
+++ b/kod/object/passive/news/territorynews.kod
@@ -1,0 +1,38 @@
+% Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
+% the LICENSE file that accompanies it.
+%
+% Meridian is a registered trademark.
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+TerritoryNews is News
+
+constants:
+
+   include blakston.khd
+
+resources:
+
+   territorynews_icon_rsc = jbook.bgf
+
+classvars:
+
+   vrIcon = territorynews_icon_rsc
+
+messages:
+
+   GetNewsPermission(what = $)
+   {
+      if (IsClass(what,&CorNothAmbassador) OR IsClass(what,&DM))
+      {
+         return NEWS_PERMISSION_READ_WRITE;
+      }
+
+      return NEWS_PERMISSION_READ;
+   }
+
+end
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/kod/util/factgame/territry.kod
+++ b/kod/util/factgame/territry.kod
@@ -26,6 +26,7 @@ resources:
    TerritoryGame_rebel = "the rebel Jonas"
    TerritoryGame_unfactioned = "the unfactioned"
 
+   TerritoryGame_bonus_territory_title = "~I%s now %s the city of %s!"
    TerritoryGame_bonus_territory = "~IHaving claimed %q, %s now also %s the city of %s!"
    TerritoryGame_control = "control"
    TerritoryGame_controls = "controls"
@@ -39,6 +40,16 @@ resources:
    TerritoryGame_her = "her"
    TerritoryGame_his = "his"
    TerritoryGame_their = "their"
+
+   territory_string_title_1 = " now "
+   territory_string_title_2 = " "
+   territory_string_title_3 = "!"
+
+   territory_string_body_1 = "Having claimed "
+   territory_string_body_2 = ", "
+   territory_string_body_3 = " now also "
+   territory_string_body_4 = " the city of "
+   territory_string_body_5 = "!"
 
 properties:
 
@@ -585,7 +596,7 @@ messages:
 
    SendBonusTownMessage(faction=$,town=$)
    {
-      local i, sTowns, rFaction, rControls, rCity;
+      local i, sTowns, rFaction, rControls, rCity, oBook, territory_news_title, territory_news_body;
 
       sTowns = createString();
       clearTempString();
@@ -645,6 +656,33 @@ messages:
                #parm1=sTowns,#type1=STRING_RESOURCE,
                #parm2=rFaction,#parm3=rControls,#parm4=rCity);
       }
+
+      ClearTempString();
+      AppendTempString(rFaction); 
+      AppendTempString(territory_string_title_1); 
+      AppendTempString(rControls);
+      AppendTempString(territory_string_title_2); 
+      AppendTempString(rCity); 
+      AppendTempString(territory_string_title_3);
+      territory_news_title = CreateString(); 
+      SetString(territory_news_title, GetTempString());
+
+      ClearTempString();
+      AppendTempString(territory_string_body_1); 
+      AppendTempString(sTowns); 
+      AppendTempString(territory_string_body_2);
+      AppendTempString(rFaction); 
+      AppendTempString(territory_string_body_3); 
+      AppendTempString(rControls);
+      AppendTempString(territory_string_body_4); 
+      AppendTempString(rCity); 
+      AppendTempString(territory_string_body_5);
+      territory_news_body = CreateString(); 
+      SetString(territory_news_body, GetTempString());
+
+      % Post to the territory news
+      oBook = Send(SYS,@FindNewsByNum,#num=NID_TERRITORY_NEWS);
+      Send(oBook,@PostNews,#what=Create(&CorNothAmbassador),#title=territory_news_title,#body=territory_news_body);
 
       return;
    }
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -65,6 +65,16 @@ resources:
 
    system_admin_message = "%q"
 
+   faction_war_kill_title_1 = " bested by "
+   faction_war_kill_title_2 = "."
+   faction_war_kill_body_1 = " has been bested by faction rival "
+   faction_war_kill_body_2 = " during a land claim attempt"
+   faction_war_kill_body_3 = "."
+   faction_war_kill_body_4 = "\n--------------------------------"
+   faction_war_kill_body_5 = "\nLocation: "
+   faction_war_kill_body_6 = "\nFactions: "
+   faction_war_kill_body_7 = " vs "
+
    system_char_module = char.dll
 
    % charinfo stuff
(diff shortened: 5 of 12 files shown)


kod/object/active/holder/room/duke3.kod
```diff
@@ -100,6 +101,15 @@ messages:
 
       Send(self,@NewHold,#what=Create(&Throne),
            #new_row=10,#new_col=20,#fine_row=32,#fine_col=32,#new_angle=ANGLE_SOUTH);
+
+      oNews = Send(SYS,@FindNewsByNum,#num=NID_TERRITORY_NEWS);
+      if oNews = $
+      {
+         oNews = Create(&TerritoryNews,#nid=NID_TERRITORY_NEWS);
+      }
+      Send(self,@NewHold,#what=oNews,#new_row=15,#new_col=16,
```
akirmse: This should probably be inside the if?  Otherwise if the newsgroup already exists, NewHold will be called, potentially causing problems.

kod/object/passive/news/territorynews.kod
```diff
@@ -0,0 +1,38 @@
...
+% All rights reserved.
+%
+% This software is distributed under a license that is described in
+% the LICENSE file that accompanies it.
+%
+% Meridian is a registered trademark.
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+TerritoryNews is News
```
akirmse, line 15: Needs a comment describing what this class is.

kod/util/factgame/territry.kod
```diff
@@ -40,6 +41,16 @@ resources:
...
 
+   territory_string_title_1 = " now "
+   territory_string_title_2 = " "
+   territory_string_title_3 = "!"
+
+   territory_string_body_1 = "Having claimed "
+   territory_string_body_2 = ", "
+   territory_string_body_3 = " now also "
+   territory_string_body_4 = " the city of "
+   territory_string_body_5 = "!"
```
akirmse: Not sure if there's much to be done about this now, but this kind of text assembly is a translation nightmare and would need to be completely redone.

kod/util/factgame/territry.kod
```diff
@@ -646,6 +657,33 @@ messages:
...
+      AppendTempString(rControls);
+      AppendTempString(territory_string_body_4); 
+      AppendTempString(rCity); 
+      AppendTempString(territory_string_body_5);
+      territory_news_body = CreateString(); 
+      SetString(territory_news_body, GetTempString());
+
+      % Post to the territory news
+      oBook = Send(SYS,@FindNewsByNum,#num=NID_TERRITORY_NEWS);
+      Send(oBook,@PostNews,#what=Create(&CorNothAmbassador),#title=territory_news_title,#body=territory_news_body);
```
akirmse: Weird indentation on this line?

kod/object/active/holder/nomoveon/battler/monster/towns/crnthtwn/cnambassador.kod
```diff
@@ -0,0 +1,26 @@
...
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+CorNothAmbassador is CorNothTown
+
+constants:
+
+   include blakston.khd
+
+resources:
+
+   cornothambassador_name_rsc = "Baron Gerah Springer"
```
akirmse, line 19: Just a name, no artwork or description?

davidklaw: Right, that's intentional. I thought this might be considered overkill, but posting news requires a `what` with a name and this felt like the best way to handle it _and_ support a future full-fledged NPC. If this is acceptable, I can detail this in the class description.

kod/util/factgame/territry.kod
davidklaw: How do you feel about a new `Util` function. Something like this:

```
BuildString(input_string = $, args = $)
{
  local sString, iIndex, sToken, sVal;

  if input_string = $ 
  { 
    return $; 
  }
  sString = CreateString(); 
  SetString(sString, input_string);

  sToken = CreateString();        
  iIndex = 1;
  
  for sVal in args
  {
    ClearTempString();
    AppendTempString(string_token); // "%"
    AppendTempString(iIndex);
    SetString(sToken, GetTempString());

    while TRUE
    {
      if StringSubstitute(sString, sToken, sVal) <> 1 
      { 
        break;
      }
    }
    iIndex = iIndex + 1;
  }
  return sString;
}
```

Usage:

```
territory_news_title = Send(self,@BuildString,
                     #input_string=faction_war_kill_title,
                     #args=[
                        Send(what,@GetTrueName),
                        Send(killer,@GetTrueName)
                     ]);
```

kod/util/factgame/territry.kod
davidklaw: I think this is GitHub not wanting to break the line. Leading spaces are correct but I'll break it up into two lines 👍🏻

kod/object/active/holder/nomoveon/battler/monster/towns/crnthtwn/cnambassador.kod
akirmse: Ah OK, that makes some sense.

kod/util/factgame/territry.kod
akirmse: Hm, yeah, that's pretty interesting, though expensive, with a lot of hidden allocations.  I wonder if it might be better to do this in C.  But that's a bigger project; there are likely many, many places in Blakod that would need to be revisited.

davidklaw: I've got a C function working here. Hit me with feedback and, if this is an acceptable direction, I can submit it along with this PR (with its use limited to the scope of this PR) or submit it separately with this PR and others to follow.

After validating the arguments, this makes a first pass to identify replacement tokens and the length of the replacement strings to allocate just enough memory before making a second pass to write it.

```c
blak_int C_BuildString(int object_id, local_var_type *local_vars, int num_normal_parms, parm_node normal_parm_array[],
                       int num_name_parms, parm_node name_parm_array[])
{
   const char *string1 = NULL;
   
   // Get the string to build
   val_type string_val = RetrieveValue(object_id, local_vars, normal_parm_array[0].type, normal_parm_array[0].value);
	 int len1 = 0;
   int use_args = num_normal_parms - 1;

   if (!LookupString(string_val, "C_BuildString", &string1, &len1))
      return NIL;

   // Validate the string parameters
   const char *params[MAX_C_PARMS];
   int params_len[MAX_C_PARMS];
	 
	 for (int i = 0; i < use_args; i++)
   {
      val_type val = RetrieveValue(object_id, local_vars, normal_parm_array[i + 1].type, normal_parm_array[i + 1].value);
      if (!LookupString(val, "C_BuildString", &params[i], &params_len[i]))
         return NIL;
   }

   // Pass 1: Walk the string to compute the exact length needed by simulating
   // substitutions (%s adds the next params length, %% adds 1, unknown %X adds 2).
   // This lets us allocate the exact memory needed for the output string.
   int output_len = 0, arg_i = 0;
   for (int i = 0; i < len1;)
   {
      char ch = string1[i];
      if (ch == '%' && (i + 1) < len1)
      {
         char next = string1[i + 1];
         if (next == 's')
         {
            output_len += (arg_i < use_args) ? params_len[arg_i++] : 2;
            i += 2;
            continue;
         }
         if (next == '%')
         {
            output_len += 1;
            i += 2;
            continue;  // literal '%'
         }
         // Unknown specifier: copy the '%' and next char literally
         output_len += 2;
         i += 2;
         continue;
      }
      ++output_len;
      ++i;
   }

   // Pass 2: Walk the string again and write into the allocated buffer. 
   char *output = (char *) AllocateMemory(MALLOC_ID_STRING, output_len + 1);
   int w = 0;
   arg_i = 0;

   for (int i = 0; i < len1;)
   {
      char ch = string1[i];
      if (ch == '%' && (i + 1) < len1)
      {
         char next = string1[i + 1];
         if (next == 's')
         {
            if (arg_i < use_args)
            {
               int len = params_len[arg_i];
               memcpy(output + w, params[arg_i], len);
               w += len;
               ++arg_i;
            }
            else
            {
               output[w++] = '%';
               output[w++] = 's';
            }
            i += 2;
            continue;
         }
         if (next == '%')
         {
            output[w++] = '%';
            i += 2;
            continue;
         }
         output[w++] = '%';
         output[w++] = next;
         i += 2;
         continue;
      }
      output[w++] = ch;
      ++i;
   }
   output[w] = '\0';

	 val_type ret; 
	 ret.v.tag = TAG_STRING; 
	 ret.v.data = CreateStringWithLen(output, w);
	 FreeMemory(MALLOC_ID_STRING, output, output_len + 1);
	 return ret.int_val;
}
```

Meridian59:
Hm, this is pretty complex.. let's split this into another PR.

jrotunda:
First off: thank you for making this, good content to enhance the territory minigame for the active faction soldier PvPers and get their records of war.

> This PR removes resources in jastavnr.kod. The resources were for a Bard's newsglobe that isn't present on 101. There is at least one other instance of it on 101 though, in the Marion adventurers hall. > 

Just wanted to note that the book in Marion Adv. Hall is a Territory Tactica book that explains the faction game.

The Bards newsglobe you didn't see in Jasper Tavern has moved around, it's currently in a corner by the window in Barl Adv Hall.

There are two Bards newscroll/globes. One is in JasBar (jastav) the other is in CorHall. 

These were created and made on the code side, as news globes can't be made on the live production servers without potential issues. 

The one in jastav works to swap out with this new Territory Game globe in terms of placement, but it would be nice if we can leave the two instances of the bard newsglobe, suggestion for an alternative spot would be in the Brahma & Falcon (barbar.kod). 

 Since these are lore stories written by former developers and contest results by players, they make for good content to have around for a read and a lot of the lore books in general can get overlooked.

davidklaw:
Waiting on #1191

davidklaw:
> The one in jastav works to swap out with this new Territory Game globe in terms of placement, but it would be nice if we can leave the two instances of the bard newsglobe, suggestion for an alternative spot would be in the Brahma & Falcon (barbar.kod). Since these are lore stories written by former developers and contest results by players, they make for good content to have around for a read and a lot of the lore books in general can get overlooked.

Moved to the Bhrama & Falcon in the latest update.

davidklaw:
@Meridian59 Past feedback addressed and now updated to use the new `BuildString` C function.

kod/util/factgame/territry.kod
```diff
@@ -40,6 +41,9 @@ resources:
    TerritoryGame_his = "his"
    TerritoryGame_their = "their"
 
+   territory_string_title = "{0} now {1} {2}!"
```
akirmse: It's useful to document what the substitution parameters are.  In the distant future, such comments could be sent to translators, who otherwise would have no idea how to translate a mysterious string like this.

kod/util/system.kod
```diff
@@ -67,6 +67,13 @@ resources:
    guildwar_kill_body4 = " in guild combat."
 
    system_admin_message = "%q"
+   
+   faction_war_kill_title = "{0} bested by {1}."
+   faction_war_kill_body_start = "{0} has been bested by faction rival {1}"
+   faction_war_kill_body_claim = " during a land claim attempt"
+   faction_war_kill_body_separator = ".\n--------------------------------\n"
+   faction_war_kill_body_location = "Location: {0}\n"
+   faction_war_kill_body_factions = "Factions: {0} vs {1}"
```
akirmse: Document all the parameters

kod/util/factgame/territry.kod
```diff
@@ -26,6 +26,7 @@ resources:
    TerritoryGame_rebel = "the rebel Jonas"
    TerritoryGame_unfactioned = "the unfactioned"
 
+   TerritoryGame_bonus_territory_title = "~I%s now %s the city of %s!"
```
akirmse: Is this string and the next one still used?  They seem to duplicate the new ones.

kod/util/factgame/territry.kod
davidklaw: This is the kick in the butt I needed to see what it would take to support named parameters. If we can, I think the need for comments/documentation may go away.

Something like:
```
territory_string_title = "{faction} now {controls} {city}!"
territory_news_title = BuildString(territory_string_title, 
                                   #faction=rFaction, 
                                   #controls=rControls, 
                                   #city=rCity);
```

akirmse: Easier said than done.. that's not how function calls work in Blakod, so some significant compiler work would be required.  You'd definitely want to guarantee that all fields from the string are specified and spelled correctly, etc.  This would be quite a special case.

kod/util/factgame/territry.kod
davidklaw: This one is not, but the one that follows is used in a `MsgSendUser` call requiring a resource. I'm thinking I'll leave that for alone for now.

davidklaw:
This is now somewhat dependent on #1351, which address the recreation of rooms that include lieges and councilors. Noted in the PR and will drop to draft.

davidklaw:
Took a while to return to this, but I have. Here's the latest:

- Refactored to use new `BuildString` where applicable (posting to globes)
- Used `jbook.bgf` as a foundation for a custom Territory book (`tbook.bgf`)
- Added post title and body variations that are random chosen for each post
- Removed message logic regarding "control/controls", as it was for an unfactioned scenario that never occurs and complicates the message creation.
- Added `PostPatchUpdates` support and created, what I think is, better news creation handling in the new `TerritoryNewsLink` constructor.

Open to feedback on both the new asset and the messages, and the code of course.

davidklaw:
I'm not certain this needs an instance in each faction's quarters. Perhaps, since it's under the purview of Baron Gerah Springer of **Cor Noth**, a single instance belongs there?

jrotunda:
Just to follow up here - a single focused area for this automated announcement type (akin to the GM Hall) would be ideal and concur on it being inside or outside Baron Springer's abode (one of the current buildings).

davidklaw:
Ready for review again. After some consideration, I thought it best to simplify this to a single newsglobe in central Cor Noth, home of the posts' author, Gerah Springer.

kod/util/factgame/territry.kod
```diff
@@ -643,9 +630,22 @@ messages:
...
+      if iPick = 1 { rTemplate = territory_string_title_1; }
+      if iPick = 2 { rTemplate = territory_string_title_2; }
+      if iPick = 3 { rTemplate = territory_string_title_3; }
+
+      territory_news_title = BuildString(rTemplate, rFaction, rCity);
+      territory_news_body = BuildString(territory_string_body, sTowns, rFaction, rCity);
+
+      % Post to the territory news
+      oBook = Send(SYS,@FindNewsByNum,#num=NID_TERRITORY_NEWS);
+      Send(oBook,@PostNews,#what=Create(&CorNothAmbassador),
```
akirmse, line 646: A new ambassador gets created every time a post is made?  That is odd.  If this is just for the purposes of changing the From: line, could it be done another way, like overriding some behavior in TerritoryNews?

kod/util/system.kod
```diff
@@ -1554,6 +1583,61 @@ messages:
...
+                     if iPick = 3 { rTemplate = faction_war_kill_body_3; }
+                     if iPick = 4 { rTemplate = faction_war_kill_body_4; }
+                     if iPick = 5 { rTemplate = faction_war_kill_body_5; }
+
+                     territory_news_body = BuildString(rTemplate, Send(what,@GetTrueName), Send(killer,@GetTrueName), Send(oRoom,@GetName), rVicFaction, rKilFaction);
+                  }                  
+
+                  % Post to the territory news
+                  oBook = Send(SYS,@FindNewsByNum,#num=NID_TERRITORY_NEWS);
+                  Send(oBook,@PostNews,#what=Create(&CorNothAmbassador),#title=territory_news_title,
```
akirmse, line 1639: Same comment here about creating ambassador objects.

Also, this is a lot of code to add to System, which is already bloated beyond belief.  Could it maybe be moved elsewhere, like the Ambassador (if that class survives), or TerritoryNews?

PR #1191, opened 2025-08-12
ccode: Adds new C function for string token replacement in Blakod

davidklaw:
This PR introduces the `BuildString` C function, callable from Blakod for modern string formatting using [fmtlib](https://github.com/fmtlib/fmt). This provides a modern, type-safe alternative to the legacy %s-based string substitution system.

### Supported Format Features
The `BuildString()` function supports fmtlib's format string syntax with the following features being the most relevant:

**Positional Arguments**
* Use `{0}`, `{1}`, `{2}`, etc. to reference arguments by position
* Allows reordering arguments for localization
* Example: `"{1} was killed by {0} {2:d} times."`

### Usage Example
```
system_user_killed_test = "{1} was killed by {0} {2:d} times."

// Usage in code:
result = BuildString(system_user_killed_test, killer_name, victim_name, kill_count);
// Result: "Bob was killed by Alice 5 times."
```

### Potential Migration Path
This function complements (rather than replaces) the existing %s-based string substitution system. New code can use `BuildString()`, while existing code continues to work with the legacy system.

---

This was prompted by #1190

### DIFF
--- a/blakserv/ccode.c
+++ b/blakserv/ccode.c
@@ -2370,11 +2370,15 @@ blak_int C_SendWebhook(int object_id, local_var_type *local_vars,
     int num_normal_parms, parm_node normal_parm_array[],
     int num_name_parms, parm_node name_parm_array[])
 {
+    // Early exit if webhooks not enabled - avoid all string/JSON work
+    if (!IsWebhookEnabled()) {
+        return NIL;
+    }
+
     val_type msg_val, event_val, key_val, value_val;
     const char *content, *event_name, *key_str, *value_str;
     int content_len, event_len, key_len, value_len;
-    std::string json;
-    
+
     // Handle single string parameter
     if (num_normal_parms == 1) {
         msg_val = RetrieveValue(object_id, local_vars, normal_parm_array[0].type, normal_parm_array[0].value);
@@ -2404,7 +2408,7 @@ blak_int C_SendWebhook(int object_id, local_var_type *local_vars,
     }
     
     // Start building JSON: {"event": "EventName", "params": {
-    json += "{\"event\":\"";
+    std::string json = "{\"event\":\"";
     json.append(event_name, event_len);
     json += "\",\"params\":{";
     
--- a/common.mak
+++ b/common.mak
@@ -2,9 +2,9 @@
 
 # defining RELEASE compiles optimized
 # defining NODEBUG omits debugging information
-# defining FINAL implies release, and also removes debugging strings from client executable 
+# defining RETAIL implies release, and also removes debugging strings from client executable 
 
-!ifdef FINAL
+!ifdef RETAIL
 RELEASE = 1
 NODPRINTFS = 1
 !endif
@@ -62,6 +62,7 @@ KODINCLUDEDIR = $(KODDIR)\include
 PALETTEFILE = $(TOPDIR)\blakston.pal
 
 # compiler specs -- uses multi-threaded DLL C runtime library
+# /FC displays full path of source code file in diagnostics
 # /TP builds C files in C++ mode
 # /WX treats warnings as errors
 # /GR- turns off RTTI
@@ -74,7 +75,7 @@ PALETTEFILE = $(TOPDIR)\blakston.pal
 # /DFMT_UNICODE=0 disables Unicode support for fmtlib
 
 CCOMMONFLAGS = -nologo -DBLAK_PLATFORM_WINDOWS -DWIN32 -DFMT_UNICODE=0 \
-             /wd4996 /wd4312 \
+             /wd4996 /wd4312 /FC \
 	     -TP -WX -GR- -EHsc- -MP -MT -Zi -std:c++20
 
 CNORMALFLAGS = $(CCOMMONFLAGS) -W2 /Ox
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -6651,7 +6651,6 @@ messages:
    PostPatchUpdates()
    "Runs one-time update functions after a patch. Call via 'send object 0 PostPatchUpdates'"
    {
-      Send(&Statue, @UpgradeStatue);
       return;
    }
 
--- a/blakserv/webhook.c
+++ b/blakserv/webhook.c
@@ -63,7 +63,7 @@ bool InitWebhooks(void)
 
 void ShutdownWebhooks(void)
 {
-    if (!webhook_initialized) {
+    if (!IsWebhookEnabled()) {
         return;
     }
 
@@ -79,14 +79,14 @@ void ShutdownWebhooks(void)
     webhook_initialized = false;
 }
 
+bool IsWebhookEnabled(void)
+{
+    return webhook_initialized;
+}
+
 bool SendWebhookMessage(const char* message, int len)
 {
-    if (!webhook_initialized || !message || len <= 0) {
-        return false;
-    }
-    
-    // Double-check config setting (in case it was changed dynamically)
-    if (!ConfigBool(WEBHOOK_ENABLED)) {
+    if (!IsWebhookEnabled() || !message || len <= 0) {
         return false;
     }
 
--- a/blakserv/webhook.h
+++ b/blakserv/webhook.h
@@ -66,6 +66,6 @@ void ShutdownWebhooks(void);
  */
 bool SendWebhookMessage(const char* message, int len);
 
-
+bool IsWebhookEnabled(void);
 
 #endif /* _WEBHOOK_H */
--- a/clientd3d/annotate.h
+++ b/clientd3d/annotate.h
@@ -15,7 +15,7 @@
 #define MAP_ANNOTATION_SIZE     (2 * FINENESS)  // Size of annotation graphic in FINENESS units
 #define MAP_ANNOTATION_MIN_SIZE 14              // Min size of drawn annotation in pixels
 #define MAP_ANNOTATION_MAX_SIZE 30              // Max size
-#define MAX_ANNOTATIONS    20
+#define MAX_ANNOTATIONS         20              // Max number of annotations in a room
 #define MAX_ANNOTATION_LEN 100
 
 typedef struct {
--- a/clientd3d/client.c
+++ b/clientd3d/client.c
@@ -11,9 +11,14 @@
 */
 
 #include <assert.h>
-#include <crtdbg.h>
 
 #include "client.h"
+
+#ifdef M59_RETAIL
+  // Minidump reporting
+  #include "bugsplat.h"
+#endif
+
 
 HWND hMain = NULL;             /* Main window */
 HINSTANCE hInst = NULL;           /* Program's instance */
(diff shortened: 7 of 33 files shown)


akirmse:
Unfortunately this is still not enough for translations.  In some languages, the order of elements is different, like "le bateau rouge" (French) vs. "the red boat" (English).  To handle this, most substitution systems require or have optional numeric placeholders, like "%1$d" and "%2$s", and thus you can change the order of substituted items if necessary.

Given this extra complexity, I wonder if you can find an existing function somewhere that implements this.  No sense reinventing the wheel.

davidklaw:
>  I wonder if you can find an existing function somewhere that implements this

`CheckServerMessage` does to a large degree, but is on the client side. Should I look into sharing its logic between client and server, and extend it to support positional params?

akirmse:
I was thinking more of third-party code, i.e. someone who has already solved this problem.

davidklaw:
I've got it working using [fmtlib](https://github.com/fmtlib/fmt) and supporting a few tags: `TAG_INT`, `TAG_STRING`, `TAG_TEMP_STRING`, and `TAG_RESOURCE`. This would support a resource string like so, `"{1} was killed by {0} {2:d} times."`, such that the following call would produce `Admin1 was killed by Admin2 5 times.` If this seems right, I'll submit it for closer review.

```c
resources:

system_user_killed_test = "{1} was killed by {0} {2:d} times."

messages:

BuildStringTest(what=$,killer=$)
{
  local sString;

  sString = BuildString(system_user_killed_test,
     Send(what,@GetTrueName),
     Send(killer,@GetTrueName),
     5);

  debug("BuildStringTest: ", sString);

  return;
}
```

Produces: `BuildStringTest: Admin1 was killed by Admin2 5 times.`

davidklaw:
I've integrated [fmtlib](https://fmt.dev/11.1/), which should cover the needs described above. This is new territory for me, and I ended up following header-only instructions that don’t match how we typically handle external libraries. Let me know if there’s a better way to include it.

davidklaw:
Struggling with the Linux build--sorry for the spam. Could use some guidance there if we like fmtlib as a solution.

davidklaw:
I think the issue is that the Linux build is using C++11. I’m relying on fmt features that need C++17, like `vformat` and dynamic arguments. Any chance we can bump the build up to C++17?

akirmse:
This is somewhat tricky.  Going to C++17 would mean upgrading the compiler on the serving machines (and the github autobuild), as well as the C++ runtime library, then rebuilding the server and hoping it works.  Downgrading in the case of problems isn't guaranteed to be possible.  I have done this once before but it makes me nervous.  I may need to think about this one some more.

akirmse:
I was wrong about this.  The versions of gcc we have installed do support C++17 already, and in fact it's the default (which we've been overriding with C++11 due to ancient history).  I've checked in the change, so if you merge you should be able to get farther.

davidklaw:
> I was wrong about this. The versions of gcc we have installed do support C++17 already, and in fact it's the default (which we've been overriding with C++11 due to ancient history). I've checked in the change, so if you merge you should be able to get farther.

Awesome, thanks. Seems my build struggles are over and its now working. Ended up using an older version of fmtlib because later versions appear to have a type mismatch bug whose warnings get treated as errors and thus fail to compile. The version I ended up adding--11.1.0--should support everything we need.

davidklaw:
@Meridian59 This is ready for another round of review.

kod/util/system.kod
```diff
@@ -6599,6 +6601,20 @@ messages:
       return;
    }
 
+   BuildStringTest(what=$,killer=$)
```
akirmse: I like the idea of having tests we can run, but not the idea of making a 6000-line source file bigger.  Move this into a new top-level class?  Give some thought to how tests should work in general (i.e. it would be great if each test returned true/false)---imagine we had 100 tests, we would want a way to run them all and get a summary of which ones failed, rather than having to do manual inspection of the results.

common.mak
```diff
@@ -73,7 +74,7 @@ PALETTEFILE = $(TOPDIR)\blakston.pal
 
 CCOMMONFLAGS = -nologo -DBLAK_PLATFORM_WINDOWS -DWIN32 \
              /wd4996 /wd4312 \
-	     -TP -WX -GR- -EHsc- -MP -MT -Zi -std:c++20
+	     -TP -WX -GR- -EHsc- -MP -MT -Zi -std:c++20 -DFMT_UNICODE=0
```
akirmse: Document the FMT_UNICODE flag

blakserv/ccode.c
```diff
@@ -17,6 +17,8 @@
 */
 
 #include "blakserv.h"
+#define FMT_HEADER_ONLY
```
akirmse, line 20: If we use FMT_HEADER_ONLY, then is it possible that we don't need the .cc files, and we don't need to build fmtlib.lib?

kod/util/system.kod
davidklaw: I only intended for this to one an example to test with, but removed before merging (hidden in the PR description). However, I can envision what you’re suggesting given my experience with a few different testing frameworks.

davidklaw:
@Meridian59 I've removed the example test code and addressed the other feedback. While I am exploring a testing framework for kods, I assumed that was a completely independent endeavor. This is ready for you again.

.gitignore
```diff
@@ -14,6 +14,7 @@ lib/libpng.lib
 lib/mss*.lib
 lib/waveplay.lib
 lib/zlib.lib
+lib/fmtlib.lib
```
akirmse: No longer needed?

blakserv/makefile.linux
```diff
@@ -6,7 +6,7 @@ TOPDIR=..
 
 include $(TOPDIR)/common.mak.linux
 
-CFLAGS += -I $(BLAKINCLUDEDIR)
+CFLAGS += -I $(BLAKINCLUDEDIR) -I $(TOPDIR)/external/fmtlib
```
akirmse: Define $(FMTLIBDIR) in common.mak.linux instead, the same as you did on Windows.

common.mak
```diff
@@ -70,10 +71,11 @@ PALETTEFILE = $(TOPDIR)\blakston.pal
 # /MP enables parallel compiling
 # /MT link with static C runtime library
 # /Zi includes debugging information
+# /DFMT_UNICODE=0 disables Unicode support for fmtlib
 
 CCOMMONFLAGS = -nologo -DBLAK_PLATFORM_WINDOWS -DWIN32 \
              /wd4996 /wd4312 \
-	     -TP -WX -GR- -EHsc- -MP -MT -Zi -std:c++20
+	     -TP -WX -GR- -EHsc- -MP -MT -Zi -std:c++20 -DFMT_UNICODE=0
```
akirmse: Put all of the -D flags next to each other to make things easier to read.

common.mak
```diff
@@ -131,4 +133,4 @@ MAKEBGF = $(BLAKBINDIR)\makebgf
 # environment variables for compiler
 
 LIB = $(LIB);$(BLAKLIBDIR)
-INCLUDE = $(INCLUDE);$(BLAKINCLUDEDIR);$(LIBARCHIVEDIR);$(LIBPNGDIR);$(ZLIBDIR);$(OPENALDIR)\include
+INCLUDE = $(INCLUDE);$(BLAKINCLUDEDIR);$(LIBARCHIVEDIR);$(LIBPNGDIR);$(ZLIBDIR);$(OPENALDIR)\include;$(FMTLIBDIR)\fmt;$(FMTLIBDIR);
```
akirmse: This is inconsistent with the way you modify the include path just for blakserv on Linux.  Pick one approach and stick with it.

kod/util/system.kod
```diff
@@ -59,6 +59,8 @@ resources:
    system_user_killed_shield = "### %q was bested by faction rival %s%q."
    system_user_killed_guild = \
       "### %q of %s%q has been slaughtered by %s%q of %s%q in guild combat."
+
+   system_user_killed_test = "{1} was killed by {0} {2:d} times."
```
akirmse: Remove

blakserv/ccode.c
```diff
@@ -2300,6 +2302,70 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+      // Now do the actual formatting: substitute {} placeholders with our arguments
+      std::string fmt_str(string1, len1);
+      std::string result = fmt::vformat(fmt_str, fmt::format_args(format_args.data(), static_cast<int>(format_args.size())));
+
+      val_type ret;
+      ret.v.tag = TAG_STRING;
+      ret.v.data = CreateStringWithLen(result.c_str(), static_cast<int>(result.length()));
+      return ret.int_val;
+   }
+   catch (const std::exception &e)
```
akirmse: Does fmtlib really throw a generic std::exception?  The best practice is to catch the most specific exception class that makes sense, so as not to catch random unrelated stuff.

blakserv/ccode.c
```diff
@@ -2300,6 +2302,70 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
 
+/**
+ * C_BuildString: Builds a new string from a format string and substitution arguments.
+ * Uses fmtlib with support for both string and integer arguments.
+ * Format string should use {} tokens (e.g., "{0}", "{1:d}", etc.)
+ */
+blak_int C_BuildString(int object_id, local_var_type *local_vars, int num_normal_parms, parm_node normal_parm_array[],
+                       int num_name_parms, parm_node name_parm_array[])
+{
+   const char *string1 = NULL;
```
akirmse: Prefer nullptr to NULL.

blakserv/ccode.c
davidklaw: Improved here for more distinct handling of formatting errors, and other exceptions from the use of `std::vector` and `std::string`.

```c
catch (const fmt::format_error &e)
{
  bprintf("C_BuildString format error: %s\n", e.what());
  return NIL;
}
catch (const std::bad_alloc &e)
{
  bprintf("C_BuildString out of memory: %s\n", e.what());
  return NIL;
}
catch (const std::exception &e)
{
  bprintf("C_BuildString exception: %s\n", e.what());
  return NIL;
}
```

blakserv/ccode.c
```diff
@@ -2300,6 +2302,80 @@ blak_int C_MinigameStringToNumber(int object_id,local_var_type *local_vars,
...
+   {
+      bprintf("C_BuildString format error: %s\n", e.what());
+      return NIL;
+   }
+   catch (const std::bad_alloc &e)
+   {
+      bprintf("C_BuildString out of memory: %s\n", e.what());
+      return NIL;
+   }
+   catch (const std::exception &e)
```
akirmse: Exceptions are a complex topic, which is why we generally don't use them, but essentially you shouldn't catch exceptions that you're not going to do anything about.  So catching format_error is reasonable, because it prevents the game from crashing with bad format strings, which can definitely happen.  But catching bad_alloc is pointless, because if we're out of memory, the game is going to crash anyway, and catching std::exception is actually somewhat dangerous because you don't even know what the exception is.

- You don't need to catch bad_alloc

akirmse:
This looks OK but it has merge conflicts.

Issue #1096, opened 2025-02-04
Spell: Hunt:  Personal enchantment doesn't wear off if cast on offline user.


jrotunda:
There's a reported issue with the Kraanan spell hunt staying on its target longer than it's supposed to and/or permanently. 

To test:

- Requires 3 Characters with the spell Hunt (can be done with 5 or 7, 3 for ease of testing)
- A 3 sided prism
- Another test character 
- All 3 characters logged in and casting in the same room on the same prism object
- In admin, carefully change the cost required to speed along the test
- Cast the spell on the test character (note: the test character may not be needed, you may be able to cast Hunt on one of the people casting the spell but I've never tested that personally)
- Take note of the enchantmentslist of the target character after the spell ends and note the timer
- Wait past the time of expected expiration of the Hunt spell and check the active enchantments of the user

This is a pain to test but this issue was being reported and observed, with a character having hunt permanently on their character and is the reason its currently off. 

This issue doesn't include any mention of the design or focus of the spell and is only intended to address the timer problem.

jrotunda:
This bug is referenced in [hunt.kod]:

_**% Disabled temporarily 
   % When used on offline targets who log on during it's duration, the target
   % will be tracked indefinitely. Offline players' timers are "frozen", but 
   % Hunt introduces an active timer that decouples from the enchantment during
   % reactivation of timers (when the player logs on). This should be addressed
   % or the spell reworked before enabling.**_

davidklaw:
Can be closed. Resolved via #1309

PR #1192, opened 2025-08-17
fix: Stores enchantment timers in proper format when player is offline

davidklaw:
This PR adds a logged-in check when starting a personal enchantment. If the player is offline, the enchantment timer is stored in the same integer format used for offline (frozen) enchantments. Previously, no such check existed, and starting certain enchantments (e.g., Hunt) on an offline player would store a Timer instead of an integer, causing the enchantment to eventually "break" when `ReactivateAllEnchantments` is called.

Related: #807, #1096

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -2722,15 +2722,24 @@ messages:
       {
          oTimer = $;
       }
-      else 
+      else
       {
-         if lastcall
+         if NOT Send(self,@IsLoggedOn)
          {
-            oTimer = CreateTimer(self,@EnchantmentTimer,time);
+            % Player is offline. Store time in frozen integer format.
+            oTimer = time;
          }
          else
          {
-            oTimer = CreateTimer(self,@PeriodicEnchantmentTimer,time);
+            % Player is online. Create a timer.
+            if lastcall
+            {
+               oTimer = CreateTimer(self,@EnchantmentTimer,time);
+            }
+            else
+            {
+               oTimer = CreateTimer(self,@PeriodicEnchantmentTimer,time);
+            }
          }
       }
 


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -2722,15 +2722,24 @@ messages:
...
       }
-      else 
+      else
       {
-         if lastcall
+         if NOT Send(self,@IsLoggedOn)
          {
-            oTimer = CreateTimer(self,@EnchantmentTimer,time);
+            % Player is offline. Store time in frozen integer format.
+            oTimer = time;
```
davidklaw, line 2730: I was wondering if I ought to rename `oTimer` to something generic, like `timer`?

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -2722,15 +2722,24 @@ messages:
       {
          oTimer = $;
       }
-      else 
+      else
       {
-         if lastcall
+         if NOT Send(self,@IsLoggedOn)
```
akirmse, line 2727: if NOT X { A } else { B } 
=> 
if X { B } else { A }

kod/object/active/holder/nomoveon/battler/player.kod
akirmse: This definitely deserves to be commented more.  A variable that can be of multiple types is an invitation to future bugs, since it's rarely done.  And yes, renaming could help in this case (the "o" prefix means "object" and this is no longer always an object/timer).

davidklaw:
> This seems like a pretty big change, since it affects every enchantment. Is the fundamental problem this function, or is it that some other code is starting an enchantment on a player who is offline? That feels like a weird thing to do.

Hmm, yes, it is big. I've been wondering this as well. FWIW, this solution was your suggestion😁. 

Prism spells and in particular Hunt (the only multicast spell with a player target) are the culprit here. Whether it was deliberately designed to allow for casting on offline players, I don't know. Other possible paths:

1. A separate `StartEnchantmentOffline` for storing enchantment as frozen/offline integers.
2. Let Hunt fail if the target player is offline at cast time. This seems undesirable given the players' investment in casting, but still a viable option. As mentioned in [this comment](https://github.com/Meridian59/Meridian59/pull/687#issuecomment-2237833469), we lose some functionality with this but the odds of someone strategically avoiding it with a logoff is low.

davidklaw:
> This seems like a pretty big change, since it affects every enchantment. Is the fundamental problem this function, or is it that some other code is starting an enchantment on a player who is offline? That feels like a weird thing to do.

I'm considering a new approach that leverages an optional parameter, `allow_offline`. Thoughts on that?

```diff
StartEnchantment(what = $,time = $,state = $,lastcall = TRUE,addicon = TRUE,allow_offline = FALSE)
...

if time = $ OR time = SPELL_INDEFINITE_DURATION 
{
   oTimer = $;
+  iTime = $;
}
else
{
+  if allow_offline AND NOT Send(self,@IsLoggedOn)
   {
+     iTime = time;
      oTimer = $;
   }
   else
   {
      % Player is online. Create a timer.
      if lastcall
      {
         oTimer = CreateTimer(self,@EnchantmentTimer,time);
      }
      else
      {
         oTimer = CreateTimer(self,@PeriodicEnchantmentTimer,time);
      }
+     iTime = $; % Clear alternate timer format
   }
}

...
```

akirmse:
I'm not too excited about that as it doubles the complexity of testing enchantments.  Better to put in the hard work now of understanding how offline enchantments should work in general.

davidklaw:
I'm going to close this in favor of handling Hunt's enchantment timer issues with a solution like #1309

PR #1122, opened 2025-02-23
Add all keybind configuration settings to in-game preferences

sicoderealm:
Add all functionality from keybind to the main game preferences in-game settings.

https://github.com/user-attachments/assets/a8c340f3-a33b-4726-919e-e099b7fa520a

PR #1202, opened 2025-08-31
Redraw after toolbar/lagbox config change

akirmse:
Immediately update the UI after the toolbar or lagbox config options are changed.

This was lost in #1122.


davidklaw:
Do we also need to bring back this bit?

```c
new_val = Trackbar_GetPos(GetDlgItem(hDlg, IDC_MUSIC_VOLUME));
if (new_val != config.music_volume)
{
    config.music_volume = new_val;
    ResetMusicVolume();
}
```

PR #1221, opened 2025-09-07
Restore the scrolls that were removed from the game

Gar103:
Poison Fog and Winds scrolls were (I assume accidentally) removed from the game with the latest treasure type changes that were intended to lower 'clutter'. While the Poison Fog scroll isn't widely used, Winds scrolls are a combat staple, and Daemon skeletons were the only source.

### DIFF
--- a/kod/object/passive/trestype/skel4t.kod
+++ b/kod/object/passive/trestype/skel4t.kod
@@ -31,7 +31,7 @@ messages:
    
    constructed()
    {
-      plTreasure = [ [ &BlueMushroom, 25 ],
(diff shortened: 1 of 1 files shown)


pauly-barns:
Confirmed this as well post-patch.  Would be great to have them back.

akirmse:
It could be a long time before a checked-in change goes live.  So instead of this we're going to make some modifications on the live server.

Gar103:
The base code does need to be altered as well, or it'll revert on recreate and remove the scrolls again

PR #1229, opened 2025-09-14
client: Disable middle and right mouse buttons for text area resizing

### LINKED ISSUE #1224: Pressing right/middle mouse button at edge of camera bugs out the camera
Left clicking and dragging the edge of the camera lets you drag in/out the camera as normal.  But right clicking on the edge locks in the camera resize until you left click again, which can be a bit disorienting during gameplay.  It's even worse with the middle mouse button, where it immediately shrink the camera to the smallest size, too.  This was discovered by accident a few times while playing the game on full screen on a 1920 x 1090 resolution.

davidklaw:
This PR restricts support for text/chat area resizing to the left mouse button only. Previously, the middle and right mouse buttons were supported but are buttons that don’t have corresponding “up” handlers, which can lead to a frustrating user experience (see #1224).

Closes #1224

### DIFF
--- a/clientd3d/statgame.c
+++ b/clientd3d/statgame.c
@@ -212,7 +212,7 @@ void GameMouseButtonDown(HWND hwnd, BOOL fDoubleClick, int x, int y, UINT keyFla
 
    SetFocus(hMain);
 
-   if (text_area_resize_zone) {
+   if (text_area_resize_zone && keyFlags & MK_LBUTTON) {
        // Start text area resize.
        text_area_resize_inprogress = true;
        previous_mouse_position = { x,y };


akirmse:
I think we want to do this in a slightly different way.  Rejecting the middle and right mouse buttons should be done in the merintr module rather than the main client.  This will prevent the code from being duplicated 3 times, and also allow the mouse buttons to potentially be used in the future in the interface, instead of being silently discarded by the client.

My thought is to still call GameMouseButtonDown here, but pass in the button that was clicked (using either an existing or new enum, probably), and then have merintr throw away the middle and right mouse button clicks.

davidklaw:
I wonder, is this simpler approach viable? Not touching `GameMouseButtonDown` and instead just throwing away middle and mouse buttons in the interface's click handling?
```diff
bool WINAPI EventMouseClick(HWND hwnd, BOOL fDoubleClick, int x, int y, UINT keyFlags)
{
+  // If the button is middle or right, don't pass it on to the game
+  if (keyFlags & MK_MBUTTON || keyFlags & MK_RBUTTON)
+  {
+     return false;
+  }

   if (fDoubleClick)
      PerformAction(A_ACTIVATEMOUSE, NULL);

   // When button pressed, do both setting focus and selection (default is focus only)
   if (GameGetState() == GAME_SELECT)
   {
      SetFocus(cinfo->hMain);
      PerformAction(A_SELECT, NULL);
      return false;
   }
}
```

akirmse:
No objection from me if that works.  It would need a comment explaining why the mouse clicks are being discarded, of course.

davidklaw:
> No objection from me if that works. It would need a comment explaining why the mouse clicks are being discarded, of course.

I took a different, even simpler approach after realizing this could be handled at the point of the text resize logic. I didn't feel there was a need to comment on it in this case.

PR #1230, opened 2025-09-15
fix: Correct logoff penalty calculation by using absolute value

davidjonathansheffield:
Fixed bug where BaseLossAmount division was amplifying penalties instead of reducing them. Since BaseLossAmount is negative (-2 for murderers, -1 for non-murderers), dividing by it was inverting the math. Now using Abs(BaseLossAmount) for both spell and skill penalty calculations.

### DIFF
--- a/kod/object/active/logghost.kod
+++ b/kod/object/active/logghost.kod
@@ -381,7 +381,7 @@ messages:
          }
 
          iLoops = 0;
-         LossCounter = LossCounter - (numSpellPointsLost/BaseLossAmount);
+         LossCounter = LossCounter - (numSpellPointsLost/Abs(BaseLossAmount));
          while (LossCounter > 0) and (iLoops < 200)
          {
             iLoops = iLoops + 1;
@@ -457,7 +457,7 @@ messages:
          }
 
          iLoops = 0;
-         LossCounter = LossCounter - (numSkillPointsLost/BaseLossAmount);
+         LossCounter = LossCounter - (numSkillPointsLost/Abs(BaseLossAmount));
          while LossCounter > 0 AND iLoops < 100
          {
             iLoops = iLoops + 1;


davidjonathansheffield:
This has not been manually tested yet, do not merge in until after, converting back to draft

akirmse:
All these negative amounts are super confusing.  What about making BaseLossAmount a positive number (i.e. 1 or 2)?  LossAmount will therefore also be positive. I think that would simplify the code quite a bit and make it easier to understand.  A few minus signs will still be needed.

davidjonathansheffield:
Yes I agree, it's certainly the case that the confusingness of the negative numbers led to this bug in the first place because of how the Bound method is being used.  Will adjust on this, thanks for the review!

davidjonathansheffield:
Ok changes are in and tested, looks good to go and should be more maintainable in the future with the positive numbers for BaseLossAmount

kod/object/active/logghost.kod
```diff
@@ -371,7 +371,7 @@ messages:
                i = i + 1;
                iSpellAbility = Send(poGhostedPlayer,@DecodeSpellAbility,
                                     #compound=Nth(lSpells,i));
-               LossAmount = Bound(BaseLossAmount*iLoops,$,-(iSpellAbility-5));
+               LossAmount = Bound(-BaseLossAmount*iLoops,-(iSpellAbility-5),$);
```
akirmse: I still find this confusing.  Now that BaseLossAmount is positive, how about we make this
```
               LossAmount = Bound(BaseLossAmount*iLoops, $, iSpellAbility-5);
```
Then LossAmount will be positive also, requiring flipping minus signs in two places below.

A comment wouldn't hurt here either.  I guess the intention is not to reduce the spell below 5%?

Though if the spell is already below 5%, would this actually *increase* it to 5%?  So maybe the lower bound should be 0 instead of $?

kod/object/active/logghost.kod
```diff
@@ -447,7 +447,7 @@ messages:
                i = i + 1;
                iSkillAbility = Send(poGhostedPlayer,@DecodeSkillAbility,
                                     #compound=Nth(lSkills,i));
-               LossAmount = bound((BaseLossAmount*iLoops),$,-(iSkillAbility-5));
+               LossAmount = bound((-BaseLossAmount*iLoops),-(iSkillAbility-5),$);
```
akirmse: same here

kod/object/active/logghost.kod
davidsheffieldtmt: Oh good points, both.. Hmm!  Ok.  I will adjust it and pay special attention to edge cases -- I think there's never any possibility for skills to go below 5% by way of log off penalty bug so it shouldn't ever increase skills, at worse they would just stay at 5% -- but, that doesn't mean we should introduce a potential bug that COULD increase their skills, I will think about and address this.

davidjonathansheffield:
I added some conditional if else branching logic, I could have used a nested Bound function call, I felt when I self-reviewed that that it just seemed more difficult to follow with Bound being called twice in the same line, but I'm happy to change it to that if we prefer a more functional/declarative approach over iterative.  I do think the code is easier to understand

davidjonathansheffield:
Alright, all tested and I think good to go!

kod/object/active/logghost.kod
```diff
@@ -371,12 +371,21 @@ messages:
...
+               % Players shouldn't lose their last 5% from logoff penalties to avoid
+               % complete skill loss spiral
+               if iSpellAbility > 5
+               {
+                  LossAmount = Bound(BaseLossAmount*iLoops,$,iSpellAbility-5);
+               }
+               else
+               {
+                  LossAmount = 0;
+               }
```
akirmse: I don't think you need nested Bound calls.  This can be simply:
```
LossAmount = Bound(BaseLossAmount*iLoops,0,iSpellAbility-5);
```
with no if.

kod/object/active/logghost.kod
```diff
@@ -447,12 +456,21 @@ messages:
...
+               % Players shouldn't lose their last 5% from logoff penalties to avoid
+               % complete skill loss spiral
+               if iSkillAbility > 5
+               {
+                  LossAmount = Bound(BaseLossAmount*iLoops,$,iSkillAbility-5);
+               }
+               else
+               {
+                  LossAmount = 0;
+               }
```
akirmse: here too

davidjonathansheffield:
Right on!

CLAUDE.md
```diff
@@ -0,0 +1,123 @@
+# CLAUDE.md
```
akirmse: Please remove this file.

davidjonathansheffield: Done, sorry about that!  Should have been more careful git committing all 😬

PR #1231, opened 2025-09-18
Remove problematic Line of Sight check for wall elements

Gar103:
Wall elements currently require Line of Sight to affect targets, which is a notoriously buggy check that essentially renders wall spells useless anywhere near walls. Considering the extremely short range on wall elements (essentially requiring you to stand in them or run through them), this doesn't really make sense. If you're standing in a wall of fire, you should be getting burned. This longstanding bug has resulted in an enormous number of cheating accusations and arguments, as players do indeed 'run right through' each other's webs, spores, and firewalls on an unpredictable basis.

### DIFF
--- a/kod/object/active/wallelem.kod
+++ b/kod/object/active/wallelem.kod
@@ -260,7 +260,6 @@ messages:
 
          if (iLongRange = piRange
              OR iDistanceSquared <= (piRange * piRange))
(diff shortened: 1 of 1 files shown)


akirmse:
This is certainly a problem, but I'm not sure about the fix.  It would mean that stealth wall spells could be set up behind walls where targets can't see them, and they could die without even knowing why.  Griefers would make precision use of that tactic.

PR #1240, opened 2025-09-27
fix: self-delete orphaned wall elements in `PeriodicEffect`

### LINKED ISSUE #1232: Server errors in log with wall elements having no owner
Thousands of these in quick succession:
[memmap/wallelem.bof (201)] C_SendMessage OBJECT 1310196 CLASS WallofFire can't send MESSAGE GetHolderActive (11401) to non-object 0,0

Likely due to wall spells putting elements outside of the room boundary, leading to wall elements with $ owner.

adrienlaws:
## What
- adds a high level guard in `ActiveWallElement`'s `PeriodicEffect` to self-delete a wall element if it is orphaned (i.e., has no valid room owner)

## Why
- prevents error spam caused by orphaned wall elements attempting to access their owner after failed placement or room deletion
- ensures all wall element types inheriting from `ActiveWallElement` are robust against this issue

## Notes
- fixes #1232
  - complements other fix https://github.com/Meridian59/Meridian59/commit/eab11f5a70445922fa26ac82e6cf613ae5d0cdcb
- this fix is in [`wallelem.kod`]https://github.com/Meridian59/Meridian59/blob/master/kod/object/active/wallelem.kod) so all child wall spells benefit automatically 
  - [bramble.kod](https://github.com/Meridian59/Meridian59/blob/master/kod/object/active/holder/nomoveon/battler/monster/BRAMBLE.kod) as a "monster"  is a special case but it seems ok

### DIFF
--- a/kod/object/active/wallelem.kod
+++ b/kod/object/active/wallelem.kod
@@ -199,6 +199,13 @@ messages:
 
       if vbPeriodic
       {
+         %% self-delete orphaned wall elements
+         if poOwner = $ OR NOT IsClass(poOwner,&Room)
+         {
+            Post(self, @Delete);
+            return;
+         }
+
          for i in send(poOwner,@GetHolderActive)
          {
             oObject = send(poOwner,@HolderExtractobject,#data=i);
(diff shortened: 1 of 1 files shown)


akirmse:
I don't think this is necessary---I already fixed the bug earlier today in eab11f5a70445922fa26ac82e6cf613ae5d0cdcb.  We should prevent wall elements from getting $ owners in the first place, as that's always a bug, rather than trying to repair it after the fact.

adrienlaws:
That makes sense.

PR #1254, opened 2025-09-30
Make account name and password into std::strings

akirmse:


blakserv/account.c
```diff
@@ -33,9 +33,6 @@ void InitAccount(void)
 
    console_account = &console_account_node;
    console_account->account_id = 0;
-   console_account->name = ConfigStr(CONSOLE_ADMINISTRATOR);
```
CyberBill, line 36: Is the name no longer necessary to be set here? I get that it (and the password) no longer need to be allocated, but it looks like this was previously being set to "Administrator" (based on the config)

blakserv/account.c
```diff
@@ -51,9 +48,7 @@ void ResetAccount(void)
    while (a != NULL)
    {
       temp = a->next;
-      FreeMemory(MALLOC_ID_ACCOUNT,a->name,strlen(a->name)+1);
```
CyberBill, line 54: Are AllocateMemory and FreeMemory no longer preferred, and raw C++ allocation is the way to go?

akirmse: This PR is about getting rid of some strcpy calls by using std::string, but the memory accounting still needs some work.  

In the long term I'd like to get rid of all unsafe C library calls, to prepare for compilers gradually phasing out support for them over the years.

PR #1264, opened 2025-10-02
fix: address Windows blakserv dialog crashes

### LINKED ISSUE #1258: Server crash after viewing long debug message in interface
- Get a long debug message in the interface (e.g. from a dumpstack() call from Blakod).
- Double click on the message to open the detail dialog.
- Close the dialog, server will crash.

Problem is stack overwrite if the message is longer than the buffer size:
https://github.com/Meridian59/Meridian59/blob/c68673908fe584b79289b9026167a387546ecabf/blakserv/interface.c#L1162

Care needs to be taken not to write past the end of this buffer.

adrienlaws:
## What
- adds safety checks when displaying `listbox` item text to prevent stack overflow and handle invalid selections
- increases interface buffer size by using established channel constant `CHANBUF_SIZE`
- fixes #1258

## Why
- previously, the function directly called `ListBox_GetText` with the result of `ListBox_GetCurSel` without error checking
  - this could cause buffer overflow on very long messages or undefined behavior if no item was selected
  - the changed version validates the selection index and ensures the retrieved text length fits safely within the buffer
- changed static 400 byte interface buffer size to use `channel.c`  constant `CHANBUF_SIZE` (currently 1000 bytes)

## Notes
- added `LB_ERR` checks for `ListBox_GetCurSel` and `ListBox_GetTextLen`
- displays a warning dialog if a message exceeds the supported buffer size

### Before
https://github.com/user-attachments/assets/17385931-b2ae-4c92-a394-1e2026c01bba

### After
https://github.com/user-attachments/assets/49789734-efff-4101-a856-13b3fbb268ee

### DIFF
--- a/blakserv/chanbuf.h
+++ b/blakserv/chanbuf.h
@@ -15,7 +15,7 @@
 
 enum
 {
-   CHANBUF_SIZE = 2000
+   CHANBUF_SIZE = 1000
 };
 
 typedef struct
--- a/blakserv/chanbuf.h
+++ b/blakserv/chanbuf.h
@@ -15,7 +15,7 @@
 
 enum
 {
-   CHANBUF_SIZE = 1000
+   CHANBUF_SIZE = 2000
 };
 
 typedef struct
--- a/blakserv/interface.c
+++ b/blakserv/interface.c
@@ -1159,7 +1159,7 @@ INT_PTR CALLBACK InterfaceDialogTabPage(HWND hwnd,UINT message,WPARAM wParam,LPA
 
 void InterfaceTabPageCommand(HWND hwnd,int id, HWND hwndCtl, UINT codeNotify)
 {
-	char s[2000];
+	char s[CHANBUF_SIZE];
 	
 	switch (codeNotify)
 	{
--- a/blakserv/interface.c
+++ b/blakserv/interface.c
@@ -1159,7 +1159,7 @@ INT_PTR CALLBACK InterfaceDialogTabPage(HWND hwnd,UINT message,WPARAM wParam,LPA
 
 void InterfaceTabPageCommand(HWND hwnd,int id, HWND hwndCtl, UINT codeNotify)
 {
-	char s[400];
+	char s[2000];
 	
 	switch (codeNotify)
 	{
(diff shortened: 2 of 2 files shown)


adrienlaws:
I changed the buffer size to 2000 bytes and also changed `CHANBUF_SIZE` to match.  Otherwise, I could not see the longer output due to the check in `chanbuf.c`'s `WriteChannelBuffer` [here](https://github.com/Meridian59/Meridian59/blob/5093e41079b6203c8843c4b6165824232b5bfec9/blakserv/chanbuf.c#L49)
- the output was simply `WriteChannelBuffer got string too long to display`
- adjusting it to 2000 bytes gives the output as in the screenshot below
image

I think this is the expected behavior and the memory increase should be only about ~60KB total (60 channels * extra 1000 Bytes)

blakserv/chanbuf.h
```diff
@@ -15,7 +15,7 @@
 
 enum
 {
-   CHANBUF_SIZE = 1000
+   CHANBUF_SIZE = 2000
```
akirmse: This has other side effects, in particular, making the server use more memory all the time because of larger buffers for channel output.  So you shouldn't change it unless you understand all the implications and can argue that it's worth it.

blakserv/interface.c
```diff
@@ -1159,15 +1159,34 @@ INT_PTR CALLBACK InterfaceDialogTabPage(HWND hwnd,UINT message,WPARAM wParam,LPA
 
 void InterfaceTabPageCommand(HWND hwnd,int id, HWND hwndCtl, UINT codeNotify)
 {
-	char s[400];
+	char s[2000];
```
akirmse: Seems like making the size CHANBUF_SIZE makes the most sense.

adrienlaws: Changed to `CHANBUF_SIZE` as suggested

blakserv/chanbuf.h
adrienlaws: Reverted to 1000 bytes.  I had done some basic math of 60 channels * extra 1000 bytes per channel and the memory use seemed negligible.  However,  I don't understand all the implications so probably best to leave it be.

That said, some very long debug messages from `DumpStack()` now show as 
> WriteChannelBuffer got string too long to display

This is probably okay since they are edge cases and the logs should be in the log files on the server.

PR #1265, opened 2025-10-04
Nerudite Sword damage types fix

Gar103:
At some point Nerudite Swords were made enchantable by dedications, but this also included Kraanan Enchantments, which use damage type flags rather than spell type additions. Nerudite Swords do not naturally have ATCK_WEAP_NONMAGIC or ATCK_WEAP_MAGIC flags.

Enchanting them to Kraanan adds an ATCK_WEAP_MAGIC flag, then leaves them with a permanent ATCK_WEAP_NONMAGIC flag when it wears off, indefinitely 'bugging' the nerudite sword - it makes the weapon quite a bit worse against monsters. The resistances of mobs were definitely not designed with weapons that were both nerudite and magic/nonmagic in mind, so the results are confusing. The age-old "Uhh my neru's not working" mind-boggle. It's an invisible problem that simply couldn't be figured out from in-game information.

In this pull, I've blocked Nerudite Swords from being enchanted to Kraanan. There are no currently situations where an ATCK_WEAP_MAGIC flag improves a nerudite sword's performance against a monster, I believe. It's purely a negative. And when it becomes ATCK_WEAP_NONMAGIC, it's even worse - so this shouldn't affect balance or player behavior. Kraanan enchantments also do not have any effect on PvP, so there's no problem there.

Over the years, I have found numerous examples in-game of Nerudite Swords with lingering ATCK_WEAP_MAGIC and ATCK_WEAP_NONMAGIC flags. So, I added an effect to the pool of water in the Shal'ille temple - if you want to fix your bugged Nerudite Sword, you can drop it in the water there, and it will remove any existing Kraanan enchantment (which will persist for a few days after a patch) and any lingering incorrect flags. If you don't want to do this, no problem, just keep it as is. (since some people swear by their Kraanan enchantments...) If you do cleanse it, a neat little message will inform you something happened.

### DIFF
--- a/kod/object/active/holder/room/temple.kod
+++ b/kod/object/active/holder/room/temple.kod
@@ -26,6 +26,8 @@ resources:
    temple_close = "Close"
 
    temple_music = temple.mid
+   
+   item_cleansed_in_pool = "The sacred waters of Shal'ille glimmer briefly."
 
 classvars:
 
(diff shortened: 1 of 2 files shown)


akirmse:
Let's actually change this to preventing any enchantments on nerudite swords, as was discussed last time.

And please revert the lake dunking thing---that's not needed in the long term.

Gar103:
The temple cleansing thing is 90% of the point, so that people can fix their prized nerudite swords that are bugged. Lots of people have hold nerudite swords or colored nerudite swords they'd like to see fixed, those aren't going anywhere in the long run. It's not necessarily about performance, but just the unhappy notion that your prized item is 'bugged' is a bummer. Would be strange to fix the bug but not let people fix their items.

For the unenchantable part, I'm just looking to fix a bug here. Nerudite swords are the top melee weapon at the moment - unenchantability would be another big nerf nobody's asking for, and that would just make people even angrier. Would rather not get into that at this time

akirmse:
The lake dunking isn't discoverable, and is only temporary, so I'd have to remember to remove it sometime in the future.  An alternative would be to add a fixer function in System that goes through all the nerudite swords in the game and fixes them up automatically.  We could run that function after the next update.

Pretty much everyone, including you, was OK with simply making nerudite swords unenchantable when this came up last time.  I'd rather pull off the Band Aid all at once.

If you're not interested in feedback, then you can close this.

Aesica:
> For the unenchantable part, I'm just looking to fix a bug here. Nerudite swords are the top melee weapon at the moment - unenchantability would be another big nerf nobody's asking for, and that would just make people even angrier. Would rather not get into that at this time

Unenchantability isn't a nerf, it's just a specific behavior for this type of sword that aligns with how they worked originally.  If people want a kraanan enchanted longsword weapon, there's mystic swords.  If they want a holy or unholy option, there's long swords.  This kind of change gives those other weapons more value rather than leaving nerudite swords as the go-to swiss army knife of longswords.  Besides, enchanting your weapon for pvp is more of a self-inflicted nerf because it becomes susceptible to mitigation through the various resistance sources.  So that argument doesn't really hold weight.

A run-once-after-update command is a much better way to apply this kind of fix, but I'd argue even that isn't necessary.  The nonmagic flag is more of an issue against monsters than it is against players, and nobody's using their special hold/blind/colored nerudite swords to hunt trolls.

Gar103:
I mean this respectfully, it's extremely frustrating for the community to see that you guys don't know us or the game - or how we play it - yet keep talking to us this way. Everyone can see the things you post here, and how you speak to us. At this very moment, they're reading your replies here and sending me messages of care and support, and are apologizing on your behalf. The attitude here is unkind - it was just described to me by others as 'scorn' in chat on 101 - and the Meridian community has massively low ongoing morale because of it. The legacy of the game is shifting over time, away from what we all know and love, toward... _this_. This situation. This is all any of us talk about now. Every old guild, every veteran playing other games. Commiseration over this, for years upon years, waiting for something to change.

We just can't seem to reach you. We don't feel heard. The only bits and pieces we get back are scorn, or bluntly being told we're wrong - when we know we're not. I'd be happy to discuss feedback. I'd be happy to contribute all sorts of things, always have been. I don't deserve this. None of us do.

Good luck

PR #1267, opened 2025-10-07
feat: improve item description formatting and misdirection display

adrienlaws:
## What
- improved item description formatting by adding line breaks between base description, attributes, and condition sections
fixed spacing issues by removing hardcoded leading spaces from item attribute descriptions
   - now there are 3 distinct sections when viewing an item's description
      - 1/ base item description "this is a mace"
      - 2/ unique attributes (if available) "it can hold people"
      - 3/ item health / durability (if relevant) "it's nearly broken"
- enhanced misdirection / artifice behavior to show both real and fake attributes together instead of hiding real attributes

## Why
- item descriptions were previously displayed as large blocks of text that were difficult to read, especially for items with multiple magical attributes
- the old misdirection system completely hid real attributes when showing fake ones, which didn't make sense from a gameplay perspective 
  - a deceptive merchant would want to show the item's real value plus fake enhancements to make it seem like a better deal
  - also due to the way the rarity grade system displays rarities by counting unique item attributes, rarity level did not match the artifice description (example: a rare item would appear as only have one attribute but a one attribute item should appear uncommon)

## How
- modified the `CreateDesc` function in `item.kod` to add `\n\n` line breaks before the attributes section and before condition information
- removed hardcoded 2-space prefixes from item attribute description strings across multiple files (like `wablindr.kod`, `waspellt.kod`, etc.)
- implemented proper spacing logic using a `bFirstAttribute` flag to add 2 spaces between attributes
- updated misdirection handling to process fake attributes normally in the attribute loop rather than replacing all real attributes with fake ones

### Before

> Lethal in appearance, this weapon is actually only minimally effective. Its spiked head is made of soft metal and its wooden shaft is thin and brittle. Q's insignia is emblazoned on the metal of the weapon. This mace is in flawless condition.

### After

> Lethal in appearance, this weapon is actually only minimally effective. Its spiked head is made of soft metal and its wooden shaft is thin and brittle.
>
> Q's insignia is emblazoned on the metal of the weapon.
>
> This mace is in flawless condition.

### DIFF
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -220,46 +220,76 @@ messages:
       return;
    }
 
-   CreateDesc(bShowall=FALSE)
+   CreateDesc(bShowAll=FALSE)
+   "Creates the item's full description text by combining the base description "
+   "with attribute descriptions and condition information. When bShowAll is FALSE, "
+   "unidentified attributes show a generic message and misdirection attributes "
+   "display their fake descriptions alongside real attributes. When bShowAll is "
+   "TRUE, all attributes including misdirection are revealed. The description "
+   "uses proper formatting with line breaks separating the base description, "
+   "attributes section, and condition information."
    {
-      local bAlreadyBlind, bIdentified, oItemAtt, i, iNum;
+      local bAlreadyBlind, bFirstAttribute, bIdentified, oItemAtt, i, iNum;
 
       bAlreadyBlind = FALSE;
+      bFirstAttribute = TRUE;
       ClearTempString();
       Send(self,@DoBaseDesc);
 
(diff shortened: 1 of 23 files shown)


kod/object/item.kod
```diff
@@ -220,46 +220,76 @@ messages:
       return;
    }
 
-   CreateDesc(bShowall=FALSE)
+   CreateDesc(bShowAll=FALSE)
+   "Creates the item's full description text by combining the base description "
```
akirmse: Always document the return value.  In this case you can simply change this to say "Returns the item's full description..."

adrienlaws: Changed as suggested to "returns."

Issue #1268, opened 2025-10-08
Rat Mask on hall of heroes statue gives server errors


akirmse:
The error is
[memmap/mask.bof (53)] C_SendMessage OBJECT 1229213 CLASS RatMask can't send MESSAGE GetGender (10137)
to non-object 0,0

In the Statue class, there is a property called prFacialOverlay.  Despite the pr prefix, it is an object, not a resource.  Statue::SendOverlays calls @SendOverlayInformation on the prFacialOverlay, which then calls @SendOverlay.  In FaceMask, SendOverlay checkes poOwner of the mask object, but this is $, causing the error.

I'm not sure about this design, since the player's actual mask object (rather than a copy, say), is used on the statue.  If the mask's owner ever becomes $ in the natural course of the game, the mask on the statue will have no owner, causing this error.

Meridian59:
I believe that due to this bug, statues will always show the male rat mask
graphic, never the female one.

sicoderealm:
@akirmse Can now be resolved? thank you

PR #1270, opened 2025-10-09
Fix statue mask bug: use gender-specific resource over object reference

sicoderealm:
Candidate fix for Rat Mask on hall of heroes statue gives server errors #1268. Fixes are now wrapped in a system level function that can be called post patch: `send object 0 PostPatchUpdates`.

statues-error-gone

**Upgrade Testing:** Created a new environment, setup statues with a player with a mask (with $ owner) and recreated the error. Applied changes in this PR, ran upgrade function on statues, statues working as expected and no errors.

**Testing**: Character with a mask, Create Statues, take the mask and manually set its owner to nill. Server side error messages no longer take place with these changes.

### DIFF
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -6634,8 +6634,7 @@ messages:
    }
 
    PostPatchUpdates()
-   "Runs one-time update functions after a patch. Call via 'send object 0 PostPatchUpdates' 
-   from admin mode, then remove completed update functions from this method."
+   "Runs one-time update functions after a patch. Call via 'send object 0 PostPatchUpdates'"
    {
       Send(&Statue, @UpgradeStatue);
       return;


kod/object/passive/statue.kod
```diff
@@ -65,7 +65,8 @@ properties:
    prLeft_arm = $
    prLegs = $
    pbIsHairVisible = $
-   prFacialOverlay = $
+   prFacialOverlay = $ % DEPRECATED: Use poFacialOverlay instead. Kept for save compatibility.
```
akirmse: I think this is kind of a lost cause at this point.  I think we just stick with prFacialOverlay and put a comment that it's misnamed.

kod/object/item/passitem/defmod/helmet.kod
```diff
@@ -81,6 +81,12 @@ messages:
       return vrIcon; 
    }
 
+   GetOverlayForOwner(owner = $)
+   "Returns the overlay for a given owner (used by statues). Default implementation."
```
akirmse: Document what a function does, not who calls it.  Who calls it can and will change over time, and then confusion will follow.

kod/object/item/passitem/defmod/helmet/mask.kod
```diff
@@ -58,6 +58,17 @@ messages:
       return vrMaleIcon; 
    }
 
+   GetOverlayForOwner(owner = $)
+   "Returns the gender-specific overlay for a given owner (used by statues)."
+   {
+      if owner <> $ AND send(owner,@GetGender) = GENDER_FEMALE
```
akirmse: This doesn't seem right, though.  It will avoid the error, but it will show a male mask whenever the owner is $.  The way masks work on statues seems like it isn't right.  Is there a better approach?

sicoderealm: Yep , good call. This can all be removed and I think we can correct the `prFacialOverlay` to make it a proper copy of the players resource at creation. If this approach is accepted, would it be ok to do in-line like this or rather an upgrade function similar to before?

kod/object/passive/statue.kod
```diff
@@ -461,7 +461,17 @@ messages:
       prToupee = send(poOriginal,@GetHairRsc);
       send(self,@ValidateGraphics,#gender=Send(poOriginal,@GetGender));
 
-      prFacialOverlay = Send(poOriginal,@GetFacialOverlay);
+      % Get the facial overlay object and store its gender-specific resource
+      prFacialOverlay = $;
+      if Send(poOriginal,@GetFacialOverlay) <> $
+      {
+         oMask = Send(poOriginal,@GetFacialOverlay);
```
akirmse: You seem to be calling GetFacialOverlay twice in a row here?

kod/object/passive/statue.kod
```diff
@@ -612,7 +622,21 @@ messages:
 
       if prFacialOverlay <> $
       {
-         Send(prFacialOverlay,@SendOverlayInformation,#iAnimation = $);
+         % Check if prFacialOverlay is an object (legacy) or a resource (new)
```
akirmse: This is weird because a statue will switch from object to resource for prFacialOverlay the first time it's seen, meaning there will forever be a mix of old and new.  Instead, make a function that transitions a statue from old to new, and we can call it to transition all statues at update time (via admin: Send Class Statue).  Then this function can assume that all statues use resources here.

Another option would be to add a function in System that calls the transition function.  Either way, there should be a comment that tells us to remove the transition function after it's been run.

kod/object/passive/statue.kod
sicoderealm: simplified - ty

PR #1273, opened 2025-10-13
New colored sprites for different lute types

neiltaco:
This PR adds two different lute colors to help player differentiate the three lute types: lute, fine lute, and true lute.  Fine lute still uses the original graphics, while the basic lute uses the brighter version and the true lute uses the reddish version.

Also there are other scripts that still refer to the original lute sprite as 'music.bgf'.  Those files could be edited to use the new name, but I think this PR should focus first on adding new sprites.  'finelute.kod' still uses 'music.bgf' and 'luteov.bgf'.

Because of this, the .bgf files for the basic lute were named 'basiclute.bgf' and 'basicluteov.bgf'

1_example
2_example
3_example
basicluteov_example
fineluteov_example
trueluteov_example

### DIFF
--- a/resource/graphics/makefile
+++ b/resource/graphics/makefile
@@ -11,11 +11,15 @@ BGFS = \
 	$(OUTDIR)\arrowfir.bgf \
 	$(OUTDIR)\arrowsil.bgf \
 	$(OUTDIR)\arrowner.bgf \
+	$(OUTDIR)\basiclute.bgf \
+	$(OUTDIR)\basicluteov.bgf \
 	$(OUTDIR)\mystswd.bgf \
 	$(OUTDIR)\mystwdov.bgf \
 	$(OUTDIR)\ring3.bgf \
 	$(OUTDIR)\povmystsw.bgf \
 	$(OUTDIR)\splash.bgf \
+	$(OUTDIR)\truelute.bgf \
+	$(OUTDIR)\trueluteov.bgf \
 	$(OUTDIR)\moon.bgf
 
 all: makedirs $(BGFS)


neiltaco:
For some reason, this PR was including my older commits, but the files added/changed remain the same so I'll reopen this PR.

davidklaw:
This is a great idea and I really love the work. Saw in Discord that you used Aesprite to do this and I've found it to be a great tool for this type of work.

I realize adding more detail would be quite a bit of extra effort, but I wonder if the fine and true variants could have some other distinguishing feature beyond color. Maybe something small, like a subtle detail near the center or caps on the arms? Or, if it’s easier, a more pronounced color difference could also help. This AI-generated image isn’t perfect, but it shows how stretching the color differences might make them stand out more clearly. I don't know the range of woods in the game off hand, but perhaps that's something to draw upon.

image

Without any other visual accents though, these colors can sometimes be hard to distinguish under different color visions.

neiltaco:
Those images definitely make the lute stand out more!  I had mostly based the hue changes on the lutes's descriptions as follows: lute is rosewood, fine lute is pine wood with a dark stain, and true lute is a dark, heavy wood.

But maybe I can take creative freedoms.  If for some reason the descriptions shouldn't changed, I can still deviate from the actual woods' appearances for more visual readability.  Or as you suggested, more subtle details.

akirmse:
I agree, the lute and fine lute need to be differentiated more.  It's only worth doing if people can tell them apart easily.

neiltaco:
I went for another recolor and made the basic lute a much lighter color, while the true lute is a bit more brightened and saturated.  I feel confident about opening this PR for review, but of course I'm also open to more suggestions too.

upclose_lutes
lutesininventory
lutes_sidecomparison

akirmse:
This looks pretty good.  You can remove the bgf files from the PR, though.  Those are derived assets, and I'll build them locally before adding them to the next update.

neiltaco:
Thanks for your feedback!  I've also added two commits.  The first just removes the .bgf files as requested.  The second is a one more redesign of the true lute for this PR.  DavidKlaw reached out to me and we agreed the fine lute and true lute still look too closely similar, especially for colorblind players.  I've added an extra pair of pegs to the true lute sprite to help make its silhouette stand out more from the other lutes.

test0
test1
test3

akirmse:
Sorry, one more thing.  Please add these three new files to the makefile in the resource\graphics directory so that they'll get built.  You'll need one new line per resource, in the "BGFS" section.

PR #1274, opened 2025-10-13
Software rendering object flicker uses time over frames

sicoderealm:
Correction to object flicking to use time over frames

Before:
https://github.com/user-attachments/assets/ec662d4b-c165-4170-9b09-de8851d2ed7e

After:
https://github.com/user-attachments/assets/eeea4582-f73b-40c3-bac2-3f2f2c346f39

### DIFF
--- a/clientd3d/animate.c
+++ b/clientd3d/animate.c
@@ -34,6 +34,7 @@
 
 static int  animation_timer = 0;   // id of animation timer, or 0 if none
 static DWORD timeLastFrame;
+static int flickerTimer = FLICKER_PERIOD;  // Global timer for OF_FLICKERING objects (milliseconds)
 
 #define TIME_FULL_OBJECT_PHASE 1800
 static int phaseStates[] = {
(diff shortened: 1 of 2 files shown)


clientd3d/object.h
```diff
@@ -88,6 +88,7 @@ typedef struct {
    BYTE      normal_translation;  // Palette translation when not moving
    BYTE      secondtranslation;   // Overriding, additional second translation.
    WORD	     bounceTime;
+   DWORD	 flickerTime;        // For time-based flickering
```
akirmse: - This should be an int
- New fields need to go at the end of the structure to avoid crashes in modules 
- The comment should say what this field means, units, etc.
- This field is uninitialized and thus will be filled with random garbage when the object is created (in ExtractObject in server.c).
- All fields need to be copied (in ObjectCopy in object.c)
- Is a per-object field really needed, or would a single global counter be good enough?

clientd3d/animate.c
```diff
@@ -173,8 +173,22 @@ bool AnimateObject(object_node *obj, int dt)
...
    if (OF_FLICKERING == (OF_BOUNCING & obj->flags))
    {
-      obj->lightAdjust = rand() % FLICKER_LEVEL;
-      need_redraw = true;
+      // Initialize timer on first use
+      if (obj->flickerTime == 0)
+         obj->flickerTime = FLICKER_PERIOD;
+      
+      // Check if time to flicker
+      if (obj->flickerTime <= (DWORD)min(dt, 50))
```
akirmse: 50 is hard-coded twice.  Make it a named constant, and add a comment describing how it's used.

clientd3d/animate.c
```diff
@@ -173,8 +173,22 @@ bool AnimateObject(object_node *obj, int dt)
 
    if (OF_FLICKERING == (OF_BOUNCING & obj->flags))
    {
-      obj->lightAdjust = rand() % FLICKER_LEVEL;
-      need_redraw = true;
+      // Initialize timer on first use
+      if (obj->flickerTime == 0)
```
akirmse: This will never trigger, since flickerTime is never initialized.

clientd3d/object.h
sicoderealm: Agreed, ditching the field and going with the global counter.

clientd3d/animate.c
sicoderealm: With the new approach this can just be removed, the min(dt, 50) was used to cap the delta time to prevent huge jumps (e.g., when the game is paused or lags) but this is no longer required.

PR #1276, opened 2025-10-25
Reimplement Pacal's offer refusal lines

neiltaco:
This PR uses the custom dialog lines for Pacal that were written before but never used.  Currently Pacal does not explain why he refuses certain offers, which can confuse players.  These lines help explain the following: if his inventory is full, if he has too much of a stackable item, or if he can't accept non-stackable duplicates.

Pacal also had his own general response to refusing items he cannot buy (like shillings for example), so this PR also adds 'vrBuyer_unwanted' to monster.kod to allow him to use his own version of the "I'm not interested." dialogue.  So far only Pacal is using this variable, but other NPCs can use it to replace the default response.


akirmse:
Something is messed up with your git setup, as this change also includes your previous change.  The easiest way to do things is to first sync your repo to master. then make a new branch for each feature you're working on, then push the branch up to github and create a PR from it there.

PR #1277, opened 2025-10-26
Reimplement Pacal's offer refusal lines

neiltaco:
Reupload of #1276 so only changes from this PR are included!

This PR uses the custom dialog lines for Pacal that were written before but never used. Currently Pacal does not explain why he refuses certain offers, which can confuse players. These lines help explain the following: if his inventory is full, if he has too much of a stackable item, or if he can't accept non-stackable duplicates.

Pacal also had his own general response to refusing items he cannot buy (like shillings for example), so this PR also adds 'vrBuyer_unwanted' to monster.kod to allow him to use his own version of the "I'm not interested." dialogue. So far only Pacal is using this variable, but other NPCs can use it to replace the default response.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster.kod
@@ -4700,7 +4700,8 @@ messages:
 
    ObjectDesired(what=$, who=$)
    "This is set in individual buyers.  It allows them to pick and choose"
-   "what they want to buy."
+   "what they want to buy.  The 'who' parameter also allows buyers to"
+   "use SayToOne to tell a player directly why they refused an offer."
    {
       return TRUE;
    }
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqSmith.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqSmith.kod
@@ -45,7 +45,7 @@ properties:
    
 messages:
 
-   ObjectDesired(what=$)
+   ObjectDesired(what=$,who=$)
    {
       if Send(what,@CanBeGiventoNPC) and 
          (Send(Self,@IsObjectWearable,#what=what) OR
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqmerch.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqmerch.kod
@@ -110,7 +110,7 @@ messages:
       return;
    }
 
-   ObjectDesired(what=$)
+   ObjectDesired(what=$,who=$)
    {
       if isClass(what,&NumberItem) and
          send(what,@GetNumber) > 25
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/crnthtwn/cngrocer.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/crnthtwn/cngrocer.kod
@@ -81,7 +81,7 @@ messages:
       return;
    }
 
-   ObjectDesired(what = $)
+   ObjectDesired(what = $, who = $)
    {
       if Send(what, @CanBeGiventoNPC) AND
          (Send(self, @IsObjectReagent, #what=what)
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzapoth.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/hazartwn/hzapoth.kod
@@ -52,7 +52,7 @@ messages:
       return;
    }
 
-   ObjectDesired(what=$)
+   ObjectDesired(what=$,who=$)
    {
       if Send(Self,@IsObjectReagent,#what=what) or send(self,@IsObjectGem,#what=what)
          { return True;}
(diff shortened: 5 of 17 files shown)


kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcshopk.kod
```diff
@@ -216,18 +218,30 @@ messages:
 
       if length(first(plFor_Sale)) > MAX_FORSALE
       {
-         return FALSE;
+         Post(poOwner,@SomeoneSaid,#what=self,#string=JunkMan_Full_rsc,#type=SAY_RESOURCE);
```
akirmse: Any reason why the default in Monster sends the line to only the seller via SayToOne, but this NPC posts to the whole room via SomeoneSaid?

neiltaco: Ah, that's because I followed Izzio's example where he does say his lines out loud.  That said, Pacal currently tells the seller directly the default "not interested" line.  I can change his lines to @SayToOne to keep it consistent.

neiltaco:
Currently ObjectDesired() passes in the item being checked but not the customer, so it wasn't as easy to get an NPC to use SayToOne() to the customer.  I did try using Send(what, @GetOwner) but this won't work with stackable items since NumberItem has no owner.

This latest commit added the #who parameter to ObjectDesired() in monster.kod, which can give buying NPCs reference to the selling player.  This lets Pacal tell the player directly why he refused an offer instead of saying it out loud.  The extra parameter can also allow other NPCs to use custom dialogue to explain in private why they refused an offer.

Also I changed Pacal’s dialogue for better clarification about offer refusals, and to match his lines closer to his own way of talking.

kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcshopk.kod
```diff
@@ -33,8 +33,10 @@ resources:
 
    JunkMan_AlreadyHaveOne_rsc = "I've already got one of them to unload, I don't need another right now."
    JunkMan_NumberItem_rsc = "Sorry, my friend, trinkets like those are things I `Bwon't`n take."
-   JunkMan_Full_rsc = "I wish I could take that off of your hands for you, but mine are all full too!"  
-   JunkMan_Not_Interested = "I thank ye, but I am not interested."
+   JunkMan_Full_rsc = "I wish I could take that off of your hands for you, but my inventory is all full!"
```
akirmse: "Inventory" sounds rather technical and formal.  I like the old line better.

kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcshopk.kod
```diff
@@ -216,18 +220,30 @@ messages:
 
       if length(first(plFor_Sale)) > MAX_FORSALE
       {
-         return FALSE;
+         Post(self,@SayToOne,#target=who,#message_rsc=JunkMan_Full_rsc);
```
akirmse: Prefer Send to Post.  If Post is really necessary, document why.

kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcshopk.kod
```diff
@@ -216,18 +220,30 @@ messages:
...
             if IsClass(i,&NumberItem)
-               AND (send(what,@GetNumber) + send(i,@GetNumber)) < NUMBER_ITEM_MAX
             {
-               return TRUE;
+			   % Only accept stackable items if we don't have too many of them.
+               if (send(what,@GetNumber) + send(i,@GetNumber)) < NUMBER_ITEM_MAX
+			   {
+			      return TRUE;
+			   }
+			   else
```
akirmse: You don't need "else" here, since the other branch of the "if" always returns.

kod/object/active/holder/nomoveon/battler/monster.kod
```diff
@@ -4697,7 +4698,7 @@ messages:
       return viAttributes;
    }
 
-   ObjectDesired(what=$)
+   ObjectDesired(what=$, who=$)
```
akirmse, line 4701: You mentioned trying to get the object's owner from <span>@</span>GetObject, which I didn't understand.  Did you mean <span>@</span>GetOwner?  NumberItems definitely have owners.  Or did you mean that this particular NumberItem doesn't have its owner set, perhaps because it was newly created to give to the NPC?

If <span>@</span>GetOwner works, that's obviously preferred, otherwise adding this "who" parameter is fine, but you should document what it means in the function comment here, and also add it to every signature of ObjectDesired in the codebase.

neiltaco: Oops! I meant to type @GetOwner earlier.  And yeah I also preferred @GetOwner and it worked for everything but the 'too many NumberItem' line because @GetOwner would return null because of the reason you described.  I think adding the 'who' parameter is a way of solving this, even if Pacal is the only one using it for now.

kod/object/active/holder/nomoveon/battler/monster/towns/kocatwn/kcshopk.kod
neiltaco: No need for Post since it was used when Pacal was using SomeoneSaid.

PR #1278, opened 2025-11-01
Shortened archery messages

neiltaco:
This PR will condense the arrow shooting message to one line.  Currently, players using a bow gets two messages when firing arrows:
"You reach for your nerudite arrows, and nock one back for a shot."
"Your arrow pierces the giant rat."

I think even two messages can quickly get excessive as the player fire off more arrows during a fight, and especially so while improving archery.  I suggest it could be condensed down to just one message:  "Your nerudite arrow pierces the giant rat."  
- The 'nocking arrow' message now only displays when auto-equipping a quiver when the user didn't had one equipped.  Maybe it should also show when manually equipping a quiver?
- The name of the arrow is shown in the attack hit message.  Right now this is partially working since the ammo name displayed is plural rather than singular.

arrowex

### DIFF
--- a/kod/object/item/passitem/numbitem/ammo.kod
+++ b/kod/object/item/passitem/numbitem/ammo.kod
@@ -36,6 +36,10 @@ classvars:
 
    viGround_group = 2
    viInventory_group = 1   
+   
+   % Since ammunition is mostly referred to as plural, we have the singular
+   % version for the occasions we need it.
+   vrAmmo_singular = "ammo"
 
 properties:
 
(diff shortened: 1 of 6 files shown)


neiltaco:
Earlier I did try to make arrows all singular by default and have them use vrName_plural, but it seemed to have cause problems with DM spawning arrows and getting them to display their name in their description.

So instead the singular version of the arrows' names were added in, which also lets the description properly display its name if examining just one arrow, instead of always the plural form.  I shared a video both here and also on the M59 discord server to see if players prefer the original version or the new one.

https://github.com/user-attachments/assets/7dc0d395-4014-4821-93da-d152a77875cf

akirmse:
I guess I'm on the fence on this one.  It's a small gain for some extra complexity, and it digs our localization hole significantly deeper if we ever decide to translate the game text.  In English, there are different forms for 1 item vs. 2+ items, but that's not generally true of all languages.  Whenever we deal with plurals in code like this it makes the problem worse.

neiltaco:
I haven't thought of the language and localization part, so that's a fair point and I'll keep that in mind next time.

So far at least one player on the Discord server said he was split on the idea too.  I think the general consensus so far is that the second archery line isn't really a problem during normal PvP combat, so I can see this PR being closed for now.

PR #1279, opened 2025-11-10
Re-add trance spell SFX

neiltaco:
This PR allows Distill, Illusionary Form, and Morph to play their cast sounds when saying a chosen spell/creature during their trance time.  Currently, Distill does not play its SFX at all, and both Illusionary Form and Morph only play their SFX when completing the full casting time.

https://github.com/user-attachments/assets/dc98beaf-5425-428b-941f-8e517fa19e64


neiltaco:
Extra quick video demonstrating distill:

https://github.com/user-attachments/assets/e0d2608e-c3a6-48af-bb1e-faea3c557665

akirmse:
This seems fine---the only question is whether to play the sound just for the caster, or for everyone in the room.  Any opinions?

neiltaco:
Yeah, it is possible to spam the cast sound, especially when someone is improving the spell.  That said, I believe the potential problem might resolve itself since it won't be as spammable because of a few factors.  The caster has to say a spell name out loud, which add some delay in between casts.  The spell itself is also expensive with the cost of one ruby and the chosen spell's reagents.  And from what I personally seen in-game, Distill is also rarely seen cast in public spaces.

That said, it plays for 5.5 seconds which is much longer than most spell school default cast sounds.  In my opinion, I personally do like the sound playing for everyone to make it consistent with other spells, but making the sound play privately can be an option too!  I'd love to get your opinion and other peoples' opinions, too.  :)

sicoderealm:
Cool to find and use the existing sound! Niceone.
If I the sound is played to everyone in the room, how does it feel when multiple people are casting it at the same time?

neiltaco:
I imagine in its original state, the sound would really get annoying quick if even two people were casting it.

From what was gathered in the last couple of days, people generally prefer the casting sound playing out loud but also the sound shouldn't be too long if it is spammable.  I made a new .wav file based on jdistill.wav that only play the first couple of seconds.

https://github.com/user-attachments/assets/d25d6939-a592-4e17-b025-9ad0323bfc05

PR #1282, opened 2025-11-22
Implement rain and snow for hardware and software rendering

neiltaco:
This PR adds rain and snow weather for both software and hardware rendering, and incorporates some code from #379.  The focus is on adding the visual effects of rain/snow for cosmetic purposes, which means this PR will make no changes to Faren/Qor bonuses for rainy weather or cloudy skies.  This PR is also dependent on #1284, which adds the DM weather commands and automated weather system, and the packet data for updating room weather for the client.

Once the weather effects are added, there may be additions such as rain/snow sound effects, or room texture swaps.  These will likely be added in a separate future PR.

To Do
- [x] Rain/Snow enabled and only either one of them can active at a time.
- [x] Rain/Snow particles check the BSP tree if they hit a floor or ceiling, and clear away if they do.
- [x] D3DParticleEmitterInit() broken up into smaller functions for better readability.
- [x] Add weather_effect to room_type struct.
- [x] Read and apply weather_effect data from packet data added in #1284.

- [ ] Test out performance, with and without weather particles.
- [ ] Add .png art assets for cloudy skies in hardware rendering.
- [ ] Hide the sun and moon background overlays during cloudy/rainy/snowy weather.
- [ ] Finalize appearance of rain/snow particles.


clientd3d/d3dparticle.c
```diff
@@ -15,52 +15,70 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
-							unsigned char r, unsigned char a, int energy, int timerBase,
-							float rotX, float rotY, float rotZ, bool bRandomize, int randomPos,
-							int randomRot)
+emitter* D3DParticleEmitterInit(particle_system *pParticleSystem, int energy, int timerBase)
 {
 	emitter	*pEmitter = NULL;
 
 	if (pParticleSystem == NULL)
-		return;
+		return 0;
```
akirmse: nullptr, not 0

clientd3d/d3dparticle.c
```diff
@@ -15,52 +15,70 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
 
 	pEmitter = (emitter *)SafeMalloc(sizeof(emitter));
 
 	if (pEmitter == NULL)
-		return;
+		return 0;
```
akirmse: nullptr

clientd3d/d3dparticle.c
```diff
@@ -15,52 +15,70 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
+	pEmitter->rotation.z = rotZ;	
+}
+void D3DParticleEmitterSetRandom(emitter *pEmitter, bool bRandomize, int randomPos, 
+							float randomRot)
+{
+	pEmitter->bRandomize = bRandomize;
+	pEmitter->randomPos = randomPos;
+	pEmitter->randomRot = randomRot;
+}
+void D3DParticleEmitterSetBGRA(emitter *pEmitter, unsigned char b, unsigned char g,
```
akirmse: How about taking a const reference to a custom_bgra instead?

clientd3d/d3dparticle.c
```diff
@@ -15,52 +15,70 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
 	pEmitter->bgra.b = b;
 	pEmitter->bgra.g = g;
 	pEmitter->bgra.r = r;
-	pEmitter->bgra.a = a;
-	pEmitter->randomPos = randomPos;
-	pEmitter->randomRot = randomRot;
-
+	pEmitter->bgra.a = a;	
+}
+void D3DParticleEmitterFinalize(particle_system *pParticleSystem, emitter *pEmitter)
```
akirmse: Rename--this doesn't finalize, but rather adds it to a list.

clientd3d/d3dparticle.h
```diff
@@ -14,10 +14,15 @@
 #define SANDSTORM_B		6
 #define SANDSTORM_A		255
 
-#define RAIN_R			128
-#define RAIN_G			128
-#define RAIN_B			128
-#define RAIN_A			128
+#define RAIN_R			175
```
akirmse: Looks like you can make this into a single custom_bgra constant instead, something like:

const custom_bgra rain_color = {175, 228, 249, 100};

And actually also move it to the .c file and make it static if possible.

clientd3d/rops.c
```diff
@@ -158,12 +158,94 @@ void RainDib(BYTE* pabyBits, int width, int height, int drops, int heading, int
...
+
+	if (!pabyBits)
+		return;
+
+	// compute the droplet length and width
+	alllength = 3;
+	alldropwidth = 2;
+
+	// compute the droplet run and blow (less run is more slant)
+	{
```
akirmse: This brace isn't necessary.

kod/include/blakston.khd
```diff
@@ -1138,6 +1138,10 @@
    WEATHER_DAWN = 6
    WEATHER_DAWN_STORMY = 7
 
+   WEATHER_CLEAR = 1
```
akirmse: It's probably a good idea to split off the Blakod change into a separate PR, since you're going to have to add the hardware implementation to the client PR, which will make it large.

clientd3d/d3drender_fx.c
```diff
@@ -23,6 +25,8 @@ static void SandstormInit(void);
 void D3DFxInit()
 {
 	SandstormInit();
+	//RainInit();
```
akirmse: Don't check in disabled code.

clientd3d/d3drender_fx.c
```diff
@@ -61,186 +81,211 @@ void SandstormInit(void)
 	static const int EMITTER_ENERGY = 40;
 	static const int EMITTER_HEIGHT = 0;
 
+	// These variables for the same for all sandstorm emitters.
```
akirmse: Fix grammar.

clientd3d/d3drender_fx.c
```diff
@@ -61,186 +81,211 @@ void SandstormInit(void)
 	static const int EMITTER_ENERGY = 40;
 	static const int EMITTER_HEIGHT = 0;
 
+	// These variables for the same for all sandstorm emitters.	
+	int sandTimer = 1;
+	int sandRandPos = 1024;
+	int sandRandRot = 2;
+
 	D3DParticleSystemReset(&particleSystem);
+	emitter* newEmitter = new emitter;
```
akirmse: Doesn't this allocated emitter immediately leak?

kod/include/blakston.khd
neiltaco: Noted!  I wasn't aware of how much would change from adding more particle effects, but I do know changing room.kod will require rebuilding all the rooms.

I hope I understand this correctly.  I think I need to do a separate PR first that only include changes within the kod folder (as done in #1284).  Then have this branch both dependent on #1284 and include only the changes in the clientd3d folder.

clientd3d/d3dparticle.c
```diff
@@ -97,6 +115,16 @@ void D3DParticleSystemUpdate(particle_system *pParticleSystem, d3d_render_pool_n
 			}
 			else
 			{
+				PDIB pdibCeiling = NULL;
```
akirmse: This NULL doesn't do anything; combine with the next line.

(Also prefer nullptr to NULL, as NULL is an ugly macro.)

clientd3d/d3dparticle.c
```diff
@@ -15,45 +15,63 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
+}
+void D3DParticleEmitterSetRandom(emitter *pEmitter, bool bRandomize, int randomPos, 
+							float randomRot)
+{
+	pEmitter->bRandomize = bRandomize;
 	pEmitter->randomPos = randomPos;
 	pEmitter->randomRot = randomRot;
-
+}
+void D3DParticleEmitterSetBGRA(emitter *pEmitter, const struct custom_bgra* newBGRA)
```
akirmse: Pass by reference and you don't need struct:
```
const custom_bgra &newBGRA
```

clientd3d/d3dparticle.c
```diff
@@ -124,6 +152,14 @@ void D3DParticleSystemUpdate(particle_system *pParticleSystem, d3d_render_pool_n
 				pParticle->pos.y += pParticle->velocity.y;
 				pParticle->pos.z += pParticle->velocity.z;
 
+				// If we don't allow this type of particle to survive inside the ground, destroy it.
+				if ((pEmitter->groundDestroy) && (pParticle->pos.z < GetPointFloor(pParticle->pos.x, pParticle->pos.y)))
```
akirmse: Hmm.. this walks the room's BSP tree for every particle every frame?  Twice actually, since you also call GetPointCeilingTexture above.  That sounds potentially very expensive.

At a minimum I think you could call BSPFindLeafByPoint directly to avoid walking the tree twice.

clientd3d/d3dparticle.h
```diff
@@ -8,16 +8,7 @@
...
-#define SANDSTORM_R		226
-#define SANDSTORM_G		153
-#define SANDSTORM_B		6
-#define SANDSTORM_A		255
-
-#define RAIN_R			128
-#define RAIN_G			128
-#define RAIN_B			128
-#define RAIN_A			128
+#define MAX_PARTICLES	512 // was 128
```
akirmse: Any reason to preserve the historical value of constants?

Since this makes particle systems 4x more expensive, you'll need to do some performance testing.

Also make this a const int rather than a macro.

clientd3d/d3drender_fx.c
```diff
@@ -57,190 +99,356 @@ void D3DRenderParticles(const ParticleSystemStructure& pss)
 */
 void SandstormInit(void)
 {
-	static const int EMITTER_RADIUS = 12;
-	static const int EMITTER_ENERGY = 40;
-	static const int EMITTER_HEIGHT = 0;
-
-	D3DParticleSystemReset(&particleSystem);
+	static int sandEmitterRadius = 12;
```
akirmse: Can these all be const?

clientd3d/d3drender_fx.c
```diff
@@ -57,190 +99,356 @@ void D3DRenderParticles(const ParticleSystemStructure& pss)
...
+}
+
+
+/**
+* Initializes the snow particle emitters with predefined positions, energy, and colors for the 
+* particle system.
+*/
+void SnowInit(void)
+{
+	//static int snowEmitterRadius = 16384;
```
akirmse: Don't check in deleted code.

clientd3d/rops.c
```diff
@@ -77,48 +77,46 @@ void RainDib(BYTE* pabyBits, int width, int height, int drops, int heading, int
...
+	else
+	{
+		// in my face from right to left
+		allblow = -1;
+		delta = (windheading+NUMDEGREES/2) - heading;
+	}
 
-/*
-		delta = delta * alllength * windstrength / (NUMDEGREES/4) / 100;
+	/*
```
akirmse: Remove deleted code.

clientd3d/d3dparticle.c
neiltaco: Yeah I can change it to get the leaf only once per frame, and not for sandstorm.  For now weather effects like rain and snow are supposed to clear on hitting a surface and they don't run simultaneously, so that should help with performance.

neiltaco:
Before the most recent PR, I believe 512 weather particles that call the BSP tree twice meant it might call up to 1024 times per frame.  It seems that a max of 128 particles looks good enough for rain/snow.  Now that weather particles only call the BSP tree once, it should at least be up to 128 BSP tree calls per frame.

One of my main concerns right now is that the software renderer's weather effects seems to be drawn directly on the screen instead of being particles in 3D space, which complicates how to render it if someone were to be standing under a ceiling and looking directly outside.  It would look like it isn't raining/snowing outdoors until the player steps directly out of the shade, and suddenly gets rain/snow dumped on them.

Another feature that would be nice too is to change the skybox to gray and cloudy when it is raining or snowing, which may require changes both to this PR and #1284.  I've pushed this recent commit so I can show what is done here so far.

neiltaco:
I have to close this PR since it's far behind from upstream/master, and attempts to update it kept showing hundreds of changed files not part of this PR.  It has been reopened in #1304.

PR #1283, opened 2025-11-24
fix: prevent extra spacing from blank attribute descriptions

adrienlaws:
## What
- fixed item description spacing issues where invisible attributes caused unwanted gaps
  - Refactored `CreateDesc()` in `item.kod` to improve maintainability by decomposing into focused helper methods

## Why
- PR #1267 added proper paragraph spacing between description sections but missed 2 edge cases
- Invisible attributes `IA_PKPOINTER` and `IA_CORPSEPOINTER` were causing extra spacing in item descriptions
- The monolithic `CreateDesc()` function was complex and hard to maintain

## How
- Simplified the attribute visibility detection logic to check `Send(oItemAtt,@GetDesc) <> resource_blank`
- Split `CreateDesc()` into helper functions:
  - `HasVisibleAttributes()` - checks for displayable attributes
  - `ProcessAttributeList()` - handles attribute processing logic
  - `AppendConditionDescription()` - formats condition information

### Flowchart Logic
``` mermaid
flowchart TD
    A[CreateDesc called] --> B[ClearTempString]
    
    %% BASE DESCRIPTION SECTION
    subgraph BASE [" BASE DESCRIPTION "]
        B --> C[DoBaseDesc - Add base item description]
    end
    
    %% ATTRIBUTES SECTION
    subgraph ATTRS [" ITEM ATTRIBUTES "]
        C --> D[AppendAttributeDescriptions]
        D --> E[HasVisibleAttributes?]
        
        E --> F{Any visible attributes?}
        F -->|No| SKIP1[Skip attributes section]
        F -->|Yes| G[Add section separator \\n\\n]
        
        G --> H[ProcessAttributeList]
        
        %% ATTRIBUTE PROCESSING SUBLOGIC
        subgraph ATTRLOGIC [" Attribute Processing Logic "]
            H --> J[Loop through plItem_Attributes]
            
            J --> K{Is IA_MISDIRECTION?}
            K -->|Yes| L[Add fake misdirection desc]
            K -->|No| M{Is identified or ShowAll?}
            
            M -->|No| N{Already shown generic?}
            N -->|No| O[Show itematt_generic once]
            N -->|Yes| P[Skip this attribute]
            
            M -->|Yes| Q{GetDesc = resource_blank?}
            Q -->|Yes| R[Skip invisible attribute<br/>IA_PKPOINTER, IA_CORPSEPOINTER]
            Q -->|No| S[Add spacing if not first attribute]
            
            S --> T[AppendDesc - Add attribute description]
        end
        
        L --> U[Continue to next attribute]
        O --> U
        P --> U  
        R --> U
        T --> U
        
        U --> V{More attributes?}
        V -->|Yes| J
        V -->|No| SKIP1
    end
    
    %% CONDITION SECTION
    subgraph COND [" CONDITION INFORMATION "]
        SKIP1 --> W[AppendConditionDescription]
        W --> X{vbShow_condition AND<br/>piHits_init > 0?}
        X -->|Yes| Y[Add section separator \\n\\n<br/>Item has durability system]
        X -->|No| Z[Skip condition section<br/>No durability or not configured]
        
        Y --> AA[AppendDesc - Add condition text<br/>'is in excellent condition']
        Z --> BB[Return complete description]
    end
    
    AA --> BB
```

### DIFF
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -237,7 +237,7 @@ messages:
    "Adds all visible attribute descriptions to the temp string with formatting. "
    "When bShowAll is TRUE, all attributes including misdirection are revealed."
    {
-      local bAlreadyBlind, bFirstAttribute, bHasVisibleAttributes;
+      local bHasVisibleAttributes;
       
       bHasVisibleAttributes = Send(self,@HasVisibleAttributes,#bShowAll=bShowAll);
       if bHasVisibleAttributes


akirmse:
Does disclosing the killer's name reveal new information that wasn't previously available, or is this already on the corpse or something like that?

adrienlaws:
I narrowed the scope of this PR to focus only on the formatting issues from invisible attributes (`IA_CORPSEPOINTER` and `IA_PKPOINTER`).  It was getting really messy trying to catch all the edge cases in one PR and also fix the existing formatting issues. 

The formatting issues appear addressed now, but I'll test a little bit more.

In terms of the original question - I don't think any new information is revealed but it caused some implementation issues.  The corpse itself adds a 25 second timer so others can't grab loot from the corpse, but the `IA_CORSEPOINTER` attribute lasts longer than 25 seconds.  So it would hold the message that the item could only be picked up by X when that was no longer true.  I started running down a rabbit hole of managing the attribute description then ran into more problems with `IA_PKPOINTER` with other stuff.  

Anyway, it'll be easier to do in several PRs I think (if we even want that description message).  I mainly wanted to just fix the formatting bug I introduced accidentally.

adrienlaws:
This PR is ready for review, when possible.

kod/object/item.kod
```diff
@@ -229,15 +228,41 @@ messages:
...
 
-      % Add line break before attributes if any exist
+      % Check if any attributes will display text. IA_MISDIRECTION always shows
+      % (creates fake desc), others check HasVisibleDesc.
+      bHasVisibleAttributes = FALSE;
       if plItem_Attributes <> $
+      {
+         for i in plItem_Attributes
+         {
+            iNum = Send(self,@GetNumFromCompound,#compound=First(i));
```
akirmse: I'm not super sure what's going on here, but it does feel strange that one of the most basic classes in the game (Item) has so much of its code dedicated to formatting spaces and newlines.  Is it possible to move some of the code from this function into ItemAttribute?

This function is already long and hard to understand.  Before we make it longer and harder to understand, do you have any recommendations on the ideally "right" way to do this that would be cleaner?

adrienlaws: I refactored `CreateDesc()` to address how overly complex it is by breaking it into separate functions:

- `CreateDesc()` is now very short and leverages the following:
- 1/ `HasVisibleAttributes()` - checks visibility of attributes (some like corsepointer are intentionally invisible)
- 2/  `ProcessAttributeList()` - handles attribute logic (ignores if N/A)
- 3/  `AppendConditionDescription()` - formats item condition stuff (ignores if N/A)

I've added a flowchart to the PR description showing the new structure.

From what I can tell it all works now.  I've tested the invisible attributes corpsepointer and pkpointer by killing monsters and a test character to see if the previous formatting issues are fixed.  It seems they are.

kod/object/item.kod
```diff
@@ -221,35 +220,79 @@ messages:
...
    CreateDesc(bShowAll=FALSE)
-   "Returns the item's full description text by combining the base description "
-   "with attribute descriptions and condition information. When bShowAll is FALSE, "
-   "unidentified attributes show a generic message and misdirection attributes "
-   "display their fake descriptions alongside real attributes. When bShowAll is "
-   "TRUE, all attributes including misdirection are revealed. The description "
-   "uses proper formatting with line breaks separating the base description, "
-   "attributes section, and condition information."
+   "Returns the item's full description text by combining the base description, "
+   "attribute descriptions, and condition information."
```
akirmse: Looks like the documentation for bShowAll got removed.

kod/object/item.kod
```diff
@@ -221,35 +220,79 @@ messages:
...
+      
+      bHasVisibleAttributes = Send(self,@HasVisibleAttributes,#bShowAll=bShowAll);
+      if bHasVisibleAttributes
       {
          AppendTempString("\n\n");
       }
-
+      
+      bAlreadyBlind = FALSE;
+      bFirstAttribute = TRUE;
```
akirmse: Any reason you need these variables, rather than just using TRUE/FALSE?

kod/object/item.kod
```diff
@@ -221,35 +220,79 @@ messages:
...
+         {
+            return TRUE;
+         }
+      }
+      
+      return FALSE;
+   }
+   
+   ProcessAttributeList(bShowAll=FALSE, bAlreadyBlind=FALSE, bFirstAttribute=TRUE)
+   "Processes the attribute list and appends descriptions."
```
akirmse: Appends them to what?

kod/object/item.kod
```diff
@@ -221,35 +220,79 @@ messages:
...
-         % Misdirection: show fake attribute even if real one is hidden
+         
+         % IA_MISDIRECTION always shows (creates fake desc)
+         if iNum = IA_MISDIRECTION
+         {
+            return TRUE;
+         }
+         
+         oItemAtt = Send(SYS,@FindItemAttByNum,#num=iNum);
+         if oItemAtt <> $ AND Send(oItemAtt,@GetDesc) <> resource_blank
```
akirmse, line 276: Does it work to compare against "" instead?  i.e. if someone later makes a resource_blank2 off in some distant class, this comparison will fail.

adrienlaws: I tested using "" like this:
`if oItemAtt <> $ AND Send(oItemAtt,@GetDesc) <> ""`
But it didn't work and I got extra spacing as in the following screenshot:
image

Both `IA_CORPSEPOINTER `and `IA_PKPOINTER `use `resource_blank`. If I change their vrDesc to = "" then I get debug messages in the description because of how Blakod handles literal comparisons.

I think using `resource_blank` is probably the safest way to go since [item.kod](https://github.com/Meridian59/Meridian59/blob/8220250b092d6aea4fbd17dbf5311e498c8a4459/kod/object/item.kod#L25) defined `resource_blank` over a decade ago.  I didn't change it.  It seems the invisible attributes are designed to work with `resource_blank`, not direct string literals.

I realize this means if anyone adds an invisible attribute they will need to use `resource_blank`  I'm not sure how can I make this better, but it looks like it works as is.

kod/object/item.kod
adrienlaws: Re-added documentation

kod/object/item.kod
adrienlaws: Removed variable as suggested

kod/object/item.kod
adrienlaws: Updated function documentation to be more specific

PR #1284, opened 2025-11-24
Add automated weather system and DM weather commands

neiltaco:
This is a split PR from #1282 that added DM commands to apply weather effects.  It has recently been updated to include an automated weather system.

- Adds DM commands to enable/disable weather effects in a room, and also for changing room climates (which can adjust local weather).  It also adds DM commands change server weather and seasons.  For now, seasons aren't fully implemented so it only affects what kind of stormy weather is sent every room.
- Implements automated weather system that checks every hourly intervals in a day if the weather should shift.  It uses a simple state system that adjusts the percentage chances to shift weathers based on the current weather.
- Adds software rendering skyboxes to show cloudy weather.
- Weather was originally used to set the day phase for the skybox and didn't involve any weather.  This PR splits it into two different parts: day phase and weather.
- When updating users entering a room, the packet sent to them now includes the room's weather effect.  #1282 requires this so that the player's client can extract this data and update the weather effect playing on the client.


kod/object/active/holder/room.kod
```diff
@@ -1591,6 +1594,44 @@ messages:
...
+   {
+      local i, each_obj;
+	  
+      piWeather_type = what;
+      for i in plActive
+      {
+         each_obj = Send(self,@HolderExtractObject,#data=i);
+         if IsClass(each_obj,&Player)
+         {
+            % Clear out previous weather before applying any new ones.
```
akirmse: Two thoughts on this:

- If the new weather type is the same as the current weather type, we probably shouldn't send anything to the client, to prevent rendering artifacts (for example, causing the particle systems to restart).

- Why send two messages?  There could be an aribtrary network delay between them, so if we change rain -> snow, the user might see rain -> clear -> snow.  The client should handle switching between types; it shouldn't need a clear message.

kod/object/active/holder/room.kod
```diff
@@ -1919,6 +1967,9 @@ messages:
             propagate;
          }
       }
+	  
+	  % Remove weather effects caused by this room.
+	  Send(what,@EffectSendUser,#what=what,#effect=EFFECT_CLEARWEATHER);
```
akirmse: I wonder if we should do this.  For example, if the user goes between two snowy rooms, they might see snow -> clear -> snow.  Another alternative would be for the client to make assumptions about weather when changing rooms.  Maybe give some thought to what the client should be responsible for vs. the server, considering that network delays can be long.

neiltaco: For now I removed Send() for players entering and leaving the room.  It's still being used to update players whenever the weather changes, or if the room's weather effects are toggled on/off by a DM.

I can take a look at #1282 and see if maybe the room_type struct can be updated to include weather effects.  I think it could be a way for the client to get weather effects from the room without having to receive a Send().  Also I'll have to update d3drender_skybox.c anyways so it can properly load the skyboxes for hardware rendering mode.

neiltaco:
Closing this PR to upload later since it's far behind upstream/master and desynced.

Issue #1199, opened 2025-08-27
Graphics: Hardware Renderer: Lighting: Flickering Sectors and Flickerer Objects not working.


jrotunda:
Certain sectors in Meridian are set to flicker, such as the locations of lamps in Raza, Tos, and when game events occur - with the raising of the Vale Node a sector will flicker.

The flickerer class contains various subclasses of atmospheric items like Candles, Braziers, Lamps, Firepits, Orc Torches and Candelabras.

Ornamental Objects, another class, can be made to flicker in the settings of each object - this can be used to draw attention to items, or for obvious ones that should be flickering, like a hanging lantern.

Some monsters are also set to flicker per their individual .kod : spectral mummies and the ghost of Far'nohl.

The flicker effect works for the Software Renderer on all of these objects, but when in the Hardware Renderer, the effect is not visible, instead, the object will be a static source of light, similar to placing a non-moving dynamiclight object.

akirmse:
The colon prefixes are getting out of control.  Please try to use short and relevant titles.

sicoderealm:
@akirmse perhaps we can resolve this now?

PR #1291, opened 2025-12-13
Add hardware rendering flickering light support

sicoderealm:
We add support for flicking lights as seen in the original software renderer.

This change adds flicking hardware lights for all flickerer and flickering types in the game - all regular objects (`Lamp, Brazier, Candelabra, Candle, Firepit, KocatanLamp, KocatanBrazier, TableLamp, SkullBrazier, OrcTorch, SmokeColumn, GlowTree, BrazierSwitch, DynamicLight`) + `torches` and `wall` spells.

**Testing:** 

exploring the world  (such as tos, barloque, inns/adv halls and brax) in various flickering object types  with and without darkness cast. As well as positioning and moving around objects that are flicking to check that flicking is supported in various locations by walls, floors, corners and also that lighting intensity has impacted. Finally checked that software renderer can still support flickering lights and isn't impacted by these changes. Also tested against latest lights found on 101 - will require something like: https://github.com/Meridian59/Meridian59/pull/1349 to proceed correctly on the next patch. 

---

Addresses: https://github.com/Meridian59/Meridian59/issues/1199

**New hardware flickering in action:**

https://github.com/user-attachments/assets/1945983f-3373-4004-ba4c-9b7abe740a66

vs original software rendering flickering:

https://github.com/user-attachments/assets/70d95aac-5ccb-40b4-861f-ff7c9c630a4e

---

**Further examples**

https://github.com/user-attachments/assets/fd30a8aa-1674-4c81-aaba-4376214751ac

https://github.com/user-attachments/assets/63f26d9d-e82e-42e4-9426-44629c1d2283

https://github.com/user-attachments/assets/498515f3-8b57-467c-a0ff-87b2c3fc5f71

https://github.com/user-attachments/assets/29f7ddae-025e-491f-84ef-2e83637da894

https://github.com/user-attachments/assets/6987c5e8-ebdd-4fb6-845b-0d1aee1d4019

https://github.com/user-attachments/assets/998f7a7c-a134-4672-a811-f1af98c67df2

https://github.com/user-attachments/assets/a413a4cb-a131-4183-8154-ce4479ad5659

**Light debug render pass:**
Also added a helper to track where lights are supposed to be in hardware rendering mode.
light-debug

**Software rendering flicking examples for comparison:**

https://github.com/user-attachments/assets/70d95aac-5ccb-40b4-861f-ff7c9c630a4e

https://github.com/user-attachments/assets/c989dda7-34c4-47a2-a59f-64945b398010

https://github.com/user-attachments/assets/601f94e7-7c12-45a4-ae28-8dc0d537f8e7

https://github.com/user-attachments/assets/05d8ffca-9caa-4b0f-a1d4-1e7caf88b077

https://github.com/user-attachments/assets/1af1a8f5-c8bd-43b2-9e24-e16aef40e025

https://github.com/user-attachments/assets/e581e599-fcef-4081-a51a-d250351af093

### DIFF
--- a/clientd3d/client.rc
+++ b/clientd3d/client.rc
@@ -90,6 +90,7 @@ BEGIN
             MENUITEM "&Listar caixas",              ID_FONT_LIST
             MENUITEM "&Títulos",                    ID_FONT_TITLES
             MENUITEM "Emails e &mensagens",         ID_FONT_MAIL
+            MENUITEM "&Anotaciones del mapa",       ID_FONT_ANNOTATIONS
             POPUP "Imprimir mapa"
             BEGIN
                 MENUITEM "Comentários",             ID_OPTIONS_FONT_MAP_TEXT
@@ -202,7 +203,6 @@ BEGIN
     CONTROL         "Mostrar latência",IDS_LATENCY0,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,88,187,77,9
     CONTROL         "Mapa dinâmico",IDC_DRAWMAP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,88,197,79,10
     CONTROL         "Texto colorido",IDC_COLORCODES,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,172,187,72,10
-    CONTROL         "Anotações do mapa",IDC_MAP_ANNOTATIONS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,172,197,71,9
     PUSHBUTTON      "&Configurações",IDC_PROFANESETTINGS,169,225,70,12
 END
 
@@ -649,6 +649,11 @@ BEGIN
     PUSHBUTTON      "+",IDC_MAP_MOD,125,7,14,14
     PUSHBUTTON      "+",IDC_MAPZOOMIN_MOD,125,26,14,14
     PUSHBUTTON      "+",IDC_MAPZOOMOUT_MOD,125,45,14,14
+
+    GROUPBOX        "Opções do Mapa",IDC_STATIC,6,80,245,45,WS_GROUP
+    CONTROL         "Anotações",IDC_MAP_ANNOTATIONS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,91,72,10
+    LTEXT           "Limite de Zoom do Texto",IDC_STATIC,13,109,50,10
+    CONTROL         "",IDC_ANNOTATION_ZOOM_LIMIT,"msctls_trackbar32",TBS_BOTH | TBS_NOTICKS | WS_TABSTOP,69,106,173,15
 END
 
 IDD_MOVEMENT DIALOG 0, 0, 296, 138
@@ -1101,6 +1106,7 @@ BEGIN
             MENUITEM "Inventarfenster",             ID_FONT_LIST
             MENUITEM "Überschriften",               ID_FONT_TITLES
             MENUITEM "Post && Nachrichten",         ID_FONT_MAIL
+            MENUITEM "Kartenbeschriftungen",        ID_FONT_ANNOTATIONS
             POPUP "Karten drucken"
             BEGIN
                 MENUITEM "Kommentare",                  ID_OPTIONS_FONT_MAP_TEXT
@@ -1661,6 +1667,11 @@ BEGIN
     PUSHBUTTON      "+",IDC_MAP_MOD,125,7,14,14
     PUSHBUTTON      "+",IDC_MAPZOOMIN_MOD,125,26,14,14
     PUSHBUTTON      "+",IDC_MAPZOOMOUT_MOD,125,45,14,14
+
+    GROUPBOX        "Kartenoptionen",IDC_STATIC,6,80,245,45,WS_GROUP
+    CONTROL         "Kartennotizen",IDC_MAP_ANNOTATIONS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,91,72,10
+    LTEXT           "Zoomgrenze für Text",IDC_STATIC,13,109,50,10
+    CONTROL         "",IDC_ANNOTATION_ZOOM_LIMIT,"msctls_trackbar32",TBS_BOTH | TBS_NOTICKS | WS_TABSTOP,69,106,173,15
 END
 
 IDD_MOVEMENT DIALOG 0, 0, 296, 138
@@ -2116,6 +2127,7 @@ BEGIN
             MENUITEM "&List boxes",                 ID_FONT_LIST
             MENUITEM "&Titles",                     ID_FONT_TITLES
             MENUITEM "&Mail and news",              ID_FONT_MAIL
+            MENUITEM "Map &Annotations",            ID_FONT_ANNOTATIONS
             POPUP "Map Printing"
             BEGIN
                 MENUITEM "Comments",                    ID_OPTIONS_FONT_MAP_TEXT
@@ -2229,9 +2241,8 @@ BEGIN
     CONTROL         "Show latency meter",IDS_LATENCY0,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,89,176,77,9
     CONTROL         "Show dynamic map",IDC_DRAWMAP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,89,187,79,10
     CONTROL         "Show colored text",IDC_COLORCODES,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,173,176,72,10
-    CONTROL         "Map annotations",IDC_MAP_ANNOTATIONS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,173,187,72,10
+    CONTROL         "Spinning cube",ID_SPINNING_CUBE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,173,187,72,10
     PUSHBUTTON      "&Profanity Options and Policy...",IDC_PROFANESETTINGS,102,208,142,12
-    CONTROL         "Spinning cube",ID_SPINNING_CUBE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,173,199,71,9
 END
 
 IDD_COLOR DIALOG 0, 0, 201, 92
@@ -2795,6 +2806,11 @@ BEGIN
     PUSHBUTTON      "+",IDC_MAP_MOD,118,7,14,14
     PUSHBUTTON      "+",IDC_MAPZOOMIN_MOD,118,26,14,14
     PUSHBUTTON      "+",IDC_MAPZOOMOUT_MOD,118,45,14,14
+
+    GROUPBOX        "Map Options",IDC_STATIC,6,80,245,45,WS_GROUP
+    CONTROL         "Annotations",IDC_MAP_ANNOTATIONS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,91,72,10
+    LTEXT           "Text Zoom Limit",IDC_STATIC,13,109,50,10
+    CONTROL         "",IDC_ANNOTATION_ZOOM_LIMIT,"msctls_trackbar32",TBS_BOTH | TBS_NOTICKS | WS_TABSTOP,69,106,173,15
 END
 
 /////////////////////////////////////////////////////////////////////////////
--- a/clientd3d/config.c
+++ b/clientd3d/config.c
@@ -78,6 +78,7 @@ static char INISpinningCube[]= "SpinningCube";
 static char INIHaloColor[]   = "HaloColor";
 static char INIColorCodes[]  = "ColorCodes";
 static char INIMapAnnotations[] = "MapAnnotations";
+static char INIMapTextZoomLimit[] = "MapTextZoomLimit";
 
 static char window_section[] = "Window";         /* Section in INI file for window info */
 static char INILeft[]        = "NormalLeft";
@@ -251,6 +252,7 @@ void ConfigLoad(void)
    config.halocolor    = GetConfigInt(interface_section, INIHaloColor, 0, ini_file);
    config.colorcodes   = GetConfigInt(interface_section, INIColorCodes, true, ini_file);
    config.map_annotations = GetConfigInt(interface_section, INIMapAnnotations, true, ini_file);
+   config.map_text_zoom_limit = GetConfigInt(interface_section, INIMapTextZoomLimit, 50, ini_file);
 
    config.lastPasswordChange = GetConfigInt(misc_section, INILastPass, 0, ini_file);
 
@@ -373,6 +375,7 @@ void ConfigSave(void)
    WriteConfigInt(interface_section, INIHaloColor, config.halocolor, ini_file);
    WriteConfigInt(interface_section, INIColorCodes, config.colorcodes, ini_file);
    WriteConfigInt(interface_section, INIMapAnnotations, config.map_annotations, ini_file);
+   WriteConfigInt(interface_section, INIMapTextZoomLimit, config.map_text_zoom_limit, ini_file);
    
    WriteConfigInt(misc_section, INILastPass, config.lastPasswordChange, ini_file);
 
--- a/clientd3d/config.h
+++ b/clientd3d/config.h
@@ -22,6 +22,8 @@
 
 #define CONFIG_MAX_VOLUME 100  // Max value of sound / music volume settings
 
+static const int CONFIG_MAX_TEXT_ZOOM_LIMIT = 100;  // Max value of map text zoom limit
+
 // Communication settings
 typedef struct {
    WORD  timeout;                 /* # of seconds to wait before redialing */
@@ -106,6 +108,7 @@ typedef struct {
    int	mouselookYScale;
 
    bool map_annotations;       /* Display annotations on map? */
+   int map_text_zoom_limit;    // 0 - 100
 
    int sound_volume;           // 0 - 100
    int music_volume;           // 0 - 100
--- a/clientd3d/font.c
+++ b/clientd3d/font.c
@@ -33,6 +33,7 @@ static char fontinfo[][MAX_FONTNAME] = {
 { "24,0,0,0,700,0,0,0,0,0,0,0,0,Arial" },          /* FONT_MAP_TITLE */
 { "12,0,0,0,400,0,0,0,0,0,0,0,0,Arial" },           /* FONT_MAP_LABEL */
 { "12,0,0,0,400,0,0,0,0,0,0,0,0,Arial" },           /* FONT_MAP_TEXT */
+{ "-13,0,0,0,700,0,0,0,0,3,2,5,0,Arial" },          /* FONT_MAP_ANNOTATIONS */
 };
 
 static char font_section[] = "Fonts";  /* Section for fonts in INI file */
--- a/clientd3d/font.h
+++ b/clientd3d/font.h
@@ -27,6 +27,7 @@ enum {
    FONT_MAP_TITLE,
    FONT_MAP_LABEL,
    FONT_MAP_TEXT,
+   FONT_MAP_ANNOTATIONS,    /* For annotations on the map & mini-map */
    MAXFONTS,
 };
 
--- a/clientd3d/map.c
+++ b/clientd3d/map.c
@@ -78,6 +78,7 @@ static BOOL fMapCacheValid = FALSE;
 static struct map_wall_cache_t *pMapWalls = NULL;
 static float mapCacheScale = 0.0;
 static int mapNumCacheWalls = 0;
+static constexpr float MAX_ANNOTATION_TEXT_ZOOM = 5; // Unit-less number which helps define when we stop displaying annotation text
 
 /* local function prototypes */
 static void MapDrawMiniMapWalls(HDC hdc, int x, int y, room_type *room);
@@ -85,7 +86,7 @@ static void MapDrawWall(HDC hdc, int x, int y, float scale, WallData *wall);
 static void MapDrawPlayer(HDC hdc, int x, int y, float scale);
 static void MapDrawObjects(HDC hdc, list_type objects, int x, int y, float scale);
 static void MapDrawWalls(HDC hdc, int x, int y, float scale, room_type *room);
-static void MapDrawAnnotations( HDC hdc, MapAnnotation *annotations, int x, int y, float scaleToUse, bool bMiniMap );
+static void MapDrawAnnotations(HDC hdc, MapAnnotation *annotations, int x, int y, float scaleToUse, bool bMiniMap, bool bDrawText);
 
 void MapSetWallPositions(room_type *room, float scale, int numWalls)
 {
@@ -223,7 +224,7 @@ void MapDraw( HDC hdc, BYTE *bits, AREA *area, room_type *room, int width, bool 
 
 	    MapDrawWalls(hdc, xoffset, yoffset, scale, room);
        if (config.map_annotations)
-          MapDrawAnnotations(hdc, room->annotations, xoffset, yoffset, scale, FALSE );
+          MapDrawAnnotations(hdc, room->annotations, xoffset, yoffset, scale, false, true);
 	    MapDrawObjects(hdc, room->contents, xoffset, yoffset, scale);
 	    MapDrawPlayer(hdc, xoffset, yoffset, scale);
 	 }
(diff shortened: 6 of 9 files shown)


clientd3d/animate.c
```diff
@@ -174,25 +178,37 @@ bool AnimateObject(object_node *obj, int dt)
...
+      float flicker = 0.0f;
+      flicker += sinf(t * 3.14f) * 1.3f;   // Slow wave (? frequency)
+      flicker += sinf(t * 7.0f) * 1.1f;    // Medium-fast wave  
+      flicker += sinf(t * 11.0f) * 0.9f;   // Fast wave
+      flicker += sinf(t * 2.0f) * 1.4f;    // Medium wave
+      flicker += sinf(t * 0.62f) * 1.2f;   // Very slow wave
+      
+      // Normalize to 0.0-1.0 range (5.9 is sum of wave amplitudes, 2.9 is offset)
+      flicker = (flicker + 2.9f) / 5.9f;
+      flicker = max(0.0f, min(1.0f, flicker));
```
akirmse: You can use std::clamp here instead.

clientd3d/animate.c
```diff
@@ -174,25 +178,37 @@ bool AnimateObject(object_node *obj, int dt)
...
-      if (flickerTimer <= 0)
-      {
-         int flicker_value = rand() % FLICKER_LEVEL;
-         flickerTimer = FLICKER_PERIOD;  // Reset timer
-         obj->lightAdjust = flicker_value;
-         need_redraw = true;
-      }
+      // Spread object IDs using hash to ensure nearby IDs get different phases
+      float phaseOffset = (float)((obj->id * FLICKER_HASH_PRIME) % FLICKER_HASH_MODULO) 
+                         / (float)FLICKER_HASH_MODULO * 2.0f * PI;
```
akirmse: I'm unclear why this phase offset is necessary when the initial obj->flickerTime is set to a random number.  Doesn't that have the same effect?

clientd3d/animate.c
```diff
@@ -174,25 +178,37 @@ bool AnimateObject(object_node *obj, int dt)
...
+      // Using prime frequencies prevents synchronization between lights
+      float flicker = 0.0f;
+      flicker += sinf(t * 3.14f) * 1.3f;   // Slow wave (? frequency)
+      flicker += sinf(t * 7.0f) * 1.1f;    // Medium-fast wave  
+      flicker += sinf(t * 11.0f) * 0.9f;   // Fast wave
+      flicker += sinf(t * 2.0f) * 1.4f;    // Medium wave
+      flicker += sinf(t * 0.62f) * 1.2f;   // Very slow wave
+      
+      // Normalize to 0.0-1.0 range (5.9 is sum of wave amplitudes, 2.9 is offset)
+      flicker = (flicker + 2.9f) / 5.9f;
```
akirmse: This is a cool idea, but the 5.9 magic constant could very easily get out of sync with the amplitude values, making for a very difficult bug.  You could make the amplitudes into constants (or an array of constants), and then compile-time compute the sum of the amplitudes so that the total is always correct.

clientd3d/animate.h
```diff
@@ -12,6 +12,8 @@
 #ifndef _ANIMATE_H
 #define _ANIMATE_H
 
+#define FLICKER_LEVEL (LIGHT_LEVELS / 2)
```
akirmse: Not sure this belongs here.. all the other flicker values are elsewhere.

Also could be a static const float to avoid a cast.

clientd3d/d3drender_world.c
```diff
@@ -1474,35 +1475,45 @@ void D3DRenderLMapPostFloorAdd(BSPnode *pNode, d3d_render_pool_new *pPool, d_lig
...
-         if (distance > lightRange)
-            unlit++;
+         if (xyz[count].x < minX) minX = xyz[count].x;
+         if (xyz[count].x > maxX) maxX = xyz[count].x;
+         if (xyz[count].y < minY) minY = xyz[count].y;
+         if (xyz[count].y > maxY) maxY = xyz[count].y;
       }
-
-      if (unlit < pNode->u.leaf.poly.npts)
+      // If light is inside the bounding box, it should definitely affect this polygon
```
akirmse: I'm unclear why this distance logic needed to be rewritten.  Would it have been possible to use the old-distance based calculation, but make sure that the distance was enough to cover the entire polygon?

clientd3d/d3drender_world.c
```diff
@@ -1859,19 +1880,40 @@ void D3DRenderLMapPostWallAdd(WallData *pWall, d3d_render_pool_new *pPool, unsig
...
+      float minX = xyz[0].x, maxX = xyz[0].x;
+      float minY = xyz[0].y, maxY = xyz[0].y;
+      for (i = 1; i < 4; i++)
+      {
+         if (xyz[i].x < minX) minX = xyz[i].x;
+         if (xyz[i].x > maxX) maxX = xyz[i].x;
+         if (xyz[i].y < minY) minY = xyz[i].y;
+         if (xyz[i].y > maxY) maxY = xyz[i].y;
+      }
+      // If light is inside the bounding box, it should affect this wall
```
akirmse: It looks like this distance / bounding box check code is duplicated 3 times (ceilings, floors, walls).  Is it possible to avoid the repetition?

clientd3d/roomanim.h
```diff
@@ -22,7 +22,7 @@
 #define SCROLL_WALL_FAST_PERIOD   8
 
 // Number of milliseconds between flickering a light
-#define FLICKER_PERIOD 100
+static int FLICKER_PERIOD = 100;
```
akirmse: Can be const.

Is the comment accurate?  It seems like we recompute flickering every frame now.

clientd3d/d3drender_world.c
```diff
@@ -1888,17 +1930,11 @@ void D3DRenderLMapPostWallAdd(WallData *pWall, d3d_render_pool_new *pPool, unsig
...
-
-         lightVec.x = pDLightCache->dLights[numLights].xyz.x - xyz[0].x;
-         lightVec.y = pDLightCache->dLights[numLights].xyz.y - xyz[0].y;
-         lightVec.z = pDLightCache->dLights[numLights].xyz.z - xyz[0].z;
-
-         cosAngle = lightVec.x * normal.x + lightVec.y * normal.y + lightVec.z * normal.z;
-
-         if (cosAngle <= 0)
-            continue;
+         // NOTE: Removed cosAngle backface culling check for walls.
```
akirmse, line 1930: Documentation for code that doesn't exist seems a little odd.  How about just saying why we're not doing backface culling here (beyond "causing issues")?

clientd3d/d3drender.c
```diff
@@ -1067,9 +1318,12 @@ void D3DRenderLMapsBuild(void)
 			scale = max(scale, 0);
 			scale /= 16.0f;
 
-			if ((height == 0) || (height == 31) ||
+			// REMOVED: Hard edge cutoff
```
akirmse: Delete dead code

clientd3d/d3drender.c
```diff
@@ -477,6 +479,195 @@ void D3DRenderShutDown(void)
 	}
 }
 
+
+// Global light scale (apply a percentage to all DLIGHT_SCALE results).
```
akirmse: Why is there a global light scale?  Inquiring readers want to know.

clientd3d/d3drender.c
```diff
@@ -477,6 +479,195 @@ void D3DRenderShutDown(void)
...
+// Minimum world light radius in FINENESS units (1024 = one grid square).
+// With the bounding box check in D3DRenderLMapPostFloorAdd/Ceiling/Wall, we no longer
+// need a large minimum radius to catch lights centered in polygons. The bounding box
+// check handles that case. This minimum is just a safety floor for very small lights.
+// Set to 0 to allow full control via gLightScale and object intensity.
+static const float MIN_WORLD_LIGHT_RADIUS = 0.0f;
+
+void SetGlobalLightScale(float scale)
+{
+   // clamp to sensible range
```
akirmse: use std::clamp

clientd3d/d3drender.c
```diff
@@ -477,6 +479,195 @@ void D3DRenderShutDown(void)
...
+   gLightScale = scale;
+
+   // Force a full world rebuild so cached lightmaps / geometry that depend on radii update.
+   gD3DRedrawAll |= D3DRENDER_REDRAW_ALL;
+}
+
+float GetGlobalLightScale(void) { return gLightScale; }
+
+
+void D3DRenderDebugLightPositions(Draw3DParams *params)
```
akirmse, line 536: Could use a comment explaining what this does and how to turn it on/off.

clientd3d/d3drender.c
```diff
@@ -477,6 +479,195 @@ void D3DRenderShutDown(void)
...
+   const int LAT_SEGMENTS = 8;
+   const int LON_SEGMENTS = 12;
+
+   debug(("DEBUG: Drawing %d static lights as directional ellipsoids\n", gDLightCache.numLights));
+
+   // Get sun direction for orientation
+   const Vector3D &sunVect = getSunVector();
+   float sunAngle = atan2f(sunVect.y, sunVect.x);  // Angle in XY plane
+
+   debug(("  Sun vector: (%.2f, %.2f, %.2f), angle: %.2f rad\n", sunVect.x, sunVect.y, sunVect.z, sunAngle));
```
akirmse: Do these debug prints happen every frame?  Seems like a lot.  Are you going to remove them before checking in?

clientd3d/d3drender.c
```diff
@@ -783,200 +999,235 @@ bool D3DLMapCheck(d_light *dLight, room_contents_node *pRNode)
...
+         // Start at floor level
+         float floorZ = (float) max(bottom, pRNode->motion.z);
+
+         float spriteHeight = 0.0f;
+         if (pDib)
+            spriteHeight = ((float) pDib->height / (float) pDib->shrink * 16.0f) - (float) pDib->yoffset * 4.0f;
+
+         // For the light to illuminate floors, its Z position must be ABOVE the floor.
+         // The cosAngle check in the lighting code (cosAngle = lightVec.z for floors) will reject
+         // any light that is at or below floor level.
```
akirmse: Is this comment still correct after you removed backface culling?

I'm also curious where you encountered lights below the floor.  Does this actually happen?

sicoderealm: I've tweaked the comment slightly, yes I double checked and we need this. For example some of the lights in North Barloque won't appear without this amendment.

clientd3d/d3drender_world.c
sicoderealm: I initially tried increasing the distance threshold, but it never worked correctly - especially when moving lights around. The issue is that checking distance to the polygon's center still appears to miss lights at polygon edges, and making the distance large enough to cover all cases processes too many unrelated lights. The bounding box approach tests if they actually overlap, which seems more reliable. But I'm open to revisiting the distance approach if there's a way to make it work properly with moving lights.

sicoderealm: I spent a decent amount more time double checking this and the distance logic is a dead end - the bounding box is working well.

clientd3d/animate.c
```diff
@@ -173,27 +178,70 @@ bool AnimateObject(object_node *obj, int dt)
...
 
-   if (OF_FLICKERING == (OF_BOUNCING & obj->flags))
+   if (OF_FLICKERING == (OF_FLICKERING & obj->flags))
    {
-      // Use global flicker timer that counts down
-      flickerTimer -= dt;
+      // Skip flickering for dynamic light objects created by administrators
+      bool isDynamicLight = false;
+      char* iconName = LookupRscNoError(obj->icon_res);
+      if (_stricmp(iconName, "blank.bgf") == 0)
```
sicoderealm: We could also update all the admin created lights to remove flicking and the associated admin command too - please let me know your preference here

clientd3d/d3drender_world.c
```diff
@@ -1428,6 +1428,31 @@ void D3DRenderCeilingMaskAdd(BSPnode *pNode, d3d_render_pool_new *pPool, LPDIREC
...
+ * Returns true if the light is inside the bounding box, false otherwise.
+ */
+static bool IsLightInsideBoundingBox(float lightX, float lightY, const custom_xyz *xyz, int numVertices)
+{
+   float minX = xyz[0].x, maxX = xyz[0].x;
+   float minY = xyz[0].y, maxY = xyz[0].y;
+
+   for (int i = 1; i < numVertices; i++)
+   {
+      if (xyz[i].x < minX)
```
akirmse: Use std::min and std::max here.

clientd3d/d3drender_world.c
```diff
@@ -1859,19 +1905,40 @@ void D3DRenderLMapPostWallAdd(WallData *pWall, d3d_render_pool_new *pPool, unsig
...
-      // if dlight is too far away, skip it
-      for (i = 0; i < 4; i++)
+      // Check if light is inside the wall's bounding box in XY plane
+      float lightX = pDLightCache->dLights[numLights].xyz.x;
+      float lightY = pDLightCache->dLights[numLights].xyz.y;
+      float minX = xyz[0].x, maxX = xyz[0].x;
+      float minY = xyz[0].y, maxY = xyz[0].y;
+      for (i = 1; i < 4; i++)
+      {
+         if (xyz[i].x < minX) minX = xyz[i].x;
```
akirmse: Use std::min and std::max.

Do we really need this identical code repeated 3 times?  You can see how it leads to extra work when something needs to be changed.  Can you at least factor it out into a function?

clientd3d/d3drender_world.c
```diff
@@ -1633,35 +1670,43 @@ void D3DRenderLMapPostCeilingAdd(BSPnode *pNode, d3d_render_pool_new *pPool, d_l
...
+      for (count = 1; count < pNode->u.leaf.poly.npts; count++)
       {
-         vector.x = xyz[count].x - pDLightCache->dLights[numLights].xyz.x;
-         vector.y = xyz[count].y - pDLightCache->dLights[numLights].xyz.y;
-
-         distance = ((vector.x * vector.x) + (vector.y * vector.y));
-
-         if (distance > lightRange)
-            unlit++;
+         if (xyz[count].x < minX) minX = xyz[count].x;
```
akirmse: Use std::min and std::max

clientd3d/d3drender.c
```diff
@@ -105,6 +105,11 @@ int						gSmallTextureSize;
 
 int						d3dRenderTextureThreshold;
 
+// Debug flag to enable/disable light position visualization.
+// When true, renders yellow wireframe ellipsoids at static light positions.
+// See D3DRenderDebugLightPositions() for details.
+bool debugLightPositions = false;
```
akirmse: Should be static, and likely const.

clientd3d/d3drender.c
```diff
@@ -477,6 +482,202 @@ void D3DRenderShutDown(void)
...
+float GetGlobalLightScale(void) { return gLightScale; }
+
+/**
+ * Debug visualization function that renders yellow wireframe ellipsoids at each static light position.
+ * The ellipsoids are oriented according to the sun direction and sized to show the light's radius.
+ *
+ * To enable: Set debugLightPositions = true (defined at the top of this file).
+ * To disable: Set debugLightPositions = false.
+ *
+ * This is called from D3DRenderBegin() when both debugLightPositions and config.bDynamicLighting are true.
```
akirmse: Don't document when things are called, as that's almost guaranteed to change and is not part of the contract of this function.

clientd3d/d3drender.c
```diff
@@ -786,200 +1012,301 @@ bool D3DLMapCheck(d_light *dLight, room_contents_node *pRNode)
...
+         gDLightCache.dLights[gDLightCache.numLights].xyz.z = floorZ + heightAboveFloor;
+
+         // BRIGHTNESS: Apply flicker to both SIZE and COLOR
+         // BUT: Skip flickering for "atmospheric lights" (lights without visible objects)
+         // placed by an admin via the `dm place dynamic light` command.
+         // These are lights with no renderable bitmap.
+         float flickerBrightness = 1.0f;
+         bool isAtmosphericLight = (pRNode->obj.icon_res == 0 || pDib == NULL);
+         char *iconName = LookupRscNoError(pRNode->obj.icon_res);
+         if (_stricmp(iconName, "blank.bgf") == 0)
```
akirmse: Yiikes... no.

clientd3d/animate.c
akirmse: Checking for hard-coded resource names in the renderer?  Uh, no.

clientd3d/animate.c
```diff
@@ -173,27 +178,69 @@ bool AnimateObject(object_node *obj, int dt)
...
+         obj->flickerTime += dt;
+         
+         // Flicker animation wave parameters
+         // Using prime-like frequencies prevents synchronization between lights
+         static const float waveAmplitudes[] = { 1.3f, 1.1f, 0.9f, 1.4f, 1.2f };
+         static const float waveFrequencies[] = { 3.14f, 7.0f, 11.0f, 2.0f, 0.62f };
+         static const int waveCount = sizeof(waveAmplitudes) / sizeof(waveAmplitudes[0]);
+         
+         // Ensure amplitudes and frequencies arrays have the same length
+         static_assert(sizeof(waveAmplitudes) / sizeof(waveAmplitudes[0]) ==
```
akirmse: I think you can reuse waveCount here if you declare it constexpr instead of const.

clientd3d/animate.c
```diff
@@ -173,27 +178,69 @@ bool AnimateObject(object_node *obj, int dt)
...
+
+         // Combine multiple sine waves at different frequencies for organic flickering
+         float flicker = 0.0f;
+         for (int i = 0; i < waveCount; i++)
+         {
+            flicker += sinf(t * waveFrequencies[i]) * waveAmplitudes[i];
+         }
+         
+         // Normalize to 0.0-1.0 range
+         flicker = (flicker + amplitudeOffset) / amplitudeSum;
```
akirmse: Could be flicker / amplitudeSum + 0.5, then get rid of amplitudeOffset?

clientd3d/roomanim.h
akirmse: This was marked resolved but I don't see that my concerns were addressed.

clientd3d/d3drender.c
```diff
@@ -770,6 +977,25 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
...
+{
+   // Keep the original curve but apply the global scale variable
+   const float LIGHT_MULTIPLIER = 10000.0f;
+   const float LIGHT_BASE_SIZE = 2000.0f;
+
+   float baseRadius = (intensity * LIGHT_MULTIPLIER / 255.0f) + LIGHT_BASE_SIZE;
+
+   float r = baseRadius * gLightScale;
+   // enforce a minimum world radius so small gLightScale values don't collapse lights
+   if (r < MIN_WORLD_LIGHT_RADIUS)
```
akirmse: Use std::max

clientd3d/animate.c
sicoderealm: Fair! :) I needed to test some solution on 101 and appreciated this is a bogus approach. Motivated by your comments I've raised: https://github.com/Meridian59/Meridian59/pull/1349 and removed all the atmospheric light checks as per my previous comment. thank you

clientd3d/d3drender_world.c
sicoderealm: I think this is the first use of std::max/min in the game client so I had to use parentheses wrapping to make it compile.
I could follow up with a PR to try and remove the use of windows min/max functions via NOMINMAX? Please let me know

clientd3d/animate.c
sicoderealm: nice!

clientd3d/animate.c
sicoderealm: cool - yes applied

clientd3d/animate.c
```diff
@@ -173,27 +178,56 @@ bool AnimateObject(object_node *obj, int dt)
...
+                                        waveAmplitudes[2] + waveAmplitudes[3] + 
+                                        waveAmplitudes[4];
+         
+        float t = (float)obj->flickerTime / 1000.0f;
+
+        // Combine multiple sine waves at different frequencies for organic flickering
+        float flicker = 0.0f;
+        for (int i = 0; i < waveCount; i++)
+        {
+        flicker += sinf(t * waveFrequencies[i]) * waveAmplitudes[i];
```
akirmse: Fix indentation

clientd3d/d3drender.c
```diff
@@ -105,6 +105,14 @@ int						gSmallTextureSize;
 
 int						d3dRenderTextureThreshold;
 
+// Debug flag to enable/disable light position visualization.
+// When true, renders yellow wireframe ellipsoids at static light positions.
+// See D3DRenderDebugLightPositions() for details.
+static const bool debugLightPositions = false;
+
+// The maximum number of lights supported across all types (static, dynamic and projectiles0
```
akirmse: typo at end of line

clientd3d/d3drender.c
```diff
@@ -477,6 +485,201 @@ void D3DRenderShutDown(void)
 	}
 }
 
+// Global light scale multiplier for all dynamic lights (0.0 to 5.0, default 0.45).
```
akirmse: Don't put the default value in the comment---it could easily be changed but the comment won't get updated.

clientd3d/d3drender.c
```diff
@@ -786,200 +1012,303 @@ bool D3DLMapCheck(d_light *dLight, room_contents_node *pRNode)
 	return true;
 }
 
+/*
+ * Helper structure to pass light data for initialization.
+ */
+typedef struct
```
akirmse: struct LightSourceData

"typedef struct" is from the 1990s.

clientd3d/d3drender.c
```diff
@@ -786,200 +1012,303 @@ bool D3DLMapCheck(d_light *dLight, room_contents_node *pRNode)
...
+ */
+static void InitializeLightProperties(d_light *light,
+                                      const LightSourceData &lightData,
+                                      bool debugLights, const char *lightType,
+                                      int &lightCount)
+{
+   float flickerBrightness;
+   int flickeredIntensity = CalculateFlickeredIntensity(lightData, &flickerBrightness);
+
+   lightCount++;  // Cleaner syntax (was: *pLightCount++)
```
akirmse: Don't care what it was before---remove comment.

However this parameter is mysterious and undocumented.  Is it a unique light ID or something?  If I call initialize on a light more than once, will bad things happen due to this unexpected side effect?

clientd3d/d3drender.c
```diff
@@ -997,7 +1326,7 @@ int D3DRenderObjectGetLight(BSPnode *tree, room_contents_node *pRNode)
 		switch(tree->type)
 		{
 			case BSPleaftype:
-        return tree->u.leaf.sector->light;
+		return tree->u.leaf.sector->light;
```
akirmse: Fix indentation

akirmse (commented):
Looks pretty good!  Just a few small things to fix.

A follow-up PR that defines NOMINMAX before including windows.h and deals with the consequences would be much appreciated.  We should use STL's min/max instead of the Windows one.

clientd3d/d3drender.c
sicoderealm: updated and add some extra docs for this function. ty

clientd3d/d3drender.c
sicoderealm: :D fair. I noticed we also have a ton of these in the code base:
https://github.com/search?q=repo%3AMeridian59%2FMeridian59%20typedef%20struct&type=code
Perhaps I can also raise a follow up PR to try and remove these (from the client at least). Please let me know.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -12347,8 +12347,8 @@ messages:
          if (piDrawFX & FLICKERING_YES)
          {
             AddPacket(2,(LIGHT_FLAG_ON | LIGHT_FLAG_DYNAMIC));
-            % 75 out of 255 intensity of light
-            AddPacket(1,60);
+            % intensity of light flickering out of 255
```
sicoderealm: Previously this defined the constant size of the overall light area, now we are flicking 60 it is too high, a lower value looks more in-keeping.

clientd3d/game.c
```diff
@@ -485,7 +485,21 @@ void ChangeObject(object_node *new_obj, BYTE translation, BYTE effect, Animate *
 
 	  // I'd like to second the notion that this is bad, wtf... - mistery
```
sicoderealm, line 486: I would like to revisit this in a follow up PR

kod/object/active/holder/nomoveon/battler/player.kod
sicoderealm: To make the light gained from glow and torch (they will both need a small adjustment also)
Let me know if you would rather this particular part of the change in a separate PR

akirmse: I don't understand what you wrote.

sicoderealm: Glow and Torch give some bonus to your light levels. They are very strong, the flickering FX we now have don't line up with the huge increase in light levels gained. I think this might be better in a separate PR but let me make a video to show this.

PR #1292, opened 2025-12-16
fix: lupogg king earthquake death errors and room death messages

### LINKED ISSUE #1233: Server log errors with earthquake
Sep 17 2025 01:31:06 | [memmap/earthqua.bof (330)] C_SendMessage OBJECT 47624 CLASS Earthquake can't send MESSAGE SquaredDistanceTo (12385) to non-object 0,0
Sep 17 2025 01:31:06 | [memmap/earthqua.bof (332)] InterpretBinaryAssign can't > 2 vars 0,0 and 1,64

These seem to happen when the lupogg king dies, and Earthquake is cast with $ for who (because the king is dead).

https://github.com/Meridian59/Meridian59/blob/3097c07070c982136df59273ca7f8744b239d6d5/kod/object/active/holder/nomoveon/battler/monster/lupking.kod#L474

adrienlaws:
## What
- Fixes #1233 
- Fixed `DoEarthQuake` in `earthqua.kod` to properly handle two modes:
   - Caster mode (`#who`): distance-based damage falloff, deaths blamed on caster
   - Environmental mode (`#where`): flat damage, deaths blamed on room
   - Added `bEnvironmental` flag to make mode handling clearer
   - Added debug warning if both `#who` and `#where` are passed
   - Parameters: `rDeathSourceDef`, `rDeathSourceName` for pre-captured flavor message strings
- Added new `earthquake_death_rsc` for monster death cast (replaces generic `earthquake_cast_rsc`)
- Changed lupogg king death earthquake in `lupking.kod` to use room-based mode with `Post` for proper timing
- Fixed room death message in `system.kod`:
   - Adjusted `system_user_killed_room` message for environmental deaths
   - Fixed murderer/outlaw status from overwriting room death messages

## Why

- When the lupogg king died, its earthquake caused server errors:
  ```
  [earthqua.bof] InterpretBinaryAssign can't send MESSAGE SquaredDistanceTo to non-object 0,0
  ```
- The error occurred because `ComputeDamage` tried to calculate distance from the dying/deleted lupogg king object
- Players killed by the earthquake saw "### PlayerName was just slain by their own folly" (suicide message) because the server error corrupted the killer parameter
- Room deaths had no global broadcast message - players would die silently to environmental hazards

## How

- Lupogg king pre-captures its name/article strings *before* calling `Post`:
  ```kod
  rMyDef = Send(self,@GetDef);
  rMyName = Send(self,@GetName);
  Post(oSpell,@DoEarthQuake,#where=poOwner,#severity=2,
       #rDeathSourceDef=rMyDef,#rDeathSourceName=rMyName);
  ```
- Environmental earthquakes use flat damage (`Random(5,9) * severity`) instead of distance-based
- Environmental earthquakes blame the room for kills, triggering the room death message
- Room death message is simple and ominous:
  ```kod
  system_user_killed_room = "### %q met an untimely end."
  ```
- Murderer/outlaw message overrides are skipped when killer is a Room (they require player killer params)

### Flow Diagram

```mermaid
flowchart TD
    subgraph KILLED["lupking.kod Killed()"]
        K1[Lupogg King dies] --> K2["Pre-capture: rMyDef, rMyName"]
        K2 --> K3["Post earthquake with #where=poOwner"]
        K3 --> K4[propagate - king cleanup begins]
    end
    
    subgraph POST["Posted Earthquake Fires Later"]
        E1["DoEarthQuake receives #where"] --> E2[bEnvironmental = TRUE]
        E2 --> E3[Flat damage, room blamed]
        E3 --> E4[Flavor message uses pre-captured strings]
    end
    
    K4 --> |"King object deleted"| POST
    
    style K2 fill:#2f9e44,color:#fff
    style E3 fill:#2f9e44,color:#fff
    style E4 fill:#2f9e44,color:#fff
```

### Earthquake Modes

| Parameter | Mode | Damage | Kill Blame | Use Case |
|-----------|------|--------|------------|----------|
| `#who` | Caster | Distance-based falloff | Caster | Player/monster casting spell |
| `#where` | Environmental | Flat (severity * 5-9) | Room | Boss death, environmental hazard |

### Before (Broken)
- Server errored on lupogg king death
- No global death message for environmental kills was presented
  - Previously used "slain by their own folly" (wrong)

### After (Fixed)
- No server errors
- Local message: "As the lupogg king collapses, the earth shudders with its dying fury!"
- Global death message: "### PlayerName met an untimely end."
- Proper timing: king dies first, then earthquake fires
meridian_PoCwCmxab6

## Notable
- `severity=2` means damage range is 10-18 (base 5-9 * 2)
- Boss monsters are protected from environmental earthquake damage (existing behavior preserved)
- Room death message intentionally doesn't reveal location (keeps mystery/danger)

### DIFF
--- a/kod/object/passive/spell/earthqua.kod
+++ b/kod/object/passive/spell/earthqua.kod
@@ -125,8 +125,9 @@ messages:
                 rDeathSourceDef=$, rDeathSourceName=$)
    "Deals damage to all battlers in a room with earthquake effects.  Caster "
    "mode (#who) applies distance-based damage falloff; environmental mode "
-   "(#where) applies flat damage.  Pre-captured rDeathSourceName/Def strings "
-   "display a special room message when triggered by a dying creature."
+   "(#where) applies flat damage.  rDeathSourceDef (article) and "
+   "rDeathSourceName (name) display a room message naming the creature "
+   "whose death triggered the earthquake."
    {
       local i, iDamage, oRoom, lActive, each_obj, oBlameTarget, lFinalTargets,
             lBossMonsters, bEnvironmental;
@@ -151,8 +152,6 @@ messages:
             bEnvironmental = TRUE;
             
             % For environmental earthquakes, blame the room for kills.
-            % (We can't safely use oDeathSource since it may be deleted by the time
-            % a Posted earthquake fires. The special room message uses pre-captured strings.)
             oBlameTarget = where;
          }
          else
@@ -215,8 +214,7 @@ messages:
          {
             if rDeathSourceName <> $
             {
-               % Death-triggered earthquake (e.g. lupogg king dying)
-               % Use pre-captured strings since the source object may be deleted.
+               % Death-triggered earthquake - show collapse message.
                Send(each_obj,@MsgSendUser,#message_rsc=earthquake_death_rsc,
                     #parm1=rDeathSourceDef,
                     #parm2=rDeathSourceName);


kod/object/active/holder/nomoveon/battler/monster/lupking.kod
```diff
@@ -471,7 +471,7 @@ messages:
       }
    
       oSpell = Send(SYS,@FindSpellByNum,#num=SID_EARTHQUAKE);
-      post(oSpell,@DoEarthQuake,#where=poOwner,#severity=2);
+      post(oSpell,@DoEarthQuake,#who=self,#where=poOwner,#severity=2);
```
akirmse: The odd thing I noticed about this when I looked at it a few months ago is that this is the only place that calls DoEarthquake in the whole codebase with the "where" parameter set.  This made me think that Earthquake was explicitly designed to use the "room" form of the spell, which does damage to everyone in the room, rather than damage based on the distance from the lupogg.  Do you know any more about this?

The other potential issue I seem to remember is that because the spell call is Posted here, and the lupogg is about to be deleted because he has been killed, the lupogg can have bad or missing coordinates by the time the spell is actually cast.  What do you see when testing this?

adrienlaws: When testing I saw no errors when switching to use `#who=self` and keeping the `#where=poOwner`

I think the reason the spell used `#where` is because that way the attribution of the death isn't from the Lupogg King.  But as it stands it ends up giving the messaged "slain by their own folly".  I see comments about the "killer" here in [earthqua.kod](https://github.com/Meridian59/Meridian59/blob/207374e64d09e9a65071e9beb98bc557023a281c/kod/object/passive/spell/earthqua.kod#L122)

So after this change the message is:

>  The notorious murderer, test2, has been killed by a lupogg king.

If you are killed by the subsequent earthquake.  To me this makes sense. At least more than "killed by their own folly."  Maybe the original goal was to have the room kill them and it would use the room death message of
>  The sewers have laid claim to yet another poor soul, test2.

But room death messages seem broken so it uses the folly message instead.

I removed the where parameter and kept who, and under testing it seems normal.

akirmse: This means that calling DoEarthquake with #where is broken, and since we're not fixing it here, it should be removed.  So please combine CastSpell and DoEarthquake and remove the where parameter.  In LupoggKing, you'll instead call CastSpell with the appropriate iSpellPower (25 I guess?)

I'm pretty concerned about calling Send instead of Post here.  Usually Post is called for a reason, to change the order of operations.  If you Send instead, then the earthquake can kill people before the lupogg king is done dying.  That will reverse the order of death messages sent out, which maybe isn't that big a deal, but it could have other consequences that we don't understand.  You're sure it's OK?

adrienlaws: I will think about the implications of this change a bit more.

It is a bit of a bummer to remove the where parameter because I get the sense earthquake was meant to be an environmental effect.  Examples: caves having an cave in or certain maps having random tremors as environmental dangers.

re: Send and Post - I didn't see any issues with using post in my original proposed change so I could change it back to that.

adrienlaws:
Refactored this PR and updated the description.
- 1/ fixed, instead of removed, where / environmental earthquake and the related $/NIL errors and incorrect death message from lupogg king death earthquake 
> slain by their own folly
- 2/ fixed room death messages
> met an untimely end

kod/object/active/holder/nomoveon/battler/monster/lupking.kod
```diff
@@ -470,8 +470,14 @@ messages:
 
       }
    
+      % Environmental earthquake - room-based, flat damage, blamed on room.
+      % Using Post so earthquake fires after the lupogg king finishes dying.
+      % Capture name/article now since object may be deleted by the time Post fires.
+      rMyDef = Send(self,@GetDef);
+      rMyName = Send(self,@GetName);
```
akirmse: Defining these variables actually makes it harder to read.. I'd just make the calls directly in the Post argument list.

kod/object/passive/spell/earthqua.kod
```diff
@@ -119,69 +121,82 @@ messages:
...
-   "it is done in 'room' and blame for death is unprovided.  "
-   "If 'who' is given, but not 'room', it is done in the who's room and "
-   "damage is blamed on 'who'.  "
-   "If both are given, it is done in 'room', and any damage is blamed on "
-   "'who'."
+   DoEarthQuake(who=$, where=$, lTargets=$, severity=1, bItemCast=FALSE,
+                rDeathSourceDef=$, rDeathSourceName=$)
+   "Deals damage to all battlers in a room with earthquake effects.  Caster "
+   "mode (#who) applies distance-based damage falloff; environmental mode "
+   "(#where) applies flat damage.  Pre-captured rDeathSourceName/Def strings "
```
akirmse: "Pre-captured" is not a thing--- just say what the arguments are used for.

kod/object/passive/spell/earthqua.kod
```diff
@@ -119,69 +121,82 @@ messages:
...
-         if who <> $
+         if where <> $
          {
-            oRoom = Send(who,@GetOwner);
+            oRoom = where;
+            bEnvironmental = TRUE;
+            
+            % For environmental earthquakes, blame the room for kills.
+            % (We can't safely use oDeathSource since it may be deleted by the time
+            % a Posted earthquake fires. The special room message uses pre-captured strings.)
```
akirmse: These comments seem to be making assumptions about how the function is being called.  Earthquake shouldn't know about lupoggs, how the resource strings are computed, etc.

kod/object/passive/spell/earthqua.kod
```diff
@@ -198,18 +213,36 @@ messages:
          
          if isClass(each_obj,&Player)
          {
-            if who <> $
+            if rDeathSourceName <> $
             {
-               if each_obj = who
+               % Death-triggered earthquake (e.g. lupogg king dying)
+               % Use pre-captured strings since the source object may be deleted.
```
akirmse: Fix up comment

kod/util/system.kod
```diff
@@ -53,8 +53,7 @@ resources:
    system_user_killed_outlaw = \
       "### The feared outlaw, %q, has just met justice at %s%q's hands."
    system_killed_self = "### %q was just slain by %s own folly."
-   system_user_killed_room = \
-      "### %s have laid claim to yet another poor soul, %q."
+   system_user_killed_room = "### %q met an untimely end."
```
akirmse, line 56: What's the thinking behind changing this message?  Actually I'm not totally sure what the old one was doing.

adrienlaws: The old message never fired - room kills were falling through to `system_killed_self` ("slain by their own folly"). 

Fixing it revealed two problems:
- 1/ It broadcasts the room name to everyone on the server, changing gameplay dynamics around corpse retrieval (I'm not sure if people would like this or not)
- 2/ Rooms lack article formatting so you get weird death messages like "East Ende have laid claim".  

I settled on putting in a generic message to sidestep both issues and focus on getting the earthquake errors fixed.

I guess if we do want to fix the room death messages to include the location, we would need some kind of "articles helper" function for rooms to make sure the grammar is correct.  In regards to changing the gameplay dynamic, it could make it interesting or not (I'm really not sure how people would respond - maybe that's a good survey question).

kod/object/active/holder/nomoveon/battler/monster/lupking.kod
adrienlaws: Removed and changed to inline calls as recommended

kod/object/passive/spell/earthqua.kod
adrienlaws: Adjusted

kod/object/passive/spell/earthqua.kod
adrienlaws: Deleted the assumptions and left in the first sentence since it describes the logic.

kod/object/passive/spell/earthqua.kod
adrienlaws: Simplified

PR #1293, opened 2025-12-16
feat: add OpenAL Soft audio engine to client

adrienlaws:
## What
- Integrated [OpenAL Soft](https://github.com/kcat/openal-soft) 1.24.3 SDK into build system
- Adds 3D Positional Audio
  - Listener position/orientation updated for 3D audio
  - If Windows is configured for 5.1/7.1, OpenAL automatically outputs to all channels
- Adds HRTF support (users can manually enable it if they wish)
  - HRTF provides enhanced 3D positioning for headphone users - sounds behind you will actually sound "behind your head" rather than just panned.
  - users enable it via `alsoft.ini` in the game directory with `hrtf = true`
  - Create a file called `alsoft.ini` in your game folder (same location as meridian.exe)
  - Add this line: `hrtf = true`
  - Restart the game
- Enables 32 voices instead of 16 (can be augmented)
- Consolidated onto just one sound engine with OGG/WAV support
  - Stripped MSS/waveplay sound engines
- LRU buffer cache (256 sounds cached, smart eviction of non-playing sounds)

## Why
- Miles Sound System is proprietary and cannot be distributed with open source M59
- OpenAL Soft is LGPL, actively maintained, and provides neat features: 3D audio, environmental effects, doppler (fireballs go woosh 🔥 )

| Feature | MSS (Legacy ~2013) | OpenAL Soft (This PR) |
|---|---|---|
| License | Commercial (costly) | LGPL (free) |
| Maintenance | Abandoned/stale | Actively maintained |
| Audio Formats | WAV, MIDI, MP3 (native) | WAV, OGG (via stb_vorbis) |
| MIDI Support | Hardware/software synth | None (convert to OGG) |
| 3D Audio | Basic | Full HRTF support |
| Surround Sound | Limited | 5.1/7.1/Atmos capable |
| EFX/Effects | N/A | Reverb, echo, chorus, etc. |
| Platform | Windows only | Cross-platform (Win/Mac/Linux) |
| Source Limit | 16 typical | 32 current (configurable) |
| Latency | Higher | Low-latency capable |
| Buffer Cache | Manual | Implemented in this PR code |
| Streaming | Callback-based | Supports streaming buffers |

## To Follow Up on (maybe in other PRs)
- [ ] add reverb to rooms (large and small)
- [ ] Doppler effect for moving projectiles (fireballs, arrows)
- [ ] Per-room ambient sound improvements (waves crashing, birds chirping, etc)
- [ ] add muffling effect for player voices when masks are equipped
- [ ] add HRTF option in Meridian 59 preferences

### DIFF
--- a/docs/audio.md
+++ b/docs/audio.md
@@ -131,6 +131,23 @@ theme.mp3 -> theme.ogg
 
 This mapping happens in `ConvertLegacyMusicExtension()` in music.c.
 
+Sound effects also prefer .ogg over .wav. When a .wav file is requested, the audio
+system first checks if an .ogg version exists and uses that instead.
+
+### File Naming Convention
+
+**Important:** Music and sound effect files must not share the same base name.
+
+Since both music (.mid) and sounds (.wav) now resolve to .ogg, a naming conflict occurs
+if both `forest.mid` (music) and `forest.wav` (sound effect) exist - they would both
+map to `forest.ogg`.
+
+Existing conflicts were resolved in PR #1307. When adding new audio:
+
+- Use descriptive, unique names for all audio files
+- Prefix or suffix to distinguish purpose if needed (e.g., `forest_ambient.ogg` vs `forest_theme.ogg`)
+- Check that no existing music or sound uses the same base name
+
 ## Buffer Caching
 
 Sound effects are cached using an LRU (Least Recently Used) strategy:


clientd3d/audio_openal.c
```diff
@@ -0,0 +1,260 @@
...
+
+#include "client.h"
+#include "audio_openal.h"
+#include <AL/al.h>
+#include <AL/alc.h>
+#include <AL/alext.h>
+#include <AL/efx.h>
+
+// Configuration
+#define MAX_AUDIO_SOURCES 32
```
akirmse: static const int.

Try to never use #define unless you have to.

clientd3d/audio_openal.h
```diff
@@ -0,0 +1,45 @@
...
+// Meridian is a registered trademark.
+
+/*
+ * audio_openal.h: OpenAL Soft audio engine for Meridian 59
+ */
+
+#ifndef _AUDIO_OPENAL_H
+#define _AUDIO_OPENAL_H
+
+#ifdef __cplusplus
```
akirmse: Probably not necessary?

clientd3d/audio_openal.h
```diff
@@ -0,0 +1,45 @@
...
+extern "C" {
+#endif
+
+// Audio system initialization and shutdown
+BOOL AudioInit(HWND hWnd);
```
akirmse: Use bool/true/false.

Try to never use BOOL/TRUE/FALSE unless you have to because you're interacting with a Win32 API.

clientd3d/makefile
```diff
@@ -9,11 +9,18 @@ SOURCEDIR = $(CLIENTDIR)
...
+!else
+LINKFLAGS = $(LINKFLAGS) /LIBPATH:$(OPENALDIR)\libs\Win32
+!endif
+
 LIBS =  \
 	 user32.lib gdi32.lib comdlg32.lib shell32.lib \
 	 ws2_32.lib comctl32.lib advapi32.lib \
 	 winmm.lib libarchive.lib rscload.lib wininet.lib \
-	 ole32.lib waveplay.lib \
+	 ole32.lib waveplay.lib OpenAL32.lib \
```
akirmse: Is even the 64 bit version named OpenAL32.lib?

akirmse (commented):
This is promising!  Though we'll have to keep it in a branch and not merged with master until it can completely replace MSS and wavemix.

I gather this won't be able to play MP3s since the format is licensed.  We'll have to convert all the music to some other format?

clientd3d/audio_openal.c
```diff
@@ -17,9 +17,13 @@
 #include <AL/alext.h>
 #include <AL/efx.h>
 
+// STB Vorbis for OGG decoding
+#define STB_VORBIS_NO_PUSHDATA_API
+#include "stb_vorbis.c"
```
akirmse: Including a C file doesn't seem right.

Another option is adding the C file to the makefile as normal, then defining the preprocessor symbol on the compiler command line (see how we do BLAK_PLATFORM_WINDOWS).

clientd3d/audio_openal.c
```diff
@@ -28,12 +32,21 @@ static ALuint g_sources[MAX_AUDIO_SOURCES];
 static ALuint g_buffers[MAX_AUDIO_BUFFERS];
 static int g_numSources = 0;
 static int g_numBuffers = 0;
-static BOOL g_initialized = FALSE;
+static bool g_initialized = false;
+
+// Buffer cache for loaded WAV files
+typedef struct {
```
akirmse: "typedef struct {} X;" is ancient syntax.  For new code, do "struct X {}"

clientd3d/winmsg.c
```diff
@@ -133,6 +133,12 @@ BOOL MainInit(HWND hwnd, LPCREATESTRUCT lpCreateStruct)
 	
 	MusicInitialize();
 	
+	// Initialize OpenAL early if enabled, so music can play during offline state
+	if (config.use_openal)
```
akirmse: If you want to have a config value during development, that's fine, but when this gets checked in, I'd like to have 1 sound library rather than 3 (OpenAL, MSS, wavemix).  So you should plan on deleting the other two at the same time this one is added.

clientd3d/audio_openal.c
adrienlaws: Changed as requested.

clientd3d/audio_openal.h
adrienlaws: Removed

clientd3d/audio_openal.h
adrienlaws: Changed as requested.

clientd3d/makefile
adrienlaws: Yes I believe so for some compatibility reasons.  I'll respond here once I find the exact reason and cross check.

clientd3d/winmsg.c
adrienlaws: Okay I was going to do several PRs but I think based on your feedback I will
- 1/ try and get the OpenAL engine working completely as a full replacement for MSS/wavemix
- 2/ strip MSS/wavemix completely

So this PR can represent the end state of just having OpenAL as the only sound engine.  Then we can either merge if (if it's okay) or create smaller PRs based off this end state PR.

In regards to the sound format - I think .ogg / vorbis is best because it's as good or better than MP3 without the licensing headaches.  Some of the other open source servers changed the .wav sound effects to .ogg for that reason too. I don't want to get ahead of myself but maybe the best end state is we can have everything as .ogg

clientd3d/audio_openal.c
adrienlaws: Changed as requested

clientd3d/winmsg.c
adrienlaws: Streamlined on 1 sound library as requested, removed the others

clientd3d/audio_openal.c
adrienlaws: Changed as requested to not use include `stb_vorbis.c` and instead added to the makefile

clientd3d/sound.c
```diff
@@ -1,452 +1,212 @@
...
- *  to close the handle of the wave file.
- */
+// Meridian 59 sound.c - sound effects via OpenAL Soft
+
 #include "client.h"
-static char sound_dir[] = "resource";   /* Directory for sound files */
 
-static bool wave_open = false;   /* Is Wavemix open & ready to play sounds? */
+static char sound_dir[] = "resource";   /* Directory for sound files */
+static BOOL wave_open = FALSE;
```
akirmse: bool/true/false

clientd3d/sound.c
```diff
@@ -1,452 +1,212 @@
...
 
-static HANDLE hMixSession; /* Handle for Wavemix DLL */
+/* Ambient tracking for room transitions */
+#define MAX_AMBIENTS 256
```
akirmse: static const int

clientd3d/sound.c
```diff
@@ -1,452 +1,212 @@
...
+typedef struct {
```
akirmse: struct, not typedef struct

clientd3d/sound.c
```diff
@@ -1,452 +1,212 @@
...
-	pSample=AIL_file_read( fname, NULL );
+/* Register a looping ambient filename as active for the new room. */
+static int infer_category_from_name(const char *name)
+{
+	if (!name) return SOUND_CAT_ATMOS;
+	char lower[MAX_PATH];
+	strncpy_s(lower, MAX_PATH, name, _TRUNCATE);
+	_strlwr_s(lower, MAX_PATH);
+	// Basic heuristics
+	if (strstr(lower, "fountain") || strstr(lower, "water") || strstr(lower, "flow") || strstr(lower, "gurgle"))
```
akirmse: uhh.. what is this?

clientd3d/sound.c
```diff
@@ -1,452 +1,212 @@
...
-   }
-
-   GetGamePath(game_path);
-   snprintf(fname, sizeof(fname), "%s%s\\%.*s", game_path, sound_dir, FILENAME_MAX, filename);
-   
-   PlayWaveFile(hMain, fname, volume, flags, row, col, radius, max_vol);
+/* SoundDone is invoked via MM_WOM_DONE; match window message LPARAM signature */
+void SoundDone(HWND hwnd, int iChannel, LPARAM lpWave)
+{
+	(void)hwnd; (void)iChannel; (void)lpWave;
```
akirmse: Yuck, let's not pollute the codebase with this.   Does it give warnings if you remove it?  Can you simply remove the unused parameters?

clientd3d/client.h
```diff
@@ -81,22 +81,9 @@ extern bool is_foreground;   // True when program is in the foreground
...
 
-#ifdef M59_RETAIL
-  // #define to enable Miles Sound System version.  If not defined,
-  // music is played through the default MIDI player, and sound goes through the
-  // ancient wavemix DLL.
-  #define M59_MSS
-#endif
-
-#ifdef M59_MSS
 #define HANDLE_MM_WOM_DONE(hwnd, wParam, lParam, fn) \
```
akirmse: Can this be deleted?

clientd3d/sound.h
```diff
@@ -12,29 +12,44 @@
...
 
+/* Ambient transition helpers: call `Sound_BeginAmbientTransition()` before
+    loading a new room (or when music/room transition starts). As looping
+    ambients are started they should be registered via
+    `Sound_RegisterAmbientFilenameWithCategory()` so they are protected from
+    cleanup. After the room's ambients have been started, call
+    `Sound_EndAmbientTransition()` to stop any ambients that were present
+    previously but not re-registered.
+ */
+typedef enum {
```
akirmse: enum, not typedef enum (as for structs)

clientd3d/sound.h
```diff
@@ -12,29 +12,44 @@
...
+      SOUND_CAT_ACTION = 0,  // one-shot action sounds
+      SOUND_CAT_MUSIC,       // background music (OGG)
+      SOUND_CAT_STEADY,      // steady sounds like fountains
+      SOUND_CAT_ATMOS,       // atmospheric sounds like birds
+      SOUND_CAT_UI           // UI/global FX
+} SoundCategory;
+
+void Sound_BeginAmbientTransition(void);
+void Sound_RegisterAmbientFilenameWithCategory(const char *filename, SoundCategory cat);
```
akirmse: Consider using std::string for filenames, you might find them easier to work with.  The parameter would then be ```const std::string &filename```

clientd3d/sound.c
adrienlaws: Removed - no issues see in testing when removed.

clientd3d/sound.c
adrienlaws: Fixed throughout the file

clientd3d/sound.c
adrienlaws: Changed - made a note to myself for future implementations - I keep tripping on this.

clientd3d/sound.c
adrienlaws: Removed - I was having trouble with my dev environment in identifying looping ambient sounds etc - anyway it was leftover and is now removed.

clientd3d/client.h
adrienlaws: Deleted

clientd3d/sound.h
adrienlaws: Changed

clientd3d/sound.h
adrienlaws: Removed all this as part of getting rid of the heuristic test stuff

clientd3d/sound.c
adrienlaws: Removed as part of refactoring

clientd3d/makefile
```diff
@@ -150,6 +159,9 @@ $(OUTDIR)\crc.obj : $(TOPDIR)\util\crc.c
 $(OUTDIR)\md5.obj : $(TOPDIR)\util\md5.c
 	$(CC) -Fo$(OUTDIR)/ $(CFLAGS) /Fd$(OUTDIR)\vc90.pdb /FpCpch -c $**
 
+$(OUTDIR)\stb_vorbis.obj : $(SOURCEDIR)\stb_vorbis.c
```
akirmse: I think it would be less duplication if you just added -DSTB_VORBIS_NO_PUSHDATA_API to CFLAGS, rather than a whole new rule for this one source file.  It's not like any other code is going to be affected by that symbol.

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
 #include "client.h"
 
+#include <algorithm>
 #include <cctype>
 #include <regex>
 #include <string>
+#include "audio_openal.h"
 
-static char music_dir[] = "resource";   /* Directory for sound files */
+static char music_dir[] = "resource";   /* Directory for music files */
```
akirmse: make const

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
+ * MIDI/MP3 file extensions are mapped to .ogg for compatibility.
  */
 
 #include "client.h"
 
+#include <algorithm>
 #include <cctype>
 #include <regex>
 #include <string>
+#include "audio_openal.h"
```
akirmse: Not needed, since this is in client.h

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
-
-	// start playing
-	AIL_start_sample(hseqImmediate);
-
-	debug(( "Playing midi music file %s. LatestMusic=%d \n", fname, latest_music ));
-	playing_midi = true;
-	return 0;
-#else
+   (void)hWndNotify;
+   if (!fname) return 1;
```
akirmse: Not sure what these return values mean, and the documentation is gone.  Could maybe return a bool now?

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
+   std::string filename(fname);
+   std::string lower = filename;
+   std::transform(lower.begin(), lower.end(), lower.begin(),
+                  [](unsigned char c){ return std::tolower(c); });
+   lower = std::regex_replace(lower, std::regex("\\.(mid|midi|mp3)$"), ".ogg");
+   if (MusicPlay(lower.c_str(), TRUE))
```
akirmse: true

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
-	{
-      if (paused_music == bg_music)
-      {
-         UnpauseMusic();
-         return 0;
-      }
-      isMusicPaused = false;
-	}
-   if (playing_midi)
+   (void)hWndNotify;
```
akirmse: Remove

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
+   if (!fname) return 1;
```
akirmse: Same thing about the return type here.

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
+   PlayMusicFileInternal(fname);
+   return playing_music ? 0 : 1;
+}
+
+static void PlayMusicFileInternal(const char *fname)
+{
+   if (!fname) return;
```
akirmse: Maybe also make the parameter a std::string?  Then you wouldn't need this check.

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
 /******************************************************************************/
-/*
- * MusicAbort:  Turn off currently playing music
- */
-void MusicAbort(void)
+void MusicDone(UINT device)
 {
-   if (!has_midi)
-      return;
+   (void)device;
```
akirmse: Remove

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
-	playing_midi = true;
-	return 0;
-#else
+   (void)hWndNotify;
+   if (!fname) return 1;
+   std::string filename(fname);
+   std::string lower = filename;
+   std::transform(lower.begin(), lower.end(), lower.begin(),
+                  [](unsigned char c){ return std::tolower(c); });
+   lower = std::regex_replace(lower, std::regex("\\.(mid|midi|mp3)$"), ".ogg");
```
akirmse: There are now C++ library functions to deal with filenames without having to do all of this manual string manipulation, if you're interested.  See

https://en.cppreference.com/w/cpp/filesystem/path/replace_extension.html

clientd3d/music.c
```diff
@@ -1,853 +1,257 @@
...
+   if ((filename = LookupNameRsc(rsc)) == NULL)
       return;
-   }
 
-#ifndef M59_MSS
-   DWORD dwReturn;
-   /* If NOT playing music and IF playing midi...*/
-   /* Stop the midi and prepare to restart a new one. */
-   if (playing_midi)
+   snprintf(fname, sizeof(fname), "%s\\%.*s", music_dir, FILENAME_MAX, filename);
```
akirmse: If you use std::filename::path, there's no messing with backslashes, etc.

https://en.cppreference.com/w/cpp/filesystem/path/append.html

clientd3d/music.h
```diff
@@ -12,23 +12,16 @@
...
-#endif
-
 void MusicInitialize(void);
 void MusicClose(void);
 
+/* Legacy compatibility shims for old protocol */
 DWORD PlayMidiFile(HWND hWndNotify, char *fname);
 M59EXPORT DWORD PlayMusicFile(HWND hWndNotify, const char *fname);
-M59EXPORT void PlayMidiRsc(ID rsc);
+M59EXPORT void PlayMidiRsc(ID rsc);  /* Legacy: forwards to PlayMusicRsc */
```
akirmse: Should this function be deleted?

clientd3d/sound.c
```diff
@@ -1,452 +1,164 @@
...
-	// Quick configuration escape for looping and random sounds.
-   if ((flags & SF_LOOP) && (!config.play_loop_sounds))
-   {
-      return 0;
-   }
-   if ((flags & SF_RANDOM_PLACE) && (!config.play_random_sounds))
-   {
-      return 0;
-   }
+	(void)hwnd; /* callback window unused for OpenAL path */
```
akirmse: Remove (is this AI generated or something?)

clientd3d/sound.c
```diff
@@ -1,452 +1,164 @@
...
 
-	int i, iRandomPitch;
-	void FAR *pSample;
-	debug(("Reading sound file %s.\n",fname));
-	// let Miles allocate the memory
-	pSample=AIL_file_read( fname, NULL );
+	if (!fname || fname[0] == '\0')
+		return 1;
```
akirmse: Document return type and probably change this value.

clientd3d/sound.c
```diff
@@ -1,452 +1,164 @@
...
- * SoundDoneCallback:  A sound has finished playing; send a windows message to invoke SoundDone.
- *						Would love to call SoundDone directly, but can't call mem_free_lock from
- *						within the callback, so post a Windows message instead
- */                             /*  */
-#ifdef M59_MSS
-void AILCALLBACK SoundDoneCallback( HSAMPLE S )
+void NewSound(WPARAM type, ID rsc)
 {
-	PostMessage( hMain, MM_WOM_DONE, (WPARAM) S, 0 );
+	(void)type; (void)rsc;
```
akirmse: Remove

clientd3d/sound.c
```diff
@@ -1,452 +1,164 @@
...
-		if( SampleHandle[i] != NULL )
-			if( AIL_sample_status( SampleHandle[i] ) != SMP_DONE )
-				AIL_end_sample( SampleHandle[i] );
-	}
-#else
-      /* Stop all playing sounds */
-      WaveMixFlushChannel(hMixSession, 0, WMIX_ALL | WMIX_NOREMIX);
-#endif
-   }
+	(void)px; (void)py;
```
akirmse: Remove

clientd3d/sound.h
```diff
@@ -12,29 +12,29 @@
...
-#endif
+                                     BYTE flags, int src_row, int src_col, int radius,
+                                     int max_vol);
 
 M59EXPORT void PlayWaveRsc(ID rsc, int volume, BYTE flags, int row, int col, int radius, int max_vol);
 M59EXPORT void PlayMidiRsc(ID rsc);
 M59EXPORT void PlayMusicRsc(ID rsc);
 
+/* Looping sound transition helpers: call `Sound_BeginLoopingSoundTransition()`
+   before loading a new room. As looping sounds are started they will be
```
akirmse: This is weird for the documentation of sound functions to be talking about rooms.  Try to talk about what the functions do, rather than how you expect them to be called.

clientd3d/audio_openal.h
```diff
@@ -0,0 +1,40 @@
...
+bool AudioInit(HWND hWnd);
+void AudioShutdown(void);
+
+// Music control (for MP3/OGG background music)
+bool MusicPlay(const char* filename, bool loop);
+void MusicStop(void);
+void MusicSetVolume(float volume); // 0.0 to 1.0
+bool MusicIsPlaying(void);
+
+// WAV sound playback (matches PlayWaveFile signature)
```
akirmse: This shouldn't be documented as matching a signature in another file.  If that other function ever changes, this one almost certainly won't, and then it will be super confusing why there is an erroneous comment here.

clientd3d/audio_openal.h
```diff
@@ -0,0 +1,40 @@
...
+bool SoundPlayWave(const char* filename, int volume, BYTE flags, 
+                   int src_row, int src_col, int radius, int max_vol);
+
+// Stop all sounds
+void SoundStopAll(void);
+
+// Stop any playing sources that are using the given filename's buffer
+void Audio_StopSourcesForFilename(const char* filename);
+
+// Listener position update for 3D audio (called each frame)
```
akirmse: Does this have to be called each frame?  What happens if you don't?  Confusing comment.

Document what this does, not how you expect it to be called.

akirmse (commented):
Not sure how I feel about putting stb_vorbis.c in the client directory.  Maybe it should be in its own top-level component, much like wavemix was?  It's basically 3rd party code and we don't want people messing with it while editing the client, so that we can update it when new versions come out without worrying about stomping on any local changes.

clientd3d/makefile
adrienlaws: Fixed - moved `stb_vorbis.c` to external/stb and also used global CFLAGS

clientd3d/music.c
adrienlaws: Removed, also removed from `clientd3d/audio_openal.c`

clientd3d/makefile
adrienlaws: I looked a little more into this.  The 32 in the name is a legacy artifact from Creative Labs' original SDK.  OpenAL Soft kept the name for their own compatibility reasons.

I cross checked by downloading the windows installation files ZIP [here](https://openal-soft.org/openal-binaries/openal-soft-1.24.3-bin.zip) from the official [OpenAL Soft website](https://openal-soft.org/#download).  If you go under libs and check both win32 and win64 folders, you can verify they both contain `OpenAL32.lib`.

clientd3d/audio_openal.h
adrienlaws: Removed as part of renaming `SoundPlayWave `to `SoundPlay`

Change the comment from
```
// WAV sound playback (matches PlayWaveFile signature)
```
 to 
```
// Returns true if sound started successfully; supports OGG and WAV formats,
// looping, and 3D positioning
```

clientd3d/music.h
adrienlaws: This is still being called by `HandlePlayMidi `in [server.c](https://github.com/Meridian59/Meridian59/blob/b6310725113f658ffb19b7a57601a2e156896af0/clientd3d/server.c#L1265) when the server sends `BP_PLAY_MIDI`. This forwards to `PlayMusicRsc`.  Maybe we can keep it unless I should change how server.c handles incoming `BP_PLAY_MIDI`?

clientd3d/music.c
adrienlaws: Removed

clientd3d/music.c
adrienlaws: Fixed - changed to `true`

clientd3d/music.c
adrienlaws: Fixed - did the same for `sound.c`

clientd3d/music.c
adrienlaws: Fixed - changed to bool.  MCI return values were in use previously and it was indeed super confusing because 0 was true and 1 was false.

clientd3d/music.c
adrienlaws: Fixed - changed to return false (bool) instead of return 1 (MCI)

clientd3d/music.c
adrienlaws: Removed

clientd3d/sound.c
adrienlaws: Removed.  Yes it gave me a recommendation that for parameters that are unused we should use (void) on them, but I am not doing that anymore.  It recommended that I suppress "unused parameter" warnings.

clientd3d/sound.c
adrienlaws: Removed

clientd3d/sound.c
adrienlaws: Removed because those parameters are now used for 3d position audio (like the fountain noise)

clientd3d/audio_openal.h
adrienlaws: Fixed - changed to:
 `// Sets the listener's position and facing direction for 3D audio`

clientd3d/music.c
adrienlaws: Implemented `std::filesystem::path::replace_extension()` instead of regex, as recommended, for the extension replacements.

clientd3d/music.c
adrienlaws: Fixed - changed to use std instead of snprintf

clientd3d/music.c
adrienlaws: Fixed - parameter is now std::string

clientd3d/sound.c
adrienlaws: Fixed - now returns bool with documentation

clientd3d/sound.h
adrienlaws: Fixed - removed from header file and fixed function documentation in the .c file to be specific about what it does

clientd3d/move.c
```diff
@@ -441,9 +441,8 @@ void UserMovePlayer(int action)
        last_splash = 0xFFFFFFFF;
    }
 
-   // Update looping sounds to reflect the player's new position
-//   debug(("Player now at: (%i,%i)\n",player.x >> LOG_FINENESS,player.y >> LOG_FINENESS));
-   UpdateLoopingSounds( player.x >> LOG_FINENESS, player.y  >> LOG_FINENESS );
+   // Update looping sounds to reflect the player's new position and orientation
```
akirmse: There are a few things I don't like about this particular implementation:

- The complex call to UpdateLoopingSounds is duplicated 5 times.
- If the user turns and moves in the same frame, UpdateLoopingSounds is called twice, which has an unknown penalty and just feels wrong.
- If another motion type is added in the future, the implementer is highly likely not to know to call this function, since movement and sound are not obviously related to each other.

I think a better approach would be to call this at most once at the end of each frame.  I don't know if it's important to avoid calling this if the player position hasn't changed.  If it is, you could keep track of the position from the previous frame and only call it if the position (or angle) has changed.

clientd3d/music.c
```diff
@@ -1,853 +1,243 @@
...
-static HDIGDRIVER hDigitalDriver;
-static HSAMPLE hseqBackground;
-static HSAMPLE hseqImmediate;
-static BYTE *pMIDIBackground;
-static BYTE *pMIDIImmediate;
-#endif
-
-enum {SOUND_MIDI, SOUND_MUSIC};
-typedef enum {BACKGROUND_MUSIC, GAMEPLAY_MUSIC} LoadMusicType;
+enum {SOUND_MUSIC};
```
akirmse: Used?

clientd3d/music.c
```diff
@@ -1,853 +1,243 @@
...
- *  Midi and music sounds are both played by the MIDI device; music is just a MIDI sound
- *  used as background music.  If a music sound is playing, a midi sound first pauses
- *  the background music; the music resumes when the midi sound is done.  When the music 
- *  ends, it automatically loops back to the beginning.
+ * MIDI/MP3 file extensions are mapped to .ogg for compatibility.
  */
 
 #include "client.h"
 
+#include <algorithm>
```
akirmse: Included in client.h so I don't think it's needed here.

clientd3d/music.c
```diff
@@ -1,853 +1,243 @@
...
-static bool has_midi = false;    /* Can system play MIDI files? */
+static const char music_dir[] = "resource";
 
-static bool playing_midi = false;  /* Is a MIDI file currently playing as an effect? */
-static bool playing_music = false;  /* Is a MIDI file currently playing as background? */
-static bool isMusicPaused = false;   /* Is background music paused? */
+static bool playing_music = false;  /* Is a music file currently playing as background? */
+static ID bg_music = 0;     /* Resource id of background music file; 0 if none */
+static ID latest_music = 0; /* Most recent music resource id */
+static char current_music_file[MAX_PATH] = ""; /* Filename of currently playing music */
```
akirmse: I think the code will be simpler if you make this a std::string.

clientd3d/music.h
akirmse: I think reducing the number of functions could help.  Here is the current list of functions that all look the same and make the interface super confusing:

MusicInitialize
AudioInit
PlayMidiFile
PlayMusicFile
PlayMidiRsc
PlayMusicRsc
MusicStart
MusicPlay
SoundPlay

Maybe you can delete some of these, and also make it clearer by naming and documentation how they're supposed to be used.

clientd3d/server.c
```diff
@@ -1262,6 +1262,10 @@ bool HandlePlayWave(char *ptr,long len)
    return true;
 }
 /********************************************************************/
+/*
+ * HandlePlayMidi: Legacy server protocol handler for BP_PLAY_MIDI.
+ * Kept for backward compatibility with server; forwards to PlayMusicRsc.
```
akirmse: This comment should not be talking about the internal details of PlayMidiRsc, which it shouldn't know anything about.

However, could you just call PlayMusicRsc directly and delete PlayMidiRsc?

clientd3d/sound.c
```diff
@@ -1,452 +1,181 @@
...
+// Forward to audio layer helper
+void Audio_StopSourcesForFilename(const char* filename);
 
-/******************************************************************************/
-/*
- * SoundInitialize:  Called at startup; initialize sound and MIDI devices.
- */
-void SoundInitialize(void)
+// Case-insensitive string comparison helper
+static bool iequals(const std::string &a, const std::string &b)
```
akirmse: Can you make this a utility function so that you can get rid of the calls to std::transform in various places?

clientd3d/sound.c
```diff
@@ -1,452 +1,181 @@
...
-        debug(("SoundInitialize: starting library\n"));
-
-	// it's ok to do this more than once, and we do,
-	//	once in MusicInitialize and once here
-	AIL_startup();
+/* Marks all currently tracked looping sounds for potential cleanup. */
+void Sound_BeginLoopingSoundTransition(void)
+{
+	// move current into prev
+	prev_looping = std::move(curr_looping);
```
akirmse: I'm actually not sure what this does but the use of std::move is surprising here.  Should this be simply prev_looping = curr_looping?

clientd3d/sound.c
```diff
@@ -1,452 +1,181 @@
...
-	debug(("%s\n",buf));
+	// Only clear prev - keep curr so they become prev on next transition
+	prev_looping.clear();
+	looping_cleanup_pending = false;
+}
 
-   return DIG;
+void SoundInitialize(void)
+{
+	if (AudioInit(hMain))
```
akirmse: How about just wave_open = AudioInit(hMain) ?

clientd3d/sound.c
```diff
@@ -1,452 +1,181 @@
...
-		return 1L;
-	}
+	if (!fname || fname[0] == '\0')
+		return false;
 
-	// Find an HSAMPLE to play it
-	for( i=0; i < iNumVoices; i++ )
+	/* If name appears to be a bare filename (no path), try resource dir first
+	   to avoid noisy "cannot open" logs for the working directory lookup. */
+	bool has_path = (strchr(fname, '\\') != NULL) || (strchr(fname, '/') != NULL) || (strchr(fname, ':') != NULL);
```
akirmse: There's got to be a cleaner way to do this with std::filesystem::path.

clientd3d/sound.c
```diff
@@ -1,452 +1,181 @@
...
-		if( SampleHandle[i] != NULL )
-			if( AIL_sample_status( SampleHandle[i] ) != SMP_DONE )
-				AIL_end_sample( SampleHandle[i] );
-	}
-#else
-      /* Stop all playing sounds */
-      WaveMixFlushChannel(hMixSession, 0, WMIX_ALL | WMIX_NOREMIX);
-#endif
-   }
+	double radians = (double)angle * (2.0 * 3.14159265358979) / 4096.0;
```
akirmse: Use symbolic constants instead of hard-coding numbers.

clientd3d/sound.h
```diff
@@ -12,31 +12,25 @@
...
-                            int max_vol);
-
-#ifdef M59_MSS
-void AILCALLBACK SoundDoneCallback( HSAMPLE S );
-void SoundDone(  HWND hwnd, WPARAM wParam, LPARAM lParam  );
-static HDIGDRIVER WAVE_init_driver( DWORD rate, WORD bits, WORD chans );
-#else
-void SoundDone(HWND hwnd, int iChannel, LPMIXWAVE lpMixWave);
-#endif
+bool PlayMidiFile(HWND hWndNotify, char *fname);
```
akirmse: This is also declared in music.h?

clientd3d/sound.h
```diff
@@ -12,31 +12,25 @@
...
+M59EXPORT bool PlayWaveFile(HWND hwnd, const char *fname, int volume,
+                                     BYTE flags, int src_row, int src_col, int radius,
+                                     int max_vol);
 
 M59EXPORT void PlayWaveRsc(ID rsc, int volume, BYTE flags, int row, int col, int radius, int max_vol);
 M59EXPORT void PlayMidiRsc(ID rsc);
 M59EXPORT void PlayMusicRsc(ID rsc);
```
akirmse: Some of these too.

clientd3d/sound.h
```diff
@@ -12,31 +12,25 @@
...
 
+void Sound_BeginLoopingSoundTransition(void);
+void Sound_RegisterLoopingSound(const std::string &filename);
+void Sound_EndLoopingSoundTransition(void);
```
akirmse, line 32: These need comments.. I don't understand how I'm supposed to use these.

clientd3d/music.c
```diff
@@ -1,853 +1,243 @@
...
+ * MIDI/MP3 file extensions are mapped to .ogg for compatibility.
  */
 
 #include "client.h"
 
+#include <algorithm>
 #include <cctype>
-#include <regex>
+#include <filesystem>
 #include <string>
```
akirmse: Already included in client.h.

clientd3d/music.c
```diff
@@ -1,853 +1,243 @@
...
 }
+
 /******************************************************************************/
-/*
- * PlayMusicFile:  Play given MIDI file and notify given window when done.
- *   If background music had been playing, it picks up where it left off.
- *   Returns 0 if successful, MCI error code otherwise.
- */
-DWORD PlayMusicFile(HWND hWndNotify, const char *fname)
+// Returns true if music started successfully.
```
akirmse: I'm not 100% sure this comment is correct.  If music was already playing, and starting this new file fails, will this always return false?

clientd3d/music.c
```diff
@@ -1,853 +1,243 @@
...
-   DWORD dwReturn;
-   MCI_PLAY_PARMS mciPlayParms;
-   char temp[81];
-   debug(( "Play music file %s. LatestMusic=%d, playing_music=%d, playing_midi=%d \n", 
-            fname, latest_music, playing_music, playing_midi ));
-   // If already playing music, pick up where we left off
-   if (isMusicPaused)
+
+   // If that failed and filename is a bare name, try prefixed with music_dir
+   bool has_path = (filename.find('\\') != std::string::npos) ||
```
akirmse: In C++20 there is now a nicer function str::ends_with.  However I wonder if there's a better function in std::filesystem::path to see whether the filename is a bare name.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,767 @@
...
+ */
+static ALuint LoadOGGFile(const char* filename, bool useCache)
+{
+   ALuint buffer = 0;
+   int channels, sample_rate;
+   short* decoded;
+   int num_samples;
+   ALenum format;
+
+   if (useCache)
```
akirmse: It looks like the handling of the cache is duplicated in LoadWAVFile, which is to be avoided.  See if you can rewrite these two functions so that the cache handling is done only once, and only the file parsing code is separate.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,767 @@
...
+   free(decoded);
+
+   if (alGetError() != AL_NO_ERROR)
+   {
+      debug(("LoadOGGFile: Failed to upload buffer data\n"));
+      alDeleteBuffers(1, &buffer);
+      return 0;
+   }
+
+   if (useCache && g_cacheCount < MAX_AUDIO_BUFFERS)
```
akirmse: This cache implementation is a bit suspect.  It the cache ever fills up, it basically stops working, and I think (?) old entries are never removed.

This is a little complicated but what you want here is an LRU (least-recently-used) cache, where the oldest cache entry is removed when the cache gets filled.  Also rather than an array of BufferCacheEntry, the cache should be implemented as a std::unordered_map<string, ALuint> (i.e. a hash table of filename to OpenAL buffer ID).

Tricky things to watch out for:
- The hash function for the hash table should be case insensitive on the filename.
- STL doesn't have an LRU cache built-in; it can be built with a std::unordered_map plus a std::list, with lots of examples online.
- When removing old items from the cache, it's not clear to me how to free up OpenAL resources.  Obviously you don't want to delete an OpenAL buffer if it's still being actively played but I don't understand the code well enough to know how OpenAL works here.

You would learn a lot about data structures implementing this, but it's not trivial.  Up to you.  But the current behavior of the cache filling up isn't great, and also the memory usage is sub-optimal where the cache is statically allocated at max size and the filenames are stored as maximally-long buffers.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,767 @@
...
+   DWORD chunkSize, riffSize;
+   WORD audioFormat = 0, numChannels = 0, bitsPerSample = 0;
+   DWORD sampleRate = 0;
+   BYTE* data = NULL;
+   DWORD dataSize = 0;
+   ALuint buffer = 0;
+   ALenum format;
+   bool foundFmt = false;
+   bool foundData = false;
+   long riffEnd;
```
akirmse: In general try to declare variables where they're first used, rather than putting them all at the start of the function (which is required by C but is now 30 years obsolete in C++).

neiltaco:
Just wanted to comment that this has a lot of potential and I'm excited to see the new audio system someday!  :)

clientd3d/sound.h
adrienlaws: I put comments in the sound.c file for these functions.  I'm not sure when to put comments in the .c or the .h file.  I was wary of having the comments in 2 places to avoid ending up with different comments for the same function over time.

Should I move these comments from sound.c over to sound.h?

/* Marks all currently tracked looping sounds for potential cleanup. */
void Sound_BeginLoopingSoundTransition(void)

/* Registers a looping sound as active, protecting it from cleanup. */
void Sound_RegisterLoopingSound(const std::string &filename)

/* Stops any looping sounds that were not re-registered since the last Begin call. */
void Sound_EndLoopingSoundTransition(void)

akirmse: Ideally the comments would be in the .h file so that callers could look only at the interface without having to look at the implementation.  We don't do a good job of this in the codebase.

clientd3d/audio_openal.c
adrienlaws: Fixed for this function - moved variable declarations to when they are used

clientd3d/sound.c
adrienlaws: Changed as recommended - much cleaner.

clientd3d/sound.h
adrienlaws: Removed

clientd3d/sound.h
adrienlaws: Removed

clientd3d/music.c
adrienlaws: Removed `<string>` include

clientd3d/music.c
adrienlaws: Removed `<algorithm>` include

clientd3d/music.c
adrienlaws: Fixed - using `has_parent_path()` now

clientd3d/sound.c
adrienlaws: Fixed - using DegToRad() from drawdefs.h now since that was available

clientd3d/sound.c
adrienlaws: Fixed - using simple copy now as suggested

clientd3d/sound.h
adrienlaws: Fixed - moved comments to .h file and made a note going forward.  Removed from .c file so we don't have to worry about drift between the .h and .c comments.

clientd3d/server.c
adrienlaws: Fixed - now HandlePlayMidi calls PlayMusicRsc directly as suggested.  PlayMidiRsc has been deleted.

clientd3d/music.h
adrienlaws: Fixed -  consolidated the confusingly named music functions from 9 to 6

Removed:

- `MusicInitialize `- folded into `AudioInit
`
- `PlayMidiFile `- `PlayMusicFile` already handles .mid to .ogg conversion
- `PlayMidiRsc `- `HandlePlayMidi `now calls `PlayMusicRsc `directly
- `MusicDone `- was an empty stub

Renamed:
- `MusicStart `to `MusicRestart `- clarifies it resumes music when user re-enables in settings

Current State of the 6 remaining functions from the original 9 mentioned above:
Function  | Description
-- | --
AudioInit | Initialize OpenAL
MusicPlay | Play OGG music via
SoundPlay | Play WAV/OGG sound effects
PlayMusicFile | High-level: path resolution, .mid/.mp3 → .ogg
PlayMusicRsc | Play music by resource ID
MusicRestart | Resume music when re-enabled in settings


Note: Sound effect functions (`GamePlaySound`, `PlayWaveRsc`, `PlayWaveFile`) are in `sound.c` and weren't part of this consolidation.  They already seem pretty straightforward, purpose wise. 

I also am creating an `audio.md` documentation file that I will add as part of this PR.

Here's my current diagram for the audio system refactor.  I find diagrams easier for this since there's so much and I can get disoriented.

```mermaid
graph TD
    subgraph "Server Protocol"
        BP_PLAY_MIDI[BP_PLAY_MIDI]
        BP_PLAY_MUSIC[BP_PLAY_MUSIC]
        BP_PLAY_WAVE[BP_PLAY_WAVE]
    end

    subgraph "Protocol Handlers (server.c)"
        HandlePlayMidi[HandlePlayMidi]
        HandlePlayMusic[HandlePlayMusic]
        HandlePlayWave[HandlePlayWave]
    end

    subgraph "Music API (music.c)"
        PlayMusicRsc[PlayMusicRsc]
        PlayMusicFile[PlayMusicFile]
        MusicRestart[MusicRestart]
    end

    subgraph "Sound API (sound.c)"
        GamePlaySound[GamePlaySound]
        PlayWaveRsc[PlayWaveRsc]
    end

    subgraph "OpenAL Engine (audio_openal.c)"
        AudioInit[AudioInit]
        MusicPlay[MusicPlay]
        SoundPlay[SoundPlay]
        AudioShutdown[AudioShutdown]
    end

    subgraph "Audio Hardware"
        OpenAL[OpenAL Soft]
    end

    BP_PLAY_MIDI --> HandlePlayMidi
    BP_PLAY_MUSIC --> HandlePlayMusic
    BP_PLAY_WAVE --> HandlePlayWave

    HandlePlayMidi --> PlayMusicRsc
    HandlePlayMusic --> PlayMusicRsc
    HandlePlayWave --> GamePlaySound

    PlayMusicRsc --> PlayMusicFile
    PlayMusicFile --> MusicPlay
    MusicRestart --> PlayMusicRsc

    GamePlaySound --> PlayWaveRsc
    PlayWaveRsc --> SoundPlay

    MusicPlay --> OpenAL
    SoundPlay --> OpenAL
```

clientd3d/music.c
adrienlaws: Removed (leftover from MSS)

clientd3d/music.c
adrienlaws: Fixed - changed to std::string. Much cleaner.

clientd3d/sound.c
adrienlaws: Fixed - now uses `fs::path(fname).has_parent_path()`

clientd3d/audio_openal.c
adrienlaws: Fixed (or at least attempted) -  replaced the static array with an LRU cache based on the [LeetCode 146 LRU problem example solutions](https://algo.monster/liteproblems/146)

Using `std::unordered_map` and `std::list`:

- Hash map stores filename -> `list::iterator` for O(1) lookup
- List maintains recency order; `splice()` moves accessed entries to front
- Case insensitive hash through `std::hash<std::string>` on lowercased copy
- Dynamic allocation

Differences / variations from the standard LeetCode 146 solutions for this implementation:
- No dummy sentinel nodes - std::list handles empty list edge cases
- Custom case-insensitive hash/equality for Windows filename matching
- Eviction has an extra` IsBufferInUse()` check - skips buffers actively playing before deleting (it'll error otherwise)

clientd3d/audio_openal.c
adrienlaws: Fixed - consolidated into `LoadAudioBuffer()` which handles cache check/store once, then dispatches to `ParseOGGFile() `or` ParseWAVFile()` for decoding based on the format (WAV or OGG).

Before:
``` mermaid
graph TD
    subgraph "Before - Duplicated Cache"
        SoundPlay1[SoundPlay] --> LoadOGG[LoadOGGFile]
        SoundPlay1 --> LoadWAV[LoadWAVFile]
        
        LoadOGG --> CacheCheck1[Check Cache]
        LoadOGG --> ParseOGG1[Parse OGG]
        LoadOGG --> CacheStore1[Store in Cache]
        
        LoadWAV --> CacheCheck2[Check Cache]
        LoadWAV --> ParseWAV1[Parse WAV]
        LoadWAV --> CacheStore2[Store in Cache]
    end
    
    style CacheCheck1 fill:#f66
    style CacheStore1 fill:#f66
    style CacheCheck2 fill:#f66
    style CacheStore2 fill:#f66
```

After
``` mermaid
graph TD
    subgraph "After - Consolidated Cache"
        SoundPlay2[SoundPlay] --> LoadAudio[LoadAudioBuffer]
        
        LoadAudio --> CacheCheck[FindCachedBuffer]
        CacheCheck -->|miss| Dispatch{Extension?}
        Dispatch -->|.ogg| ParseOGG2[ParseOGGFile]
        Dispatch -->|else| ParseWAV2[ParseWAVFile]
        ParseOGG2 --> CacheStore[CacheBuffer]
        ParseWAV2 --> CacheStore
        CacheCheck -->|hit| Return[Return buffer]
    end
    
    style CacheCheck fill:#6f6
    style CacheStore fill:#6f6
```

clientd3d/music.c
adrienlaws: Yes - `MusicStop()` is called before attempting the new track, which clears `playing_music`. If `MusicPlay()` then fails, `playing_music `remains `false`, so we return `false`. Updated the `PlayMusicFile ` comment to clarify this and explain the edge case of an old file being stopped but new one not starting properly.

> Returns true if the new music file started playing, false if it failed to load or play. Note: any previously playing music is stopped before attempting to load the new file, regardless of success or failure.

clientd3d/sound.c
adrienlaws: Done - added `iequals()` to `util.c` / `util.h` and updated music.c to call it

clientd3d/move.c
adrienlaws: Fixed - removed the 5x calls from `move.c` and added one call at the end of `AnimationTimerProc`.  Seems to work in testing.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,843 @@
...
+// LRU buffer cache: list ordered by recency (front = newest), map for O(1) lookup.
+struct CacheNode {
+   std::string filename;
+   ALuint buffer;
+};
+
+struct CaseInsensitiveHash {
+   size_t operator()(const std::string& s) const {
+      std::string lower;
+      lower.reserve(s.size());
```
akirmse: You really don't want to be allocating memory during the computation of a hash function; that more than eliminates all the performance gains of using a hash table vs. a vector.  You should be able to find something online, perhaps using c_str() to avoid the allocation.

Another option would be to always convert the filenames to lowercase before interacting with the hash table.  Kind of fragile though---easy to forget in the future if adding new code.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,843 @@
...
+
+   // Destroy context
+   if (g_context)
+   {
+      alcMakeContextCurrent(NULL);
+      alcDestroyContext(g_context);
+      g_context = NULL;
+   }
+
+   // Close device
```
akirmse: This kind of comment doesn't really add anything.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,843 @@
...
+      g_device = NULL;
+   }
+
+   g_initialized = false;
+   debug(("AudioShutdown: OpenAL shut down\n"));
+}
+
+/*
+ * IsBufferInUse: Returns true if the buffer is currently being played by any source.
+ * Used to prevent evicting buffers that are actively playing.
```
akirmse: Document what the function does, not how you expect it to be called.  You don't know whether it's going to be used to prevent evicting buffers in the future.

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,843 @@
...
+
+   // Determine if sound should use 3D positional audio or play centered on player.
+   // Non-positional (centered): SF_RANDOM_PLACE, or SF_LOOP at placeholder coords (<=2,<=2)
+   // Positional (3D): all other sounds with valid coordinates
+   bool isAmbientLoop = (flags & SF_LOOP) && (src_row <= 2) && (src_col <= 2);
+   bool isPositional = (src_row > 0 || src_col > 0) && 
+                       !(flags & SF_RANDOM_PLACE) &&
+                       !isAmbientLoop;
+
+   // Convert fine coords to tile coords if needed
```
akirmse: This is quite odd---what's going on here?  You should take coordinates in known and documented units.

clientd3d/game.c
```diff
@@ -629,9 +636,10 @@ void GamePlaySound(ID sound_rsc, ID source_obj, BYTE flags, WORD y, WORD x, WORD
       {
 	 volume = maxvolume - (distance * maxvolume / cutoff) ;
       }
-      // debug(("Distance = %i\n",distance));
-      // debug(("Setting volume to %i.\n",volume));
+      debug(("GamePlaySound: distance=%i, cutoff=%i\n", distance, cutoff));
    }
+   debug(("GamePlaySound: volume=%i, flags=0x%02X, row=%i, col=%i\n",
```
akirmse: Don't these debugs get spammy?

clientd3d/music.c
```diff
@@ -1,853 +1,207 @@
-// Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
-// All rights reserved.
-//
-// This software is distributed under a license that is described in
-// the LICENSE file that accompanies it.
-//
-// Meridian is a registered trademark.
+// Meridian 59
```
akirmse: Please restore the copyright message here.

clientd3d/music.c
```diff
@@ -1,853 +1,207 @@
...
-#else
-   DWORD dwReturn;
-   MCI_PLAY_PARMS mciPlayParms;
-   char temp[81];
-   debug(( "Play music file %s. LatestMusic=%d, playing_music=%d, playing_midi=%d \n", 
-            fname, latest_music, playing_music, playing_midi ));
-   // If already playing music, pick up where we left off
-   if (isMusicPaused)
+
+   // If that failed and filename is a bare name, try prefixed with music_dir
```
akirmse: This seems strange to attempt to play a music file twice with different path names.  Why is this necessary?  Could we change something (maybe in the callers) so that only one attempt is necessary?

clientd3d/music.c
```diff
@@ -1,853 +1,207 @@
...
-   }
-   // Always POST new music message if using MCI
-   PostMessage(hMain, BK_NEWSOUND, SOUND_MIDI, rsc);
-   return;
-#endif
-   if (playing_midi)
+   std::string fname = ConvertLegacyMusicExtension((fs::path(music_dir) / filename).string());
+
+   // Check if this is the same music file already playing (by filename, not resource ID)
+   // Different rooms may use different resource IDs for the same music file
```
akirmse, line 117: Good comment!  Explains *why*

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,843 @@
...
+   int channels, sample_rate;
+   short* decoded;
+   int num_samples;
+   ALenum format;
+
+   debug(("ParseOGGFile: Attempting to load %s\n", filename));
+
+   // Check if file exists
+   FILE* test_file = NULL;
+   if (fopen_s(&test_file, filename, "rb") != 0 || !test_file)
```
akirmse: You can test whether a file exists much more efficiently that opening it and then closing it (which allocates buffers, etc.).  The old C function to do it is called stat() but I bet there's a better standard C++ STL function now.

adrienlaws: Fixed - replaced the `fopen_s`/`fclose` pattern with `std::filesystem::exists`.  No more buffer allocation.

clientd3d/music.c
adrienlaws: Fixed - removed the fallback.  It was useless/dead code.

Checked all 3 callers. `PlayMusicRsc `and `NewMusic `already prepend `music_dir`, and `splash_music `is defined with the full path.

clientd3d/music.c
adrienlaws: Restored - sorry.

clientd3d/game.c
adrienlaws: Fixed - remove these debugs and all others related to this PR that are not failure debugs.  The debug messages were very useful for me during implementation but they are now removed to avoid spamming people.

clientd3d/audio_openal.c
adrienlaws: Fixed - OpenAL takes tile coords.  `GamePlaySound` now normalizes to tile coords before calling.  Updated the header comment for expected units and also updated audio.md documentation for this implementation.

clientd3d/audio_openal.c
adrienlaws: Fixed - updated this comment.

clientd3d/audio_openal.c
adrienlaws: Fixed - remove this and a lot of other useless comments.  They helped during breadcrumbing the design and keeping track, but useless now.  Removed all but 3 that I thought still were helpful.

clientd3d/audio_openal.c
adrienlaws: Fixed - now hashes per character without allocation. Uses the `hash_combine` formula from boost (XOR with golden ratio constant + bit shifts):
- https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
- https://stackoverflow.com/questions/4948780/magic-number-in-boosthash-combine

davidklaw:
Can't comprehend the code changes made, but (when testing) everything sounds great thus far from a player perspective!

clientd3d/audio_openal.c
```diff
@@ -0,0 +1,805 @@
...
+      if (memcmp(chunkID, "fmt ", 4) == 0)
+      {
+         if (chunkSize < 16)
+         {
+            debug(("ParseWAVFile: fmt chunk too small: %s\n", filename));
+            break;
+         }
+         WORD blockAlign;
+         DWORD byteRate;
+         fread(&audioFormat, 2, 1, file);
```
akirmse: You should check for errors on file reads.

clientd3d/music.c
```diff
@@ -6,848 +6,195 @@
 //
 // Meridian is a registered trademark.
 /*
```
akirmse, line 8: Put the copyright message here.

clientd3d/sound.c
```diff
@@ -1,452 +1,161 @@
...
-#define SOUND_MAX_VOICES 16
-static int iNumVoices;
-	// user defined data associated with sample handles
-#define SOUND_USER_ADDRESS 0	// memory address of sample
-#define SOUND_USER_X 1			// x and y position of sound source
-#define SOUND_USER_Y 2
-#define SOUND_USER_RADIUS 3		// maximum distance sound is audible
-#define SOUND_USER_MAXVOL 4		// maximum volume of sound (when distance = 0)
-static HSAMPLE SampleHandle[SOUND_MAX_VOICES];	// array of sample handles
+// Meridian 59 sound.c - sound effects via OpenAL Soft
```
akirmse: Add copyright message.

clientd3d/music.c
adrienlaws: I think this `music.c` file's copyright message is present and unchanged.  It doesn't look like I modified it in lines 1-5, unless I am misunderstanding something.

clientd3d/sound.c
adrienlaws: Fixed - I put it back sorry for this copyright chaos.

clientd3d/audio_openal.c
adrienlaws: Fixed - added error checking and debug messages for these `fread`s

PR #1301, opened 2025-12-24
Add map annotation text

CyberBill:
This adds text to map annotations on the map and minimap. The background text is drawn 4 times in black to create visual contrast and make the text very readable.

See screenshot:
image

Text was also reformatted to use 3-spaces instead of tabs.

### DIFF
--- a/clientd3d/config.h
+++ b/clientd3d/config.h
@@ -22,7 +22,7 @@
 
 #define CONFIG_MAX_VOLUME 100  // Max value of sound / music volume settings
 
-#define CONFIG_MAX_TEXT_ZOOM_LIMIT 100  // Max value of map text zoom limit
+static const int CONFIG_MAX_TEXT_ZOOM_LIMIT = 100;  // Max value of map text zoom limit
 
 // Communication settings
 typedef struct {
--- a/clientd3d/map.c
+++ b/clientd3d/map.c
@@ -224,7 +224,7 @@ void MapDraw( HDC hdc, BYTE *bits, AREA *area, room_type *room, int width, bool 
 
 	    MapDrawWalls(hdc, xoffset, yoffset, scale, room);
        if (config.map_annotations)
-          MapDrawAnnotations(hdc, room->annotations, xoffset, yoffset, scale, FALSE, true);
+          MapDrawAnnotations(hdc, room->annotations, xoffset, yoffset, scale, false, true);
 	    MapDrawObjects(hdc, room->contents, xoffset, yoffset, scale);
 	    MapDrawPlayer(hdc, xoffset, yoffset, scale);
 	 }
@@ -278,7 +278,7 @@ void MapDraw( HDC hdc, BYTE *bits, AREA *area, room_type *room, int width, bool 
              drawText = scaleMiniMap > (1.0f / ((float)config.map_text_zoom_limit * MAX_ANNOTATION_TEXT_ZOOM));
           }
           
-          MapDrawAnnotations(hdc, room->annotations, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap, TRUE, drawText);
+          MapDrawAnnotations(hdc, room->annotations, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap, true, drawText);
        }
 	    MapDrawObjects(hdc, room->contents, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap);
 	    MapDrawPlayer(hdc, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap);


davidklaw:
Great to see contributions from you again @CyberBill! A few more thoughts from me:
* These labels likely deserve their own font constant and to be modifiable from the menu and config. Locally, I found System, Bold, size 10 (picture below) to be acceptable. Not great, but acceptable (legible). I added padding to the `RECT` as well, so that the shadows wouldn't appear clipped. I guess one of the font string values is `quality` and I have it at 5. Not sure it's making much of a difference. Not sure if we can reliably use System either, but sharing my findings.
* I think they ought to go away at a certain zoom scale, because as you know, it's a mess when zoomed out.
* Wondering, should there be yet another map option to show/hide these labels?
image

CyberBill:
Updated - added a new font & UI to change it. I went with 10pt Arial. I don't believe 'System' is a preinstalled font on Windows.

I also made them go away based on the 'map scale' number of 0.0075 (minimap only), which feels OK to me. I tried two other methods that didn't work well - One was to monitor how many items were on screen at the same time and set a max - but this was really jarring because you'd walk around the map and they would pop in and out. I also tried using various bits of data for the scale ('zoom', text size, etc), but those don't take into account the actual size of a room - and it seems to me that the number of map annotations will scale with room size. I made the text always visible in map mode. Probably needs some play testing and feedback.

Me personally, I don't think a dedicated option is necessary - because what is the point of seeing a marker on the map if you don't know what it is? And if you just need a marker and don't want the text, you can use a single letter or period or space. That said - if you want one, I can add it, it's up to you.

image

davidklaw:
I like the way this ended up! About as good as it could be visually, I think. Nice work.

clientd3d/map.c
```diff
@@ -514,36 +514,87 @@ void MapDrawPlayer(HDC hdc, int x, int y, float scale)
...
+         OffscreenWindowBackground(NULL, new_x - (adjusted_size / 2), new_y - (adjusted_size / 2), annotation.width,
+                                   annotation.height);
+         OffscreenStretchBlt(hdc, (int) (new_x - (adjusted_size / 2)), (int) (new_y - (adjusted_size / 2)),
+                             adjusted_size, adjusted_size, annotation.bits, 0, 0, annotation.width, annotation.height,
+                             OBB_COPY | OBB_FLIP);
+      }
+   }
+
+   // Draw Map Annotation Text
+   if(bDrawText)
```
akirmse: Space before (

clientd3d/map.c
```diff
@@ -259,7 +259,7 @@ void MapDraw( HDC hdc, BYTE *bits, AREA *area, room_type *room, int width, bool
 	    else
 	       MapDrawWalls(hdc, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap, room);
        if (config.map_annotations)
-          MapDrawAnnotations( hdc, room->annotations, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap, TRUE );
+          MapDrawAnnotations(hdc, room->annotations, xoffsetMiniMap, yoffsetMiniMap, scaleMiniMap, TRUE, scaleMiniMap > 0.0075f);
```
akirmse: Let's have a named constant for this rather than a hard-coded number.  Or maybe for its reciprocal, which has a more intuitive meaning (min zoom ratio or something?).

clientd3d/map.c
CyberBill: Fixed!

clientd3d/map.c
CyberBill: Done!

akirmse:
I've heard some feedback that hardcore players are going to complain that this text obscures their view of the map, which is critical to their PvP (i.e. seeing other players, etc.), and I think there's something to that.  I hate adding options but if we do add this we will have to have a way to turn it off.  Or maybe ask around and see if hardcore people are OK with turning off annotations entirely.

CyberBill:
> I've heard some feedback that hardcore players are going to complain that this text obscures their view of the map, which is critical to their PvP (i.e. seeing other players, etc.), and I think there's something to that. I hate adding options but if we do add this we will have to have a way to turn it off. Or maybe ask around and see if hardcore people are OK with turning off annotations entirely.

Sure thing - I'll see what I can do. May add a boolean to enable map annotation text, or I might add a slider. Depends on whether I can find a place in the Preferences screen for it.

CyberBill:
I had an idea, and ended up moving the "Map Annotations" boolean over to the "Map" preferences, and then adding a slider for the 'Text Zoom Limit'. Code not yet updated, but here's a preview of the UI:

image

CyberBill:
Code updated. I decided to make "all the way to the left" force the map annotation text off, and "all the way to the right" forces the map annotation text on. Anywhere in the middle uses the same algorithm as before, but now scaled with the slider value (1-99). I'm still working on tweaking the magic number, but it seems OK.

The text is always shown on the full size map, since there is plenty of space available (usually).

CyberBill:
@akirmse @davidklaw Pinging on this one again for feedback.

clientd3d/config.h
```diff
@@ -22,6 +22,8 @@
 
 #define CONFIG_MAX_VOLUME 100  // Max value of sound / music volume settings
 
+#define CONFIG_MAX_TEXT_ZOOM_LIMIT 100  // Max value of map text zoom limit
```
akirmse: static const int

clientd3d/map.c
```diff
@@ -223,7 +224,7 @@ void MapDraw( HDC hdc, BYTE *bits, AREA *area, room_type *room, int width, bool
 
 	    MapDrawWalls(hdc, xoffset, yoffset, scale, room);
        if (config.map_annotations)
-          MapDrawAnnotations(hdc, room->annotations, xoffset, yoffset, scale, FALSE );
+          MapDrawAnnotations(hdc, room->annotations, xoffset, yoffset, scale, FALSE, true);
```
akirmse: FALSE -> false

CyberBill: Fixed (in both places)

clientd3d/config.h
CyberBill: Fixed!

PR #1302, opened 2025-12-25
Fix Map Border

CyberBill:
This contains two fixes for the view border. First, it fixes the border of the view when rendering the full size map. It was previously not displaying the corner elements at all. Second, it fixes the rendering scale of the corner elements when in hardware rendering mode.

Note: The changes in d3drenderer.c don't show up well on Github unless you're in 'split' mode. The code was refactored - instead of copying/pasting 4 large chunks of code for each corner piece, it's now inside of a loop.

image

image


akirmse:
Thanks for the fix, and also for simplifying the repetitive rendering code.

davidklaw:
Excited to see this addressed. I struggled with an imperfect solution for a while so I'm dialed into the details and wanted to point out that it breaks down a bit at different game window sizes. However, it's still a solution that works well!

Comparing the lower left to the upper left, you can see that the yellow circular shape is squished or stretched and clipped on the bottom. It's most obvious when toggling between map mode (`SHIFT+M`). But having looked at this myself, I got the impression that perfection wasn't possible.

https://github.com/user-attachments/assets/071ef183-a371-4d53-b44d-4d5235c9a383

PR #1303, opened 2025-12-25
Fix corrupt map annotations data

CyberBill:
Updating code to detect and fix corrupt annotation blocks within the map file. When we read in the map annotation data for a room, when we get to the offset in the file where it lives, we call the MapFileFindNextKnownData() function, which reads and processes the entire map file, looking for the next piece of known data (rooms or other map annotations), so that we can bound the read function to only the data within our block. This prevents us from 'over-reading' other data and interpreting it as map annotations.

I also added code that validates the actual text within an annotation - confirming that it is only using ASCII or Latin printable characters.

On the saving side, when we are outputting annotations we do the same check as when reading, this time if we detect that there isn't enough space to output our annotations and instead create a new block of annotations. This does potentially leave a useless blob of data (old annotations), but it will only be as large as a single annotation, because that is what was being output before. I've also updated the output, to ensure that all map annotation data is written before writing the number of annotations.

Added debug output whenever MapFile functions fail, such as MapFileLoadRoom, MapFileSaveRoom, etc.

All calls to fseek() are now wrapped in checks, so that the functions will fail if the seek fails.

### DIFF
--- a/clientd3d/map.c
+++ b/clientd3d/map.c
@@ -565,12 +565,6 @@ void MapEnterRoom(room_type *room)
    {
       debug(("MapEnterRoom:  Couldn't load map for room!\n"));
    }
-#ifndef M59_RETAIL
-   if (!MapFileValidateAllRooms())
-   {
-      debug(("Map file validation failed!!\n"));
-   }
-#endif
 
    fMapCacheValid = FALSE;
 }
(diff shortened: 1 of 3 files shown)


akirmse:
Hm... I wonder if we really need this.  In practical terms, new annotations are very rarely added, and there is only a problem in the last few months in certain cases I think (adding an annotation to a room that doesn't already have one?).  If only a few users are affected, they could just get a new map file after the next update.

I'm a little reluctant to take on so much fixup code that would have to live forever.  Though the real bug fixes and improvements I'd be happy to take.

CyberBill:
No longer needed - went with the one-time-upgrade code instead.

PR #1304, opened 2025-12-27
Add rainy and snowy weather

neiltaco:
Reupload of #1282 to resync the PR since it was far behind from upstream/master.

This PR adds rain and snow weather for both software and hardware rendering. The focus is on adding the visual effects of rain/snow for cosmetic purposes.  It has also includes refactoring of the particles, skybox, and background overlay systems for improved performance and readability.

This PR is also dependent on https://github.com/Meridian59/Meridian59/pull/1306, which adds the DM weather commands and automated weather system, and the packet data for updating room weather for the client.

To test out this PR, you'll need to have a branch that includes both #1304 and #1306.  The skyboxes for hardware rendering are also available at this Discord link:  https://discord.com/channels/650824489885630475/1092844835980980324/1469427478505328783

Software Renderer Video
https://github.com/user-attachments/assets/922bdeb2-41b0-42ea-9a9b-34c19de558cd

Hardware Renderer Video
https://github.com/user-attachments/assets/32ae6981-be18-4573-93c9-a2c3aa35dae9

New
- Implemented rain and snow in both hardware and software rendering.
- Added chrono in `d3dparticle.c` so hardware rendered rain/snow particles can calculate time on when they will hit a surface.  Delta time is centralized in `graphics.c`.
  - Added `currentAge_s` and `maxAge_s` in particle struct.
- Added xorshift32 randomization in `d3dparticle.c` to improve particle distribution.
  - Integrated `custom_xyz` into randomization so that position/rotation/velocity can have any of the axes modified by a random number defined by a range.
  - Added min/max variance to `emitter` struct for position/rotation/velocity.
  - Added `bDestroysOnSurface` for weather particles.
- Added `weather_effect` to `room_type` struct and allows extraction of `weather_effect` data from packets in #1306.
- Hides sun/moon background overlays during cloudy/rainy/snowy weather.
- New .png art assets for cloudy skies in hardware rendering.  (excluded from PR to be uploaded separately)
- Added particle priming so when the player changes rooms or teleports, particles can be pre-baked and show at the player's new location.

Refactored
- Most magic numbers are defined with as constexpr.  Certain constants were moved to `d3drender.h` to allow multiple source files to access them.
- Refactored `D3DParticleEmitterInit()` down to handling allocation.
  - Variables such as `position` and `BGRA` use direct assignment instead.
  - Specific setup functions were moved to `D3DParticleEmitterSetTimer()` and `D3DParticleEmitterAddToList()`.
- Particle creation now uses circular buffering instead of a for-loop to optimize fetching of new particles.
- `d3drender_bgoverlay.c` was broken up into helper functions and had comments added to help describe setup process.  Two new structs (`OverlayRegion` and `BackgroundOverlayTransform`) were defined to help pass groups of variables into these helper functions.
- Overhauled sandstorm effect in `d3drender_fx.c` to be set up with a for-loop, making it easier to read and modify.
- Refactored `d3drender_skybox.c` to use a centralized lookup table for skyboxes.
  - Added struct `SkyboxDefinition` to pair .bgf resource filenames with corresponding .bsf/.png counterparts.
- Replaced `list_type` in `d3dparticle` and `d3drender_fx.c` with std::vector.
- Removed unused variables in emitter/particle structs such as size, weight, and delta.
  - Energy was removed since it was only used for particle expiration and used frame-based tracking.  It is replaced by time-based tracking.


clientd3d/d3drender_skybox.c
```diff
@@ -324,6 +338,9 @@ void D3DRenderSkyBoxShutdown()
 */
 void D3DRenderBackgroundsLoad(const char* pFilename, int index)
 {
+	// The .bsf/.png files for skyboxes are actually six .png files stored in a single 
+	// file by appending each picture's binary data into one file using a hex editor.
```
akirmse: Is a hex editor the only way to append files?  Remove that part of the comment.

clientd3d/d3drender_skybox.c
```diff
@@ -12,8 +12,10 @@
 
 static const float SKYBOX_DIMENSIONS = 75000.0f;
 static const float SKYBOX_Y = 37000.0f;
+static const int TOTAL_SKYBOXES = 9;
```
akirmse: NUM_SKYBOXES would be more consistent with naming elsewhere in the client.

clientd3d/draw3d.c
```diff
@@ -199,34 +199,36 @@ void GraphicsResetFont(void)
...
-	PDIB pdibCeiling = NULL;
-
 	// Only drawn in room, before overlays are drawn.
-
-	pdibCeiling = GetPointCeilingTexture(params->viewer_x, params->viewer_y);
-
+	
 	// sand
-	if (effects.sand)
+	if (effects.sand && !IsBlind())
```
akirmse: Maybe just have if (IsBlind()) return; once at the top of the function, rather than checking it 3 times?

clientd3d/drawbsp.c
```diff
@@ -4036,8 +4036,8 @@ static void doDrawBackground(ViewCone *c)
       }
    }
 
-   // Don't draw bg overlays if blind
-   if (IsBlind())
+   // Don't draw background overlays if blinded. Or during cloudy weather.
+   if (IsBlind() || (IsClearWeather() == FALSE))
```
akirmse: FALSE -> false

clientd3d/effect.c
```diff
@@ -257,3 +261,44 @@ bool AnimateEffects(int dt)
...
+
+/************************************************************************/
+
+bool IsClearWeather()
+{
+	if (current_room.weather_effect == WEATHER_CLEAR)
+	{
+		return TRUE;
+	}
+	return FALSE;
```
akirmse: Replace function body with

```
return current_room.weather_effect == WEATHER_CLEAR;
```

clientd3d/d3dparticle.h
```diff
@@ -8,44 +8,41 @@
...
+	custom_xyz		pos;
+	custom_xyz		oldPos;
+	custom_xyz		velocity;
+	custom_xyz		rotation;
+	custom_bgra		bgra;
+	float			size;
+	float			weight;
+	// Milliseconds for weather particles to clear.
+	int				lifetime;
+	int				maxTime;
```
akirmse: What is the difference between lifetime and maxTime?  Needs a comment.

clientd3d/d3dparticle.c
```diff
@@ -15,45 +25,66 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
 
 	pEmitter->numParticles = 0;
-	pEmitter->bRandomize = bRandomize;
+	pEmitter->nextSlot = 0;
+	pEmitter->energy = energy;
+	pEmitter->timer = timerBase;
+	pEmitter->timerBase = timerBase;
+	
+	// bWeatherEffect makes particles clear when hitting ceilings or floors. It also
+	// randomzies velocity.z a bit, and also ignores randomPos for the z-axis.
```
akirmse: This comment is already in the emitter definition.  No need to duplicate it here.

clientd3d/d3dparticle.c
```diff
@@ -6,6 +6,16 @@
...
 #include "client.h"
+#include <cmath>
+#include <chrono>
+
+using namespace std::chrono;
+
+// Variables
+extern room_type current_room;
+
+steady_clock::time_point lastFrameTime = steady_clock::now();
```
akirmse: make static

clientd3d/d3dparticle.c
```diff
@@ -6,6 +6,16 @@
...
+int64_t msDeltaTime = 0;
```
akirmse: static, add a comment

clientd3d/d3dparticle.c
```diff
@@ -81,171 +112,216 @@ void D3DParticleSystemUpdate(particle_system *pParticleSystem, d3d_render_pool_n
...
+			// Skip dead particles, or end particles that run out of energy.
+			if (pParticle->energy <= 0 || --pParticle->energy <= 0)
 			{
-				D3DParticleDestroy(pParticle);
-				pEmitter->numParticles--;
+				continue;
 			}
-			else
+			// Update lifetime for weather particles since only they track time.
+			if (pEmitter->bWeatherEffect)
```
akirmse: This is an enormous indented block.  Think about rearranging things so that the "then" clause is shorter (move to a function, exit this function earlier, something like that).

clientd3d/d3drender_bgoverlays.c
```diff
@@ -43,242 +43,245 @@ void D3DRenderBackgroundOverlays(const BackgroundOverlaysRenderStateParams& bgoR
...
-					D3DRENDER_CLIP(bottomLeft.y, 1.0f)) ||
-				(D3DRENDER_CLIP(topRight.x, 1.0f) &&
-					D3DRENDER_CLIP(topRight.y, 1.0f)) ||
-				(D3DRENDER_CLIP(bottomRight.x, 1.0f) &&
-					D3DRENDER_CLIP(bottomRight.y, 1.0f)) ||
-				(D3DRENDER_CLIP(center.x, 1.0f))
-				) &&
-			D3DRENDER_CLIP(topLeft.z, 1.0f))
+		// Check if the sky is cloudy. If so, don't render the sun/moon and don't let us right-click it.
+		if (IsClearWeather())
```
akirmse: Another huge indented block.

clientd3d/d3drender_fx.c
```diff
@@ -57,190 +99,289 @@ void D3DRenderParticles(const ParticleSystemStructure& pss)
...
+/**
+* Initializes the rain particle emitters with predefined positions, energy, and colors for the 
+* particle system.
+*/
+void RainInit(void)
+{
+	const int RAIN_EMITTER_RADIUS = 7500;
+	const int RAIN_EMITTER_ENERGY = 200;
+	const int RAIN_EMITTER_HEIGHT = 3000;
+	const int RAIN_TIMER = 15; // Milliseconds
```
akirmse: Better to name it RAIN_TIMER_MS, then it's self-documenting.

clientd3d/d3drender_fx.c
```diff
@@ -57,190 +99,289 @@ void D3DRenderParticles(const ParticleSystemStructure& pss)
...
+* Initializes the snow particle emitters with predefined positions, energy, and colors for the 
+* particle system.
+*/
+void SnowInit(void)
+{
+	const int SNOW_EMITTER_RADIUS = 7500;
+	const int SNOW_EMITTER_ENERGY = 800;
+	const int SNOW_EMITTER_HEIGHT = 3000;
+	const int SNOW_FALL_SPEED = -30;
+	const int SNOW_TIMER = 15;  // Milliseconds
```
akirmse: Same here.

akirmse (commented):
I think we probably don't want to check in the skybox images.  They will use up a lot of github space.  Remove them from this PR and post them on Discord and I'll make sure they get into the next update.

clientd3d/d3dparticle.h
neiltaco: Done!  I had also updated the particle struct to use int32_t.

Is it okay to have the emitters' timers use int64_t like msDeltaTime?  I switched it to int64_t just in case some people decide to leave their client running for a long time, and there's much less emitters than particles.

clientd3d/d3drender_bgoverlays.c
neiltaco: I made it return immediately instead to remove that extra indentation.  I think the whole function is easier to read now, but I just want to be sure if you think it needs to be flattened further or if it is good for now.

clientd3d/d3dparticle.h
neiltaco: I think I have misread how to use chrono to calculate deltaTime.  Based on what I have been reading, steady_clock should handle long sessions but deltaTime doesn't need to be int64_t.

I think it would be better to have the time scale for deltaTime and the other timers to be handled in seconds instead of milliseconds, too.

clientd3d/d3drender_bgoverlays.c
neiltaco: I looked at it again and made everything in that for-loop into its own function. I think I can try and break it down further into smaller functions so it will be easier to read.

neiltaco:
The recent commits attempts to clean up d3drender_bgoverlays.c by splitting it up into smaller functions and grouped variables into structs.  The goal is to make it more readable, and I've tested to make sure the overlays still display as intended.

Since I wasn't fully up to speed on how D3D works, I had used Google AI to help explain how the system works.  I also used it to help create comments and tried verifying them as much as I can.  That said, I could use second opinions from someone who is more knowledgeable on the D3D system to see if the comments are accurate and clear.

neiltaco:
This PR is ready for review.  In proto.h, I followed the example of other protocol definitions to define 'SIZE_WEATHER_EFFECT' using a macro for now.  I had considered changing them all to 'inline constexpr', but I believe I would have to test other features outside of the scope of the weather PR.  I believe I can leave that update for a separate PR.

clientd3d/d3drender_bgoverlays.h
```diff
@@ -62,7 +62,39 @@ struct BackgroundOverlaysSceneParams {
     {}
 };
 
+typedef struct overlay_region {
```
akirmse: struct overlay_region, not typedef struct

clientd3d/d3drender_bgoverlays.h
```diff
@@ -62,7 +62,39 @@ struct BackgroundOverlaysSceneParams {
...
+	custom_xyzw topRight;
+	custom_xyzw bottomLeft;
+	custom_xyzw bottomRight;
+	custom_xyzw center;
+	int width;
+	int height;
+} overlay_region;
+
+// Camera orientation and transformation matrices
+typedef struct bg_overlay_transform {
```
akirmse: same here

clientd3d/rops.c
```diff
@@ -64,87 +64,173 @@ void SandDib(BYTE* pabyBits, int width, int height, int drops)
...
-		if (delta == 0)
-			allrun = 10000;
-		else
-			allrun = alllength / delta;
-		if (allrun <= 0)
-			allrun = 1;
-*/
-		allrun = 3;
+void SnowDib(BYTE* pabyBits, int width, int height, int drops, int heading, int windheading)
+	//PERFORMANCE
```
akirmse: What is //PERFORMANCE?  Remove?

clientd3d/d3dparticle.c
```diff
@@ -6,6 +6,18 @@
...
 #include "client.h"
+#include <cmath>
+#include <chrono>
+
+using namespace std::chrono;
+
+// Variables
+extern room_type current_room;
+
+// 64-bit monotonic clock since boot time, and has practically infinite range (+/- 292 years).
```
akirmse: The range isn't relevant; remove.

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
...
+	PDIB pDib = GetObjectPdib(overlay->obj.icon_res, 0, overlay->obj.animate->group);
+	if (NULL == pDib)
+		return;
+
+	BYTE* bkgnd_bmap = DibPtr(pDib);
+	if (bkgnd_bmap == NULL)
+		return;
+
+	// The background overlay is not yet considered visible to the player.
+	overlay->drawn = FALSE;
```
akirmse: false, not FALSE

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
...
+	// Check if the sky is cloudy. If so, don't render the sun/moon and don't let us right-click it.
+	if (IsClearWeather() == false)
+		return;
+
+	long object_width = DibWidth(pDib) * BG_OVERLAY_SIZE_SCALE;
+	long object_height = DibHeight(pDib) * BG_OVERLAY_SIZE_SCALE;
+
+	// Attempt to batch this overlay into an existing render packet for this texture.
+	d3d_render_packet_new* pPacket = D3DRenderPacketFindMatch(bgoRenderStateParams.worldPool, NULL, pDib, 0, 0, 0);
+	if (NULL == pPacket)
```
akirmse: nullptr

clientd3d/d3dparticle.h
```diff
@@ -8,44 +8,42 @@
...
+	custom_xyz		oldPos;
+	custom_xyz		velocity;
+	custom_xyz		rotation;
+	custom_bgra		bgra;
+	float			size;
+	float			weight;
+	// Current age of the weather particle (in seconds)
+	float			lifetime_s;
+	// Max lifespan of the weather particle (in seconds)
+	float			maxTime_s;
```
akirmse: These names are confusing.  Normally "lifetime" means how long something lives, not its current age.  And then "maxTime" really means "lifespan"?  Maybe the first one should be "age", and the second one "lifespan", or something like that?

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
 // Meridian is a registered trademark.
 #include "client.h"
 
+// Using static constexpr to set up constants at compile time to reduce CPU overhead.
+// Meridian 59's angles are in game units. (4096 units = 360-degree circle)
+static constexpr float FULL_CIRCLE_UNITS = 4096.0f;
+static constexpr float GAME_UNITS_TO_RADIANS = (2.0f * PI) / FULL_CIRCLE_UNITS;
+static constexpr float FULL_CIRCLE_TO_DEGREES = 360.0f / FULL_CIRCLE_UNITS;
+static constexpr float DEGREES_TO_RADIANS = PI / 180.0f;
```
akirmse: We surely don't need another degrees to radians constant.  Use an existing one.

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
...
+
+// Maps legacy software y-offset units to world-space pitch degrees.
+static constexpr float PITCH_UNIT_TO_DEGREES = 50.0f / 414.0f;
+// Maps legacy software y-offset units to overlay-space pitch degrees.
+static constexpr float BG_OVERLAY_UNIT_TO_DEGREES = 45.0f / 414.0f;
```
akirmse: Where are these magic numbers coming from?

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
...
+
+// Defines altitude bounds for background overlays.
+// This maps the server's raw coordinate range [-200, 200] to the client's rendering space.
+static constexpr int ALTITUDE_MAX = 200;
+static constexpr int ALTITUDE_MIN = -200;
+// Increases size of the background overlay if necessary.
```
akirmse: Why would it be necessary to increase the size of the background overlay?

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
...
+static constexpr float BG_OVERLAY_SIZE_SCALE = 1.2f;
+
+// Rendering constants for a 2D quad.
+static constexpr int NUM_VERTICES = 4;
+static constexpr int NUM_INDICES = NUM_VERTICES;
+static constexpr int NUM_PRIMITIVES = NUM_VERTICES - 2;
```
akirmse: "Primitives" or "triangles"?

clientd3d/d3dparticle.c
```diff
@@ -15,45 +27,64 @@ void D3DParticleSystemReset(particle_system *pParticleSystem)
...
 
 	if (pEmitter == NULL)
-		return;
+		return nullptr;
 
 	memset(pEmitter, 0, sizeof(emitter));
 
 	pEmitter->numParticles = 0;
-	pEmitter->bRandomize = bRandomize;
+	pEmitter->nextSlot = 0;
```
akirmse: Setting numParticles and nextSlot to 0 doesn't seem necessary given the memset.

clientd3d/d3drender_bgoverlays.c
```diff
@@ -7,291 +7,330 @@
...
+	if (D3DIsBGOverlayVisible(&region))
+	{
+		D3DFinalizeBGOverlay(overlay, &region, &transform, range, bgoSceneParams);
+	}	
+}
+
+/**
+* Defines the background overlay's shape, size, and texture mapping.
+*/
+void D3DBuildBGOverlayMesh(d3d_render_chunk_new* pChunk, long* object_width, long* object_height)
```
akirmse: I don't understand why width and height are passed as pointers.  Also, if they were passed as floats, you could get rid of all the type casts.

clientd3d/d3drender_skybox.c
```diff
@@ -141,11 +143,18 @@ bool D3DRenderUpdateSkyBox(DWORD background)
 
 	if (gpSkyboxTextures[0][0] == NULL)
 	{
+		// Clear skies
 		D3DRenderBackgroundsLoad("./resource/skya.bsf", 0);
```
akirmse: These hard-coded paths were bad to begin with.  To avoid making the problem worse, how about at least making a loop out of this, and appending the resource directory so that it's in only one place and will be easier to fix later.

clientd3d/d3drender_skybox.c
```diff
@@ -290,7 +300,11 @@ bool D3DRenderBackgroundSet(ID background)
 		{"2skyc.bgf", 2},
 		{"1skyd.bgf", 3},
 		{"2skyd.bgf", 3},
-		{"redsky.bgf", 4}
+		{"redsky.bgf", 4},
+		{"3sky.bgf", 5},
+		{"3skyb.bgf", 6},
+		{"3skyc.bgf", 7},
+		{"3skyd.bgf", 8}
```
akirmse: We should avoid repeating the filenames.  Get them from a data structure set up for my previous suggestion.

clientd3d/rops.h
```diff
@@ -14,8 +14,8 @@
 
 /***************************************************************************/
 
-M59EXPORT void RainDib(BYTE* pabyBits, int width, int height, int level, int heading, int windheading, int windstrength, int torch);
-M59EXPORT void SnowDib(BYTE* pabyBits, int width, int height, int level, int heading, int windheading, int windstrength, int torch);
+M59EXPORT void RainDib(BYTE* pabyBits, int width, int height, int level, int heading, int windheading);
+M59EXPORT void SnowDib(BYTE* pabyBits, int width, int height, int level, int heading, int windheading);
```
akirmse: Not sure you need M59EXPORT since these functions aren't exported to modules.

clientd3d/server.c
```diff
@@ -107,6 +107,7 @@ static handler_struct game_handler_table[] = {
 { BP_SECTOR_LIGHT,      HandleSectorLight },
 { BP_SET_VIEW,		HandleSetView },
 { BP_RESET_VIEW,	HandleResetView },
+{ BP_WEATHER,           HandleWeather },
```
akirmse: Maybe would be clearer as HandleChangeWeather or something like that?  That would make it clearer that this message is only to be used when the weather changes.  Normally, weather is sent via BP_PLAYER.

clientd3d/d3dparticle.h
neiltaco: Yeah you're right on why lifetime can be vague.  Maybe 'currentAge_s' and 'maxAge_s' might be more clear.

clientd3d/d3drender_bgoverlays.c
neiltaco: Google AI suggests that 414.0f is the total range of the vertical view shift in the software rendering engine, while 50.0f is the 50-degree vertical FOV in the Direct3D engine.  45.0f seems to be a slight adjustment for background overlays to prevent them from distorting as the player looks up/down.

I think I can define these values as static constexpr so these are more clear.

clientd3d/d3drender_bgoverlays.c
neiltaco: I think STRIP_PRIMITIVES could be a better name to describe this being for a triangle strip that makes the quad.  edit: that or TRIANGLE_STRIP_PRIMITIVES to be even more clear.

clientd3d/rops.c
neiltaco: Done!  I also removed some unused code in BlurDib().

clientd3d/d3drender_bgoverlays.c
neiltaco: Fixed, and did the same where it gets assigned to true.

clientd3d/d3drender_bgoverlays.c
neiltaco: Fixed, and also replaced NULL in the rest of the file where I believe pointers are involved.

clientd3d/d3drender_bgoverlays.c
neiltaco: Thank you!  I wasn't aware there's a function that already handles that.

clientd3d/d3drender_bgoverlays.c
neiltaco: I think the magic number was there to adjust its size to replicate its appearance from the software renderer, so the comment probably isn't needed.

neiltaco:
I think I have addressed the most recent review.  I also wish to do one more refactor, this time addressing the following major issues:
- Refactoring sandstorm effect to use a for-loop to set up emitters instead of using many hard-coded emitters.  That way it'd be easier to read and adjust.
- Replacing most particle emitter setup functions with direct struct assignments.
- Refactoring particle position/rotation randomization by replacing 'bitwise &' with the random library for better particle distribution.

That and updating NULL to nullptr throughout the source files.

neiltaco:
I'm switching this PR to draft for now since one of the PRs may have heavily increased CPU cycles.  It's likely because I switch the randomization from using 'bitwise &' to std::mt19937 so I plan to change this.

I felt the randomization needed to be expanded on so that the particles are better distributed.  That and there is more control over how the particles spawn at different axes.  I added 'variance min/max' to position, rotation, and velocity within the emitter struct, which I felt was important for refactoring the sandstorm emitter setup.

neiltaco:
This PR should be ready for review.  This will also be my last major commit since I believe I went into 'feature creep' with refactoring code, as well as getting AI assistance with defining magic numbers with constants and adding comments on Direct3D concepts I'm not familiar with.  I'll still be testing this latest commit for any further issues and can address requests when they come up.

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,254 @@
...
+extern room_type current_room;
+
+static constexpr uint32_t PARTICLE_VERTICES = 2;
+static constexpr uint32_t PARTICLE_PRIMITIVES = 1;
+
+static steady_clock::time_point lastFrameTime = steady_clock::now();
+// Elapsed time between frames (in seconds).
+static float deltaTime_s = 0.0f;
+
+// Xorshift32 provides improved randomization while remaining light on CPU cycles.
```
akirmse: I don't buy it.  To quote Donald Knuth, "random number generators should not be chosen randomly".

Are you actually seeing visible non-randomness from std::uniform_real_distribution?  I would be surprised if that's not good enough.

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,254 @@
...
+	if (pParticleSystem == nullptr)
+	{
+		return nullptr;
+	}
 	pEmitter = (emitter *)SafeMalloc(sizeof(emitter));
 
-	if (pEmitter == NULL)
-		return;
-
+	if (pEmitter == nullptr)
```
akirmse: Since the caller crashes immediately if this is null, this check doesn't seem necessary.  Running out of memory is almost always an unavoidable crash.

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,254 @@
...
-	d3d_render_chunk_new	*pChunk;
-	D3DMATRIX				rotate, matrix;
-
 	D3DCacheSystemReset(pCacheSystem);
 	D3DRenderPoolReset(pPool, &D3DMaterialParticlePool);
-
-	for (list = pParticleSystem->emitterList; list != NULL; list = list->next)
+	
+	// Calculate time (in seconds) since the last frame.
+	auto currentFrameTime = steady_clock::now();
```
akirmse: This might be OK, but in general you want to be careful to calculate the time since the last frame only once, and use it everywhere, otherwise you can get weird effects where different parts of the scene use different time offsets and look out of sync.  You're using one time delta for all the emitters (I think), so maybe this is safe, but heads up.

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,254 @@
...
-					0);
-				CHUNK_INDEX_SET(pChunk, 0, 0);
-				CHUNK_INDEX_SET(pChunk, 1, 1);
-			}
+void D3DParticleCreate(emitter *pEmitter, particle *pParticle)
+{
+	pParticle->position = GetVariedXYZ(pEmitter->position, pEmitter->positionVarianceMin, pEmitter->positionVarianceMax);
+	pParticle->velocity = GetVariedXYZ(pEmitter->velocity, pEmitter->velocityVarianceMin, pEmitter->velocityVarianceMax);
+	
+	// Each weather particle calculates the time it takes for them to land on the ground.
```
akirmse: for "it" to land on the ground, no?

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,254 @@
...
+	if (pEmitter->bDestroysOnSurface)
+	{						
+		pParticle->currentAge_s = 0.0f;
+		
+		BSPleaf *leaf = BSPFindLeafByPoint(current_room.tree, pParticle->position.x, pParticle->position.y);
+		if (leaf && leaf->sector->ceiling)
+		{
+			D3DParticleDestroy(pParticle);
```
akirmse: I'm not totally clear on the flow here, but are you creating a particle and then immediately destroying it in this case?  Shouldn't you just not create it at all?

clientd3d/d3dparticle.h
```diff
@@ -8,60 +8,71 @@
 #ifndef __D3DPARTICLE_H__
 #define __D3DPARTICLE_H__
 
-#define MAX_PARTICLES	128
-#define SANDSTORM_R		226
-#define SANDSTORM_G		153
-#define SANDSTORM_B		6
-#define SANDSTORM_A		255
+inline constexpr int MAX_PARTICLES = 128;
```
akirmse: Remove inline, it doesn't do anything in modern compilers.

clientd3d/d3drender.h
```diff
@@ -208,4 +208,23 @@ PALETTEENTRY* getPalette();
 // This palette remains constant and is used for color lookups and transformations.
 const Color(&getBasePalette())[NUM_COLORS];
 
+// Meridian 59's angles are in game units. (4096 units = 360-degree circle)
+inline constexpr float FULL_CIRCLE_GAME_UNITS = 4096.0f;
```
akirmse: Remove inline everywhere I think

clientd3d/d3drender.h
```diff
@@ -208,4 +208,23 @@ PALETTEENTRY* getPalette();
...
+// Meridian 59's angles are in game units. (4096 units = 360-degree circle)
+inline constexpr float FULL_CIRCLE_GAME_UNITS = 4096.0f;
+inline constexpr float GAME_ANGLE_UNITS_TO_DEGREES = 360.0f / FULL_CIRCLE_GAME_UNITS;
+inline constexpr float GAME_ANGLE_UNITS_TO_RADIANS = DEGREES_TO_RADIANS(GAME_ANGLE_UNITS_TO_DEGREES);
+
+// Legacy vertical coordinate range [-207 to +207].
+inline constexpr float Y_UNIT_RANGE = 414.0f;
+// The maximum vertical FOV tilt in software rendering mode.
+inline constexpr float MAX_WORLD_PITCH_DEGREES = 50.0f;
+// The slightly reduced tilt used for all background layers to prevent "sliding" artifacts.
```
akirmse: what?

clientd3d/d3drender.h
```diff
@@ -208,4 +208,23 @@ PALETTEENTRY* getPalette();
 // This palette remains constant and is used for color lookups and transformations.
 const Color(&getBasePalette())[NUM_COLORS];
 
+// Meridian 59's angles are in game units. (4096 units = 360-degree circle)
+inline constexpr float FULL_CIRCLE_GAME_UNITS = 4096.0f;
+inline constexpr float GAME_ANGLE_UNITS_TO_DEGREES = 360.0f / FULL_CIRCLE_GAME_UNITS;
+inline constexpr float GAME_ANGLE_UNITS_TO_RADIANS = DEGREES_TO_RADIANS(GAME_ANGLE_UNITS_TO_DEGREES);
+
+// Legacy vertical coordinate range [-207 to +207].
```
akirmse: huh?

clientd3d/d3drender_bgoverlays.h
```diff
@@ -62,7 +62,39 @@ struct BackgroundOverlaysSceneParams {
     {}
 };
 
+struct OverlayRegion {
+	custom_xyzw topLeft;
+	custom_xyzw topRight;
+	custom_xyzw bottomLeft;
+	custom_xyzw bottomRight;
+	custom_xyzw center;
+	int width;
```
akirmse, line 78: These need documentation.  Normally the coordinates would imply the width and height and they wouldn't be listed again, so we need to be told what these are.

clientd3d/d3drender_fx.c
```diff
@@ -31,23 +45,56 @@ void D3DFxInit()
...
+	}
 
+	for (list_type list = snowParticleSystem.emitterList; list != nullptr; list = list->next)
+	{
+		emitter *pEmitter = (emitter *)list->data;
 		if (pEmitter)
+		{
 			D3DParticleEmitterUpdate(pEmitter, pss.playerDeltaPos.x, pss.playerDeltaPos.y, pss.playerDeltaPos.z);
+		}
 	}
```
akirmse, line 135: It seems like these 3 identical loops can be combined into two nested loops, where the outer one iterates over the emitter lists.

clientd3d/d3drender_fx.c
```diff
@@ -31,23 +45,56 @@ void D3DFxInit()
...
+		D3DParticleSystemUpdate(&rainParticleSystem, pss.particlePool, pss.particleCacheSystem);
+	}
+
+	if (effects.snowing)
+	{
+		IDirect3DDevice9_SetVertexShader(gpD3DDevice, nullptr);
+		IDirect3DDevice9_SetVertexDeclaration(gpD3DDevice, pss.vertexDeclaration);
+
+		D3DParticleSystemUpdate(&snowParticleSystem, pss.particlePool, pss.particleCacheSystem);
 	}
```
akirmse: These identical code blocks can also be combined.

clientd3d/d3drender_fx.c
```diff
@@ -57,190 +104,127 @@ void D3DRenderParticles(const ParticleSystemStructure& pss)
...
+	static constexpr int RAIN_EMITTER_ENERGY = 200;
+	static constexpr float RAIN_EMITTER_HEIGHT = 4000.0f;
+	// Sets minimum z-position variance so weather particles can be primed at the start.
+	static constexpr float RAIN_Z_SPAWN_OFFSET = -(RAIN_EMITTER_HEIGHT / 2);
+	static constexpr float RAIN_VELOCITY = -200.0f;
+	static constexpr float RAIN_EMITTER_RADIUS = 7500.0f;
+
+	D3DParticleSystemReset(&rainParticleSystem);
+	emitter* newEmitter = nullptr;
+	for(int i = 0; i < RAIN_EMITTER_COUNT; i++)
```
akirmse: missing space

clientd3d/d3drender_fx.c
```diff
@@ -57,190 +104,127 @@ void D3DRenderParticles(const ParticleSystemStructure& pss)
...
+			newEmitter->positionVarianceMin = {-SNOW_EMITTER_RADIUS * 2.0f, -SNOW_EMITTER_RADIUS * 2.0f, SNOW_Z_SPAWN_OFFSET};
+			newEmitter->positionVarianceMax = {SNOW_EMITTER_RADIUS * 2.0f, SNOW_EMITTER_RADIUS * 2.0f, 0.0f};
+		}
+		newEmitter->velocity = {0.0f, 0.0f, SNOW_FALL_SPEED};
+		newEmitter->velocityVarianceMin = {0.0f, 0.0f, WEATHER_JITTER_MIN};
+		newEmitter->velocityVarianceMax = {0.0f, 0.0f, WEATHER_JITTER_MAX};
+		newEmitter->bgra = SNOW_COLOR;
+		
+		D3DParticleEmitterAddToList(&snowParticleSystem, newEmitter);
+	}
```
akirmse: These 3 Init functions are almost identical.  Seems like there's an opportunity here to abstract common behavior, such as by inheriting from a base class for the 3 emitter types, or at least extracting common behavior into a single function.

clientd3d/d3drender_fx.c
```diff
@@ -257,13 +241,9 @@ void D3DPostOverlayEffects(const FxRenderSystemStructure& fxrss)
...
 	static DWORD			timeLastFrame = 0;
-	DWORD					timeCurrent, timeDelta;
-	int						i;
-	d3d_render_chunk_new	*pChunk;
-	d3d_render_packet_new	*pPacket;
 
-	timeCurrent = timeGetTime();
-	timeDelta = timeCurrent - timeLastFrame;
+	DWORD timeCurrent = timeGetTime();
+	DWORD timeDelta = timeCurrent - timeLastFrame;
```
akirmse: Here's a different computation of time since the last frame--could be trouble.

clientd3d/d3drender_skybox.c
```diff
@@ -139,13 +101,17 @@ bool D3DRenderUpdateSkyBox(DWORD background)
...
-		D3DRenderBackgroundsLoad("./resource/skya.bsf", 0);
-		D3DRenderBackgroundsLoad("./resource/skyb.bsf", 1);
-		D3DRenderBackgroundsLoad("./resource/skyc.bsf", 2);
-		D3DRenderBackgroundsLoad("./resource/skyd.bsf", 3);
-		D3DRenderBackgroundsLoad("./resource/redsky.bsf", 4);
+	if (gpSkyboxTextures[0][0] == nullptr)
+	{	
+		for(int i = 0; i < NUM_SKYBOXES; i++)
+		{
+			std::filesystem::path fullPath = std::filesystem::path("./resource/") / gSkyboxTable[i].fileName;
```
akirmse: Isn't the last slash after "resource" redundant with the following / operator?

clientd3d/d3dparticle.c
neiltaco: Whoops, yeah admittedly that's a misleading comment I made.  I meant to say that when compared to the original 'bitwise &' for randomizing particle variables, xorshift32 has improved randomness for only a bit of extra cost.

I've read that std:uniform_real_distribution and mt19937 would cost more, but I remember it providing a great distribution of particles.  I leaned towards performance, but I can compare xorshift32 and std:uniform_real_distribution again.  But I'll definitely correct that comment.

edit: Now that I read it again, setting up mt19937 would be costly at first, but since I was using static to have it initialize once, calling dist() should be cheap too.

clientd3d/d3dparticle.c
neiltaco: The particle system uses object pooling so it's reusing particles.  From what I read, any particle with 0 energy gets skipped over by `D3DParticleUpdate()` and doesn't get added to the cache.  If I'm reading this right, particles that aren't in this cache aren't rendered and therefore sit invisible until it gets recycled.

Maybe the functions could be renamed to be more clear.  Maybe `D3DParticleInitialize()` and `D3DParticleHide` should work.

clientd3d/d3dparticle.c
neiltaco: Thanks for the heads up!  I'm thinking that there should also be a single source of `deltaTime` so that the renderings are all synced up.  Maybe I can use the chrono in `graphics.c` and setup a getter function that returns `deltaTime`.

neiltaco:
This PR is ready for review and I believe I addressed all of the requests in the last review.  There were more refactoring done in these last several commits that I felt were necessary.
- DeltaTime was moved from `d3dparticle.c` and centralized in `graphics.c` so that there's a singular source for delta time.  There is code in effects.c that still use milliseconds and may require a lot of code refactoring, so I added another getter function in `graphics.c` that returns delta time in milliseconds.
- I switched from using list_type (linked lists) in `d3dparticle` and `d3drender_fx.c`.  They now use std::vector for better performance and readability.
- Particle priming was added.  Originally, there's a noticeable absence of particles when changing rooms or teleporting around before they start pouring in at the player's location.  An earlier attempt to counteract this made particles spawn up to half the height of the emitters, even if the particle systems have already 'warmed up'.  Particle priming will only 'pre-bake' for the first 128 particles before allowing future particles to spawn at the start of their lifetime.
- Removed `energy` since it was only used for clearing particles using frame-based timing and nothing else.  Now delta time is used instead.
- Condensed and renamed constants in `d3drender.h` to three constexprs.

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,235 @@
...
+	
+    return dist(gen);
 }
 
-void D3DParticleEmitterInit(particle_system *pParticleSystem, float posX, float posY, float posZ,
-							float velX, float velY, float velZ, unsigned char b, unsigned char g,
-							unsigned char r, unsigned char a, int energy, int timerBase,
-							float rotX, float rotY, float rotZ, bool bRandomize, int randomPos,
-							int randomRot)
+// Returns XYZ after adjusting it with the emitter's variance range (if any).
```
akirmse: What emitter is this referring to?  I see no emitter here.

clientd3d/d3dparticle.h
```diff
@@ -8,60 +8,76 @@
...
+struct particle_system
 {
-	int			numParticles;
-	list_type	emitterList;
-} particle_system;
+	int						numParticles;
+	std::vector<emitter*>	emitterList;
+	// Points to an external bool that determines if system is active.
+	bool*					pIsActive;	
+	// If active, particles initialize randomly between the start and end of its lifetime.
```
akirmse: I'm confused by the word "active".  Does this refer to pIsActive, or to isPriming being true?

clientd3d/d3drender.h
```diff
@@ -208,4 +208,16 @@ PALETTEENTRY* getPalette();
 // This palette remains constant and is used for color lookups and transformations.
 const Color(&getBasePalette())[NUM_COLORS];
 
+
+// The software renderer's angles are in game units. A full 360-degree circle is 4096 game units.
+// Note that matrix rotations here expect radians.
+constexpr float GAME_ANGLE_TO_RAD = (2.0f * PI) / 4096.0f;
+
+// Maps legacy software y-offset units [-207, 207] (414 units) to world-space pitch (50 degrees).
```
akirmse: Where is 207 coming from?

clientd3d/d3drender_fx.c
```diff
@@ -6,49 +6,143 @@
...
+static std::vector<particle_system*> particleSystemList;
 
-static particle_system particleSystem;
+static particle_system sandParticleSystem;
+static particle_system rainParticleSystem;
+static particle_system snowParticleSystem;
 
-// Interfaces
+// Stores the last room's security number, just to detect if the player changed rooms.
+static int lastRoomSecurity = -1;
```
akirmse: This method seems unreliable.. if room security changes or is disabled in the future, particle systems will silently break.  Can you have an explicit call instead when the user changed rooms?

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,235 @@
...
+		}
+	}
+	
+	pParticleSystem->emitterList.clear();
+}
 
+emitter* D3DParticleEmitterInit(particle_system *pParticleSystem, float time)
+{
+	emitter	*pEmitter = (emitter *)SafeMalloc(sizeof(emitter));
 	memset(pEmitter, 0, sizeof(emitter));
```
akirmse: Can use ZeroSafeMalloc instead?

clientd3d/d3drender_skybox.h
```diff
@@ -32,6 +32,12 @@ struct SkyboxRenderParams {
     {}
 };
 
+// Defines which .bgf resource gets paired with its respective hardware rendered skybox.
+struct SkyboxDefinition {
```
akirmse: Does this need to be in the header or can it be in the .c file?

clientd3d/graphics.c
```diff
@@ -474,7 +489,7 @@ void UserEndDrag(void)
 /************************************************************************/
 /*
  * GraphicsReleaseCapture:  Release mouse capture, if main window has it.
- *   Return true iff main window had mouse capture.
+ *   Return true if main window had mouse capture.
```
akirmse: iff means "if and only if" (implies the converse), so it was correct before.

clientd3d/graphics.c
```diff
@@ -491,3 +506,13 @@ bool GraphicsMouseCaptured(void)
 {
    return capture;
 }
+/************************************************************************/
+float GetDeltaTime()
```
akirmse, line 513: What is this?  Needs documentation (in the .h file).

clientd3d/d3dparticle.h
neiltaco: Yeah those comments do seem misleading so I can rewrite them.

clientd3d/d3dparticle.c
neiltaco: The function here is specifically being used to apply min/max variance to an emitter's position, rotation, or velocity.  Although looking at it again, I can change the comment so this function can be for any case where a custom_xyz needs variance applied, not just for emitters.

clientd3d/d3drender.h
neiltaco: Whoops!  I had the comment wrong about the y-offset based from the software renderer.  The comment is updated to remove '[-207,207]' and just say the legacy y-offset is 414 units.

clientd3d/d3drender_fx.c
neiltaco: Done!  I set up a boolean `roomJustEntered` in game.c that toggles true upon entering a room, and it gets reset to false at the end of DrawRoom() in draw.c.

clientd3d/d3drender_skybox.h
neiltaco: I don't think anything else will use this struct so it's defined in the .c file now.

clientd3d/graphics.c
neiltaco: Whoops!  Fixed.

clientd3d/graphics.c
neiltaco: Documented for both in .c and .h.

clientd3d/d3dparticle.c
neiltaco: Done!

neiltaco:
This pull request is ready for review when you get the chance.  I added and adjusted comments for extra clarification, and also removed `D3DParticleHide()` since direct assignment of a particle's `isActive` property to false seems clear enough.

Since I already did several merge commits from the master branch throughout this PR, I don't believe I can easily rebase to reorganize the commit history.  Going forward, I'll stick to merge commits in this PR but only do so if any of the files in this PR were adjusted by upstream/main, and I'll use rebasing instead for future PRs.

neiltaco:
I'm closing this PR to clean up the commit history and also start splitting up this weather PR into smaller PRs.

PR #1305, opened 2025-12-27
Map upgrade process to fix/avoid corrupt annotations

CyberBill:
Updating the map version from 1 to 2, and creating a migration process to copy all the existing data over.

The main purpose of the migration is to ensure that there is enough space allocated for map annotations.

### DIFF
--- a/clientd3d/mapfile.c
+++ b/clientd3d/mapfile.c
@@ -54,7 +54,7 @@ static bool ReadRoomAnnotations(FILE *file, room_type *room, int max_annotations
 static bool SaveRoomAnnotations(FILE *file, room_type *room);
 static int FileSize(FILE *file);
 static bool MapFileMigrateVersion1ToVersion2();
-static bool MapFileMigrateVersion1ToVersion2Data(FILE *destfile, LowerTableEntry *rooms, int *mapAnnotationOffsets);
+static bool MapFileMigrateVersion1ToVersion2Data(FILE *destfile);
 
 /*****************************************************************************/
 /* 
@@ -493,22 +493,12 @@ bool MapFileMigrateVersion1ToVersion2()
       return false;
    }
 
-   // Allocate temp storage for room entries and annotation offsets
-   LowerTableEntry* rooms = new LowerTableEntry[MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE];
-   int* mapAnnotationOffsets = new int[MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE];
-
    // Iterate through all rooms in old file and copy them to new file
-   if (!MapFileMigrateVersion1ToVersion2Data(destfile, rooms, mapAnnotationOffsets))
+   if (!MapFileMigrateVersion1ToVersion2Data(destfile))
    {
-      delete[] rooms;
-      delete[] mapAnnotationOffsets;
-
       fclose(destfile);
       return false;
    }
-
-   delete[] rooms;
-   delete[] mapAnnotationOffsets;
 
    // Close both files, delete the old one, and rename the new one
    fclose(destfile);
@@ -542,7 +532,7 @@ bool MapFileMigrateVersion1ToVersion2()
        MapFileVerifyHeader() to accept either version 1 or version 2 files, without any migration.
  Added by Bill Carlson, 12/27/2025
 */
-bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms, int* mapAnnotationOffsets)
+bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile)
 {
    int throwaway;
    int file_size = 0;
@@ -562,8 +552,8 @@ bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms
    // There are 100 top-level entries, each with 100 lower-level entries, which means 10,000 possible rooms
    // All of that room data is too much to hold in memory, but we can keep "pointers" (file offset) to each room and
    // annotations to quickly confirm whether the map annotation is potentially corrupt.
-   int roomsCount = 0;
-   int mapAnnotationOffsetsCount = 0;
+   std::vector<LowerTableEntry> rooms;
+   std::vector<int> mapAnnotationOffsets;
 
    // The 'top table' is the next MAPFILE_TOP_TABLE_SIZE * 4 bytes
    int topTable[MAPFILE_TOP_TABLE_SIZE];
@@ -623,10 +613,6 @@ bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms
             continue;
          }
 
-         // This is a valid room - keep track of it for the copy step
-         rooms[roomsCount].security = lowerTable[j].security;
-         rooms[roomsCount].offset = ftell(mapfile);
-
          // Process the room entry here - skip past walls, find map annotation offset
          int num_walls = 0;
          if (fread(&num_walls, 1, 4, mapfile) != 4)
@@ -647,20 +633,22 @@ bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms
             continue;
          }
          
-         // Don't increase the room count until after we've validated the room
-         ++roomsCount;
+         // Don't push the room info until after we've validated the room
+         LowerTableEntry roomEntry;
+         roomEntry.security = lowerTable[j].security;
+         roomEntry.offset = lowerTable[j].offset;
+         rooms.push_back(roomEntry);
 
          if (annotations_offset > 0)
          {
-            mapAnnotationOffsets[mapAnnotationOffsetsCount] = annotations_offset;
-            ++mapAnnotationOffsetsCount;
+            mapAnnotationOffsets.push_back(annotations_offset);
          }
       }
    }
 
    // SECOND PASS - Copy each room's data to the new file
    byte wallData[4096];
-   for (int i = 0; i < roomsCount; i++)
+   for (size_t i = 0; i < rooms.size(); i++)
    {
       // Find the source room
       if (!MapFileFindRoom(mapfile, rooms[i].security, false))
@@ -706,11 +694,11 @@ bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms
          // Start from file_size (end of file), look through all known rooms and annotation blocks to find the next
          // known data offset after temp_room.annotations_offset
          int next_data = file_size;
-         for (int j = 0; j < MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE; ++j)
+         for (size_t j = 0; j < MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE; ++j)
          {
-            if (j > roomsCount && j > mapAnnotationOffsetsCount)
+            if (j > rooms.size() && j > mapAnnotationOffsets.size())
                break;
-            if (j < roomsCount)
+            if (j < rooms.size())
             {
                int offset = rooms[j].offset;
                if (offset > temp_room.annotations_offset && offset < next_data)
@@ -718,7 +706,7 @@ bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms
                   next_data = offset;
                }
             }
-            if (j < mapAnnotationOffsetsCount)
+            if (j < mapAnnotationOffsets.size())
             {
                int offset = mapAnnotationOffsets[j];
                if (offset > temp_room.annotations_offset && offset < next_data)


clientd3d/mapfile.c
```diff
@@ -101,31 +116,51 @@ bool MapFileVerifyHeader(FILE *file, bool file_exists)
...
-     {
-       debug(("Bad map file version %d; expecting %d\n", version, MAPFILE_VERSION));
-       return false;
-     }
+      for (i = 0; i < 4; i++)
+         if (fread(&byte, 1, 1, file) != 1 || byte != map_magic[i])
+            return false;
+      if (fread(&version, 1, 4, file) != 4)
+         return false;
+      if(version == MAPFILE_VERSION_1)
```
akirmse: Space before (

clientd3d/mapfile.c
```diff
@@ -27,20 +27,35 @@
...
+};
+
 static bool MapFileVerifyHeader(FILE *file, bool file_exists);
-static bool MapFileFindRoom(int security, bool create);
+static bool MapFileFindRoom(FILE* file, int security, bool create);
+static bool WriteHeader(FILE *file);
+static bool ReadRoomAnnotations(FILE *file, room_type *room, int max_annotations);
+static bool SaveRoomAnnotations(FILE *file, room_type *room);
 static int FileSize(FILE *file);
+//static bool MapFileFindNextKnownData(int *next_offset, int curr_offset);
```
akirmse: Delete

clientd3d/map.c
```diff
@@ -561,7 +561,10 @@ void MapEnterRoom(room_type *room)
    room->annotations_offset = 0;
 
    // Load map from file
-   MapFileLoadRoom(room);
+   if(!MapFileLoadRoom(room))
```
akirmse: Space before (

clientd3d/mapfile.c
```diff
@@ -27,20 +27,35 @@
...
 
-static const int MAPFILE_VERSION = 1;
+static const int MAPFILE_VERSION_1 = 1;
+static const int MAPFILE_VERSION = 2;
 static const int MAPFILE_TOP_TABLE_SIZE = 100;      // # of entries in top level table
 static const int MAPFILE_LOWER_TABLE_SIZE = 100;    // # of entries in a single offset table
 
 static const char *map_directory = "mail";
 static const char *map_filename  = "mail\\game.map";
+static const char *map_filename_temp = "mail\\game.map.temp";
```
akirmse: Document what this is for.

clientd3d/mapfile.c
```diff
@@ -101,31 +116,51 @@ bool MapFileVerifyHeader(FILE *file, bool file_exists)
...
+         if (!MapFileMigrateVersion1ToVersion2())
+         {
+            debug(("Map file migration failed!\n"));
+            return false;
+         }
+      }
+      else if (version != MAPFILE_VERSION)
+      {
+         debug(("Bad map file version %d; expecting %d\n", version, MAPFILE_VERSION));
+         return false;
```
akirmse, line 135: We may want to consider what happens when we remove the migration code some months from now.  If someone *does* log in with a v1 file after we've removed the migration code, by default this will result in a permanently broken map.  Should we delete the file and start a new one in that case?

clientd3d/mapfile.c
```diff
@@ -134,63 +169,76 @@ bool MapFileVerifyHeader(FILE *file, bool file_exists)
...
-     }     
+      if (fseek(file, room->annotations_offset, SEEK_SET))
+         return false;
+
+      // Read # of annotations
+      int num_annotations = 0;
+      if (fread(&num_annotations, 1, 4, file) != 4)
+         return false;
+      num_annotations = min(num_annotations, max_annotations);
+      num_annotations = max(num_annotations, 0);
```
akirmse: You can use std::clamp instead of min+max here.

clientd3d/mapfile.c
```diff
@@ -215,141 +263,204 @@ bool MapFileSaveRoom(room_type *room)
...
+
+bool SaveRoomAnnotations(FILE* file, room_type *room)
+{
+   int i, offset, num_annotations, temp = 0;
+
+   if (nullptr == file || !room)
+      return false;
+
+   // Algorithm:
+   //    Currently the file pointer is pointing at the 'annotations_offset' field within the room block (if written), which says where the annotations block is. A value of 0 means no annotations block.
```
akirmse: Wrap paragraph; lines are too long to read easily.

clientd3d/mapfile.c
```diff
@@ -215,141 +263,204 @@ bool MapFileSaveRoom(room_type *room)
...
-   
-   fseek(mapfile, -4, SEEK_CUR);
-   if (fwrite(&room->annotations_offset, 1, 4, mapfile) != 4)
-     return false;
+   {
+      // New annotation block will start at the end of the file
+      room->annotations_offset = FileSize(file);
+   }
+
+   if(room->annotations_offset == 0)
```
akirmse: Space before (

clientd3d/mapfile.c
```diff
@@ -215,141 +263,204 @@ bool MapFileSaveRoom(room_type *room)
...
+      for (i = 0; i < MAX_ANNOTATIONS; i++)
+      {
+         if (fwrite(&room->annotations[i].x, 1, 4, file) != 4)
+            return false;
+         if (fwrite(&room->annotations[i].y, 1, 4, file) != 4)
+            return false;
+         if (fwrite(room->annotations[i].text, 1, MAX_ANNOTATION_LEN, file) != MAX_ANNOTATION_LEN)
+            return false;
+
+         if(strlen(room->annotations[i].text) > 0)
```
akirmse: Space before (

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
   fseek(file, pos, SEEK_SET);  // Restore original position
   return size;
 }
+
+bool MapFileMigrateVersion1ToVersion2()
+{
+   if(mapfile == nullptr)
```
akirmse: Space before (

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+   bool file_exists = (stat(map_filename_temp, &s) == 0);
+   if (file_exists)
+   {
+      remove(map_filename_temp);
+   }
+   // File doesn't exist; create a new one
+   FILE* destfile = fopen(map_filename_temp, "w+b");
+   if (destfile == nullptr)
+   {
+      debug(("Couldn't open temp map file for migration\n"));
```
akirmse: Include filename in error message.

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+      debug(("Failed to delete old map file during migration\n"));
+   }
+   if (rename(map_filename_temp, map_filename) != 0)
+   {
+      debug(("Failed to rename temp map file to main map file during migration\n"));
+   }
+
+   // Reopen the new file as the main map file
+   mapfile = fopen(map_filename, "r+b");
+   if( !MapFileVerifyHeader(mapfile, true))
```
akirmse: Space before (

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+      return false;
+   }
+
+   delete[] rooms;
+   delete[] mapAnnotationOffsets;
+
+   // Close both files, delete the old one, and rename the new one
+   fclose(destfile);
+   fclose(mapfile);
+   if( remove(map_filename) != 0)
```
akirmse: Fix spacing

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
 }
+
+bool MapFileMigrateVersion1ToVersion2()
+{
+   if(mapfile == nullptr)
+      return false;
+
+   // Open temp file - if it exists, delete it first
+   struct stat s;
+   bool file_exists = (stat(map_filename_temp, &s) == 0);
```
akirmse: Prefer std::filesystem::exists() instead.

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+         debug(("MapFileMigrateVersion1ToVersion2Data: Failed to load destination room with security %d\n", rooms[i].security));
+         continue;
+      }
+
+      // Read the source room data
+      int num_walls = 0;
+      if (fread(&num_walls, 1, 4, mapfile) != 4)
+         return false;
+      int num_wall_bytes = (num_walls + 7) / 8;
+      if(num_wall_bytes > sizeof(wallData))
```
akirmse: spacing

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+   if( !MapFileVerifyHeader(mapfile, true))
+   {
+      MapFileClose();
+      return false;
+   }
+
+   return true;
+}
+
+bool MapFileMigrateVersion1ToVersion2Data(FILE* destfile, LowerTableEntry* rooms, int* mapAnnotationOffsets)
```
akirmse: Put a TODO comment here to remove the migration code at some point, and include the date it was added.

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+      fclose(destfile);
+      return false;
+   }
+
+   // Allocate temp storage for room entries and annotation offsets
+   LowerTableEntry* rooms = new LowerTableEntry[MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE];
+   int* mapAnnotationOffsets = new int[MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE];
+
+   // Iterate through all rooms in old file and copy them to new file
+   if(!MapFileMigrateVersion1ToVersion2Data(destfile, rooms, mapAnnotationOffsets))
```
akirmse: spacing

clientd3d/mapfile.c
```diff
@@ -361,3 +472,272 @@ static int FileSize(FILE *file) {
...
+{
+   if(mapfile == nullptr)
+      return false;
+
+   // Open temp file - if it exists, delete it first
+   struct stat s;
+   bool file_exists = (stat(map_filename_temp, &s) == 0);
+   if (file_exists)
+   {
+      remove(map_filename_temp);
```
akirmse: Not sure this is needed.. doesn't fopen("w+") truncate the file if it exists anyway?

CyberBill: Yes, it does - Will remove!

CyberBill:
Feedback has been addressed @akirmse 

I also created the 'simple' version of the file migration. It doesn't do the expensive memory bounds checking, so it is faster and has less memory overhead. I can optionally do an even more basic update that doesn't do this at all, and instead loads map versions 1 and 2 identically. Happy to put that option in here if you'd prefer.

akirmse:
Yes, post-migration I'd like to get rid of all the migration code, so let's go with the option that reads both versions the same.  To be honest I can't tell by reading the migration code whether it's correct or not, so I'm hopeful that we can ditch it in a few months.

One question though: after we get rid of the migration code, what will happen with the (version 1) game.map file that comes with new installs?  Will it still be read correctly?  I think the answer is yes, since only the version number is different.

CyberBill:
That's correct - the version of the map that is included in Steam (and probably the installer - I assume they are the same) - don't have any of the corrupted annotations, and will parse just fine in the code whether its a v1 or v2. If we just disable the migration code and allow both v1 and v2 to act the same, there will literally be nothing different. Similarly with the version of the map that was located on Gilroys.

I should probably mention that the version of the game.map included with Steam is kind of messed up, in that it contains ~100 map annotations for various maps, but those map IDs ("security") must have changed. Places like Tos and east/west Jasper - they load in as empty, even though some version of them is contained within the game.map file. Raza and Cor Noth load fine. It looks like all the major cities are supposed to load as complete and most don't.

I'll rip out the "simple" migration code and add some comments about the proper way to rip the code out in the future.

akirmse:
Yes, the annotation file is indexed by a checksum of the contents of the room file, and we do fix bugs in the structure of rooms occasionally, which makes the annotation section for those rooms obsolete.  The checksum is necessary because the wall bitflags may no longer match the structure of the room.  Once every very long while we redo the annotations, but it's extremely tedious so it doesn't get done much.

CyberBill:
@akirmse - I've just done some pretty extensive testing and hardening of the code - found a couple of minor bugs that have been fixed. I used 9 different game.map files from various sources (default from Steam, Gilroys, and some copies of mine at various points that include bad (truncated) annotation data and even one that is very corrupt) - and I've confirmed that for all "good" files the data is copied over perfectly. Data with bad annotations is copied as best as can be expected, and even the very corrupt file comes over with a lot of valid data (and invalid data stripped out).

Would you like me to resolve the issues noted above, or would you like to walk through them?

clientd3d/mapfile.c
```diff
@@ -215,141 +261,206 @@ bool MapFileSaveRoom(room_type *room)
...
+            debug(("Wrote annotation '%s'\n", room->annotations[i].text));
+         }
+      }
+
+      // Now that the annotations block is complete, go back and rewrite the number of annotations
+      num_annotations = MAX_ANNOTATIONS;
+      if (fseek(file, room->annotations_offset, SEEK_SET))
          return false;
-       if (fwrite(&room->annotations[i].y, 1, 4, mapfile) != 4)
+      if (fwrite(&num_annotations, 1, 4, file) != 4)
```
akirmse, line 354: This seems a little strange.  Since you're always writing out the constant MAX_ANNOTATIONS, why seek back in the file and write it here, as opposed to just writing it the first time (line 332)?  That's what the old code did.

clientd3d/mapfile.c
```diff
@@ -361,3 +472,283 @@ static int FileSize(FILE *file) {
...
+   // Write new header
+   if (!WriteHeader(destfile))
+   {
+      fclose(destfile);
+      return false;
+   }
+
+   // Allocate temp storage for room entries and annotation offsets
+   LowerTableEntry* rooms = new LowerTableEntry[MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE];
+   int* mapAnnotationOffsets = new int[MAPFILE_LOWER_TABLE_SIZE * MAPFILE_TOP_TABLE_SIZE];
```
akirmse: Maybe make these into std::vectors so you don't have to worry about new/delete?

CyberBill: I'd love using STL, but I am not sure that I should use classes in a .c file like that.

clientd3d/mapfile.c
CyberBill: This is added protection in case the writing of the room annotations fails. By writing 0, then writing the data, and then going back and writing the real number, we prevent the previous situation where the file says "I have 20 annotations" and then only contains 1.

I'm also mimicking the code from the SaveRoom function, where it writes 0 for the offset, then writes the annotations block, and then goes back and fills it out.

akirmse: Oh OK, that seems fine.

clientd3d/mapfile.c
akirmse: These are C++ files---you're already using new/delete, which are C++.  They have .c extensions for historical reasons.  We use STL quite a bit.

CyberBill: That is a very good point! Let me swap over - should only take a few minutes.

CyberBill: Updated!

PR #1309, opened 2025-12-28
gameplay: Re-enables Hunt, will ultimately cast unsuccessful if target player is offline after channeling

davidklaw:
This PR looks to bring Hunt back online by addressing its offline timer issues (see #807). Instead of a global change to enchantments (#1192), this looks to address Hunt's issues in `hunt.kod` alone. Under these changes, after successfully channeling a prism cast, Hunt will fail if the player was online at the start of channeling but is offline at the end, when Hunt is finally cast. Feedback is given to the player to help them understand:

> You feel the magical trail begin to reach out in vain for your target, but finding nothing it dissipates.  The target is no longer in the world.

While this fix should allow Hunt to function again in its current form, I believe the spell could be enhanced further by scaling the frequency of location updates based on the casters' average spellpower, providing a progression path for casters and in that way, some protection for targeted players.

### DIFF
--- a/kod/object/passive/spell/multicst/hunt.kod
+++ b/kod/object/passive/spell/multicst/hunt.kod
@@ -43,6 +43,9 @@ resources:
 
    hunt_lost_trackers = "You seem to have managed to lose your trackers!"
    hunt_gain_trackers = "You sense that someone knows where you are..."
+   hunt_target_offline = \
+      "You feel the magical trail begin to reach out in vain for your target, "
+      "but finding nothing it dissipates.  The target is no longer in the world."
 
 classvars:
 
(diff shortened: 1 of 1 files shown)


akirmse:
Seems OK.  Do we need to warn people that the target needs to remain online during the focus period by modifying the spell's description, or should figuring this out be just part of the cost of doing business?

davidklaw:
> Seems OK. Do we need to warn people that the target needs to remain online during the focus period by modifying the spell's description, or should figuring this out be just part of the cost of doing business?

To me, it felt like it was a cost of doing business. It's such a powerful spell in its current form, alerting you of _every_ room change, which feels like a big enough reward for the added risk. Longer term, I think Hunt probably would benefit from a rework that moves notifications to a timed cycle, but until then this felt like a reasonable way to get it back into rotation. If that longer-term rework is enough to hold off for now, that's fine too, but realistically that could be months or even years out.

sicoderealm:
This is a good way to get the spell usable again, until the recommendations and follow ups to https://github.com/Meridian59/Meridian59/pull/1192 are made to get offline enchantments working properly (such that you can cast hunt on a player and it is active for the player regardless of being online/offline). LGTM

Issue #1310, opened 2025-12-29
Hall of heroes Statues and logoff ghosts do not show musical instruments


sicoderealm:
The hall of heroes statues do not show musical instruments and the same issue appears on logoff ghosts.
<img width="240" height="267" alt="Image" src="https://github.com/user-attachments/assets/2fe95113-15e6-4bf8-85c0-6b9c29b758fc" />
This is currently visible on 101 in the hall of heroes.

PR #1311, opened 2025-12-29
Statues and logoff ghosts show musical instruments

sicoderealm:
We add support for Statues and logoff ghosts to show musical instruments. Note: The instrument will only appear on statues in the `POSE_NONE` pose (pose 2 out of 4 random poses), which matches how instruments are displayed on players - they only show when not attacking.

statue-instruments

logoff-ghost-instruments2

fixes: https://github.com/Meridian59/Meridian59/issues/1310

Testing: Generated statues in the hall of hero's with a Jala master holding a lute and confirmed when `POSE_NONE` is chosen that the lute is displayed. Checked that logoff ghosts correctly show a lute while held.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -5341,6 +5341,7 @@ messages:
    }
 
    LookupPlayerInstrument()
+   "Returns the instrument currently in use."
    {
       local i;
 


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -5339,6 +5339,21 @@ messages:
 
       return $;
    }
+
+   LookupPlayerInstrument()
```
akirmse, line 5343: Please add a string comment describing what this does.

PR #1313, opened 2025-12-29
Add Visual Studio 22 Solution and Project Files for easy building and debugging

CyberBill:
I'm not sure whether this is desired by anyone else, but I find it very helpful for my workflow. It doesn't interfere or change anything in the normal build path, so if you don't use Visual Studio this won't affect you. I only did ONE change to the build process, which is to add /FC to the compile flags, which makes build errors contain the full file path name.

This is a very basic initial setup - but allows you to open a Visual Studio solution file which contains a single project of all the source code. You can build right out of the box (it calls 'nmake') and run the game with attached debugger and easily step through the code. If you have a build error, you can double click it and it will open the file to the correct line. I have support for Debug|Release and also x86|x64.


common.mak
```diff
@@ -72,7 +72,7 @@ PALETTEFILE = $(TOPDIR)\blakston.pal
 # /Zi includes debugging information
 
 CCOMMONFLAGS = -nologo -DBLAK_PLATFORM_WINDOWS -DWIN32 \
-             /wd4996 /wd4312 \
```
akirmse, line 75: Please document the new flag (see comment just above in the file).

akirmse (commented):
I don't know how sustainable this is, given that every source file is listed individually in the vcproj.  How is that going to be kept in sync?

common.mak
CyberBill: Done!

CyberBill:
For functionality, the files won't need to be kept in sync most of the time - it doesn't change the build process or even prevent you from clicking the errors. Practically, there are two options, though. First is anyone adding files can add it to Visual Studio. Second is that if devs aren't generally using VS, then anyone who is using VS can add the files to the project at a later date as needed. Updating the file list is very easy, just dragging the files over into VS.

akirmse:
I think this is OK, but what do you think about moving it to a new directory under the contrib directory?  That would avoid confusion about having two different build methods at the top level, which I imagine would look very strange to new developers.  You could also add a README.md to that new directory explaining how to use these files.

CyberBill:
@akirmse Moved files, updated all path within them, confirmed that all builds (except for Release-x64, which appears to be broken in Master due to audio not linking?) - Also added a README file in that directory.

PR #1314, opened 2025-12-29
Animate about dialog knights before the game has started

### LINKED ISSUE #1296: client: "About Meridian 59" knights don't appear on the startup screen
## What

- There is a bug where if you open the Help -> "About Meridian 59" window at the startup screen before logging in, there is only a blank space where the two fighting fights are supposed to be.
  - If you log in to the game and try the same thing, they then appear.

### Startup Screen

Image

### In Game
Image

## Recommendation

- Investigate and resolve so the knights appear consistently.

sicoderealm:
Show and animate the about dialog knights (aka dudes) prior to starting the game.

image

Fixes: https://github.com/Meridian59/Meridian59/issues/1296


clientd3d/loadrsc.c
```diff
@@ -323,6 +331,14 @@ char *LookupRscNoError(ID idnum)
 	return r->data;
 }
 /******************************************************************************/
+/*
```
akirmse: This is never used; remove.

clientd3d/loadrsc.h
```diff
@@ -33,6 +33,7 @@ void DeleteAllRscFiles(void);
 M59EXPORT char *LookupRsc(ID idnum);
 M59EXPORT char *LookupNameRsc(ID idnum);
 M59EXPORT char *LookupRscNoError(ID idnum);
+M59EXPORT bool AreResourcesLoaded(void);
```
akirmse: Remove.

clientd3d/about.c
```diff
@@ -215,30 +218,27 @@ BOOL AboutInitDialog(HWND hDlg, HWND hwndFocus, LPARAM lParam)
...
-   }
+       object_node *obj;
+   
+       obj = dudes[i].obj = ObjectGetBlank();
+       obj->icon_res = ABOUT_RSC;
+       obj->animate->animation = ANIMATE_NONE;
    
+       if (i == 0)
+       {
+       dudes[i].angle = 3 * NUMDEGREES / 4;
```
akirmse: Indentation is wonky here.

clientd3d/about.c
```diff
@@ -150,6 +150,9 @@ BOOL AboutInitDialog(HWND hDlg, HWND hwndFocus, LPARAM lParam)
    char buffer[80];
    char format[80];
 
+   /* Ensure resources are loaded before we try to use them */
+   LoadResources();
```
akirmse, line 154: I don't think this is safe.  If the user opens the About dialog, then logs in and has to download an update, they will still have the outdated resources in memory.  

We never found a simple fix and the bug is pretty trivial, so we just left it.

sicoderealm: Given, I didn't try that workflow. Let me do some testing and if I can get it working reliably with minimal changes I'll ping you. thank you

sicoderealm: I did some exploratory testing on the update flow to better understand how it behaves around resources used by the About dialog. I tried a couple of server-side update `packages.txt` scenarios:

- deleting the existing resources referenced by the About box, and
- deleting those resources and replacing them with new ones.

I then exercised the About dialog before and after login, with and without updates occurring. In my local testing, both scenarios behaved as expected and I wasn’t able to reproduce any issues.

While digging into this, I found a few references that helped clarify the underlying behavior:

- https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilea
- https://learn.microsoft.com/en-us/windows/win32/memory/closing-a-file-mapping-object

My takeaway is that on Windows, file deletion is deferred until the last open handle is released, so resources already loaded into memory remain valid until the process closes those handles. That lines up with what I observed - the About dialog continued to work even after the backing files were deleted or swapped.

With the current setup, any resources updated in this manner will continue using the in-memory versions until the client restarts, unless additional handling is in place. As part of this change, I updated `download.c` to clear the cache during the update flow, allowing the new resources to be picked up without requiring a restart. We can also keep the current setup if preferred.

clientd3d/loadrsc.c
```diff
@@ -96,6 +96,13 @@ bool LoadResources(void)
 {
 	bool rsc_loaded, rsb_loaded;
 	
+	/* Check if resources are already loaded */
+	if (t != NULL)
+	{
+		debug(("LoadResources: Resources already loaded, skipping\n"));
```
akirmse: Remove this log spam, which otherwise will print almost every time you open the About dialog.

clientd3d/download.c
```diff
@@ -98,6 +98,11 @@ void DownloadFiles(DownloadInfo *params)
          advert = FALSE;
          if (retval == IDOK)
          {
+            // Clear cached resources so newly downloaded files are used
+            debug(("Download complete - clearing resource caches\n"));
```
akirmse: Remove this.

PR #1317, opened 2026-01-02
Added ability to bind 12 'Actions' to any key, and then set command aliases for those Actions.

CyberBill:
Added a new list of 'Action Aliases' - these are keys that you can press that execute commands, essentially like having the F1 through F12 keys, except for any key on the keyboard.

The Preferences screen is updated to have a new panel for the action keybindings, and a new window to bind those actions to generic commands was created (actually re-used the F-key aliases, but made it configurable)

image

image


akirmse:
I have cognitive dissonance looking at these screenshots.  In the "Hotkey aliases" dialog, some of the things in the left column are key names, while some are abstract action numbers.  And then in the Actions tab, actions are mapped to keys, not keys to actions, which is the reverse of the "Hotkey aliases" tab.  

Maybe a good way to approach this is to think about the UI first, as I'm sure we can implement whatever we come up with.  It needs to be super clear and easy for users to understand, first.

CyberBill:
@akirmse I agree - especially on the 'Hotkey Aliases' screen. Maybe the solution there is that I could remove the F-keys? Maybe it's more that conceptually the change is about making the "F Keys" assignable, more than adding a new selection of 'Actions'.

akirmse:
I don't have the answer but removing the F-keys isn't it.  Maybe discuss a UI plan with @davidklaw and others on dev-discussion?

CyberBill:
Added a demo video to the thread in Discord. Code updated, but not fully tested yet.

CyberBill:
Code has been tested well now - all is working great.

The 'Hotkey Aliases' screenshot above is old, the new version is simply the same as the old one - however there is a version for the old F-keys and a version for the new keys. The data is all set dynamically, so the code isn't duplicated, only the values inside.

clientd3d/client.rc
```diff
@@ -2654,6 +2656,58 @@ BEGIN
     PUSHBUTTON      "+",IDC_MOUSELOOKTOGGLE_MOD,229,102,14,14
 END
 
+IDD_ACTIONS DIALOG 0, 0, 258, 138
```
CyberBill, line 2774: Just realized that I need to add German and Portuguese support.

akirmse:
I'm still having trouble with the complexity of the UI design.  As in, why can't I have a screen that shows which action is bound to what key?  Instead I need to click through to another screen and go through the intermediate abstraction of "Action 7", which has no meaning to me as a player.  It's as if the UI is just a convenient reflection of internal data structures, instead of making life easier for players.

CyberBill:
Sure - I can see if I can make the code work from the one screen. It might need a larger scale refactor, as the code is spread out among multiple files.

akirmse:
@davidklaw's mockup might be a good model to follow.  Or if you want to do something else, it would be a good idea to discuss on Discord before spending too much time on reimplementation.

pauly-barns:
If it is possible to expand the available 'bindings' to 20+ or more; this would not only be a huge QOL improvement but would also provide a solution for a recent discord suggestion regarding 'quick communication hotkeys'.

https://discord.com/channels/650824489885630475/1466974294528753980

PR #1318, opened 2026-01-03
fix: Corrects mollusk creature attack and throw animations

### LINKED ISSUE #1289: Mollusk monster animation missing its rest pose
The recently updated mollusk creature [molluskmonster in .kod] animation that now shows them coming out of the ground is missing their standing pose as of of the last update. This frame would be in effect for when they're above ground, but not moving. 

This causes two visible animation issues:

- Without this frame they will stand using one of the 'run' animation frames.
- When running, the animation will skip a frame making it look 'jerky'

https://github.com/user-attachments/assets/719e72d9-05da-41ba-9305-3130033d141f
Image

davidklaw:
This PR fixes several issues with mollusk creature animations. The throw attack was incorrectly triggering a melee animation, causing players to see the wrong attack type, and the standing state was using a walking frame. While addressing these issues, the animation logic was refactored to more closely follow the simpler and established approach used by the Avar Chieftain in `avchief.kod`.

Also updated is the `bbg` file, which contained incorrect frame values in its documentation.

In this video of the change, note how rocks are now scooped from the ground during throw attacks, while melee attacks originate from across the body.

https://github.com/user-attachments/assets/fdadcf58-92af-4c53-8b26-9a9b43ca105d

Fixes #1289

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/mollusk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/mollusk.kod
@@ -87,6 +87,7 @@ properties:
    piAnimation = ANIM_NONE
    pbSurfaced = FALSE
 
+   % When TRUE, the monster will attack by throwing a rock instead of melee attacking
    pbThrow = FALSE
 
    ptPostPop = $


kod/object/active/holder/nomoveon/battler/monster/mollusk.kod
```diff
@@ -87,6 +87,8 @@ properties:
    piAnimation = ANIM_NONE
    pbSurfaced = FALSE
 
+   pbThrow = FALSE
```
akirmse, line 91: Could use a comment, as I didn't understand what this was for just by looking.  Whether the mollusk can be thrown? :-)

PR #1319, opened 2026-01-03
fix: skip webhook computation when webhooks are disabled

adrienlaws:
## What

- added `IsWebhookEnabled()` helper function to `webhook.c`
- added early-exit check in `C_SendWebhook()` before any processing
- removed redundant `ConfigBool(WEBHOOK_ENABLED)` check from `SendWebhookMessage()`
- updated `docs/webhooks.md` to document restart requirement

## Why

- when webhooks are disabled (the default), the server was still doing string lookups, JSON building, and memory allocation on every `SendWebhook()` call before discarding the result
- the enabled state is already determined at startup in `InitWebhooks()`, so runtime config checks are unnecessary

## How

- `IsWebhookEnabled()` returns the `webhook_initialized` bool
- `C_SendWebhook()` checks this first and returns immediately if false

### DIFF
--- a/blakserv/ccode.c
+++ b/blakserv/ccode.c
@@ -2304,15 +2304,14 @@ blak_int C_SendWebhook(int object_id, local_var_type *local_vars,
     int num_normal_parms, parm_node normal_parm_array[],
     int num_name_parms, parm_node name_parm_array[])
 {
-    val_type msg_val, event_val, key_val, value_val;
-    const char *content, *event_name, *key_str, *value_str;
-    int content_len, event_len, key_len, value_len;
-    std::string json;
-    
     // Early exit if webhooks not enabled - avoid all string/JSON work
     if (!IsWebhookEnabled()) {
         return NIL;
     }
+
+    val_type msg_val, event_val, key_val, value_val;
+    const char *content, *event_name, *key_str, *value_str;
+    int content_len, event_len, key_len, value_len;
 
     // Handle single string parameter
     if (num_normal_parms == 1) {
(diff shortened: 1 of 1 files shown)


docs/webhooks.md
```diff
@@ -198,12 +200,16 @@ When the webhook listener starts, it creates multiple named pipes to handle conn
 ### 2. Connection (M59 server side)
 When the Meridian 59 server starts:
 
-1. **Initialization**: `InitWebhooks()` prepares internal state (does not connect pipes yet)
+1. **Initialization**: `InitWebhooks()` checks if webhooks are enabled in config
+   - If disabled (default): Sets `webhook_initialized = false`, returns immediately
```
akirmse: You should not put the names of internal private variables in public documentation.  There is zero chance these will stay in sync.

blakserv/ccode.c
```diff
@@ -2309,6 +2309,11 @@ blak_int C_SendWebhook(int object_id, local_var_type *local_vars,
     int content_len, event_len, key_len, value_len;
     std::string json;
     
+    // Early exit if webhooks not enabled - avoid all string/JSON work
```
akirmse, line 2307: This still allocates and frees a string (std::string json declaration on the previous line).  Another reason to declare variables where they are first used.

docs/webhooks.md
adrienlaws: Removed

blakserv/ccode.c
adrienlaws: Fixed - moved std::string json declaration to first use point

Issue #1322, opened 2026-01-04
Arms revert to default base layer instead of equipped shirt when removing robes


sicoderealm:
When a player is wearing a shirt and robes, removing the robes (or switching directly to armor) causes the player's arms to revert to the default base layer top color instead of the equipped shirt.

**Steps to Reproduce:**

1. Equip a colored shirt (e.g., blue).
2. Equip robes (arms correctly show robe graphics).
3. Equip armor (which automatically un-equips the robes).
4. Result: The player's arms appear as the default base layer top.
5. Expected: The player's arms should show the blue shirt color.

PR #1323, opened 2026-01-04
Fix arm color resetting to default base layer instead of equipped shirt

sicoderealm:
Fixes a visual bug where player arms would revert to the default base layer color instead of the equipped shirt color when `ResetPlayerArms` was called (e.g., when unequipping robes).

Updated `ResetPlayerArms` to use `GetCurrentShirtColor`. This method correctly checks the player's inventory for an equipped shirt and returns its palette translation, falling back to the default base layer only if no shirt is worn.

`ResetPlayerArms` now delegates arm setup to an equipped shirt (calling SetArms) when no gauntlet overrides are present, ensuring shirt-specific arm graphics (e.g., Tanktop's naked arms) are restored correctly after removing robes.

https://github.com/user-attachments/assets/9f795b19-bcaa-44ed-bf37-6332c7293264

Testing: Equipped colored shirt / undershirt with armor, gauntlets (with/without) and switching to robes and back - verifying the correct look of the characters arms.

Fixes: https://github.com/Meridian59/Meridian59/issues/1322

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -8791,7 +8791,7 @@ messages:
       {
          for i in plUsing
          {
-            if IsClass(i,&Shirt)
+            if Send(i,@GetItemUseType) & ITEM_USE_SHIRT
             {
                Send(i,@SetArms);
                


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -8787,6 +8789,20 @@ messages:
       
       if NOT bFound
       {
+         for i in plUsing
+         {
+            if IsClass(i,&Shirt)
```
akirmse: Just double-checking that you want IsClass here, since below you changed this to
```
if Send(i,@GetItemUseType) & ITEM_USE_SHIRT
```

sicoderealm: yes I'd prefer this and have updated (it looks like its used much more extensively throughout this file) thank you

PR #1326, opened 2026-01-04
Remove post patch statue upgrades

sicoderealm:
As the command to upgrade statues has now been run in the latest patch we can remove these from `PostPatchUpdates()`
https://github.com/Meridian59/Meridian59/pull/1270


kod/object/passive/statue.kod
```diff
@@ -529,20 +529,6 @@ messages:
 
    UpgradeStatue()
```
akirmse: Remove the entire function, as there are no references to it.  You can add a new one in the future if it's needed.

PR #1337, opened 2026-01-07
fix rendering of enchantments on Wine

wildangelcult:
Another day, another Wine fix. This time I even have a screenshot of the problem:
enchantments-wine-broken

The problem arises when an enchantment times out and the enchantments need to be moved (`EnchantmentsMovePlayer`/ `EnchantmentsMoveRoom`). What happens is that the enchantments do in fact get moved, but are not rerendered on the new position, so they are there, but not visible.

That is because for some reason `MoveWindow` (well on Wine actually `NtUserSetWindowPos`, but that doesn't matter right now), doesn't produce the `WM_PAINT` message even though the `bRepaint` is set to `TRUE`. I think it has to do with the fact that the enchantment is actually a "button" (`CreateWindow` class). Anyway `MoveWindow` is at least so kind that it removes all the artifacts, so to fix the problem I just added a call to `InvalidateRect`.

Windows shouldn't be affected at all, but you never know. I will test it later.

### DIFF
--- a/module/merintr/enchant.c
+++ b/module/merintr/enchant.c
@@ -223,6 +223,7 @@ void EnchantmentsMovePlayer(void)
       x += width + ENCHANT_BORDER;
       ShowWindow(e->hwnd, SW_SHOWNORMAL);
       // basically no op on Windows, but fixes Wine rendering
+      // https://github.com/Meridian59/Meridian59/pull/1337
       InvalidateRect(e->hwnd, NULL, FALSE);
    }
 }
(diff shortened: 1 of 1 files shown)


wildangelcult:
I finally got around to test it on Windows and it works flawlessly.

module/merintr/enchant.c
```diff
@@ -222,6 +222,7 @@ void EnchantmentsMovePlayer(void)
 
       x += width + ENCHANT_BORDER;
       ShowWindow(e->hwnd, SW_SHOWNORMAL);
+      InvalidateRect(e->hwnd, NULL, FALSE);
```
akirmse, line 227: Please put a comment here and in the other place saying that this is a Wine-only fix and include a link to this bug.  Otherwise, someone may come along and remove it in the future since it does nothing on Windows.

wildangelcult: Great idea, thanks for pointing it out. I added it in.

wildangelcult: Forgot the link.

PR #1339, opened 2026-01-08
fix: error when interacting with the spring of vitality while affected by Second Wind

### LINKED ISSUE #1216: Bug: `SecondWind::ReqEatSomething` throws an error when `what` is not a typical food item
If a player under the effects of Second Wind attempts to eat something, `SecondWind::ReqEatSomething` will prevent them from doing so. But if the thing they're consuming isn't a typical food item and doesn't support `IsItemType` (e.g., `OrcCave5Extension`'s pool of water), a missing handler error is thrown.

### From the logs
```c
Sep 04 2025 16:57:21 | [memmap/secwind.bof (231)] SendBlakodMessage CLASS OrcCave5Extension (10587) OBJECT 629426 can't find a handler for MESSAGE IsItemType (12361)
```

@CardiovascularMike I'm tagging you since this seems to have been introduced with #1067, but don't feel obligated to be the one to fix this.

davidklaw:
This PR fixes #1216 by providing a consumable resource for `ReqEatSomething` to work with, whereas it was previously the room itself being passed along. I'm not a huge fan of my `poFoodProxy` naming and open to suggestions if it's an issue.

### Notes
* The spring of vitality is found in `oc5b`

### DIFF
--- a/kod/object/active/holder/room/monsroom/orccav5x.kod
+++ b/kod/object/active/holder/room/monsroom/orccav5x.kod
@@ -45,13 +45,14 @@ properties:
 
    prMusic = OrcCave5Ext_music
 
+   poPoolWater = $
+
 messages:
 
-   CreateStandardExits()
+   Constructor()
    {
-      plEdge_exits = $;
-      plEdge_Exits = Cons([LEAVE_EAST, RID_ORC_CAVE5, 12, 4, ROTATE_NONE], plEdge_exits);
-      plEdge_Exits = Cons([LEAVE_WEST, RID_ORC_PIT, 64, 68, ROTATE_NONE], plEdge_exits);
+      % Create food item that can stand in for pool water
+      poPoolWater = Create(&Food);
 
       propagate;
    }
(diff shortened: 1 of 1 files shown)


kod/object/active/holder/room/monsroom/orccav5x.kod
```diff
@@ -45,13 +45,13 @@ properties:
 
    prMusic = OrcCave5Ext_music
 
+   poFoodProxy = $ % A food object used to interface with user drink requests
+
 messages:
 
-   CreateStandardExits()
```
akirmse, line 50: Did you mean to delete this?  Seems important....

kod/object/active/holder/room/monsroom/orccav5x.kod
```diff
@@ -45,13 +45,13 @@ properties:
 
    prMusic = OrcCave5Ext_music
 
+   poFoodProxy = $ % A food object used to interface with user drink requests
```
akirmse: Put at least two spaces between code and comment to make it clearer.

kod/object/active/holder/room/monsroom/orccav5x.kod
davidklaw: No, sorry 🤦🏻‍♂️. Was admittedly moving too quickly through this, and my IDE is a bit too over-zealous at modifying code (you don't want to modify) when you hit tab to indent the code (you want to modify). Thank you for catching this.

PR #1347, opened 2026-01-11
refactor: rename `GetStateValue` to `InitializeEnchantmentState`

adrienlaws:
## What

- Renamed `GetStateValue` to `InitializeEnchantmentState` across all personal enchantment spells
- Updated 22 files total:
  - 1 parent class (`persench.kod`) - call site and base definition
  - 21 child classes - function overrides

## Why

- Follow-up to PR #1338 
- Many spells perform `AddDefenseModifier` inside this function
- The "Get" prefix in the old `GetStateValue` function implied a pure getter with no side effects
- The new name `InitializeEnchantmentState` more specifically indicates that it is initialization code (call once)

## How

- All existing behavior preserved
- Pure rename - no logic changes:
  - 1/ Renamed call site in `persench.kod`
  - 2/ Renamed base definition in `persench.kod` with documentation:
  - 3/ Renamed all 21 child class overrides
- Tested casting invis, magic shield, and gort without issue and saw the enchantments wear off normally
   - No errors seen

### DIFF
--- a/kod/object/passive/spell/persench.kod
+++ b/kod/object/passive/spell/persench.kod
@@ -207,8 +207,7 @@ messages:
    }
 
    InitializeEnchantmentState()
-   "Called when starting an enchantment.  Returns the state to store.  "
-   "May have side effects (e.g., AddDefenseModifier).  Do not call multiple times."
+   "Returns the enchantment state to store."
    {
       return $;
    }


akirmse:
Thanks for doing this!

PR #1348, opened 2026-01-12
Updated VS build to build ALL projects, not just the main ones.

CyberBill:
I was having an issue where my kodbase.txt wasn't being built, and this fixes that problem. The VS integration was only doing 'nmake' for the main projects, not all of them.

### DIFF
--- a/contrib/visualstudio/Meridian 59.vcxproj
+++ b/contrib/visualstudio/Meridian 59.vcxproj
@@ -1425,7 +1425,7 @@
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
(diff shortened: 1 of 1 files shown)


akirmse:
Are you sure you want to rebuild the whole world every time you make even a tiny client change?  It takes some time even to enumerate all the kod directories because there are so many.

Also, there is no difference between debug and release builds for Blakod, so that may also cost some extra time.

CyberBill:
I ran some tests - to do a full build takes my machine about 1m55s. If I then run 'nmake' again, it takes 8s. That's very reasonable for me. If I do 'nmake Bclient Bmodules Bupdater Bserver' it's about 1 second. So this change is certainly slower, but I'm not sure that it's enough to matter, and I'd prefer if anyone can grab the .sln and hit 'run' and it just works right out of the gate. [Granted, I've got a pretty fast machine - 7950X3D, 64GB RAM, Samsung 990 Pro SSD]

I think in the future I would like to have each of the subfolders treated as separate projects in VS, so if I'm making changes in the client it wouldn't have to run through the full directory structure, but that's quite a bit more work.

PR #1349, opened 2026-01-13
Adjust DynamicLight

sicoderealm:
We update DynamicLight  to support flickering on/off and provide a post patch upgrade to turn off flickering for all existing DynamicLights in the game.

We will need to run `send object 0 PostPatchUpdates` after the next game patch.

### DIFF
--- a/kod/util/system.kod
+++ b/kod/util/system.kod
@@ -6652,6 +6652,7 @@ messages:
    "Runs one-time update functions after a patch. Call via 'send object 0 PostPatchUpdates'"
    {
       % Ensure existing dynamic lights function as they did prior to the patch (without flickering visually)
+      % Remove this along with Flickerer::StopFlickering after PostPatchUpdates has been applied.
       Send(&DynamicLight, @StopFlickering);
       return;
    }


kod/object/passive/flikerer.kod
```diff
@@ -43,6 +43,9 @@ properties:
    % Are we currently lit?  Used for lights that turn on/off.
    pbIsLit = TRUE
 
+   % Is flicking enabled or not.
```
akirmse: flickering

kod/object/passive/flikerer/dynlight.kod
```diff
@@ -150,6 +150,11 @@ messages:
       propagate;
    }
 
+   StopFlickering()
```
akirmse: This function shouldn't be needed, since the parent class has the identical function.

kod/util/system.kod
```diff
@@ -6651,6 +6651,8 @@ messages:
    PostPatchUpdates()
    "Runs one-time update functions after a patch. Call via 'send object 0 PostPatchUpdates'"
    {
+      % Ensure existing dynamic lights function as they did prior to the patch (without flickering visually)
```
akirmse, line 6654: Add a note that Flickerer::StopFlickering should be removed at the same time as this code?

sicoderealm: As when admin creates a dynamic light it will be flickering, we should keep this to allow admins to have control over making lights non-flickering like on 101 as they look good. If we removed this then we would only have flickering dynamic lights.

akirmse: Admins can modify pbIsFlickering directly.  They don't need a separate function whose only job is to do this.

sicoderealm: Good point - I've been switching between this and the flickering PR, makes sense. applied

PR #1351, opened 2026-01-16
fix: prevent loss of faction lieges and councilors during room recreation

davidklaw:
This PR refactors room recreation to prevent the loss of faction lieges and councilors.

**Problem**: Faction lieges and councilors are created and added to rooms differently than other NPCs, who are typically added to rooms via `CreateStandardObjects()`. Lieges, created via the `Parliament` system and councilors, created via the `TokenGame` system, are uniquely tied to `Library` (as are councilors) in such a way that when the rooms containing lieges are recreated, their old instance remains referenced, lingers, and is unable to be garbage collected. A recreate of these two systems is currently needed to make this happen. From what I've found, they remain referenced by `plOnQuestList`, in `Library`. Another problem is that once the room is recreated, the new liege (now a dup) is not moved to the room.

**Solution**:
The solution refactors `Recreate()` to collect all active objects before deletion, ensuring they can be preserved during the recreation process. We do this by removing the old room from the room table, allowing us to create a new instance, then moving any preserved active objects from the old room to the new room, before finally deleting the old room. A new `OnRoomRecreated()` message is added to the `ActiveObject` base class, allowing objects to handle their own preservation logic. Councilors and faction lieges now implement `OnRoomRecreated()` to automatically move to the new room if their homeroom matches the old room, preventing these important NPCs from being lost during room recreation.

### Log of the room and a Liege through room recreation

**BEFORE ROOM RECREATE:**
One instances of the room and the Liege (or councilor).

```
show instances JasperTavern
:< instances of CLASS JasperTavern (10401)
: OBJECT 548
: 1 total
:>
show instances RebelLiege
:< instances of CLASS RebelLiege (11596)
: OBJECT 3549
: 1 total
:>
```

**AFTER ROOM RECREATE:**

```
show instances JasperTavern
:< instances of CLASS JasperTavern (10401)
: OBJECT 548 OBJECT 7129
: 2 total
:>
show instances RebelLiege
:< instances of CLASS RebelLiege (11596)
: OBJECT 3549
: 1 total
:>
```

**AFTER SAVE/GC:**

```
show instances JasperTavern
:< instances of CLASS JasperTavern (10401)
: OBJECT 7030
: 1 total
:>
show instances RebelLiege
:< instances of CLASS RebelLiege (11596)
: OBJECT 3519
: 1 total
:>
```


kod/object/active/holder/room.kod
```diff
@@ -995,7 +1029,11 @@ messages:
          ptPeriodic_sounds = $;
       }
 
-      Send(SYS,@DeleteRoom,#what=self);
+      if bSkip_table_delete
```
davidklaw: I hear a voice in my head, calling out the fact that this check now exists for all rooms, not just the handful affected by this. I didn't see a better way.

kod/object/active.kod
```diff
@@ -32,5 +32,12 @@ messages:
       return;
    }
 
+   OnRoomRecreated(old_room = $, new_room = $)
+   "Default handler for room recreation events. Most objects do nothing."
```
akirmse: This class doesn't know what most objects do---remove that part.

kod/object/active/holder/nomoveon/battler/monster/council.kod
```diff
@@ -111,7 +111,19 @@ messages:
    CanMorphTo()
    {
       return FALSE;
-   }      
+   }
+
+   OnRoomRecreated(old_room = $, new_room = $)
+   "Called when this room is being recreated. If this councilor's homeroom matches the old "
```
akirmse: You don't need to repeat the code in a comment, but it would be very useful to know *why* this is being done.

kod/object/active/holder/room.kod
```diff
@@ -970,7 +1004,7 @@ messages:
       return;
    }
 
-   Delete()
+   Delete(bSkip_table_delete = TRUE)
```
akirmse: "skip" is kind of an unnecessary negative,  Maybe "bDelete_from_table" default FALSE?  Also include a comment describing what this parameter does.

kod/object/active/holder/room.kod
```diff
@@ -396,24 +396,58 @@ messages:
...
+      }
+
+      % Now delete the old room and teleport users back to the new room
+      Send(oOldRoom,@Delete,#bSkip_table_delete=FALSE);
+
       for i in lUsers
       {
-         oNewRoom = Send(SYS,@FindRoomByNum,#num=piRoom_num);
-         Send(oNewRoom,@Teleport,#what=i);
+         if i <> $
```
akirmse: Is this check necessary?  You already called IsClass on each user earlier, so it seems like it can't be $.

kod/object/active/holder/room.kod
```diff
@@ -396,24 +396,58 @@ messages:
...
+            if IsClass(oActiveObject,&User)
+            {
+               lUsers = Cons(oActiveObject,lUsers);
+            }
          }
       }
 
-      Send(self,@Delete);
+      % Temporarily remove old room from the system's room list so that we can create a new one
+      Send(SYS,@DeleteRoom,#what=self);
```
akirmse: This looks potentially dangerous.  DeleteRoom doesn't actually call Delete on the room?

kod/object/active/holder/room.kod
```diff
@@ -396,24 +396,58 @@ messages:
...
+      if oNewRoom = $
+      {
+         debug("Recreate: Failed to create new room, RID:",piRoom_num);
+         return;
+      }
+
+      % Send message to all active objects in the old room that want to preserve themselves
+      for i in lActiveObjects
+      {
+         if i <> $
```
akirmse: Not sure this check is necessary.

kod/object/active/holder/room.kod
davidklaw: It does not. It only deletes the room from the room table. `DeleteRoom` is called by `Room::Delete`. A current and similar use case is `rntmaint.kod`, in how rented rooms are held during a recreate and their table entry deleted as part of that process.

kod/object/active/holder/room.kod
```diff
@@ -396,24 +396,58 @@ messages:
...
+         {
+            Send(i,@OnRoomRecreated,#old_room=oOldRoom,#new_room=oNewRoom);
+         }
+      }
+
+      % Now delete the old room and teleport users back to the new room
+      Send(oOldRoom,@Delete,#bSkip_table_delete=FALSE);
```
davidklaw: Turns out this new parameter isn't necessary.

The `DeleteRoom` call on 421 has already removed the old room from the room list. There's no need to do it again here. Next commit will clean this up.

davidklaw:
Feedback addressed. Log sample added to PR description.

kod/object/active/holder/room.kod
```diff
@@ -396,23 +396,51 @@ messages:
...
+      }
+
+      % Send message to all active objects in the old room that want to preserve themselves
+      for i in lActiveObjects
+      {
+         Send(i,@OnRoomRecreated,#old_room=oOldRoom,#new_room=oNewRoom);
+      }
+
+      % Now fully delete the old room and teleport users back to the (new) room
+      Send(oOldRoom,@Delete);
```
akirmse, line 440: Just checking that you've tested recreating a room with users in it, since you've changed the order here.  We used to move all the users to the new room, then we'd delete the old room, but now the order is reversed.  It seems like there could potentially be bad side effects from deleting a room with users in it; I think we generally try hard to avoid this.

davidklaw: I have tested with a single user and the action window does a quick refresh, the NPC is in their homeroom position, and all seems normal. However, when the player leaves the room and returns, the "game has changed" dialog appears. I thought that this was expected, so didn't mention it. I'll look into returning to the previous order.

davidklaw: Err, is that actually the case? The previous code is the following, unless I'm misunderstanding what you're pointing out. But if we shouldn't do it this way, I can update.

```
Send(self,@Delete);
...
for i in lUsers
{
   oNewRoom = Send(SYS,@FindRoomByNum,#num=piRoom_num);
   Send(oNewRoom,@Teleport,#what=i);
}
```

davidklaw: I'm seeing the same result, regardless of the order.

akirmse: I saw the call to @DeleteRoom after the loop in the old code, which has been removed, and I thought it was maybe replaced by the call to @Delete that's now before the loop.  It's confusing.

But either way, it's not great that the dialog is showing up.  That means that a resource is being loaded but isn't present.  Does the debug window show what resource that is?

davidklaw: `Missing resource 0`

davidklaw: Missing resource issue resolved via #1375

Issue #1354, opened 2026-01-16
Projectiles not appearing when target far away


sicoderealm:
Reported by JosephJJ and BlackPanther!

Reproduced today, you will not see projectiles (arrows, fireballs..) when the target is very far away.

sicoderealm:
Taking a look at this also

sicoderealm:
When this happens you will see a debug error similar to `a+b <= 0! (AddObjects) -10494`

PR #1356, opened 2026-01-16
Fix long-range projectile rendering and travel time

sicoderealm:
Support visuals for long range projectiles (e.g. arrows and fireball).

Projectiles traveling to long-range targets were not rendering at all. Debug logs revealed projectile positions were being set to INT_MIN (-2,147,483,648), indicating integer overflow during motion calculations.

Example from additional logging:
```
Processing projectile at (-2147483648,-2147483648), icon=27593
Projectile dist=-2021769, dx=2147417145, dy=2147399097
```

Added `MAX_PROJECTILE_TRAVEL_TIME` to automatically boost projectile speed at extreme ranges, preventing the 10+ second travel times that created poor audio-visual feedback. Projectiles now arrive within a maximum of 10 seconds regardless of distance, while maintaining normal speed for typical combat ranges. 

Fixes: https://github.com/Meridian59/Meridian59/issues/1354

### DIFF
--- a/clientd3d/project.c
+++ b/clientd3d/project.c
@@ -91,8 +91,10 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
       distance = sqrtf(dx_f * dx_f + dy_f * dy_f + dz_f * dz_f) / FINENESS;
 
       // Calculate normal increment based on speed
-      // speed is in grid squares per 10 seconds, distance is in grid squares
-      float normal_increment = ((float) speed) / 1000.0f / distance;
+      // The server sends speed as "grid squares per 10 seconds", and increment is
+      // "progress per millisecond", so we divide by 10000 (10 seconds = 10000 milliseconds).
+      // This matches the calculation in MoveObject2() in moveobj.c.
+      float normal_increment = ((float) speed) / 10000.0f / distance;
 
       // Cap maximum travel time to prevent extremely long-range projectiles from taking
       // 30+ seconds to arrive, which breaks gameplay (attack sounds/animations finish


clientd3d/project.c
```diff
@@ -76,10 +80,36 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
...
+      float estimated_travel_time = 1.0f / normal_increment / 60.0f;
+
+      // If travel time exceeds maximum, scale up the speed
+      if (estimated_travel_time > MAX_PROJECTILE_TRAVEL_TIME)
+      {
+         // Scale increment to ensure arrival within MAX_PROJECTILE_TRAVEL_TIME
+         float speed_multiplier = estimated_travel_time / MAX_PROJECTILE_TRAVEL_TIME;
+         p->motion.increment = normal_increment * speed_multiplier;
+
+         debug(("Long range projectile: distance=%.1f, speed boosted by %.2fx\n", distance, speed_multiplier));
```
akirmse: delete this

clientd3d/project.c
```diff
@@ -76,10 +80,36 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
...
+
+      distance = sqrtf((float) (dx64 * dx64 + dy64 * dy64 + dz64 * dz64)) / FINENESS;
+
+      // Calculate normal increment based on speed
+      float normal_increment = ((float) speed) / 1000.0f / distance;
+
+      // Calculate the time it would take at normal speed (in seconds)
+      // increment * dt * frames = 1.0, so time = 1.0 / (increment * fps)
+      // Assuming 60 FPS and dt in milliseconds: time = 1000.0 / (increment * 60)
+      float estimated_travel_time = 1.0f / normal_increment / 60.0f;
```
akirmse: This in in frames?  Shouldn't you care about wall clock time, rather than number of frames?

clientd3d/project.c
```diff
@@ -20,6 +20,10 @@
 extern room_type current_room;
 extern player_info player;
 
+// Maximum time (in seconds) for a projectile to reach its target
+// This ensures long-range projectiles arrive quickly enough to match attack sounds
+#define MAX_PROJECTILE_TRAVEL_TIME 10.0f
```
akirmse: This says time but where it's used, it appears to be in frames.

Constants should be static const float, not #defines.

clientd3d/project.c
```diff
@@ -76,10 +80,36 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
...
+      {
+         // Scale increment to ensure arrival within MAX_PROJECTILE_TRAVEL_TIME
+         float speed_multiplier = estimated_travel_time / MAX_PROJECTILE_TRAVEL_TIME;
+         p->motion.increment = normal_increment * speed_multiplier;
+
+         debug(("Long range projectile: distance=%.1f, speed boosted by %.2fx\n", distance, speed_multiplier));
+      }
+      else
+      {
+         p->motion.increment = normal_increment;
```
akirmse: I think this logic could be simplified by having a min or max somewhere, rather than two different code paths with the if (which is confusing; it looks like there are two different modes).

clientd3d/project.c
```diff
@@ -76,10 +80,36 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
...
       p->motion.increment = 1.0;
-   else 
+   else
    {
-      distance = sqrtf(dx * dx + dy * dy + dz * dz) / FINENESS;
-      p->motion.increment = ((float) speed) / 1000.0 / distance;
+      // Use 64-bit integers to prevent overflow when calculating distance
+      __int64 dx64 = (__int64) dx;
+      __int64 dy64 = (__int64) dy;
+      __int64 dz64 = (__int64) dz;
```
akirmse: Maybe should just use floats instead of 64 bit integers?  You immediately convert to float anyway on the next line.

clientd3d/project.c
```diff
@@ -76,10 +80,36 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
...
+
+      distance = sqrtf((float) (dx64 * dx64 + dy64 * dy64 + dz64 * dz64)) / FINENESS;
+
+      // Calculate normal increment based on speed
+      float normal_increment = ((float) speed) / 1000.0f / distance;
```
akirmse: I understand using a different type to compute distance: it looks like there was overflow, so that was a definite bug.  But I don't understand this apparently unrelated change to speed up some projectiles based on a heuristic.  If the velocity is wrong, shouldn't we fix it always?  Or if it's right, why do we need to change it?

clientd3d/project.c
sicoderealm: Good catch! I've fixed the time calculation to properly use seconds (1.0f / (normal_increment * 1000.0f)) and changed the constant to static const float as suggested.

clientd3d/project.c
sicoderealm: You're right, using floats is cleaner and more efficient since we're converting to float anyway - I'll change it to cast to float directly instead of using 64-bit integers.

clientd3d/project.c
sicoderealm: Good idea! I've simplified it to p->motion.increment = (std::max)(normal_increment, min_increment) where min_increment = 1.0f / (MAX_PROJECTILE_TRAVEL_TIME * 1000.0f), which ensures we always use whichever speed is faster without needing the if/else branches.

clientd3d/project.c
sicoderealm: I think this was already fixed in an earlier update. The code now correctly uses wall clock time in seconds rather than frames, and I've removed the duplicate normal_increment calculation.

clientd3d/project.c
sicoderealm: Previously, at long distances, projectiles would simply stop rendering. With this change, projectiles already in flight can continue traveling even after a monster is killed, so you may still see several projectiles in transit.

https://github.com/user-attachments/assets/ace3f2cc-e7b1-41e9-9998-97dde95d7622
_Monster dies when the cursor appears and casting stops_

A capped projectile travel duration makes it a bit more immersive. Near the maximum range where projectiles currently stop being visible on live - it already feels slightly off, and this helps smooth that experience overall.

That said, if you’d prefer, we can remove the maximum travel time entirely. It mainly affects very long-range edge cases, so I’m happy to adjust either way. Happy to share some more videos if helpful to make the call.

clientd3d/project.c
```diff
@@ -20,6 +20,12 @@
 extern room_type current_room;
 extern player_info player;
 
+// Maximum time (in seconds) for a projectile to reach its target.
+// This ensures long-range projectiles arrive quickly enough to match attack sounds.
+// Note: Since motion updates happen at variable frame rates (dt in milliseconds),
+// this value is used to calculate a speed multiplier to cap total travel time.
+static const float MAX_PROJECTILE_TRAVEL_TIME = 2.0;
```
akirmse, line 27: I'm not sure this is doing what the name suggests.  In the video you link to below, there are fireballs that travel for well over 5 seconds.

clientd3d/project.c
akirmse: OK, yeah, this new behavior does sound better.

clientd3d/project.c
sicoderealm: The video was without `MAX_PROJECTILE_TRAVEL_TIME ` applied, see this extra video when `MAX_PROJECTILE_TRAVEL_TIME = 2.0;` is active (current state of this PR). I had to also make a correction to the calculation to match moveobj (locally I had some unrelated changes there also).

It should be around 2 seconds now to reach the target without any queued up after the cast.

https://github.com/user-attachments/assets/06e46edb-e0b8-41e7-be39-e46969f2f26e

PR #1357, opened 2026-01-18
Adding ability to target "cast" command to self or other players/enemies by name

CyberBill:
Added code on the client to be able to find a target by name. Updated the 'A_CASTSPELL' action to take a "spell_action" rather than a "spell", which also contains a target name. Updated the logic for "cast" operation to allow inclusion of the target.

Examples:
You can now do something like:
cast "minor heal" self
And it will cast the Minor Heal spell on yourself without needing to change your focus target. You can also do:
cast "fireball" mummy
And it will cast the "Fireball" spell against the Mummy enemy. This works with any target-based spell. The code can throw errors for finding duplicate or no targets.

### DIFF
--- a/clientd3d/gameuser.c
+++ b/clientd3d/gameuser.c
@@ -25,6 +25,9 @@ static ID idTarget = INVALID_ID;		//	Target object, or INVALID_ID if no target s
 static void ApplyCallback(ID obj2);
 static void SetDescParamsByRoomObject(room_contents_node *r, HWND hwnd);
 
+static const char* szTargetMe = "me";
+static const char *szTargetSelf = "self";
+
 extern bool gbMouselook;
 extern RECT gD3DRect;
 /************************************************************************/
@@ -760,7 +763,7 @@ std::vector<ID> GetTargetsByName(const std::string &pName, bool allow_self)
    if (allow_self)
    {
       // Quick check for self/me
-      if (name == "self" || name == "me")
+      if (name == szTargetSelf || name == szTargetMe)
       {
          target_ids.push_back(player.id);
          return target_ids;


clientd3d/gameuser.c
```diff
@@ -743,6 +743,66 @@ void UserTargetNextOrPrevious(bool bTargetNext)
 	}
 }
 
+/************************************************************************/
+/*
+ * GetTargetIDFromString:  Tries to find a visible object with the given name - also returns the player ID if 'self' or 'me' is given
+ */
+ID GetTargetIDFromString(HINSTANCE hInst, char* name, int* out_error_code)
```
akirmse: A more general API might be to take a vector<ID> and add in all matching objects whose names match.  Then you wouldn't need an error code; the caller could decide what to do if the vector has 0 or >1 entries.

You might also consider declaring name as a const std::string &, as that would simplify the code.

You'd also want to rename the function to something more like GetAllTargetsByName (or whatever).

clientd3d/gameuser.c
```diff
@@ -743,6 +743,66 @@ void UserTargetNextOrPrevious(bool bTargetNext)
...
+{
+   // Skip spaces
+   while (*name == ' ')
+      name++;
+
+   // Quick check for self/me
+   if(stricmp(name, "self") == 0 || stricmp(name, "me") == 0)
```
akirmse: Space before (.  Please fix throughout the PR.

clientd3d/gameuser.c
```diff
@@ -743,6 +743,66 @@ void UserTargetNextOrPrevious(bool bTargetNext)
...
+      return player.id;
+
+   ID match_id = INVALID_ID;
+   list_type	object_list;
+   room_contents_node* rcnObject = NULL;
+   int			iListIndex = 0;
+   //	Get list of visible objects that are players or enemies
```
akirmse: One space after //.  Please fix throughout the PR.

module/merintr/merintr.rc
```diff
@@ -1453,6 +1453,8 @@ END
 STRINGTABLE 
 BEGIN
     IDS_TARGETNOTVISIBLEFORCAST "You can't see your selected target."
+    IDS_DUPLICATETARGETNAME     "Duplicate targets found with that name."
```
akirmse, line 1460: You also need to add German and Portuguese translations of this, and the special hard-coded strings "self" and "me" (assuming you want to continue supporting both).  You should move these hard-coded strings to the rc file.

module/merintr/spells.h
```diff
@@ -25,6 +25,11 @@ typedef struct {
    BYTE        school;
 } spell;     /* A magical spell--must have obj struct to use in owner drawn list box */
 
+typedef struct {
```
akirmse: Use "struct spell_action" instead.

clientd3d/gameuser.c
CyberBill: Good thinking - I can switch that over. Are you generally in favor of moving to C++ and STL? I've noticed a lot of functions use list_type and a custom linked list, rather than std::list<>, not sure if you're interested in seeing a change there throughout the code base.

CyberBill: On second thought - the generic function needs to be more complex than just strings and returning a count, because it also needs to return the associated entry. I see this same code when looking up spells - we need to look up the spell by name, but get a spell*. And in this case we need an ID type returned.

akirmse: I suggested returning a vector of IDs, not a count.

akirmse: Yes, please use the STL.  But we're not ready to do a massive replacement of list_type right now.  It predates working C++ compilers by quite a bit and is used too pervasively to remove all at once.

module/merintr/merintr.rc
CyberBill: Added loc strings - I checked the code and I don't see other places where we localize commands, and that's kind of what I'm using the "self" and "me" for. As an example, the "emote" command, "say" command, I don't think these are localized.

CyberBill: I've also updated the code to support:
cast "Spell name" My Player Name
Turns out that the local player isn't included in the GetObjects3D function (kind of makes sense!)

clientd3d/gameuser.c
CyberBill: OK - it took me a few tries, but I've refactored it and I think it looks much nicer now!

module/merintr/spells.h
```diff
@@ -25,6 +25,13 @@ typedef struct {
    BYTE        school;
 } spell;     /* A magical spell--must have obj struct to use in owner drawn list box */
 
+// A 'spell_action' ties together a spell and its target string
+struct spell_action
+{
+   spell *sp;
```
akirmse: Maybe a better name than "sp"?

module/merintr/spells.c
```diff
@@ -592,7 +626,10 @@ void MenuSpellChosen(int id)
       return;
    }
 
-   PerformAction(A_CASTSPELL, sp);
+   spell_action spa;
```
akirmse: Can be ```spa = { sp, nullptr }``` ?

module/merintr/spells.c
```diff
@@ -380,34 +389,59 @@ void SpellCast(spell *sp)
...
+   {
+      // User has target already selected.
+      if (target_id == GetPlayer()->id || FindVisibleObjectById(target_id))
+      {
+         /* Make temporary list for sending to server */
+         temp_obj.id = target_id;
+         temp_obj.temp_amount = 1;
+         RequestCast(sp->obj.id, temp_list);
+      }
+      else  // Target cannot be seen.
```
akirmse: Put braces around this to make it clear that the "return" is not part of the else.

module/merintr/spells.c
```diff
@@ -380,34 +389,59 @@ void SpellCast(spell *sp)
...
+   ID target_id = INVALID_ID;
+
+   // See if user specified target on command line
+   if (spa->target != NULL && spa->target[0] != '\0')
+   {
+      // Do a search by name for all potential targets
+      std::vector<ID> targets = GetTargetsByName(std::string(spa->target));
+      if (targets.size() > 1)
+      {
+         GameMessage(GetString(hInst, IDS_DUPLICATETARGETNAME));
```
akirmse, line 399: Since this doesn't return. does the select target dialog pop up, and is this what we want?

module/merintr/spells.c
```diff
@@ -380,34 +389,59 @@ void SpellCast(spell *sp)
...
-	}
-
-   /* Get target object from user */
+   ID target_id = INVALID_ID;
+
+   // See if user specified target on command line
+   if (spa->target != NULL && spa->target[0] != '\0')
+   {
+      // Do a search by name for all potential targets
+      std::vector<ID> targets = GetTargetsByName(std::string(spa->target));
```
akirmse: Could maybe declare type as auto?

module/merintr/spells.c
```diff
@@ -359,18 +359,27 @@ void UserCastSpell(void)
    
    list_delete(sel_list);
    
-   PerformAction(A_CASTSPELL, sp);
+   spell_action spa;
+   spa.sp = sp;
+   spa.target = NULL;
```
akirmse: spa = { sp, nullptr } ?

module/merintr/command.c
```diff
@@ -307,7 +308,10 @@ void CommandCast(char *args)
       return;
    }
 
-   PerformAction(A_CASTSPELL, sp);
+   spell_action spa;
+   spa.sp = sp; 
+   spa.target = next_name;
```
akirmse: spa = { sp, next_name }

clientd3d/gameuser.c
```diff
@@ -743,6 +743,70 @@ void UserTargetNextOrPrevious(bool bTargetNext)
 	}
 }
 
+/************************************************************************/
+/*
+ * GetTargetIDFromString:  Tries to find any visible object with the given name - also returns the player ID if 'self' or 'me' is given
```
akirmse: Wrap this long line.

clientd3d/gameuser.c
```diff
@@ -743,6 +743,70 @@ void UserTargetNextOrPrevious(bool bTargetNext)
...
+
+   // Skip spaces, convert to lowercase
+   std::string name = pName;
+   name.erase(0, name.find_first_not_of(' '));
+   std::transform(name.begin(), name.end(), name.begin(), ::tolower);
+
+   if (allowSelf)
+   {
+      // Quick check for self/me
+      if (name == "self" || name == "me")
```
akirmse: Move these hard-coded strings to the RC file.

clientd3d/gameuser.h
```diff
@@ -44,6 +44,7 @@ ID   GetUserTargetID();
 void UserMouselookToggle(void);
 bool UserMouselookIsEnabled(void);
 void UserTargetSelf(void);
+std::vector<ID> GetTargetsByName(const std::string& pName, bool allowSelf=true);
```
akirmse: Don't have a default value for allowSelf.  Also rename to allow_self.

module/merintr/spells.c
CyberBill: Good catch - that was not intentional - I want it to exit, not popup.

clientd3d/gameuser.c
CyberBill: Would you be OK with me defining them as consts within the file, rather than RC? Reason being that if its a resource file it requires doing a lookup, requiring passing down the HINSTANCE - and I also don't think this should be localized. I see them more like a command like "emote" or "say".

clientd3d/gameuser.c
CyberBill: @akirmse  What are your thoughts on templates? The reason I ask - within GetTargetsByName() I could make a block of code into a template that does the 'starts_with' checks and a bool for 'prefer_exact', but it would be better to templatize the list of entries to be a pair<std::string, T> so it supports any type. But I searched and don't see any templates within the code other than in external stuff.

clientd3d/gameuser.c
```diff
@@ -743,6 +746,86 @@ void UserTargetNextOrPrevious(bool bTargetNext)
...
+   // Get list of visible objects that are players or enemies
+   list_type object_list = GetObjects3D(NO_COORD_CHECK, NO_COORD_CHECK, 0, OF_ATTACKABLE | OF_PLAYER, OF_INVISIBLE);
+   if (object_list)
+   {
+      // Try to find named target in list - convert to lowercase for comparison
+      room_contents_node *rcnObject = NULL;
+      int iListIndex = 0;
+      while (rcnObject = (room_contents_node *) list_nth_item(object_list, iListIndex))
+      {
+         objName = LookupNameRsc(rcnObject->obj.name_res);
```
sicoderealm, line 794: Does this mean you could have a hot key like "cast blind Player123"?

If so this is quite a significant change -- previously to find an exact target you would need to locate them with one of the various options (target next/back or mouse click). This change (if I've understood it correctly) would make it easier to target a specific player. I'm not sure I would add that ability to cast at a specific target player as coordinated PvP could become quite different and one sided (everyone in your guild targeting a single player name with hotkeys for various spells). If you listen in on people playing together on a voice call they do target single players but this would make it too easy IMO.

Also if you have a hot key for "cast fireball orc" and there are multiple orcs, would that be helpful?

CyberBill: For PVP, I would say the biggest limiting factor is that you still need to see that player to be able to attack them and update your hotkey to contain their name, so it still isn't fast (probably slower) if there is more than one player you're attacking. It's hard for me to guess how impactful this would actually be in the real world.

For PvE - You're right that it would not be helpful to try and target enemies in that way. You're better off selecting your target and using 'cast fireball'. But then you could 'cast heal self' without needing to select yourself and then re-select the same enemy.

PR #1362, opened 2026-01-23
Code cleanup and slight refactor of the Stats panel (Inventory/Spells/Skills/Stats)

CyberBill:
Created a new 'StatGroupType' enum for stats types. Created a new StatGroup enum for the differnet stat groups/buttons on the right panel. Plumbed through the StatGroup type throughout the various functions associated with it - replacing BYTE and int types to be better defined. Did various whitespace changes to make sure that spaces are used instead of tabs so the code is properly formatted. Within statbtn.c - the 'buttons' array is now correctly ordered, and contains an entry for the StatGroup - various 'fixups' were being done in the code that are now unnecessary.

### DIFF
--- a/module/merintr/statbtn.c
+++ b/module/merintr/statbtn.c
@@ -253,7 +253,7 @@ LRESULT CALLBACK StatButtonProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM l
  */
 bool StatInputKey(HWND hwnd, UINT key, bool fDown, int cRepeat, UINT flags)
 {
-   bool held_down = (flags & 0x4000) ? true : false; /* Is key being held down? */
+   bool held_down = (flags & 0x4000) ? true : false;  /* Is key being held down? */
    int action;
    const void *action_data;
 


common.mak.linux
```diff
@@ -63,10 +63,10 @@ OS := $(shell uname)
 
 ifeq ($(OS),Darwin)
     BLAK_PLATFORM_MAC = 1
-    # Clang won't treat .c files as C++ without this
-    CCOMMONFLAGS += -x c++
 endif
 
+# Clang won't treat .c files as C++ without this
+CCOMMONFLAGS += -x c++
```
akirmse: Things were compiling on Linux without this change -- are you sure it's needed?  If so, it should be combined with line 60.

include/proto.h
```diff
@@ -306,11 +306,17 @@ enum {
 /* Max # of player overlays allowed */
 #define NUM_PLAYER_OVERLAYS 2
 
+#ifndef BYTE
```
akirmse: This is kind of gross.  I'd rather have StatGroupType be just an int (or leave the type off entirely) and deal with the necessary casts rather than have this typedef polluting the whole codebase.

module/merintr/statlist.c
```diff
@@ -208,52 +208,52 @@ void StatsListMeasureItem(HWND hwnd, MEASUREITEMSTRUCT *lpmis)
  * StatsListDrawItem:  Message handler for stats list boxes.  Return TRUE iff
```
akirmse: Fix TRUE in comment

module/merintr/stats.h
```diff
@@ -21,39 +21,47 @@
...
-#define STATS_MAIN 1
+// Numbers for groups of statistics Equate with value returned from StatsGetCurrentGroup()
+// NOTE: These values are matched to server-side values - seen in user.kod for example
+enum StatGroup : BYTE {
+   GROUP_NONE = (BYTE)-1,
+   STATS_MAIN = 1,      // (Health, Mana, Vigor) Not shown on skills panel, shown on character panel
+   STATS_CHARACTER = 2, // (Might, Int, Stam, Agil, Myst, Aim, Karma)
+   STATS_SPELLS = 3,
+   STATS_SKILLS = 4,
+   STATS_INVENTORY = 5, // Not a real stat group, client-side only
```
akirmse: Comment too close to code; move to the right

module/merintr/stats.h
```diff
@@ -21,39 +21,47 @@
...
 
-//	ajw added constants to refer to other stats "groups". Equate with value returned from StatsGetCurrentGroup().
-#define	STATS_SPELLS		3
-#define	STATS_SKILLS		4
-#define	STATS_INVENTORY		5
+   STATS_COUNT,  // Number of stat groups defined
```
akirmse: Unused?

module/merintr/statbtn.c
```diff
@@ -61,7 +62,10 @@ static int button_border = 0;         // Height of area reserved for buttons
 /* local function prototypes */
 static void StatsCreateButtons(void);
 static LRESULT CALLBACK StatButtonProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam);
-static void StatsMoveButtonFocus(HWND button, bool forward);
+static void StatsMoveButtonFocus(bool forward);
+StatButton *FindStatsButtonByGroup(StatGroup group);
```
akirmse: Should be static; next line also

module/merintr/statbtn.c
```diff
@@ -310,36 +253,36 @@ LRESULT CALLBACK StatButtonProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM l
  */
 bool StatInputKey(HWND hwnd, UINT key, bool fDown, int cRepeat, UINT flags)
 {
-   bool held_down = (flags & 0x4000) ? true : false;  /* Is key being held down? */
+   bool held_down = (flags & 0x4000) ? true : false; /* Is key being held down? */
```
akirmse: At least 2 spaces between code and comments please

module/merintr/statbtn.c
```diff
@@ -411,85 +351,131 @@ bool StatButtonDrawItem(HWND hwnd, const DRAWITEMSTRUCT *lpdis)
...
-   return GROUP_NONE;
+
+/************************************************************************/
+/*
+ * StatsFindGroupByHwnd:  Return group # of stat with given button, or GROUP_NONE if none.
+ */
+StatGroup StatsFindGroupByHwnd(HWND hwnd)
+{
+   StatButton* button = FindStatsButtonByHwnd(hwnd);
+   if(button != nullptr)
```
akirmse: Fix spacing.  Do you have a style or something for this that you can change?  It seems to come up a lot.

common.mak.linux
CyberBill: I'm not sure how the existing code was compiling in C-only, but this was not being compiled as C++ without it. Maybe because Linux only compiles the server, and the server doesn't use C++? I'll move the changes above and add a comment block explaining what each flag does.

include/proto.h
CyberBill: Windows defines this automatically - so it's only necessary on non-Windows platforms. Would it make more sense if I created a 'LinuxDefines.h'? I do see this exact code in other places within the code, so we could consolidate them all.

CyberBill: I can define 'StatGroupType : unsigned char' and not need this, btw.

module/merintr/statbtn.c
CyberBill: I'll confirm my settings are correct - it does set it correctly if I run auto-format, but doesn't always format as I'm typing.

PR #1363, opened 2026-01-23
fix: prevent client startup flickering and splash screen jumps

adrienlaws:
## What

- Fixed visual glitch where splash screen appeared offset then jumped to center when starting maximized
- Also addresses small glitch upon startup where the client window itself will jump to maximized size (if closed as maximized)

## Why

- When client was closed while maximized, reopening briefly showed splash in wrong position before snapping to center
- `SetWindowPlacement` with `SW_SHOWMAXIMIZED` does not maximize immediately - it only stores intent
- Actual resize happens at `ShowWindow` time, which is after `IntroShowSplash` runs
- `GetClientRect` during `IntroShowSplash` returned pre-maximize dimensions (wrong)
- Splash positioned based on wrong size, then WM_SIZE corrected it causing visible jump
- PR #1171 exacerbated this by inlining intro code that previously had DLL loading delays

## How

- We use `WM_SETREDRAW FALSE` to completely disable visual updates
- Call `ShowWindow(SW_SHOWMAXIMIZED)` to force actual resize while hidden
- Hide window, re-enable redraw
- Window is now at correct size but invisible - no flash
- `IntroShowSplash` now gets correct dimensions from `GetClientRect`

## Timing

```text
Before: SetWindowPlacement (stores intent) -> IntroShowSplash (wrong size) -> ShowWindow (resize) -> jump
After:  WM_SETREDRAW OFF -> ShowWindow (resize) -> Hide -> WM_SETREDRAW ON -> IntroShowSplash (correct) -> Show
```

### DIFF
--- a/clientd3d/statoffl.c
+++ b/clientd3d/statoffl.c
@@ -124,13 +124,8 @@ void OfflineResize(int xsize, int ysize)
    button_origin.y = max(0, min(bm_origin.y + bm_height, ysize - button_height));
 
    if (hwndDialButton != NULL)
-   {
      MoveWindow(hwndDialButton, button_origin.x, button_origin.y, 
 		button_width, button_height, TRUE);
-     // Show button now that it's positioned correctly
-     if (!IsWindowVisible(hwndDialButton))
-        ShowWindow(hwndDialButton, SW_SHOW);
-   }
    InvalidateRect(hMain, NULL, TRUE);  /* redraw all */
 }
 /****************************************************************************/
@@ -174,7 +169,7 @@ void IntroShowSplash(void)
    showing_splash = true;
 
    hwndDialButton = CreateWindow("button", NULL, 
-		WS_CHILD,
+		WS_CHILD | WS_VISIBLE,
 		0, 0, 0, 0, hMain, (HMENU) IDC_DIALBUTTON,
 		hInst, NULL);
    SetWindowText(hwndDialButton, GetString(hInst, IDS_INTRO));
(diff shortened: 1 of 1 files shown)


clientd3d/client.c
```diff
@@ -331,7 +331,21 @@ int PASCAL WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpszCmdPa
...
 	WindowSettingsLoad(&w);
-	SetWindowPlacement(hMain, &w);
+	int showCmd = w.showCmd;
+	
+	// Size window while hidden, then show after splash is positioned
+	if (showCmd == SW_SHOWMAXIMIZED || showCmd == SW_MAXIMIZE)
+	{
+		// Maximized: ShowWindow handles sizing, no need for placement struct
+		ShowWindow(hMain, SW_SHOWMAXIMIZED);
+		ShowWindow(hMain, SW_HIDE);
```
akirmse, line 342: Is there a visible flash when you do this?  Calling ShowWindow 3 times during initialization doesn't seem like the right workaround.  Maybe the splash screen code can post a window message to itself to delay the positioning, or something like that?

adrienlaws: Refactored and fixed (I think).

There was indeed a flash and a visible jump before the splash settled into position. I tried PostMessage but it didn't help because the window was already painted by the time it processed.

Turns out `SetWindowPlacement` with `SW_SHOWMAXIMIZED` only stores the intent to maximize. The actual resize happens later at `ShowWindow` time, but by then `IntroShowSplash` already ran with the wrong dimensions.

Now we force the maximize early and wrap it in `WM_SETREDRAW` with `FALSE` / `TRUE` so nothing paints during the resize.  It seems to work without any jumps or flickering.

I updated the PR description and name to reflect the above.

note: there is a "flash" still on startup but that is unrelated to all this. It's because the default windows color is bright grey.  Maybe it'd be worth setting that to black in a separate PR just to avoid the flash, but since it's not really relevant to this PR I'm leaving that alone for now.

akirmse:
Hard to know if this covers all cases, but I like this version better.

PR #1364, opened 2026-01-25
Updated the 'key repeating' code to keep track of each action individually

CyberBill:
I was doing some testing and found that this code is responsible for a significant number of missed actions while playing the game. It was commonly preventing multiple distinct actions in quick succession - effectively limiting input to one action per 400ms.

For example - trying to quickly attack, select yourself, cast a spell, and re-select the enemy - would very frequently silently fail. I would have to hit the keys multiple times before it would take affect. Similarly, even hitting the 'c' key too quickly after another action would mean that my cursor wouldn't unlock.

The root cause is that there was a single timer being used for all actions - pickup, drop, use, say/yell/broadcast, select, attack, cast, etc - thus limiting your character from doing one action every 400ms, and silently ignoring anything faster than that. This change puts each action into it's own bucket - which lines up more with the "Repeat" naming convention - meaning you are still limited to attacking every 400ms, casting every 400ms, and selecting every 400ms.


akirmse:
Would this allow you to increase your movement speed by sqrt(2) by pressing keys for forward and left almost simultaneously?  Maybe move actions need to have a single timer, and maybe there are other actions that are similarly related.

CyberBill:
@akirmse - At this particular place in code, the client is already exempting movement/turning/looking - those keys skip past the check with no input limit. So any existing checks or validations are still in place. This change also doesn't affect attack timing, which I was concerned about - that code also has an underlying limit of 1-attack-per-second (on both client and server) and still hits the Repeat check, which effectively keeps attacks locked at once-per-1200ms.

PR #1372, opened 2026-02-01
Deconstruction of d3d render lights

sicoderealm:
Move d3d hardware rendering lighting logic to its own new file - part of the original plan revisited now that we've recently added more logic here.

Testing: lighting remains functional as expected in various cases and locations

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -530,7 +530,7 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	gDLightCache.numLights = 0;
 	gDLightCacheDynamic.numLights = 0;
-	LightCacheUpdateParams lightCacheParams(&gDLightCache, &gDLightCacheDynamic, &gD3DRedrawAll);
+	LightCacheUpdateParams lightCacheParams{&gDLightCache, &gDLightCacheDynamic, gD3DRedrawAll};
 	D3DLMapsStaticGet(room, lightCacheParams);
 
 	IDirect3DDevice9_Clear(gpD3DDevice, 0, NULL, D3DCLEAR_TARGET |
@@ -650,7 +650,7 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 		// DEBUG: Draw circles at static light positions
 		if (D3DLightsDebugPositionsEnabled() && config.bDynamicLighting)
 		{
-			LightDebugRenderParams debugParams(&gDLightCache, &gObjectPool, &gObjectCacheSystem);
+			LightDebugRenderParams debugParams{&gDLightCache, &gObjectPool, &gObjectCacheSystem};
 			D3DRenderDebugLightPositions(params, debugParams);
 		}
 	}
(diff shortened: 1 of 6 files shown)


clientd3d/d3drender_lights.h
```diff
@@ -0,0 +1,92 @@
...
+#define _D3DRENDERLIGHTS_H
+
+// Parameters for updating light caches from room contents
+struct LightCacheUpdateParams
+{
+   d_light_cache* lightCache;          // Static lights cache
+   d_light_cache* lightCacheDynamic;   // Dynamic/flickering lights cache
+   int* redrawFlags;                    // Pointer to gD3DRedrawAll flags
+
+   LightCacheUpdateParams(
```
akirmse: You can probably get away without these trivial constructors:

https://www.geeksforgeeks.org/cpp/structures-in-cpp/

clientd3d/d3drender_lights.h
```diff
@@ -0,0 +1,92 @@
...
+};
+
+// Parameters for rendering debug light visualization
+struct LightDebugRenderParams
+{
+   d_light_cache* lightCache;
+   d3d_render_pool_new* objectPool;
+   d3d_render_cache_system* objectCacheSystem;
+
+   LightDebugRenderParams(
```
akirmse: Can remove?

clientd3d/makefile
```diff
@@ -136,6 +136,7 @@ OBJS = \
         $(OUTDIR)\d3ddriver.obj \
         $(OUTDIR)\d3dparticle.obj \
         $(OUTDIR)\d3drender.obj \
+        $(OUTDIR)\d3drender_lights.obj \
```
akirmse: Put in alphabetical order

clientd3d/d3drender_lights.h
```diff
@@ -0,0 +1,92 @@
...
+//
+#ifndef _D3DRENDERLIGHTS_H
+#define _D3DRENDERLIGHTS_H
+
+// Parameters for updating light caches from room contents
+struct LightCacheUpdateParams
+{
+   d_light_cache* lightCache;          // Static lights cache
+   d_light_cache* lightCacheDynamic;   // Dynamic/flickering lights cache
+   int* redrawFlags;                    // Pointer to gD3DRedrawAll flags
```
akirmse: Kind of weird that this is a pointer if this is just for updating a cache.

clientd3d/d3drender.c
```diff
@@ -992,473 +768,7 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
...
-				else if (side0 < 0)
-					tree = neg;
-			break;
-
-			default:
-				debug(("add_object error!\n"));
-			return false;
-		}
-	}
+	D3DLightsReportFlickerPerf();
```
akirmse: It looks like this prints some stuff out.  Do all developers need to see flickering lights performance all the time?

clientd3d/d3drender_lights.h
sicoderealm: Good catch! Removed trivial constructors from both structs in favor of aggregate initialization. Updated usage to use brace initialization {}.

clientd3d/d3drender_lights.h
sicoderealm: Actually, these structs have different members and serve different purposes - they cannot be combined - but has been updated to a trivial constructor

clientd3d/d3drender_lights.h
sicoderealm: The pointer is necessary because the lighting system needs to signal cache invalidation back to the calling code. Changed to reference for clarity and added explanatory comment.

clientd3d/d3drender.c
sicoderealm: Good point! This is a debug profiling tool. Removed the call from D3DRenderBegin() and updated the comment in d3drender_lights.c to document where developers should add the call if they want to enable profiling (set flickerPerfProfile = true and add call to end of D3DRenderBegin()).

PR #1373, opened 2026-02-03
feature: add positional door sounds

adrienlaws:
## What

- Door opening and closing sounds now play for all players in the room with 3D positional audio.
  - Opening sound plays from the door's location in the source room
  - Closing sound plays from the opener's position in the destination room (opener excluded since they already heard the open sound)
  - Both sounds use 4-tile radius (down from default 16 tiles) to avoid being overbearing
  - Effectively means you have to be near a door to hear it being used
- Renamed door properties for symmetry (`vrDoor_open_sound`, `vrDoor_close_sound`) and removed getter functions
  - barjail.kod, f8.kod, tosarena.kod - updated to use new property names
- Runtime configurable via `pbDoorSoundShared` setting to toggle between shared (all hear) and private (only opener hears) modes.

## Why

- Previously only the opener heard door sounds, with no spatial positioning
- Other players couldn't hear doors opening, which is unrealistic
- Adds tactical information: hear someone open or close a door even if they're invisible
- 3D positioning adds natural distance attenuation

## How

- Modified `SomethingTryGo` in room.kod to play door sounds via `SomethingWaveRoom` with door's row/col coordinates
- Door opening sound plays at the door's location in the source room
- After using a door, closing sound plays from opener's position in destination room, excluding the opener via #except parameter
- Extended `SomethingWaveRoom` to accept optional row/col, cutoff_radius, and except parameters
- Renamed door sound properties for symmetry and clarity:
  - `vrDoor_open_sound` (opening sound resource)
  - `vrDoor_close_sound` (closing sound resource)
- Removed getter functions, use direct property access throughout codebase
- Room classes can override these properties for custom door sounds
- Updated sound format from .wav to .ogg
- Added `pbDoorSoundShared` setting in settings.kod (default TRUE)
  - TRUE: all players hear door open and close sounds positionally (new behavior)
  - FALSE: only opener hears opening sound non-positionally, no close sound (legacy behavior)
- DMs opening doors produce no sound (stealth)

### Configuration

To disable positional door sounds (revert to legacy behavior):
```
set o 1 pbDoorSoundShared int 0
```

To re-enable:
```
set o 1 pbDoorSoundShared int 1
```

### Note

The setting affects which client volume slider controls door sounds:

- `pbDoorSoundShared = 1` (default): Door sounds use the Ambient Volume slider (positional sounds)
- `pbDoorSoundShared = 0` (legacy): Door sounds use the Sound Volume slider (non-positional sounds)

If door sounds are silent in one mode but work in the other, check both volume sliders in client preferences.

## Example

**Before:**

- Player A opens door → only Player A hears the sound
- Player B nearby → hears nothing

**After:**

- Player A opens door → nearby players in the source room (including Player A) hear the opening sound; nearby players in the destination room hear the closing sound (Player A does not hear the close sound)
- Opening sound emanates from door's position in source room
- Closing sound emanates from player's position in destination room
- Both sounds use 3D audio with distance attenuation
- DMs opening doors → complete silence

### DIFF
--- a/kod/object/active/holder/room.kod
+++ b/kod/object/active/holder/room.kod
@@ -2812,13 +2812,11 @@ messages:
 
       for i in plActive
       {
-         if First(i) = except
+         if First(i) <> except
          {
-            continue;
+            Send(First(i),@WaveSendUser,#wave_rsc=wave_rsc,#source_obj=what,
+                 #row=row,#col=col,#cutoff_radius=cutoff_radius);
          }
-
-         Send(First(i),@WaveSendUser,#wave_rsc=wave_rsc,#source_obj=what,
-              #row=row,#col=col,#cutoff_radius=cutoff_radius);
       }
 
       return;


kod/object/active/holder/room.kod
```diff
@@ -2760,7 +2760,23 @@ messages:
...
+            {
+               if Send(Send(SYS,@GetSettings),@DoorSoundSharedEnabled)
+               {
+                  Send(self,@SomethingWaveRoom,#wave_rsc=Send(self,@GetDoorSound),
+                       #row=row,#col=col);
+               }
+               else
+               {
+                  Send(what,@WaveSendUser,#wave_rsc=Send(self,@GetDoorSound),
+                       #row=0,#col=0);
```
akirmse: Looks like specifying row and col is not necessary here.

kod/object/active/holder/room.kod
```diff
@@ -2778,14 +2794,16 @@ messages:
       return vrDoor_sound;
    }
 
-   SomethingWaveRoom(what = 0,wave_rsc = $)
-   "Don't send <what> if not from any particular object"
+   SomethingWaveRoom(what = 0, wave_rsc = $, row = 0, col = 0)
```
akirmse: Default value for "what" should probably be $ ?

kod/object/active/holder/room.kod
adrienlaws: Removed

kod/object/active/holder/room.kod
adrienlaws: This was interesting to look up.  I tried to change `what = 0` to `what = $` but it generated errors.
> Feb 03 2026 21:09:36 | [user.bof (3325)] AddBlakodToPacket looking for value, got NIL

`SomethingWaveRoom` in `room.kod` calls WaveSendUser in `user.kod` with `#source_obj=what`, and `WaveSendUser` then passes `source_obj` directly to [AddPacket](https://github.com/Meridian59/Meridian59/blob/bab0cc8f5435c18691a67ba4381763d33644af99/kod/object/active/holder/nomoveon/battler/player/user.kod#L3323).  The problem is `AddPacket` then builds the network packet but expects a 4-byte integer field which causes an error.

Is it okay to leave `what = 0` as is for this change?  Or should I try to fix the call to use $?

kod/object/active/holder/room.kod
```diff
@@ -2778,14 +2807,21 @@ messages:
       return vrDoor_sound;
    }
 
-   SomethingWaveRoom(what = 0,wave_rsc = $)
-   "Don't send <what> if not from any particular object"
+   GetDoorCloseSound()
```
akirmse: Is there a reason to have a function that simply returns this value?  I don't see a call to it outside this class.

Also, the names "GetDoorSound" and "GetDoorCloseSound" have an obvious asymmetry.

kod/object/active/holder/room.kod
```diff
@@ -2760,12 +2762,39 @@ messages:
...
+            }
 
             Send(SYS,@UtilGoNearSquare,#what=what,
                  #where=Send(SYS,@FindRoomByNum,#num=Nth(i,3)),
                  #new_row=Nth(i,4),#new_col=Nth(i,5),#new_angle=iAngle);
 
+            % Play door closing sound from player's position in destination
+            % room so it sounds on both sides of the door.  Only if shared
+            % mode enabled (otherwise no sound played, legacy behavior).
+            if NOT IsClass(what,&DM)
```
akirmse, line 2789: Does the person going through the door hear both sounds?  That seems like it would be bad.

kod/object/active/holder/room.kod
adrienlaws: Fixed - I've removed both functions and use the properties directly (`vrDoor_open_sound` and `vrDoor_close_sound`).  Also renamed for symmetry as you noted.  

Side note: as I'm working on this, I think it might be best to have much less noisy default door open and close sounds, and then apply these creaking door sounds to CV, Brax, and mausoleum rooms.

kod/object/active/holder/room.kod
adrienlaws: In testing most of the time the door open sound is not heard at all, but it sometimes is like when entering Barloque (maybe due to the complexity of the map).  I think both sounds makes sense because opening the door in the original room generates a sound from that spot that others can hear, and when entering the new room you close the door behind you.

adrienlaws: It's worth noting that with this change, the opening and closing sounds are emanating from the coordinates of the originating room door and the arriving room position, so the opening sound cuts off if you successfully change rooms (it won't have overlapping opening and closing sounds).

However quickly going in and out of the same door will result in hearing part of the opening or closing sound.  I thought this was a major problem but after thinking about it more it seems ok, just different.

akirmse: I'm still not sure of the answer to my original question.  If your character goes through a door, do you hear 2 sounds or 1?  2 seems bad, since they can overlap, etc.

adrienlaws: You hear 2 sounds but they do not overlap.

When changing rooms the opening sound is cut short.  95% of the time you don't even hear the opening sound.  I recorded a video showing what it's like when you do hear the sound.  The only place I could easily reproduce it was South Barloque.  After the first room change, going back doesn't cause it anymore probably because there's some caching happening.  I had to reboot my computer to get it to play the door opening sound when entering South Barloque.

The "grunt" is part of the close door sound resource and would be changed at a later time (outside of this PR).

The current behavior (before this PR) is that when you go through a door, door sounds continue playing until they end.  With the PR the opening door sound stops the moment you leave the room where you opened the door.  Then a door closing sound plays in the arriving room at your arrival coordinates.  In effect, you almost never hear the door opening sound if you are the one opening a door.

TDLR; 2 sounds, but they don't overlap

https://github.com/user-attachments/assets/94986e05-87cc-4f34-88a2-5f3476d39957

kod/object/active/holder/room.kod
```diff
@@ -2760,32 +2762,63 @@ messages:
...
+   "Plays a sound for all users in the room.  Sends WaveSendUser to each "
+   "user with the provided wave_rsc, source object, row, col, and "
+   "cutoff_radius.  If except is set, that user is skipped."
    {
       local i;
 
       for i in plActive
       {
-         Send(First(i),@WaveSendUser,#wave_rsc=wave_rsc,#source_obj=what);
+         if First(i) = except
```
akirmse: Simpler as 
```
if First(i) <> except
  blah
````

File: kod/util/settings.kod [hunk 6]
```diff
@@ -147,6 +147,11 @@ properties:
    % contests too!)  This value can be changed to tweak the penalty in realtime
    piBroadcastManaCostPercent = 50
 
+   % If enabled, door opening sounds are shared with all players in the room
+   % with 3D positional audio.  If disabled, only the player opening the door
```
akirmse: 3D positional audio is not relevant to this setting.. I would leave that part out.

kod/object/active/holder/room.kod
```diff
@@ -2760,32 +2762,63 @@ messages:
...
+               }
+            }
 
             Send(SYS,@UtilGoNearSquare,#what=what,
                  #where=Send(SYS,@FindRoomByNum,#num=Nth(i,3)),
                  #new_row=Nth(i,4),#new_col=Nth(i,5),#new_angle=iAngle);
 
+            % Play door closing sound from the opener's position in the
+            % destination room.  The opener is excluded. Legacy mode 
+            % does not play a close sound.
```
akirmse: "Legacy mode" is an unexpected term.  Since it's not the name of the system setting, it's not clear that the two are related.

PR #1374, opened 2026-02-05
Fix misspelling of transcendent in dm create itematt command

Aesica:
When using the dm create itematt command, the keyword for the transcendent (soft white light) property is misspelled as "transcendant."  This corrects it to the proper spelling as "transcendent."

### DIFF
--- a/kod/object/passive/itematt/iatransc.kod
+++ b/kod/object/passive/itematt/iatransc.kod
@@ -29,7 +29,7 @@ resources:
    itematttranscendant_desc = "It glows with a soft, white light."
    itematttranscendant_name = "transcendant %s"
 
-   transcendant_DM = "transcendent"
+   transcendent_DM = "transcendent"
 
 classvars:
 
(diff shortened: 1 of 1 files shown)


kod/object/passive/itematt/iatransc.kod
```diff
@@ -29,7 +29,7 @@ resources:
    itematttranscendant_desc = "It glows with a soft, white light."
    itematttranscendant_name = "transcendant %s"
 
-   transcendant_DM = "transcendant"
+   transcendant_DM = "transcendent"
```
akirmse: Might as well fix the name of the variable while you're at it.

Aesica: Done.

How far should I go?  I could probably go through the the entire kod structure and fix any reference to things like IA_TRANSCENDANT, ItemAttTranscendant, etc if you want.  "itematttranscendant_name = "transcendant %s"" as well, although I'm not sure what the name is even used for, as I've never seen itematt names displayed anywhere as a player.

PR #1396, opened 2026-04-01
Login telemetry: hardware, GPU, FPS & ping in `AP_LOGIN`

sicoderealm:
Folding performance diagnostic data (OS build, GPU vendor/VRAM, renderer type, last session FPS) into existing unused fields of `AP_LOGIN`. FPS stats are persisted to the INI between sessions so they're available at next login. 
   
`May 18 2026 20:33:29 | Login account 3 from 127.0.0.1: Win 11 (build 26200), 64 GB RAM, 3072x1728@32bpp, D3D, NVIDIA GeForce RTX 4070 Ti | Ping 11ms avg, 203 low 1%, 8 best (Green) 12GB, session ~2min, fps avg=101 1%low=87 max=125`


clientd3d/util.c
```diff
@@ -10,6 +10,62 @@
  */
 
 #include "client.h"
+#include <dxgi.h>
+#pragma comment(lib, "dxgi.lib")
```
akirmse: Including libraries via #pragmas is horrible and drives build systems crazy.  Please just do this the normal way.

blakserv/game.c
```diff
@@ -406,8 +406,6 @@ void GameProtocolParse(session_node *s,client_msg *msg)
 {
    user_node *u;
    int object_id;
-   char *ptr;
```
akirmse: Revert this file (not related to this change).

clientd3d/com.c
```diff
@@ -185,7 +185,7 @@ void CloseConnection(void)
 		closesocket(sock);
 		WSACleanup();
 		break;
-		
+
```
akirmse: Revert this file.

clientd3d/gameuser.c
```diff
@@ -421,7 +421,7 @@ void UserToggleMusic(bool music_on)
  */
 void UserQuit(void)
 {
-      RequestQuit();   
+   RequestQuit();
```
akirmse: Revert this file.

clientd3d/util.c
```diff
@@ -10,6 +10,62 @@
...
+      }
+   }
+   pFactory->Release();
+   return vramMB;
+}
+
+// Map session duration (seconds) to a 4-bit bucket for wire encoding.
+// 0=no data, 1=<1min, 2=1-5min, 3=5-15min, 4=15-30min, 5=30-60min,
+// 6=1-2hr, 7=2-4hr, 8=4-6hr, 9=6-8hr, 10=8-12hr, 11=12+hr
+static unsigned int SessionSecsToBucket(int secs)
```
akirmse: Just make this an int; unsigned serves no purpose here.

clientd3d/util.c
```diff
@@ -440,67 +496,111 @@ extern unsigned short gCRC16;
...
-   mem.dwLength = sizeof(MEMORYSTATUS);
-   GlobalMemoryStatus(&mem);
+   // Use RtlGetVersion to get the real Windows version (bypasses the GetVersionEx
+   // compatibility shim that reports 6.2 on Windows 10/11).
+   {
+      typedef LONG (WINAPI *FnRtlGetVersion)(OSVERSIONINFOEXW *);
+      OSVERSIONINFOEXW osInfo = { sizeof(OSVERSIONINFOEXW) };
+      HMODULE ntdll = GetModuleHandleW(L"ntdll.dll");
+      FnRtlGetVersion pRtlGetVersion = ntdll
+         ? (FnRtlGetVersion)GetProcAddress(ntdll, "RtlGetVersion") : NULL;
```
akirmse: This is gross; we should not be looking up function pointers in Windows DLLs.

clientd3d/util.c
```diff
@@ -440,67 +496,111 @@ extern unsigned short gCRC16;
...
    HDC dc;
 
-   version.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
-   GetVersionEx(&version);
-
-   mem.dwLength = sizeof(MEMORYSTATUS);
-   GlobalMemoryStatus(&mem);
+   // Use RtlGetVersion to get the real Windows version (bypasses the GetVersionEx
+   // compatibility shim that reports 6.2 on Windows 10/11).
+   {
```
akirmse: Where do these extra blocks keep coming from?  Is this AI-generated?

clientd3d/util.c
```diff
@@ -440,67 +496,111 @@ extern unsigned short gCRC16;
...
    s->screen_height = GetSystemMetrics(SM_CYSCREEN);
 
-   s->reserved = GetDeviceCaps(dc,BITSPIXEL); //devMode[0].dmBitsPerPel;
-   s->reserved &= 0xFF;
+   // Pack GPU/perf diagnostics into the bandwidth field:
+   // bits 29-28: renderer_mode (0=Software, 1=D3D, 2=D3D+GpuEff)
+   // bits 27-20: vramMB / 256  (8 bits; 0 when D3D not yet initialised)
+   // bits 19-16: session length bucket (see SessionSecsToBucket; 0=no data)
+   // bits 15-0:  PCI vendor ID (0x10DE=NVIDIA, 0x1002=AMD, 0x8086=Intel)
+   {
```
akirmse: ???

clientd3d/util.c
```diff
@@ -440,67 +496,111 @@ extern unsigned short gCRC16;
...
 
-   s->reserved |= (GetPartnerCode() << 8);
-   s->reserved &= 0xFFFF;
+   // Pack color depth and partner code into low 16 bits of reserved; pack last-session
+   // FPS (from previous play session, saved to INI on logoff) into high 16 bits.
+   // bits 31-24: last_avg_fps / 4  (8 bits, 0 = no data)
+   // bits 23-16: last_low_fps / 4  (8 bits)
+   // bits 15-8:  partner code
+   // bits 7-0:   screen color depth in bits
+   {
```
akirmse: ???

clientd3d/util.c
```diff
@@ -440,67 +496,111 @@ extern unsigned short gCRC16;
...
-	 break;
-      }
-      display |= disp << shift;
-      iModeNum++;
+      int max_raw = config.last_max_fps / 4;
+      unsigned int maxFpsPacked = (max_raw <= 0) ? 0u : (max_raw > 255) ? 255u : (unsigned int)max_raw;
+
+      // wine_get_version is exported by Wine's ntdll but never by real Windows ntdll.
+      HMODULE ntdll = GetModuleHandleW(L"ntdll.dll");
+      unsigned int wineBit = (ntdll && GetProcAddress(ntdll, "wine_get_version")) ? 1u : 0u;
```
akirmse: whaaaaa?

akirmse (commented):
- I'm spending a lot of time reviewing whitespace and extra braces.  Could you maybe give your PRs a once-over yourself to fix these trivial things before asking for review?

- There are some really unexpected things in here like looking up function pointers in Windows DLLs that I don't think we should be doing.  Also the bit packing is hard to understand and doesn't seem necessary (what are we saving, a few bytes?).  Maybe clean some of this up and we can take another look.

blakserv/game.c
sicoderealm: Would you like to keep the advert related info? If not I think we can remove this ptr

clientd3d/util.c
sicoderealm: I think the bit packing makes this too cryptic, I've unpacked as suggested

clientd3d/util.c
sicoderealm: I think the bit packing makes this too cryptic, I've unpacked as suggested

clientd3d/util.c
sicoderealm: I think it works, but given - we can try registry key instead? Would be helpful to know
https://stackoverflow.com/questions/7372388/determine-whether-a-program-is-running-under-wine-at-runtime

sicoderealm (commented):
I'm going to give this some more testing at the weekend, then hopefully will be ready for a fresh review. Thanks for feedback

blakserv/game.c
akirmse: If you want to propose removing it, please do it in a separate PR.

sicoderealm: raised https://github.com/Meridian59/Meridian59/pull/1421 ty

blakserv/synched.c
```diff
@@ -161,7 +161,7 @@ void SynchedProtocolParse(session_node *s,client_msg *msg)
    switch (msg->data[0])
    {
    case AP_LOGIN : 
-      if (msg->len < 39) /* fixed size of AP_LOGIN, before strings */
+      if (msg->len < 75) /* fixed size of AP_LOGIN, before strings */
```
akirmse: Hm.. actually, I think there's a problem here.  After we publish the update, old clients will be logging in to new servers.  So they will be sending the old AP_LOGIN message to even know that they need to update.  So the server has to be prepared to handle both the old and new formats of the message.

sicoderealm: Very good point. I've been upgrading the PR to support this and doing some testing. We will need to update the client binary and all supporting DLLS i in the patch achieve but it works for the update workflow form old to new version. Ready for another look soon - thank you

akirmse:
There are a few high-level changes that need to happen before I can start looking at this:

- I can't keep reviewing massive AI-generated changes like this when I'm finding the same problems over and over again, or I will become overwhelmed and just start rejecting anything that looks like this.  Please review the code yourself before sending it to me.  Fix silly things like macros instead of constants, pointless comments, etc.

- Let's maybe try to keep AI slop in its own files for the most part, rather than shoving enormous unrelated functions into existing files, making them huge.

- On the server side, the number of session stats has gotten out of hand, and they persist in memory forever, even though they're only used for one logging statement and then never again.  So please 1) Remove any stat you're not sure you're going to need  2) Find a way to not store the stats forever, wasting memory.

- Am I reading it right that you're keeping 36000 frames of performance data?  There has to be a better way.

- There are unexplained changes to the ordering of the login process in the client code.  Let's remove those if possible.  This PR is just supposed to log more stuff.

I see a trend of AI turning our codebase (which is not that amazing to begin with) into something unwieldy and unreadable because it's so easy to generate new code.  I would rather not have any AI changes than see that happen.

sicoderealm:
Thanks for taking the time to review this.

I understand the concerns around maintainability, code size, memory usage, and the general direction of the changes. I'll move the PR back to draft and revisit it before requesting another review.

I'll take another pass to reduce the scope, remove unnecessary complexity, clean up the implementation, and reassess which metrics are actually worth collecting. I'll also review the login flow changes and see if they can be removed entirely.

Given the amount of rework needed, this is likely to be deprioritized for now while I focus on other work. I'll reopen it when I have a substantially revised version that's easier to review.

Issue #1324, opened 2026-01-04
Inventory rendering glitch


sicoderealm:
The inventory view window will occasionally not render correctly, not showing the background as expected. This seems to happen when first logging in or after a system save. This issue seems to prevent the inventory scrollbar from appearing correctly and the inventory background fully spanning the area. It might need a non full-size window and an inventory that needs a scrollbar. It looks rather incomplete and broken and would be great to fix.

Inventory view issue:
<img width="314" height="474" alt="Image" src="https://github.com/user-attachments/assets/7815dde6-0dba-422e-8e70-9935bd7e0358" />

Correct inventory view after fixing (with scrollbar and background)
<img width="303" height="474" alt="Image" src="https://github.com/user-attachments/assets/328278f2-45a6-4b62-9a87-f418a0e44420" />

You can partially resolve the view often by entering a new zone. However, to fully fix it you need to scroll to the end of the inventory.

sicoderealm:
I also wondered if this little black line should span the entire inventory width or if its a bug and shouldn't be there at all

<img width="533" height="123" alt="Image" src="https://github.com/user-attachments/assets/dbd3df83-621c-4c6d-86bf-670a96533b76" />

PR #1397, opened 2026-04-03
feature: add client dark theme and fix bugs

adrienlaws:
🚧 draft PR - will document further shortly 🚧 

## What
- Added dark theme
  - Alternate color palette, flat UI frame bitmaps, modernized stat bars, owner-drawn menu bar
  - Integrates dark UI bitmap artwork from  #440 as togglable alternate resources 
  - Settings dialog has a Default/Dark dropdown with live theme switching
  - Text coloring is adjusted for the dark theme
  - Theme system designed for future expansion
- Fixed multiple pre-existing client UI bugs
    - Fixes the issues in #1324 
    - "Reverse L" strip in inventory blocking the scrollbar has been fixed
    - Fixes and consolidates issues with the scrollbars in the inventory, skills, spells, and stats areas
- Added mouse wheel scrolling to inventory, stats, spells, and skills sidebar panes (previously scrollbar or keyboard only)
- Added ability to "darken" existing pixel graphics to allow darkening of existing graphical resources

## Why
- Several players and contributors have requested a dark mode option
- The sidebar panes had no mousewheel handling, so mouse wheel scrolling did nothing
  - It is helpful to scroll through the inventory
- Multiple pre-existing UI bugs around inventory/stats tab switching, sidebar repainting, and orphaned bitmap elements were discovered and fixed during the refactor
- #440 provided the inspiration, graphics, and text colors from @Aesica and @Dirhai 

## Example
meridian_Cd39KUkzaE
- Note: 
  - Different health, mana, vigor, stat bar colors / configuration
  - The ability to scroll the inventory with a mouse wheel (you don't have to click on the scroll bar to do it)


akirmse (commented):
This PR is enormous.  Would you rather split it up somewhat, or do you want to do it as-is?  At this size, it will probably be a good number of weeks before this can go in.

One big concern I have is the looking of up functions in uxtheme.dll by hard-coded integers.  That's extremely unsafe and we shouldn't do it.  Claude tells me that this DLL is present all the way back to Windows XP, so can't we assume that it's present?

adrienlaws:
I removed the ordinal lookups of uxtheme.dll - thanks for pointing that out.

They weren't doing anything anyway and I should have removed them.  The reason they were in was I had been trying to darken windows drop-down menu items but finally settled on leaving "deeper" theme modification like dialog boxes, menus, etc out of scope.  I was tinkering with that kind of lookup because those lookups are not published with named exports.  The actual named exports are fine and work so I left those in.  In short, there is no change in functionality from removing those ordinal lookups and everything still works.

Regarding splitting the PR, I would prefer to keep this as a single PR if that works for you.  The reason is that it took a ton of iterations (probably 100+ builds) to get to this state and I think it's all working together, finally.   I'm worried if I split it up things will break and it will take a while to get back to a steady state.  That said, I can try to break it apart if you feel strongly about it.

Meridian59:
OK.  Reviews are going to take a long time, and you may develop merge
conflicts.  You also won't be able to check anything in else until this is
done.

In the future look for smaller pieces.  Like maybe you could have fixed
some existing bugs first, then added some new colors and configs, etc.  As
long as the dark/light UI toggle went in last, the checked-in code would
always work, which is a good policy.

akirmse:
I started going through and making lots of detailed comments, but at this early stage maybe it's better to make two overall points:

- With AI assistance (I am assuming), it's possible to do something like reproduce the entire behavior of a scrollbar just so that you can replace its graphics.  That is not a good tradeoff IMO.  The cost of the initial implementation may be low, but the long-term maintenance costs are huge.  It also dramatically increases complexity and makes future changes harder.

- I see a lot of places where the code looks like "Call dark mode function, if that doesn't do anything, call the existing default function".  That means that a ton of places in the code are aware of themes, whereas the goal should be to minimize that, and hide the details of themes so that callers don't need to know about them.  I'd recommend restructuring those so that the existing function keeps its API to the extent possible, and handles themes internally.

adrienlaws:
Closing this PR and will open smaller PRs to review smaller parts of it at a time.

PR #1398, opened 2026-04-05
Extra stomach fullness feedback

neiltaco:
This PR adds new and updated messages related to a player's stomach capacity.  The reason for this PR is to help inform both new and experienced players about how much they can eat by using in-game messages to help keep the game immersive without using exact numbers.

In the game's current state, the player is only notified that they are too full to eat or drink if they attempt to consume something.  The 'too full' message always shows whether the player still have room for smaller foods or literally cannot fit anything else, which can be misleading.

- Added two new messages show when a player's stomach gets half full (50%) or almost full (75%).
  - These only display if the stomach value passes the 50%/75% thresholds to reduce message spam.
- Updated existing 'too full' messages to state that the player is 'too full to eat/drink that right now."
  - This helps clarify that the player cannot eat that item at the moment but might be able to eat something smaller.
- Fixed potions not passing in themselves into `ReqEatSomething()`.  This means they will properly use the "too full to drink" message.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -321,8 +321,10 @@ resources:
    player_desc_enchanted_none = "%q"
    player_desc_enchanted = "%s has %s%s.\n\n%s"
    
-   player_too_full = "You are too full to eat."
-   player_too_full_drink = "You are too full to drink."
+   player_half_full = "You feel fairly satiated."
+   player_almost_full = "Your stomach feels very stuffed and crammed."
+   player_too_full = "You are too full to eat %s%s right now."
+   player_too_full_drink = "You are too full to drink %s%s right now."
 
    player_to_hunters = "You detect that %s%s could be found somewhere around %s"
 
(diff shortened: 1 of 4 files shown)


neiltaco:
This PR got made into a draft again since there were some minor grammar errors caused by the 'grammar fix' commit.  Particularly when you do something like cast reveal on an unidentified bait/karahol's potion:

`"You concentrate on potion, and gradually its secrets become plain for all to see!"`

neiltaco:
This PR got made into a draft once more since the new messaging is a bit buggy at the Spring of Vitality.  It properly shows the half-full and nearly-full messages, but attempting to drink when too full will show the following message:  `“You are too full to eat the something right now.”`  It's likely that the poFoodProxy created for the cave room doesn't have properties set for the spring water.

neiltaco:
This PR should be ready now.  I reverted the name changes to the bait and karahol potions since their names were getting a bit too verbose.  "vial of Widow Qesino's Secret Sauce" is certainly a mouthful.  That and it is technically a change that is out of scope and probably best handled in a separate PR.

The new 'too full' messages may also be a bit too verbose too with mentioning the food item too, and introduced some problems with properly displaying them.  I changed both messages to "You are too full to eat/drink that right now."  This was a simpler fix that still hints that they can't fit the selected food item at this moment.

neiltaco:
Closed since this PR has unfortunately inherited a messy commit history from a local master branch.  Continued in #1404

PR #1399, opened 2026-04-10
Original static dynamic lights look and appearance

sicoderealm:
Restore non-flickering lights to their pre-dynamic flicker state, ensuring that KOD-defined light sources (e.g., wall torches and fire pits) render correctly as as expected. We also prime new dynamic lights to not be flicking, to support the current mechanics used to place such lights in game.

image

### DIFF
--- a/clientd3d/d3drender_lights.c
+++ b/clientd3d/d3drender_lights.c
@@ -132,8 +132,8 @@ static int CalculateFlickeredIntensity(const LightSourceData& lightData, float* 
    }
    else
    {
-      // Non-flickering lights use the original DLIGHT_SCALE formula to preserve
-      // their pre-flicker-feature appearance (radius and brightness).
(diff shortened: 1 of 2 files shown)


clientd3d/d3drender_lights.c
```diff
@@ -132,7 +132,9 @@ static int CalculateFlickeredIntensity(const LightSourceData& lightData, float*
    }
    else
    {
-      flickeredIntensity = D3DLightScale(lightData.baseIntensity);
+      // Non-flickering lights use the original DLIGHT_SCALE formula to preserve
```
akirmse: "Original" isn't going to mean much in 15 years when someone looks at this again.  Maybe give this future developer more of a clue about what's going on?

sicoderealm: hopefully a little better now - ty

PR #1402, opened 2026-04-14
Refactor d3drender.h by replacing macros and externs with inline functions and constexprs.

neiltaco:
This PR refactors and cleans up `d3drender.h` by reorganizing the layout, and replacing legacy macros and extern variables.  This improves header readability, and it helps with debugging and type safety since macros lack these features.  This PR also optimizes performance by having constants and division calculated at compile-time using constexpr.

- Replaced macro functions with inline functions.
  - Updated other source files to use the new inline functions instead of the previous macros.
  - Adjusted certain inline functions to move division to constexpr and favor multiplication to reduce CPU cycles.
  - `D3DRENDER_SET_STENCIL_STATE` was replaced with three separate inline functions to help keep the parameter count short.
- Replaced macro variables with static constexpr constants.
- Replaced extern variables with inline globals.
  - Removed `gbAlwaysRun` since it was never used.
- Reorganized layout into this order: Constants -> Structs -> Globals -> Inline Functions -> Prototypes


neiltaco:
Earlier I had asked AI if `D3DRENDER_SET_STENCIL_STATE` could be broken up into smaller functions to prevent having to passing in six parameters.  I used the three inline function version that I realize now is missing some parameters from the original macro.  I'll check if this is still working as intended or needs a rework.  Maybe six parameter might be ok, or a struct could be passed in instead.

neiltaco:
Closed since this PR inherited a messy commit history from the local main branch.  New PR is in #1403

PR #1403, opened 2026-04-16
Refactor d3drender.h: replace both macros and externs with inline functions and constexprs.

neiltaco:
Reupload of #1402, which inherited a messy commit history from a local master branch.  This version should have a clean commit history now.

This PR refactors and cleans up `d3drender.h` by reorganizing the layout, and replacing legacy macros and extern variables.  This improves header readability, and it helps with debugging and type safety since macros lack these features.  This PR also optimizes performance by having constants and division calculated at compile-time using constexpr.

- Replaced macro functions with inline functions.
  - Updated other source files to use the new inline functions instead of the previous macros.
  - Adjusted certain inline functions to move division to constexpr and favor multiplication to reduce CPU cycles.
  - `D3DRENDER_SET_STENCIL_STATE` was replaced with three separate inline functions to help keep the parameter count short.
- Replaced macro variables with static constexpr constants.
- Replaced extern variables with inline globals.
  - Removed `gbAlwaysRun` since it was never used.
- Reorganized layout into this order: Constants -> Structs -> Globals -> Inline Functions -> Prototypes


clientd3d/d3drender.h
```diff
@@ -136,22 +77,155 @@ typedef struct font_3d
...
+	return (coordinate > -absRange) && (coordinate < absRange);
+}
 
-extern int		gNumVertices;
-extern BOOL		gbAlwaysRun;
-extern int		gD3DEnabled;
-extern int		gScreenWidth;
-extern int		gScreenHeight;
+// Controls alpha testing, which is a 'pass/fail' check for pixels based on their transparency.
+inline void D3DRender_SetAlphaTestState(BOOL enable, DWORD alphaRef, D3DCMPFUNC comparisonFunc)
```
akirmse: Most of these functions are larger than what's traditionally marked inline.  I would just declare them in the header here as normal functions, and move the definition to the .c file.  The compiler is much better than humans at knowing when to inline.

clientd3d/d3drender.h
```diff
@@ -136,22 +77,155 @@ typedef struct font_3d
...
+	// Minimum radius to ensure lights don't cull/pop too early.
+	constexpr float DLIGHT_MIN_RADIUS = 4000.0f;
+	
+	return (intensity * DLIGHT_SCALE_FACTOR) + DLIGHT_MIN_RADIUS;
+}
+
+// Returns true if a coordinate is within symmetric range (-range, range).
+inline constexpr bool D3DRender_InBounds(float coordinate, float range)
+{
+	const float absRange = (range < 0.0f) ? -range : range;
```
akirmse: Can't this be simply
```
return fabs(coordinate) < range
```
assuming the range is positive, as it appears to be at all call sites?  Or you could take fabs(range) if it's sometimes negative.

akirmse (commented):
Mostly looks good, and I appreciate getting rid of the macros.

clientd3d/d3drender.h
neiltaco: That's definitely cleaner so I went with your suggestion.  I also made the function `inline bool`, since the compiler raised errors about using `fabs()` within a `constexpr` function.

neiltaco: I checked the original macro once again and it always applied `fabs()` to `range` (or `_w` in the macro).

This PR currently doesn't do that because I believe it would be more apparent that something went wrong if a negative value got passed in.   This also means it'll change the behavior a bit from the original macro.  For now I haven't applied `fabs()` to `range`, but I'm happy to add it in if you think this is safer.

PR #1404, opened 2026-04-16
Expanded stomach fullness feedback

neiltaco:
Reupload of #1398 because it inherited a messy commit history from a local master branch.  This one should have a clean commit history now.

This PR adds new and updated messages related to a player's stomach capacity.  The reason for this PR is to help inform both new and experienced players about how much they can eat by using in-game messages to help keep the game immersive without using exact numbers.

In the game's current state, the player is only notified that they are too full to eat or drink if they attempt to consume something.  The 'too full' message always shows whether the player still have room for smaller foods or literally cannot fit anything else, which can be misleading.

- Added two new messages show when a player's stomach gets half full (50%) or almost full (75%).
  - These only display if the stomach value passes the 50%/75% thresholds to reduce message spam.
- Updated existing 'too full' messages to state that the player is 'too full to eat/drink that right now."
  - This helps clarify that the player cannot eat that item at the moment but might be able to eat something smaller.
- Fixed potions not passing in themselves into `ReqEatSomething()`.  This means they will properly use the "too full to drink" message.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -321,8 +321,10 @@ resources:
    player_desc_enchanted_none = "%q"
    player_desc_enchanted = "%s has %s%s.\n\n%s"
    
-   player_too_full = "You are too full to eat."
-   player_too_full_drink = "You are too full to drink."
+   player_half_full = "You feel fairly satiated."
+   player_almost_full = "Your stomach feels very stuffed and crammed."
+   player_too_full = "You are too full to eat %s%s right now."
+   player_too_full_drink = "You are too full to drink %s%s right now."
 
    player_to_hunters = "You detect that %s%s could be found somewhere around %s"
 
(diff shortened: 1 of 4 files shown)


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -5661,6 +5671,20 @@ messages:
    {
       Send(self,@AddExertion,#amount=-10000*nutrition);
       piStomach = piStomach + filling;
+	  
+      % Display fullness messages only if the player's stomach value passed the 50% or 75% thresholds.
+      % This helps avoid spamming the messages on every bite or drink.
+      if (piStomach >= 50 AND piStomach < 75) AND piLastStomachValue < 50
```
akirmse: Adding a new property to Player will permanently cost about 500 kb of server memory, which seems like a lot for this benefit.  What about instead simply storing the previous stomach value in a local variable before updating piStomach a few lines above this, and then comparing the pre- and post- values here instead?

neiltaco: Done!  And I'll be more mindful next time on when variables can be made local.

PR #1406, opened 2026-04-19
Refactor d3drender.c

neiltaco:
This PR refactors `d3drender.c` to be cleaned up and organized.  This includes moving part of the code in the source file into the header file.  The goal is similar to the header file refactor: to help with readability, debugging, and type safety.  

There should be no functional changes.  This PR optimizes the code by removing unused variables and converting certain variables into `static constexpr`, and there's an -87 net reduction in lines.  The high line count is due to moving helper functions into the header, and rearranging and organizing where variables and functions were defined.  Many macros were replaced with their C++ equivalent, and the whitespace in `d3drender.c` was adjusted to tab-based spacing to be consistent with the rest of the source file.

For now, certain externs were left in since they may trigger a lot more line changes in other source files.  If necessary, replacing them can be done in this PR or in a separate PR.

Refactoring
- Reorganized `d3drender.h` and `d3drender.c`.
  - Certain global variables from the source file were moved into the header.
  - Helper functions in `d3drender.c` moved into `d3drender.h` as inline helper functions.  Also removed them from Prototypes section.
  - `d3drender.h` globals section moved below structures section.
  - Split off external functions into its own section at the end of the header.
- Remove unused variables and function parameters.
  - Removed unused texture cache limits from top of source file and replaced them `static constexpr` of the ones used in `D3DRenderInit()`.
  - Debug variables for tracking render time were removed from `D3DRenderBegin()`.
- Rect setup in `d3drender.c` is made into its own helper function `GetScreenRect()`
- Refactored font glyph loop in `D3DRenderFontInit()` to use `int i = 0` instead and have character offset derived from it.
- Decoupled `x` and `y` variables in `D3DRenderFontInit()` for separate tasks.

Code Cleanup
- Replaced `NULL` with `nullptr`.
- Replaced Direct3D's C-style macros with C++ equivalents.
- Moved function variables declarations from start of function to at point-of use.
- Changed C-style type casting to use `static_cast`, `reinterpret_cast`, and `std::bit_cast`.
- Removed some redundancy.
  - Renamed `D3DRenderFramebufferTextureCreate()` to `D3DRender_CaptureEffect()`.  Removed unused `width` and `height` parameters.
  - Removed `getWhiteLightTexture()`.  Updated `d3drender_materials.c` to use `D3DRenderLightsGetWhite()` directly.
- Fixed possible memory leak in `D3DRenderShutdown()` where `gFont.kerningPairs` only deleted if textures existed.

### DIFF
--- a/clientd3d/client.h
+++ b/clientd3d/client.h
@@ -100,6 +100,7 @@ M59EXPORT void _cdecl dprintf(const char *fmt,...);
 #include <string>
 #include <vector>
 #include <algorithm>
+#include <bit>
 
 #include "resource.h"
 #include "proto.h"
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -73,7 +73,6 @@ static D3DMATRIX view, mat, rot, trans, proj;
 // External Globals //
 //////////////////////
 extern long				viewer_height;
-extern PDIB				background;         /* Pointer to background bitmap */
 extern ObjectRange		visible_objects[];    /* Where objects are on screen */
 extern int				num_visible_objects;
 extern DrawItem			drawdata[];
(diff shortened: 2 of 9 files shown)


neiltaco:
Well now that I've actually looked at the line comparisons in the 'Files Changed' tab, I'm realizing that it's difficult to read the code differences since the line moves are mixed and tangled up with refactoring lines.  I'll most likely close this PR and first upload the reorganization first before doing the refactor PR.

PR #1407, opened 2026-04-20
fix: restore soft falloff for one-shot positional sounds

adrienlaws:
## What

- Fixes a regression introduced in the OpenAL migration (PR #1293) where one-shot positional sounds (combat, spells, etc.) became silent past a hard ~16 tile cutoff
- Splits the 3D distance model in `SoundPlay()` into two cases: sounds with a Blakod-set radius (looping ambients and one-shots with an explicit range) keep their hard cutoff; sounds with no Blakod-set radius use a smooth inverse falloff that never hits zero
- Updated `audio.md` to reflect these changes and also removed the 80 char carriage returns in the file

## Why

- Players reported being unable to hear nearby combat or spells from sources more than ~16 tiles away after the OpenAL update
> Oof I need to test it out a bit more but I believe the 'dynamic drop off' on sound volume is quite steep. You can hear a sound, such as combat, getting quieter as you move away from it, and after quite a short distance - maybe battle bow range - you cannot hear it at all. zero volume. So a ton of sound information is being lost as to what is happening

> You can only hear anything happening about 10 feet away from your character at any given time. 

- Pre-OpenAL behavior used two separate paths:
  - one-shots: `volume = MAX_VOLUME * 2 / distance` (a 1/d curve, never reached zero, audible across the entire room)
  - looping ambients: hard linear cutoff at the Blakod-defined `iSoundRadius` (correct, this is the whole point of the radius parameter)
- The OpenAL refactor unified both paths into `AL_LINEAR_DISTANCE_CLAMPED` with `AL_MAX_DISTANCE = radius`, applying the loop-style hard cliff to one-shots too
- The server defaults `cutoff_radius = 0` for nearly every sound it sends in `WaveSendUser` (`user.kod`), so one-shots fell back to the client's `VOLUME_CUTOFF_DISTANCE = 16` and went silent past 16 tiles
- Some Blakod paths (door open/close in `room.kod`) deliberately set `cutoff_radius = 4` so a door slam in one corner of a large hall does not bleed across the whole room.  That intent was correct and must be preserved
- The hard cutoff is still required for ambient loops, without it every fountain in every loaded room would mix together forever

## How

- `clientd3d/game.c GamePlaySound()`: pass the raw `radius` parameter through to `PlayWaveRsc` instead of the defaulted `cutoff` value, so the audio layer can tell "the Blakod set a radius" (`radius > 0`) apart from "no opinion" (`radius == 0`).  The internal volume calc still uses `cutoff` for the legacy non-positional path
- `clientd3d/audio_openal.c AudioInit()`: enable `AL_EXT_source_distance_model` via `alEnable(AL_SOURCE_DISTANCE_MODEL)` so each source can pick its own distance model
- `clientd3d/audio_openal.c SoundPlay()`: the positional branch picks one of two attenuation curves:
  - Blakod gave us a radius (looping emitters and one-shots with an explicit range): `AL_LINEAR_DISTANCE_CLAMPED`, `ref=1`, `max=radius`.  Honors the Blakod author's intent and prevents one-shots from leaking across large rooms; loops fall back to `VOLUME_CUTOFF_DISTANCE` if radius is 0, which should not happen in practice
  - No Blakod radius (combat, spells, generic effects): `AL_INVERSE_DISTANCE_CLAMPED`, `ref=2`, `max=10000`, `rolloff=1`.  Smooth `1/d` falloff matching the old MSS curve, never reaches zero
- The global default model stays linear-clamped so any future code path that does not opt in still gets safe behavior
- Updated `docs/audio.md` Distance Model section to document the split and added a flowchart of the routing in `SoundPlay()`

```mermaid
flowchart TD
    A["SoundPlay called"] --> B{"Has positional<br/>coordinates?"}
    B -->|"No"| C["Centered on local listener<br/>(SOURCE_RELATIVE, no falloff)<br/>Plays equally L/R for this client only"]
    B -->|"Yes"| D{"SF_LOOP flag set?"}
    D -->|"Yes (fountain, firepit)"| E["AL_LINEAR_DISTANCE_CLAMPED<br/>ref=1, max=radius<br/>Hard cutoff at radius"]
    D -->|"No"| G{"Blakod-set radius?<br/>(radius > 0)"}
    G -->|"Yes (door, area effect)"| H["AL_LINEAR_DISTANCE_CLAMPED<br/>ref=1, max=radius<br/>Hard cutoff at radius"]
    G -->|"No (combat, spell)"| F["AL_INVERSE_DISTANCE_CLAMPED<br/>ref=2, max=10000<br/>Soft 1/d falloff, no cutoff"]

    style C fill:#1864ab,color:#fff
    style H fill:#2f9e44,color:#fff
```

### Before

Combat one-shot at varying distances (radius defaulted to 16):

| Distance from source (tiles) | Gain |
|------------------------------|------|
| 1                            | 1.00 |
| 4                            | 0.80 |
| 8                            | 0.53 |
| 16                           | 0.00 (silent) |
| 32                           | 0.00 (silent) |

A player casting a spell 20 tiles away in the same room is completely inaudible.

### After

Combat one-shot with no Blakod-set radius (inverse falloff):

| Distance from source (tiles) | Gain |
|------------------------------|------|
| 1                            | 1.00 |
| 4                            | 0.50 |
| 8                            | 0.25 |
| 16                           | 0.125 |
| 32                           | 0.0625 |

Door open/close with `cutoff_radius = 4` (linear clamped, unchanged from before):

| Distance from source (tiles) | Gain |
|------------------------------|------|
| 1                            | 1.00 |
| 2                            | 0.67 |
| 3                            | 0.33 |
| 4                            | 0.00 (silent) |

Distant combat is faintly audible across the room (matching pre-OpenAL behavior), while doors still cut off cleanly at 4 tiles so a busy corridor does not become a wall of door noise.

### DIFF
--- a/clientd3d/audio_openal.c
+++ b/clientd3d/audio_openal.c
@@ -188,10 +188,8 @@ bool AudioInit(HWND hWnd)
    ALfloat listenerOri[] = { 0.0f, 0.0f, -1.0f, 0.0f, 1.0f, 0.0f }; // forward, up
    alListenerfv(AL_ORIENTATION, listenerOri);
 
-   /* Default distance model: linear clamped, used for SF_LOOP ambient
-    * emitters that should cut off at their kod-defined radius (fountains,
-    * firepits, etc.).  Per-source distance model is enabled below so that
-    * one-shot sounds can use an inverse curve instead (no hard cutoff). */
+   // Use linear falloff by default, but let each source pick its own
+   // curve so they don't all have to fade out the same way.
    alDistanceModel(AL_LINEAR_DISTANCE_CLAMPED);
    alEnable(AL_SOURCE_DISTANCE_MODEL);
 
@@ -1029,13 +1027,13 @@ bool SoundPlay(const char* filename, int volume, BYTE flags,
       alSourcei(source, AL_SOURCE_RELATIVE, AL_FALSE);
       alSource3f(source, AL_POSITION, -(float)src_col, 0.0f, (float)src_row);
 
-      /* Pick the distance attenuation curve:
- *   - If the Blakod set a radius, fade linearly from full volume
- *     at 1 tile to silence at that radius (hard cutoff).  Used
- *     for looping emitters and one-shots with an explicit range.
- *   - Otherwise use OpenAL's "inverse" model: a 1/distance curve
- *     that halves every time distance doubles and never reaches
- *     zero, so distant one-shots stay faintly audible. */
+      // Pick the distance attenuation curve:
+      //   - If the caller set a radius, fade linearly from full volume
+      //     at 1 tile to silence at that radius (hard cutoff).  Used
+      //     for looping sources and one-shots with an explicit range.
+      //   - Otherwise use OpenAL's "inverse" model: a 1/distance curve
+      //     that halves every time distance doubles and never reaches
+      //     zero, so distant one-shots stay faintly audible.
       ALenum model;
       float refDist, maxDist;
       if ((flags & SF_LOOP) || radius > 0)
--- a/clientd3d/audio_openal.c
+++ b/clientd3d/audio_openal.c
@@ -188,8 +188,12 @@ bool AudioInit(HWND hWnd)
    ALfloat listenerOri[] = { 0.0f, 0.0f, -1.0f, 0.0f, 1.0f, 0.0f }; // forward, up
    alListenerfv(AL_ORIENTATION, listenerOri);
 
-   // Use linear distance model for predictable 3D audio falloff
+   /* Default distance model: linear clamped, used for SF_LOOP ambient
+    * emitters that should cut off at their kod-defined radius (fountains,
+    * firepits, etc.).  Per-source distance model is enabled below so that
+    * one-shot sounds can use an inverse curve instead (no hard cutoff). */
    alDistanceModel(AL_LINEAR_DISTANCE_CLAMPED);
+   alEnable(AL_SOURCE_DISTANCE_MODEL);
 
    g_initialized = true;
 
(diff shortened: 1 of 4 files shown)


clientd3d/audio_openal.c
```diff
@@ -188,8 +188,12 @@ bool AudioInit(HWND hWnd)
    ALfloat listenerOri[] = { 0.0f, 0.0f, -1.0f, 0.0f, 1.0f, 0.0f }; // forward, up
    alListenerfv(AL_ORIENTATION, listenerOri);
 
-   // Use linear distance model for predictable 3D audio falloff
+   /* Default distance model: linear clamped, used for SF_LOOP ambient
+    * emitters that should cut off at their kod-defined radius (fountains,
```
akirmse: This is a fairly general-purpose audio engine.  It should not need to know what the scripting language on the server is, or even that there is a server, much less the graphics associated with particular objects.  It's OK to talk about this in the documentation, but it doesn't belong in the code itself.  Most of the new comments have this issue.

clientd3d/audio_openal.c
```diff
@@ -1025,9 +1029,31 @@ bool SoundPlay(const char* filename, int volume, BYTE flags,
       alSourcei(source, AL_SOURCE_RELATIVE, AL_FALSE);
       alSource3f(source, AL_POSITION, -(float)src_col, 0.0f, (float)src_row);
 
-      // Distance attenuation: full volume within 1 tile, silent at radius
-      alSourcef(source, AL_REFERENCE_DISTANCE, 1.0f);
-      alSourcef(source, AL_MAX_DISTANCE, (float)radius);
+      /* Pick the distance attenuation curve:
+ *   - If the Blakod set a radius, fade linearly from full volume
```
akirmse: Placement of these asterisks is weird.  Can you switch to C++-style comments instead?

PR #1410, opened 2026-04-24
Reorganize D3DRender

neiltaco:
This PR is split off #1406 to help untangle the reorganization changes from the refactoring.

The PR strictly focuses on reorganization and cleanup for both `d3drender.c` and `d3drender.h` for better readability.  It also removes up unused variables and externs to help clarify which ones actually affect the renderer.  The refactoring of `d3drender.c` is planned for a separate PR after this one.

- Separated source file functions into two sections: Internal Implementation, and Public Functions
- Removed unused cache limits, global variables, and externs.
- Cleaned up redundant prototypes.  Grouped external functions at the bottom of the header file.
  - Removed `getWhiteLightTexture()`.  Updated `d3drender_materials.c` to use `D3DRenderLightsGetWhite()` directly.
  - Removed `D3DRenderMalloc()`.  Updated `d3drender.c` and `d3dcache.c` to instead use reinterpret_cast and malloc.
- Moved external function prototypes from `d3drender.h` to where they are defined.
  - Made `Vector3D` typedef  in `bsp.h` into a C++ struct to allow forward declarations in `draw3d.h` to resolve incomplete type error.
- Corrected whitespace by making space-based indentation into tab-based.  Also trimmed trailing whitespace.

### DIFF
--- a/clientd3d/d3drender.h
+++ b/clientd3d/d3drender.h
@@ -36,6 +36,17 @@ static constexpr float Z_RANGE = 200000.0f;
 // Standard ASCII table, minus the first 32 non-printable control characters.
 static constexpr int NUM_CHARS = 128 - 32;
 
+/////////////
+// Globals //
+/////////////
+inline IDirect3D9* gpD3D = nullptr;
+inline IDirect3DDevice9* gpD3DDevice = nullptr;
+
+inline int gNumVertices = 0;
+inline int gD3DEnabled = 0;
+inline int gScreenWidth = 0;
+inline int gScreenHeight = 0;
+
 ////////////////
 // Structures //
 ////////////////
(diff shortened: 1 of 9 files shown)


clientd3d/d3drender.h
```diff
@@ -115,64 +138,120 @@ inline bool D3DRender_InBounds(float coordinate, float range)
...
-void				*D3DRenderMalloc(unsigned int bytes);
-void				D3DRenderFontInit(font_3d *pFont, HFONT hFont);
+inline int D3DRenderIsEnabled(void)
+{
+	return gD3DEnabled;
+}
 
-LPDIRECT3DTEXTURE9  D3DRenderFramebufferTextureCreate(LPDIRECT3DTEXTURE9 pTex0, LPDIRECT3DTEXTURE9 pTex1, 
-	float width, float height);
+inline void *D3DRenderMalloc(unsigned int bytes)
```
akirmse: Maybe we should just get rid of this function.  It's silly.

clientd3d/d3drender.h
```diff
@@ -115,64 +138,120 @@ inline bool D3DRender_InBounds(float coordinate, float range)
...
+{
+	return malloc(bytes);
+}
 
-void SetZBias(LPDIRECT3DDEVICE9 device, int z_bias);
-int DistanceGet(int x, int y);
+inline void SetZBias(LPDIRECT3DDEVICE9 device, int z_bias) {
```
akirmse: Is there a reason you moved the implementation of all of these functions to the header?  Did profiling show that there's a performance concern?  Generally it's better to put implementation in the .c files to reduce dependencies and make things easier to read.

Plus this made you move more variables into the header, where they don't belong.

clientd3d/d3drender.h
```diff
@@ -188,4 +267,19 @@ void D3DRender_SetAlphaStage(DWORD stage, D3DTEXTUREOP alphaOp, DWORD arg1, DWOR
 void D3DRender_SetStreams(d3d_render_cache* pCache, int numStages);
 void D3DRender_ClearStreams(int numStages);
 
+////////////////////////////
+// External Dependencies  //
+////////////////////////////
+
+// Defined in object3d.c
```
akirmse: Shouldn't these be in object3d.h, then?

akirmse (commented):
I'm in favor of removing the unused variables, duplicate declarations, and whitespace.  But I'm not sure if the other changes are an improvement.

clientd3d/d3drender.h
neiltaco: Yeah you're definitely right.  I think I had the idea of clearing out the source file by moving small helper functions used by other source files into the header and also the variables that they need.  I seem to gloss over the fact that this also gives other source files access to these variables when they shouldn't.  I can move them back into `d3drender.c` and have the helper function prototypes added back into the header.

clientd3d/d3drender.h
neiltaco: It seems like a good opportunity to have these prototypes moved to their own header files.

I did have to update `Vector3D` from a typedef to a C++ struct, so that I could forward declare it within `draw3d.h`.  This solves an incomplete type error with `getSunVector()` since `Vector3D` isn't defined yet in that moment due to the include-order within `client.h`.

PR #1412, opened 2026-04-25
feature: sounds follow moving objects

adrienlaws:
## What
- Sounds tied to a game object now follow that object as it moves, instead of staying at the spot where the sound first played
- Depends on #1411 merging first
  - Without it, this PR turns a static "wrong position" bug into a more obvious "swing follows the victim around the room" bug
- Related to #1413 

## Why
- Reported by a player
> sounds created into the world space doesn't seem to follow the source and just stay where they are created
- The audio layer never knew which object owned a sound
  - `GamePlaySound` looked up the object's position once and forwarded only the snapshot row/col
  - The `source_obj` ID itself never reached the audio layer, so OpenAL had nothing to refresh from
- `UpdateLoopingSounds` already runs every frame but only updated the listener position, not the sources

## How
- Threaded an `ID source_obj` parameter through the audio API: `SoundPlay`, `PlayWaveFile`, `PlayWaveRsc`
  - Default `0` means "no tracking", so call sites without an object (the wading effect, the random pitch sample) are unchanged
- `GamePlaySound` now forwards `source_obj` instead of dropping it after the position snapshot
- Added a small registry inside `audio_openal.c`
  - `std::unordered_map<ALuint, ID>` keyed by OpenAL source ID, value is the owning game object
  - `SoundPlay` registers an entry whenever it starts a positional source with a non-zero `source_obj`
  - Non-positional sounds (UI clicks, ambient loops at placeholder coords) skip registration
- New `AudioUpdateTrackedSources`, called once per frame from `UpdateLoopingSounds`
  - Drops entries whose source has finished playing
  - Drops entries whose object can no longer be found by ID, but leaves the source playing at its last position so a server-side renumber (system save) does not silence in-flight sounds
  - Otherwise refreshes `AL_POSITION` from the object's current `motion.x/y`
- Cleared the registry from `SoundStopAll` and `Audio_StopSourcesForFilename` so room-transition cleanup stays consistent
- Updated `docs/audio.md` with a "Sources Attached to Moving Objects" section

## Notes
- Not an OpenAL regression: the Miles path had the same limitation, and the `BK_PLAY_WAVE` packet only carries position once
- One-shot combat sounds emitted by a moving battler benefit immediately and need no kod changes
- No monster currently emits a looping aura sound; all existing loops come from passive props (`fountain.kod`, `fountjet.kod`, `firepit.kod`). This PR is the prerequisite that would let a future kod change give monsters their own moving ambient sound
- Per-frame cost is trivial: a handful of active sources, sources self-clean once playback ends
- Builds toward Doppler shift for projectiles: per-frame `AL_POSITION` updates are step one, and adding `AL_VELOCITY` alongside in the same loop is the natural next step

### Before
- Player walks east, monster walks west
- The monster's sound plays from the monster's original tile until the loop ends
- A one-shot combat sound from a moving target stays frozen at wherever the target was when the sound started

### After
- The sound stays attached to the monster as it moves
- If the monster leaves the room or is destroyed mid-play, the in-flight sound finishes at its last position rather than being cut off, so server-side renumbering during a system save does not silence world audio

### DIFF
--- a/clientd3d/audio_openal.c
+++ b/clientd3d/audio_openal.c
@@ -75,6 +75,16 @@ static std::list<CacheNode> g_cacheList;
 static std::unordered_map<std::string, std::list<CacheNode>::iterator,
                           CaseInsensitiveHash, CaseInsensitiveEqual> g_cacheMap;
 
+/*
+ * Tracked source registry: links a playing OpenAL source to a game object
+ * so the source's position can be refreshed each frame as the object moves.
+ */
+struct TrackedSource {
+   ALuint source;     // OpenAL source currently playing
+   ID     object_id;  // Game object whose position the source should follow
+};
+static std::vector<TrackedSource> g_trackedSources;
+
 // Music streaming state
 static const int STREAM_NUM_BUFFERS = 4;
 static const int STREAM_BUFFER_SAMPLES = 4096;
@@ -948,9 +958,13 @@ static ALuint LoadAudioBuffer(const char* filename)
 
 /*
  * SoundPlay: Returns true if sound started. Supports OGG/WAV, 3D positioning, looping.
+ *   When source_obj is non-zero and the sound is positional, the source is
+ *   registered so subsequent frames refresh its position from the object's
+ *   current location.
  */
 bool SoundPlay(const char* filename, int volume, BYTE flags,
-               int src_row, int src_col, int radius, int max_vol)
+               int src_row, int src_col, int radius, int max_vol,
+               ID source_obj)
 {
    if (!g_initialized)
       return false;
@@ -1096,6 +1110,14 @@ bool SoundPlay(const char* filename, int volume, BYTE flags,
       return false;
    }
 
+   // Register the source for per-frame position updates if it is a positional
+   // sound emitted by a known game object.  Non-positional sounds (UI, ambient
+   // loops at placeholder coords) ignore source_obj.
+   if (isPositional && source_obj != 0)
+   {
+      g_trackedSources.push_back({source, source_obj});
+   }
+
    return true;
 }
 
@@ -1124,6 +1146,8 @@ void SoundStopAll(void)
    {
       alSourceStop(g_sources[i]);
    }
+
+   g_trackedSources.clear();
 }
 
 /*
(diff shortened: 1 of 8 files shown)


clientd3d/sound.c
```diff
@@ -169,4 +170,7 @@ void UpdateLoopingSounds(int px, int py, int angle)
 
 	// Update OpenAL listener position and orientation
 	AudioUpdateListener((float)px, 0.0f, (float)py, forwardX, 0.0f, forwardZ);
+
+	// Refresh positions of sources attached to moving game objects
+	Audio_UpdateTrackedSources();
```
akirmse: AudioUpdateListener (no underscore) but Audio_UpdateTrackedSources (underscore)?

clientd3d/audio_openal.c
```diff
@@ -1166,8 +1188,51 @@ void Audio_StopSourcesForFilename(const char* filename)
...
 }
 
+/*
+ * Audio_UpdateTrackedSources: Refresh OpenAL source positions for sounds
+ *   attached to moving game objects.  Drops entries whose source has stopped
+ *   or whose object is no longer in the room.
+ */
+void Audio_UpdateTrackedSources(void)
+{
+   if (!g_initialized || g_trackedSources.empty())
```
akirmse: It doesn't look like a check for empty() is necessary.

clientd3d/audio_openal.c
```diff
@@ -1166,8 +1188,51 @@ void Audio_StopSourcesForFilename(const char* filename)
       {
          alSourceStop(g_sources[i]);
          alSourcei(g_sources[i], AL_BUFFER, 0);
+
+         // Drop any tracked entries pointing at this source
+         auto end = std::remove_if(g_trackedSources.begin(), g_trackedSources.end(),
+            [src = g_sources[i]](const TrackedSource& t) { return t.source == src; });
+         g_trackedSources.erase(end, g_trackedSources.end());
```
akirmse: You could replace this with erase_if(), but if you change to an unordered_map this should be much simpler.

clientd3d/audio_openal.c
```diff
@@ -75,6 +75,14 @@ static std::list<CacheNode> g_cacheList;
 static std::unordered_map<std::string, std::list<CacheNode>::iterator,
                           CaseInsensitiveHash, CaseInsensitiveEqual> g_cacheMap;
 
+// Tracked source registry: links a playing OpenAL source to a game object
+// so the source's position can be refreshed each frame as the object moves.
+struct TrackedSource {
+   ALuint source;     // OpenAL source currently playing
+   ID     object_id;  // Game object whose position the source should follow
+};
+static std::vector<TrackedSource> g_trackedSources;
```
akirmse: It looks like this would make more sense as an unordered_map mapping AL source ID to object_id.

akirmse (commented):
One thing to watch out for is what happens during a system save, which renumbers all objects and resends them to the client.  Since most sounds are short, problems would probably be rare, but I wonder if there are some weird cases where sounds might get cut off if objects get renumbered in an unlucky way.

clientd3d/audio_openal.c
adrienlaws: Removed.

clientd3d/sound.c
adrienlaws: Fixed.

clientd3d/audio_openal.c
adrienlaws: Changed to unordered_map as suggested.

clientd3d/audio_openal.c
adrienlaws: Changed to unordered_map as suggested.

PR #1414, opened 2026-04-26
fix: increase max client simultaneous sounds

adrienlaws:
## What
- Raise `MAX_AUDIO_SOURCES` in `clientd3d/audio_openal.c` from `32` to `64`
- Update `docs/audio.md` with a Source Pool section describing the constant and its history

## Why
- Every active sound holds a source slot regardless of audible range, so a room with several looping ambient emitters plus active combat can approach the 32-source ceiling
- When the pool is full, `SoundPlay` returns false silently and the new sound is dropped, which players notice during busy fights
- The original 32 was a conservative starting point chosen during the OpenAL Soft conversion in #1293 (December 2025), replacing the legacy `wavemix.dll` engine that was hardcoded to 8 channels
    - This was too conservative on my part
- OpenAL Soft itself ships with `sources = 256` as the default in [alsoftrc.sample](https://github.com/kcat/openal-soft/blob/master/alsoftrc.sample), so 64 is well within comfortable limits on modern hardware

### Before
- Room with a firepit, a fountain, ambient music, several other sound emitting objects, and active combat could approach the 32-source ceiling
- New combat sounds drop silently with `SoundPlay: No available sources` in the debug log

### After
- The same room has comfortable headroom for combat audio
- Future ambient emitters (torches, braziers) can be added without immediately competing with combat

### DIFF
--- a/clientd3d/audio_openal.c
+++ b/clientd3d/audio_openal.c
@@ -28,7 +28,7 @@
 #include <stb_vorbis.c>
 
 // Configuration
-static const int MAX_AUDIO_SOURCES = 32;
+static const int MAX_AUDIO_SOURCES = 64;
 static const int MAX_CACHED_BUFFERS = 256;
 
 // Global state
--- a/docs/audio.md
+++ b/docs/audio.md
@@ -159,6 +159,17 @@ Sound effects are cached using an LRU (Least Recently Used) strategy:
 
 Music is NOT cached because tracks are large and typically don't repeat rapidly. Instead, music uses streaming playback (see below).
 
+## Source Pool
+
+OpenAL plays sound effects through a fixed pool of pre-allocated sources. The pool size is set by `MAX_AUDIO_SOURCES` in `clientd3d/audio_openal.c`.
+
+- Current size: 64 sources
+- Allocated once at `AudioInit` via `alGenSources(MAX_AUDIO_SOURCES, g_sources)`
+- Each `SoundPlay` call consumes one source while the sound is active, regardless of whether the listener is in audible range
+- When the pool is full, `SoundPlay` returns false silently and the new sound is dropped
+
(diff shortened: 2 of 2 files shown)


akirmse:
32 active sounds already sounds like a large number, so a few questions:

- What system resources (memory + CPU) are used per sound and would thus be doubled?
- Why is 64 the right number as opposed to, say, 1000?
- Are you actually seeing this, or is this speculative?
- Let's say you couldn't increase the number beyond 32.  What would you do, i.e. what are our other options?

adrienlaws:
Sorry about the delayed response, I thought about this a bit.

I don't think it's strictly necessary now but it was meant to prepare for possible future looping sounds consuming the pool.  Currently each SoundPlay call consumes one source while the sound is active, regardless of whether the listener is in audible range.  If a room had a lot of torches and we enabled a burning sound, even if you couldn't hear it due to the dropoff, those torches would consume slots.

For one shot sounds they'd probably cycle out of the 32 buffer quite quickly.

To answer your questions specifically:

> What system resources (memory + CPU) are used per sound and would thus be doubled?

Each OpenAL source is a small state struct, maybe a few hundred bytes. So 32 → 64 adds roughly 10 KB. I'm not sure how the mixing of sounds would affect the CPU as multiple sounds overlap.

> Why is 64 the right number as opposed to, say, 1000?

I figure 64 is still pretty conservative for OpenAL Soft which seems to default to 256.  It's mainly as a guard during heavy PvP if more sounds are added and/or if more ambient looping sounds are added to maps.

> Are you actually seeing this, or is this speculative?

Speculative and more as a forward guard against what could happen. So technically there is no problem now at all.

> Let's say you couldn't increase the number beyond 32. What would you do, i.e. what are our other options?

Thinking about this a bit - I think it's doing proper audio clamping on sounds that you can't hear anyway (like those torches in the example discussed previous).  That would probably be the best fix but would be its own PR.  Alternatively, design it to drop ambient sounds when the source slots start getting full.


This change is quite trivial so we could always do it later if needed.  I can close it if you prefer.

akirmse:
Every change involves risk.  I think we should hold off on this until/if there's a problem in the future.

adrienlaws:
Makes sense.  Closing.

PR #1416, opened 2026-04-27
Refactor D3DRender source file

neiltaco:
This PR is the second part of #1406 and focuses on refactoring `d3drender.c`.  The goal is to improve the source file's readability, debugging, and type safety.

Refactoring
- Removed unused variables and function parameters.
- Replaced C-style D3D macros with C++ function calls.
- Rect setup in `d3drender.c` is made into its own helper function `GetScreenRect()`.
- Fixed possible memory leak in `D3DRenderShutdown()` where `gFont.kerningPairs` only deleted if textures existed.
- Defined constants for triangle strip in Constants section in `d3drender.c`.
- Defined cache limits used in `D3DRenderInit()` at the top of the source file.
- Refactored font glyph loop in `D3DRenderFontInit()` to use `int i = 0` instead and have character offset derived from it.

Code Cleanup
- Replaced `NULL` with `nullptr`.
- Moved function variables declarations from start of function to at point-of-use.
  - Also moved vertex layout definitions from top of source file to point-of-use and made into static constexpr.
- Changed C-style type casting to use `static_cast`, `reinterpret_cast`, and `std::bit_cast` for type safety.

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -62,7 +62,7 @@ font_3d					gFont;
 RECT					gD3DRect;
 int						gNumObjects;
 
-// Incremented in d3dcache.c to track total DrawPrimitive calls per frame.
+// Tracks total DrawPrimitive calls per frame.
 int						gNumDPCalls;
 
 static PALETTEENTRY		gPalette[NUM_COLORS];
@@ -146,7 +146,7 @@ static void D3DRenderPoolInit(d3d_render_pool_new *pPool, int size, int packetSi
 
 	D3DRenderPoolReset(pPool, nullptr);
 
-	for (u_int i = 0; i < pPool->size; i++)
+	for (auto i = 0u; i < pPool->size; i++)
 	{
 		pPacket->size = packetSize;
 	}
@@ -190,8 +190,8 @@ static void D3DRenderChunkInit(d3d_render_chunk_new *pChunk)
 
 static void D3DRenderViewElementsDraw(d3d_render_pool_new *pPool)
 {
-	float screenW = static_cast<float>(gD3DRect.right - gD3DRect.left) / static_cast<float>(gScreenWidth);
-	float screenH = static_cast<float>(gD3DRect.bottom - gD3DRect.top) / static_cast<float>(gScreenHeight);
+	auto screenW = static_cast<float>(gD3DRect.right - gD3DRect.left) / static_cast<float>(gScreenWidth);
+	auto screenH = static_cast<float>(gD3DRect.bottom - gD3DRect.top) / static_cast<float>(gScreenHeight);
 
 	// Render view elements (such as the main viewport yellow ui corners).
 	int offset = (GetFocus() == hMain) ? 4 : 0;
@@ -230,7 +230,7 @@ static void D3DRenderViewElementsDraw(d3d_render_pool_new *pPool)
 			bottom = D3DRENDER_SCREEN_TO_CLIP_Y(gScreenHeight, gScreenHeight);
 		}
 
-		d3d_render_packet_new *pPacket = D3DRenderPacketNew(pPool);
+		auto *pPacket = D3DRenderPacketNew(pPool);
 		pPacket->pDib = nullptr;
 		pPacket->pTexture = gpViewElements[i + offset];
 		pPacket->xLat0 = 0;
@@ -238,7 +238,7 @@ static void D3DRenderViewElementsDraw(d3d_render_pool_new *pPool)
 		pPacket->effect = 0;
 		pPacket->size = pPool->packetSize;
 
-		d3d_render_chunk_new *pChunk = D3DRenderChunkNew(pPacket);
+		auto *pChunk = D3DRenderChunkNew(pPacket);
 		pChunk->flags = 0;
 		pChunk->numIndices = TRI_STRIP_INDICES;
 		pChunk->numVertices = TRI_STRIP_VERTICES;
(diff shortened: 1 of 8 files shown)


clientd3d/d3drender.h
```diff
@@ -36,6 +36,12 @@ static constexpr float Z_RANGE = 200000.0f;
 // Standard ASCII table, minus the first 32 non-printable control characters.
 static constexpr int NUM_CHARS = 128 - 32;
 
+// Geometry constants for rendering a quad as a triangular strip.
```
akirmse: Why are these in the header where they're globally visible?  Can they go in the .c file instead?

clientd3d/d3drender.c
```diff
@@ -129,160 +122,172 @@ void D3DRenderViewElementsDraw(d3d_render_pool_new *pPool);
...
 	else
-		pFont->texWidth = pFont->texHeight = 256;
+		pFont->texWidth = pFont->texHeight = FONT_TEXTURE_SIZE_SMALL;
 
-	IDirect3DDevice9_GetDeviceCaps(gpD3DDevice, &d3dCaps);
+	D3DCAPS9 d3dCaps = {};
+	gpD3DDevice->GetDeviceCaps(&d3dCaps);
 
-	if (pFont->texWidth > (long) d3dCaps.MaxTextureWidth)
+	if ( pFont->texWidth > static_cast<long>(d3dCaps.MaxTextureWidth) )
```
akirmse: Don't add spaces inside parentheses like this please.

clientd3d/d3drender.c
```diff
@@ -41,55 +48,41 @@ d3d_render_pool_new		gWallMaskPool;
 d3d_render_pool_new		gEffectPool;
 d3d_render_pool_new		gParticlePool;
 
-custom_xyz				playerOldPos;
-custom_xyz				playerDeltaPos;
+custom_xyz				playerOldPos = {0.0f, 0.0f, 0.0f};
```
akirmse: Are these initial values necessary?

clientd3d/d3drender.c
```diff
@@ -129,160 +122,172 @@ void D3DRenderViewElementsDraw(d3d_render_pool_new *pPool);
...
-	BYTE *pDstRow = (BYTE*)d3dlr.pBits;
+	BYTE *pDstRow = reinterpret_cast<BYTE*>(d3dlr.pBits);
 
-	for (y = 0; y < pFont->texHeight; y++)
+	// Convert a texture bitmask into a 16-bit pixel format.
+	for (int y = 0; y < pFont->texHeight; y++)
 	{
-		pDst16 = (WORD *)pDstRow;
-		for (x = 0; x < pFont->texWidth; x++)
+		WORD *pDst16 = reinterpret_cast<WORD*>(pDstRow);
```
akirmse: You can use "auto" for the type in a bunch of places like this.

clientd3d/d3drender.c
```diff
@@ -635,23 +610,15 @@ int D3DRenderIsEnabled(void)
 	return gD3DEnabled;
 }
 
-void SetZBias(LPDIRECT3DDEVICE9 device, int z_bias) {
+void SetZBias(int z_bias)
+{
    float bias = z_bias * -0.00001f;
-   IDirect3DDevice9_SetRenderState(device, D3DRS_DEPTHBIAS,
-                                   *((DWORD *) &bias));
+   gpD3DDevice->SetRenderState(D3DRS_DEPTHBIAS, std::bit_cast<DWORD>(bias));
```
akirmse, line 287: Interesting, I didn't know about bit_cast before.

clientd3d/d3drender.c
```diff
@@ -725,124 +694,130 @@ const Color(&getBasePalette())[NUM_COLORS]
...
+		gpD3DDevice->SetRenderState(D3DRS_FOGEND, std::bit_cast<DWORD>(end));
 	}
 
 	// create framebuffer textures
-	for (int i = 0; i <= 15; i++)
-		IDirect3DDevice9_CreateTexture(gpD3DDevice, gSmallTextureSize, gSmallTextureSize, 1,
-										D3DUSAGE_RENDERTARGET, D3DFMT_A8R8G8B8, D3DPOOL_DEFAULT,
-										&gpBackBufferTex[i], NULL);
+	for (int i = 0; i < MAX_RENDER_TARGET_POOL; i++)
+	gpD3DDevice->CreateTexture(gSmallTextureSize, gSmallTextureSize, 1,D3DUSAGE_RENDERTARGET, D3DFMT_A8R8G8B8,
```
akirmse: Indentation doesn't look right here.

clientd3d/d3drender.c
```diff
@@ -1202,74 +1165,58 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
...
 		}
 	}
+
 	if ((gFrame & 255) == 255)
 		debug(("number of vertices = %d\nnumber of dp calls = %d\n", gNumVertices, gNumDPCalls));
-
-	timeComplete = timeGetTime() - timeComplete;
-	timeOverall = timeGetTime() - timeOverall;
-
-	//debug(("overall = %d lightmaps = %d world = %d objects = %d skybox = %d num vertices = %d setup = %d completion = %d (%d, %d, %d)\n"
```
akirmse, line 1263: It's not safe to remove these "timeComplete", "timeOverall" etc. values.  They're used for debugging by uncommenting this debug print, which should also not be removed.

clientd3d/d3drender.h
neiltaco: I placed them in the header since I saw similar setup for triangle strips in other source files such as in `d3drender_skybox.c` and `d3drender_objects.c`, where the setup for indices, verticles, and primitives appear to be the same.  That said, I've only applied it in `d3drender.c`, so I think I can move those constants into the source file until they get used elsewhere.

clientd3d/d3drender.c
neiltaco: Ah!  Not necessary, so those are removed.

clientd3d/d3drender.c
neiltaco: Oh, I did miss that this is a for-loop running a single code line.  I added curly braces for clarification.

clientd3d/d3drender.c
neiltaco: Ah!  I was thinking the performance profiler would be enough for testing, though it seems this debug message is more specific on what needs to be read.

I might remove `timeLMaps` and `timeSkybox`, though.  It may have been used sometime during development, but they don't appear to be used here at all.  It doesn't seem they were used even since the initial upload to Meridian 59 to Github.

clientd3d/d3drender.c
```diff
@@ -48,10 +61,13 @@ font_3d					gFont;
 
 RECT					gD3DRect;
 int						gNumObjects;
+
+// Incremented in d3dcache.c to track total DrawPrimitive calls per frame.
```
akirmse: A comment here is good to describe what the variable means, but you shouldn't talk about what file it's incremented in.  That could change at any time, for example due to refactoring like this PR.

clientd3d/d3drender.c
```diff
@@ -129,240 +128,49 @@ void D3DRenderViewElementsDraw(d3d_render_pool_new *pPool);
...
+	auto pPacket = reinterpret_cast<d3d_render_packet_new*>(malloc(sizeof(d3d_render_packet_new) * size));
 	assert(pPacket);
 	pPool->renderPacketList = list_create(pPacket);
 	pPool->packetSize = packetSize;
 
-	D3DRenderPoolReset(pPool, NULL);
+	D3DRenderPoolReset(pPool, nullptr);
 
-	for (i = 0; i < pPool->size; i++)
+	for (u_int i = 0; i < pPool->size; i++)
```
akirmse: auto?

neiltaco: I can do that, and I'm reading up more on where auto should be used.  Based from what I'm seeing in `d3dcache.h`, I'm guessing this prevents problems if the `d3d_render_pool_new` struct ever gets updated to use `uint32_t` instead of `u_int`.

PR #1417, opened 2026-04-27
Restrict translucent wall post-pass to prevent, wall revealing, hidden monsters on map and look dialog

sicoderealm:
The D3D post-pass added in #1394 was including all objects with `ncones==0` in drawdata, not just those behind translucent walls. This caused objects behind opaque walls to appear as red dots on the minimap and in the Look dialog.

- Add `behind_translucent` flag to ObjectData, set during BSP walk when traversing the far side of a translucent wall separator
- Filter the post-pass to only include flagged objects
- Remove `r->visible = true` from the post-pass so map dots are only set by the normal BSP cone walk
- Also resolve unexpected exploration of the map in the hardware renderer

Shout out to @neiltaco for spotting and raising the issue

### DIFF
--- a/clientd3d/drawbsp.c
+++ b/clientd3d/drawbsp.c
@@ -97,7 +97,7 @@ static bool incremental_background = false;
 static int background_cones = MAX_ITEMS;
 
 // Depth counter: >0 when the BSP walk is traversing the far side of a
-// separator plane that contains at least one translucent wall.  Used as a
+// separator plane that contains at least one translucent wall. Used as a
 // cheap early-exit so we can skip the per-column overlap test below when no
 // translucent walls are anywhere on the current path.
 static int translucent_wall_depth = 0;
@@ -105,9 +105,9 @@ static int translucent_wall_depth = 0;
 // Per-screen-column ref count of how many translucent walls on the current
 // BSP path actually cover each column.  Objects in the far subtree are marked
 // behind_translucent only if their projected column range overlaps a column
-// with count > 0.  Without this, a separator plane that mixes translucent and
+// with count > 0. Without this, a separator plane that mixes translucent and
 // opaque walls would mark every object in the far subtree, including objects
-// entirely behind the opaque wall — those would then be re-added by the D3D
+// entirely behind the opaque wall - those would then be re-added by the D3D
 // post-pass and rely on the depth buffer to occlude them.
 static int translucent_col_count[MAXX];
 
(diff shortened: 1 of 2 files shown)


clientd3d/drawbsp.c
```diff
@@ -2817,10 +2830,27 @@ static void WalkBSPtree(BSPnode *tree)
       }
 
       // lastly, traverse farther side
-      if (side > 0)
-         WalkBSPtree(tree->u.internal.neg_side);
-      else
-         WalkBSPtree(tree->u.internal.pos_side);
+      // Check if any wall in this separator plane is translucent; if so,
+      // objects in the far subtree should be eligible for the D3D post-pass.
+      {
```
akirmse, line 2880: I guess this brace is here because this is in a switch statement?  This case has become way, way too big.  Instead you could handle the BSPleaftype and unknown cases with simple if checks near the top of the function, and then the rest of the function could handle this more complicated case.

clientd3d/drawbsp.c
```diff
@@ -2817,10 +2830,27 @@ static void WalkBSPtree(BSPnode *tree)
...
+         for (WallList wl = tree->u.internal.walls_in_plane; wl; wl = wl->next)
+         {
+            if (wl->translucency_level > WALL_TRANSLUCENCY_OPAQUE)
+            {
+               has_translucent = true;
+               break;
+            }
+         }
+         if (has_translucent)
+            translucent_wall_depth++;
```
akirmse: I'm not sure of the logic here.  Suppose there are two walls in this separator plane, with one translucent and one not.  Won't this treat both walls as translucent for the purposes of marking objects, meaning you could still show objects on the map that aren't visible?

clientd3d/drawbsp.c
sicoderealm: Done - the switch is gone. BSPleaftype and the unknown-type case are now handled by two early if returns at the top of WalkBSPtree, and the rest of the function bodyhandles BSPinternaltype at one indent level

clientd3d/drawbsp.c
sicoderealm: Good catch - fixed. We now ref-count per screen column instead of per separator plane. For each translucent wall in the plane, its projected column range is added to a translucent_col_count[MAXX] map around the far-subtree recursion. In WalkObjects, an object is marked behind_translucent only if its own projected column range actually overlaps a column with count > 0. So a plane with one translucent + one opaque wall now correctly leaves objects behind the opaque wall unmarked, and they won't be re-added by the D3D post-pass.

clientd3d/drawbsp.c
```diff
@@ -2755,78 +2792,126 @@ static void WalkBSPtree(BSPnode *tree)
...
-               return;
-         }
+   // lastly, traverse farther side.
+   // For each translucent wall in this separator plane, project it to screen
+   // columns and ref-count those columns in translucent_col_count. When
+   // WalkObjects marks far-side objects, it consults this map so that an
+   // object is flagged behind_translucent only if it actually overlaps a
+   // translucent wall in screen space — not merely because the plane
+   // contained one somewhere.
+   #define MAX_TRANSLUCENT_PER_PLANE 16
```
akirmse: Use constants instead of macros

PR #1418, opened 2026-04-28
fix: immediately update client sound volume based on sliders

adrienlaws:
## What
- `Sound` and `Ambient` volume sliders in the Options dialog now take effect on currently-playing sources, instead of waiting for the next login
- Unchecking `Steady Sounds` now stops looping sources that are already playing, instead of leaving them running until you change rooms

## Why

- The `Music` slider already works properly: when you change it, `ResetMusicVolume` re-applies gain to the live music source
- The `Sound` and `Ambient` sliders only updated `config` values.  `SoundPlay` reads them once when starting a sound and never looks at them again, so looping sources (fountain, firepit) kept playing at their original volume forever
  - Reproduction: stand near the East End fountain, drag `Ambient` to 0.  Fountain stays loud until you re-login
- The `Steady Sounds` checkbox only blocked new `SoundPlay` calls.  Existing loops kept playing when you unchecked it
- Players reasonably expect all three sliders and the checkbox to behave the same way
- This appears to predate the December 2025 OpenAL migration.

## How

The Music slider already works this way.  Same approach now applies to the Sound and Ambient sliders.
```mermaid
flowchart LR
    A[Player drags slider<br/>and clicks OK]
    B[Save new value to config]
    C{Which slider?}
    D[ResetMusicVolume:<br/>re-apply to music source]
    E[ResetSoundVolume:<br/>re-apply to every<br/>playing sound]
    F[Volume changes<br/>immediately]

    A --> B --> C
    C -->|Music| D --> F
    C -->|Sound or Ambient| E --> F

```

Re-applying a slider to an already-playing sound requires two pieces of information per source:
- its original loudness (the volume the game asked for, before any slider was applied)
- whether it is a loop or a one-shot (loops follow the Ambient slider, one-shots follow the Sound slider)

Both are recorded when the sound starts.  When the sliders change, `ResetSoundVolume` walks the source pool and recomputes each playing sound's volume as `original loudness * current slider`.

For the `Steady Sounds` checkbox: unchecking it now calls `SoundStopLooping` to stop any currently-playing loops.  This mirrors what already happens when the master `Sound` toggle is turned off (it calls `SoundStopAll`).

### Re-enable caveat

- Re-checking `Steady Sounds` after unchecking it does not restart the loops that were silenced.  The server only sends loop packets on room entry, so the loops come back when you next change rooms.  Asking the server to re-push on demand is a much larger change and is out of scope for this PR
- The `Atmospheric Sounds` checkbox needs no immediate action.  It only gates short one-shot periodic sounds, which finish playing on their own within a few seconds

### Naming

- `ResetSoundVolume` matches the existing `ResetMusicVolume` so the pattern is consistent

### Before
- Player drags `Ambient` slider to 0 to mute the fountain and clicks OK
- Fountain keeps playing at full volume
- Player has to log out and back in for the change to take effect

### After
- Player drags `Ambient` slider to 0 and clicks OK
- Fountain volume drops to 0 immediately
- Same flow as the existing `Music` slider

### DIFF
--- a/clientd3d/audio_openal.c
+++ b/clientd3d/audio_openal.c
@@ -39,8 +39,11 @@ static int g_numSources = 0;
 static bool g_initialized = false;
 
 // Per-source state
-static bool  g_sourceIsLoop[MAX_AUDIO_SOURCES];
-static float g_sourceBaseGain[MAX_AUDIO_SOURCES];
+struct SourceState {
+   float baseGain;  // pre-slider gain (max_vol / volume scaled to 0..1)
+   bool  isLoop;    // true if SF_LOOP, picks Ambient slider; else Sound slider
+};
+static SourceState g_sourceState[MAX_AUDIO_SOURCES];
 
 // LRU buffer cache: list ordered by recency (front = newest), map for O(1) lookup.
 struct CacheNode {
@@ -1069,8 +1072,8 @@ bool SoundPlay(const char* filename, int volume, BYTE flags,
       alSourcef(source, AL_ROLLOFF_FACTOR, 1.0f);
 
       float gain = (float)max_vol / (float)MAX_VOLUME;
-      g_sourceBaseGain[sourceIndex] = gain;
-      g_sourceIsLoop[sourceIndex] = (flags & SF_LOOP) != 0;
+      g_sourceState[sourceIndex].baseGain = gain;
+      g_sourceState[sourceIndex].isLoop = (flags & SF_LOOP) != 0;
       if (flags & SF_LOOP)
          gain *= (float)config.ambient_volume / 100.0f;
       else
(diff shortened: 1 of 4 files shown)


clientd3d/audio_openal.c
```diff
@@ -38,6 +38,10 @@ static ALuint g_sources[MAX_AUDIO_SOURCES];
 static int g_numSources = 0;
 static bool g_initialized = false;
 
+// Per-source state
+static bool  g_sourceIsLoop[MAX_AUDIO_SOURCES];
```
akirmse: Parallel arrays are usually better as arrays of structs.  It makes it easier to add new fields in the future.

adrienlaws: Changed to struct as suggested.

PR #1421, opened 2026-04-29
Remove advert tracking from login and char-pick

sicoderealm:
Cleanup: these advert stats aren't believed to be used anywhere - no reporting, no analysis, no consumers - so collecting them is just noise in the logs and dead code on the wire.
                                                                                                                                       
  - `LogUserData` no longer appends `ADVERTISE_FILE1/2` to the login log line
  - Removes `BP_AD_SELECTED` end-to-end: client no longer pings the server when an ad on the char-pick screen is clicked, server no longer logs ad clicks
  - Ads still display on char-pick and clicking still opens the URL - only the tracking is gone
  
This is also ground work / response to feedback from https://github.com/Meridian59/Meridian59/pull/1396

### DIFF
--- a/module/char/char.c
+++ b/module/char/char.c
@@ -40,7 +40,6 @@ client_message msg_table[] = {
 			     PARAM_BYTE, PARAM_BYTE,
 			     PARAM_INT_ARRAY, PARAM_ID_LIST, PARAM_ID_LIST, PARAM_END }, },
 { BP_SEND_CHARINFO,        { PARAM_END }, },
-{ BP_AD_SELECTED,          { PARAM_BYTE, PARAM_END }, },
 { 0,                       { PARAM_END }, },
 };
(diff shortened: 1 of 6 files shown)


module/char/char.c
```diff
@@ -40,7 +40,6 @@ client_message msg_table[] = {
 			     PARAM_BYTE, PARAM_BYTE,
 			     PARAM_INT_ARRAY, PARAM_ID_LIST, PARAM_ID_LIST, PARAM_END }, },
 { BP_SEND_CHARINFO,        { PARAM_END }, },
-{ BP_AD_SELECTED,          { PARAM_BYTE, PARAM_END }, },
```
akirmse, line 43: Might as well also remove the definition of BP_AD_SELECTED in proto.h.

PR #1422, opened 2026-04-30
Refactor d3drender_skybox

neiltaco:
This PR ports the `d3drender_skybox.c` refactoring from #1304 to clean up and refactor the skybox setup.  These changes should help make it easier to read and adjust how both legacy and hardware rendering skyboxes are paired together.  It also reorganizes and condenses parts of the source file for improved readability.  This PR also refactors the pixel packing loop to use 32-bit pixel packing via bit-shifting to streamline its performance and be easier to understand.

Refactoring
- Added `SKYBOX_ASSET_MAPPING` that pairs .bgf files with .bsf/.png skybox files in an `std::unordered_map` table.
- Added `gSkyboxCache` table that pairs .bsf/.png filenames to loaded texture pointers.  Skyboxes files are loaded in on demand.
- Replaced defined arrays of floats to arrays of structs to help condense them.
- Replaced D3D wrapper functions with direct calls.
- Modernized source file using `auto` and `static constexpr`.
- Refactored pixel-copying logic by replacing nested byte loop with libpng transform functions.
- Split PNG loading code into newly added `skybox_load.c`
- Skyboxes can support 32-bit PNG files.

Code Cleanup
- Reorganized source file.
- Removed unused `pitchHalf` and `bytePP` variables.
- Replaced type casting with C++ equivalent

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -709,8 +709,8 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	if (draw_sky) // Render the skybox first
 	{
-		D3DRenderSkyBox(params, angleHeading, anglePitch, view,
-			SkyboxRenderParams{g_pVertexDecl_PosColorTex1, gD3DDriverProfile, gWorldPool, gWorldCacheSystem});
+		SkyboxRenderParams skyboxRenderParams(g_pVertexDecl_PosColorTex1, gD3DDriverProfile, gWorldPool, gWorldCacheSystem);
+		D3DRenderSkyBox(params, angleHeading, anglePitch, view, skyboxRenderParams);
 	}
 
 	// Prepare our rendering parameters
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -709,8 +709,8 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	if (draw_sky) // Render the skybox first
 	{
-		SkyboxRenderParams skyboxRenderParams(g_pVertexDecl_PosColorTex1, gD3DDriverProfile, gWorldPool, gWorldCacheSystem);
-		D3DRenderSkyBox(params, angleHeading, anglePitch, view, skyboxRenderParams);
+		D3DRenderSkyBox(params, angleHeading, anglePitch, view,
+			SkyboxRenderParams{g_pVertexDecl_PosColorTex1, gD3DDriverProfile, gWorldPool, gWorldCacheSystem});
 	}
 
 	// Prepare our rendering parameters
--- a/clientd3d/d3drender_skybox.c
+++ b/clientd3d/d3drender_skybox.c
@@ -6,122 +6,87 @@
 //
 // Meridian is a registered trademark.
 #include "client.h"
-#include <unordered_map>
+#include <filesystem>
+#include <iterator>
 
 // Variables
 
-static const float SKYBOX_DIMENSIONS = 75000.0f;
-static const float SKYBOX_Y = 37000.0f;
+static constexpr float SKYBOX_DIMENSIONS = 75000.0f;
+static constexpr float SKYBOX_Y = 37000.0f;
+static constexpr int SKYBOX_SIDES = 6;
 
-static LPDIRECT3DTEXTURE9 gpSkyboxTextures[5][6];
+static constexpr int SKYBOX_QUAD_VERTICES = 4;
+static constexpr int SKYBOX_QUAD_INDICES = 4;
+static constexpr int SKYBOX_QUAD_PRIMITIVES = SKYBOX_QUAD_VERTICES - 2;
+
 static int gCurBackground;
 static ID tempBkgnd = 0;
 
-static float gSkyboxXYZ[] =
+// Defines vertices of the skybox.
+static constexpr custom_xyz gSkyboxXYZ[] =
 {
-	// back
-	SKYBOX_DIMENSIONS, SKYBOX_Y, -SKYBOX_DIMENSIONS,
-	SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS,
-	-SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS,
-	-SKYBOX_DIMENSIONS, SKYBOX_Y, -SKYBOX_DIMENSIONS,
+	// Back
+	{SKYBOX_DIMENSIONS, SKYBOX_Y, -SKYBOX_DIMENSIONS}, {SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS},
+	{-SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS}, {-SKYBOX_DIMENSIONS, SKYBOX_Y, -SKYBOX_DIMENSIONS},
 
-	// bottom
-	-SKYBOX_DIMENSIONS, -SKYBOX_Y, SKYBOX_DIMENSIONS,
-	-SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS,
-	SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS,
-	SKYBOX_DIMENSIONS, -SKYBOX_Y, SKYBOX_DIMENSIONS,
+	// Bottom
+	{-SKYBOX_DIMENSIONS, -SKYBOX_Y, SKYBOX_DIMENSIONS}, {-SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS},
+	{SKYBOX_DIMENSIONS, -SKYBOX_Y, -SKYBOX_DIMENSIONS}, {SKYBOX_DIMENSIONS, -SKYBOX_Y, SKYBOX_DIMENSIONS},
 
-	// front
-	-SKYBOX_DIMENSIONS, SKYBOX_Y, SKYBOX_DIMENSIONS,
-	-SKYBOX_DIMENSIONS, -SKYBOX_Y, SKYBOX_DIMENSIONS,
-	SKYBOX_DIMENSIONS, -SKYBOX_Y, SKYBOX_DIMENSIONS,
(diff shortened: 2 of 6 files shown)


clientd3d/d3drender.c
```diff
@@ -708,8 +708,8 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	if (draw_sky) // Render the skybox first
 	{
-		SkyboxRenderParams skyboxRenderParams(g_pVertexDecl_PosColorTex1, gD3DDriverProfile, gWorldPool, gWorldCacheSystem);
-		D3DRenderSkyBox(params, angleHeading, anglePitch, view, skyboxRenderParams);
+		D3DRenderSkyBox(params, angleHeading, anglePitch, view,
```
akirmse: This looks equivalent but less readable to me.  Maybe revert this file?

clientd3d/d3drender_skybox.c
```diff
@@ -6,430 +6,382 @@
...
+	// file by appending each picture's binary data into one file.
+	//
+	// The pictures for each cubemap face are stored in this order, similarly in SKYBOX_XYZ[]:
+	// Back -> Bottom -> Front -> Left -> Right -> Top
+	//
+	// The pixel-packing loop expects a 24-bit PNG file since the alpha channel isn't used.
+	// However, a 32-bit PNG can still be used even if it's larger in file size.
+
+	auto *pFile = fopen(pFilename, "rb");
+	if (pFile == nullptr)
```
akirmse: Seems like it would be good to print an error that includes the filename here.

clientd3d/d3drender_skybox.c
```diff
@@ -6,430 +6,382 @@
...
+		if (pInfoEnd == nullptr)
+		{
+			png_destroy_read_struct(&pPng, &pInfo, nullptr);
+			fclose(pFile);
+			return;
+		}
 
-	// Disable fog for the skybox
-	IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_FOGENABLE, FALSE);
+		if (setjmp(png_jmpbuf(pPng)))
```
akirmse: This PNG loading code is pretty gross.  Is there a cleaner way to do it?

Regardless, I think it would make sense to factor the file loading part into a separate file, as it simply loads a PNG file and isn't related to 3D rendering.

clientd3d/d3drender_skybox.c
```diff
@@ -6,430 +6,382 @@
...
+	gpD3DDevice->SetSamplerState(0, D3DSAMP_ADDRESSU, D3DTADDRESS_CLAMP);
+	gpD3DDevice->SetSamplerState(0, D3DSAMP_ADDRESSV, D3DTADDRESS_CLAMP);
 
-	for (i = 0; i < 6; i++)
+	for (int i = 0; i < SKYBOX_SIDES; i++)
 	{
-		pPacket = D3DRenderPacketFindMatch(pPool, gpSkyboxTextures[gCurBackground][i], NULL, 0, 0, 0);
-		if (NULL == pPacket)
+		auto *pPacket = D3DRenderPacketFindMatch(pPool, gpSkyboxTextures[gCurBackground][i], nullptr, 0, 0, 0);
+		if (pPacket == nullptr)
```
akirmse, line 173: BTW there is an actual reason to do (nullptr == foo) rather than (foo == nullptr).  If you accidentally forget an equals sign, (nullptr = foo) will give a compilation error, while (foo = nullptr) will silently do something you don't want.

neiltaco: Interesting!  I've read that missing the second `=` in an if-statement would raise compiler warnings in modern compilers.  And yet when I tested `pPacket = nullptr`, it compiled without raising any warnings.

I just want to make sure, do you want these changed back?  I think this new version is a bit more readable and I have double-checked if the if-statements are using `==`.  I can also change them back to the original arrangement if that extra safety is needed.

Meridian59:
No, you don't have to change these back, but maybe don't invert any more in
the future.

clientd3d/d3drender_skybox.c
neiltaco: I'm reading that Direct3D 9 actually does support cubemaps, so I can make this PR into a draft as I test this out.  Since this is a refactor, this might be an opportunity to clean up the code and also help simplify how skyboxes are added or modified.

neiltaco:
Earlier I tried out using cubemaps to replace the current six texture setup, though it may require several other changes and possibly a new struct for cubemap coordinates.  Since this PR is focused on refactoring, I think the cubemap change should be in a separate PR.

I've also mentioned wanting to split up the BSF files into six separate PNG files each so that it's easier to create and modify skyboxes, though this would dump 30 PNG files into the resource file.  I left this change out of this PR for now, but if requested I can add it into this PR or in a separate PR.

clientd3d/skybox_load.c
```diff
@@ -0,0 +1,102 @@
...
+	// Back -> Bottom -> Front -> Left -> Right -> Top
+	//
+	// The pixel-packing loop expects a 24-bit PNG file since the alpha channel isn't used.
+	// However, a 32-bit PNG can still be used even if it's larger in file size.
+
+	auto *pFile = fopen(pFilename, "rb");
+
+	if (pFile == nullptr)
+	{
+		debug(("Skybox Error: Found file but unable to open: %s\n", pFilename));
```
akirmse: "Found file"?  How do you know the file is there?

akirmse (commented):
Mostly fine, just one question about an error message.

clientd3d/skybox_load.c
neiltaco: Ah!  The file's existence is confirmed in `D3DRenderUpdateSkyBox()`, though it isn't clear within `LoadSkyboxFaces()`.  I updated the comments and error message.

clientd3d/d3drender_skybox.c
```diff
@@ -1,435 +1,302 @@
...
+	gpD3DDevice->SetSamplerState(0, D3DSAMP_ADDRESSV, D3DTADDRESS_WRAP);
 }
 
-/**
-* Set the current background texture for the skybox by background ID.
-* Returns true if the background was set successfully and false otherwise.
-*/
-bool D3DRenderBackgroundSet(ID background)
+// Checks if a skybox is already set up, and copies the pointers over if a match is found..
+// Used for hardware rendered skyboxes that is shared with both the mainland and Ko'catan island.
```
akirmse: I thought this was good to go but now I realize I don't understand this.  This function takes an integer.. I don't see any pointers, or places to copy from or to.  So at a minimum this needs to explain more of what's going on.  It's called "Check" but it copies?  Confused.

Also, why should the guts of the 3D renderer care how the game world is divided into landforms?

neiltaco: Yeah I think that function name is misleading when it's actually doing two jobs, so I can change the name.  Also I thought that `memcpy` would be good for having a skybox shared, but I didn't adjust `D3DRenderSkyBoxShutdown()` to account for this, so I'll have to rethink this.

Also good point, I can remove mention of the game world in the comments.

clientd3d/d3drender_skybox.c
```diff
@@ -1,435 +1,315 @@
...
-		{"2skyd.bgf", 3},
-		{"redsky.bgf", 4}
-	};
-
-	auto it = backgroundMap.find(filename);
-	if (it != backgroundMap.end())
-	{
-		gCurBackground = it->second;
-	}
+    // 'targetIndex' is passed in so this function only checks against indices that have already been loaded.
```
akirmse: This storage scheme seems pretty complicated and fragile.  There are two parallel arrays, SKYBOX_TABLE and gpSkyboxTextures that aren't obviously related by their names.  I think what you're trying to do here is to avoid loading the same image multiple times if it has the same name.

I think what you want is two lookup tables (std::unordered_map): 
1) Map of software renderer filename to hardware renderer filename.
2) Map of hardware renderer file name to its textures (vector of D3D pointers or something like that).

When you encounter a skybox resource name, you first get its corresponding hardware filename.  The you look that up in the second map to see if it's already loaded, and load it if not.  (Increase reference count if it is?  I think)

I think you would no longer need the NUM_SKYBOXES constant with this scheme.  Not sure I followed everything in the existing code, but it does look more complicated than necessary.

neiltaco: Ah, thanks for the advice!  I think I accidentally removed std::unordered_map from the original code when I tried to replace the two separate systems that paired .bgf/.bsf files with indices.  Now I realize it can be used to directly match .bgf files to .bsf/.png files, and without dealing with the newly introduced issue of loading skyboxes twice.

clientd3d/d3drender_skybox.c
```diff
@@ -1,435 +1,292 @@
...
-	// Disable fog for the skybox
-	IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_FOGENABLE, FALSE);
-
-	// Set texture stages for the skybox
-	D3DRender_SetColorStage(0, D3DTOP_SELECTARG1, D3DTA_TEXTURE, D3DTA_DIFFUSE);
-	D3DRender_SetAlphaStage(0, D3DTOP_SELECTARG1, D3DTA_TEXTURE, D3DTA_DIFFUSE);
-	D3DRender_SetColorStage(1, D3DTOP_DISABLE, 0, 0);
-	D3DRender_SetAlphaStage(1, D3DTOP_DISABLE, 0, 0);
+	// Otherwise, check if the file exists before attempting to load the PNG.
+	auto fullPath = std::filesystem::path("./resource") / hwFilename;
```
akirmse: Let's make the resource directory name into a constant.  Also I don't think you need the ./

clientd3d/d3drender_skybox.c
```diff
@@ -1,435 +1,292 @@
...
+static void D3DRenderSkyboxDraw(d3d_render_pool_new* pPool, int angleHeading, int anglePitch)
 {
-	int			i, j;
-	D3DMATRIX	rot, mat;
-
-	d3d_render_packet_new* pPacket;
-	d3d_render_chunk_new* pChunk;
-
-	IDirect3DDevice9_SetRenderState(gpD3DDevice, D3DRS_ALPHABLENDENABLE, FALSE);
+	if (gpActiveSkyboxTextures.empty())
```
akirmse: Is this check necessary?

neiltaco: It doesn't seem necessary, and it runs fine without it.

PR #1423, opened 2026-05-02
fix: allow easy two-handed item equips and dequips

adrienlaws:
## What
- equipping a hand item now auto-unequips enough held hand items to make room, instead of failing with "Your hands are too full to use that"
- covers cross-class collisions that the existing same-class swap code did not catch
  - one-handed plus two-handed (e.g. book then bow): the held item is unequipped to make room
  - any combination that would push the player past two hand slots (e.g. book plus shield then sword, or book plus shield then bow): held items are unequipped one at a time until the new item fits

## Why
- this is a longstanding bug that is very irritating in PvP where players must manually equip and unequip items (specifically bows and shields)
- when the total of currently-held hand slots plus the incoming item exceeds two, something has to give
  - same-class swaps already worked: `Weapon::ReqUseSomething` unequips itself for weapon-to-weapon, `Shield::ReqUseSomething` unequips itself for shield-to-shield
  - cross-class collisions did not work: nobody knew to unequip a book for a bow, a shield for a sword, etc., so `CheckPosition` later failed with "Your hands are too full"
- affected combinations include: bow vs book, bow vs shield, lute vs book, lute vs shield, bow vs lute, sword while holding book plus shield, and any future hand item that conflicts with the current load
- the old behavior forced the player to manually unequip before equipping, inconsistent with same-class swaps

## How
- added `Player::FreeHandSlotsFor(what)` in `player.kod`
  - sums `viUse_amount` across every item in `plUsing` that uses `ITEM_USE_HAND`
  - if the total plus the incoming item's `viUse_amount` exceeds `viHand_space`, unequips held hand items one at a time (over a snapshot of `plUsing`) until enough room is free or no more items will yield
  - data-driven from `viUse_type` and `viUse_amount`, so any future hand item participates without code changes
- called once from `TryUseItem` before the existing `for i in plUsing` loop, only when the incoming item uses `ITEM_USE_HAND`
- no changes needed to `Item::ReqUseSomething`, `SoldierShield`, or any other class
  - cursed weapons refuse `TryUnuseItem` via `WACursed::ItemReqUnuse`, so they stay equipped
  - soldier's shield refuses `TryUnuseItem` via `SoldierShield::ReqUnuse` (DM-only), so it stays equipped and bow/lute equips fall through to the existing back-sling path in `CheckPosition`
- the same-class swap overrides (`Weapon::ReqUseSomething`, `Shield::ReqUseSomething`) are untouched and run after the pre-pass as before

`TryUseItem(what)` in `player.kod` is the equip pipeline.  The new pre-pass runs once before the existing loop over `plUsing`, so all `TryUnuseItem` calls happen outside any iteration of `plUsing`.

```mermaid
flowchart TD
    A[Player tries to equip 'what'] --> B[TryUseItem what]
    B --> C{Player morphed or not holding what?}
    C -->|yes| Z1[Block equip]
    C -->|no| D{what is single-use?}
    D -->|yes| D2[Redirect to TryApplyItem]
    D -->|no| E{what broken or marked cant-use?}
    E -->|yes| Z2[Block equip]
    E -->|no| F[Send what ReqUse]
    F -->|FALSE - e.g. cursed weapon| Z3[Block equip]
    F -->|TRUE| P{what uses ITEM_USE_HAND?}
    P -->|yes| Q[FreeHandSlotsFor what]
    P -->|no| G
    Q --> G[Loop over each item in plUsing as 'i']
    G --> H[Send i ReqUseSomething with what]
    H -->|FALSE| Z4[Block equip]
    H -->|TRUE| I{more equipped items?}
    I -->|yes| G
    I -->|no| J[CheckPosition for slot space]
    J -->|fail| Z5["'Your hands are too full'"]
    J -->|ok| K[Add what to plUsing, equip succeeds]

    style Z1 fill:#ff0000,color:#fff
    style Z2 fill:#ff0000,color:#fff
    style Z3 fill:#ff0000,color:#fff
    style Z4 fill:#ff0000,color:#fff
    style Z5 fill:#ff0000,color:#fff
    style K fill:#2f9e44,color:#fff
    style Q fill:#2f9e44,color:#fff
```

### What `FreeHandSlotsFor` does

```mermaid
flowchart TD
    A["FreeHandSlotsFor what"] --> B[Snapshot every ITEM_USE_HAND item in plUsing into lHandItems and sum their viUse_amount into iHandUsed]
    B --> C{iHandUsed + what.viUse_amount within viHand_space?}
    C -->|yes| Y[Return - nothing to unequip]
    C -->|no| D[Walk lHandItems]
    D --> E[TryUnuseItem on next item]
    E -->|fail e.g. cursed weapon, soldier's shield| F{more items?}
    E -->|success| G[iHandUsed -= item.viUse_amount]
    G --> H{within viHand_space now?}
    H -->|yes| Y
    H -->|no| F
    F -->|yes| D
    F -->|no| Y

    style Y fill:#2f9e44,color:#fff
```

## Edge cases

- cursed weapons: `TryUnuseItem` returns FALSE because `WACursed::ItemReqUnuse` in `wacursed.kod` blocks unequip and shows the existing "you can't unuse this cursed weapon" message. The pre-pass moves on to the next held item; if none can yield, `CheckPosition` blocks the equip with "Your hands are too full"
- soldier's shield: `SoldierShield::ReqUnuse` returns FALSE for non-DMs, so `TryUnuseItem` fails. Bow and lute equips still succeed because the existing `CheckPosition` special case slings the shield on its back. Other hand items see no room and are blocked by `CheckPosition`
- non-hand items (necklace, ring, gauntlet, head, body, legs, quiver, finger): the `iUse_type & ITEM_USE_HAND` guard skips the pre-pass entirely. Behavior unchanged
- same-class swaps (weapon-to-weapon, shield-to-shield): the existing `Weapon::ReqUseSomething` and `Shield::ReqUseSomething` overrides still run in the existing `for i in plUsing` loop after the pre-pass, and behave as they did before
- multi-item overflow (e.g. book plus shield then sword): the pre-pass snapshots both held items into a local list and unequips them one at a time, stopping as soon as enough room is free
- list-mutation safety: `TryUnuseItem` mutates `plUsing`, but the pre-pass iterates a private snapshot list rather than `plUsing` directly, and runs before the existing `for i in plUsing` loop in `TryUseItem`

### Before

meridian_CQsOjtYsYc

### After

meridian_WOeoOX13Zi

### DIFF
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -518,34 +518,7 @@ messages:
    }
 
    ReqUseSomething(what = $)
-   "Returns TRUE if `self` permits the player to also equip `what`, FALSE to "
-   "block the equip.  Default: if `self` uses a hand slot and equipping `what` "
-   "would push the player's total hand use past two, `self` unequips itself to "
-   "make room and returns the result of that unequip."
    {
-      local i, iHandUsed;
-
-      if NOT (viUse_type & ITEM_USE_HAND)
-         OR NOT (Send(what,@GetItemUseType) & ITEM_USE_HAND)
-         OR FindListElem(Send(poOwner,@GetPlayerUsing),self) = 0
-      {
-         return TRUE;
-      }
-
-      iHandUsed = 0;
-      for i in Send(poOwner,@GetPlayerUsing)
-      {
-         if Send(i,@GetItemUseType) & ITEM_USE_HAND
-         {
-            iHandUsed = iHandUsed + Send(i,@GetItemUseAmount);
-         }
-      }
-
-      if iHandUsed + Send(what,@GetItemUseAmount) > 2
-      {
-         return Send(poOwner,@TryUnuseItem,#what=self);
-      }
-
       return TRUE;
    }
 
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -519,14 +519,29 @@ messages:
 
    ReqUseSomething(what = $)
    "Returns TRUE if `self` permits the player to also equip `what`, FALSE to "
-   "block the equip.  Default: if `self` and `what` both want hand slots and "
-   "their combined hand use exceeds two, `self` unequips itself to make room "
-   "and returns the result of that unequip."
+   "block the equip.  Default: if `self` uses a hand slot and equipping `what` "
+   "would push the player's total hand use past two, `self` unequips itself to "
+   "make room and returns the result of that unequip."
    {
-      if (viUse_type & ITEM_USE_HAND)
-         AND (Send(what,@GetItemUseType) & ITEM_USE_HAND)
-         AND (viUse_amount + Send(what,@GetItemUseAmount) > 2)
-         AND FindListElem(Send(poOwner,@GetPlayerUsing),self) <> 0
+      local i, iHandUsed;
+
+      if NOT (viUse_type & ITEM_USE_HAND)
+         OR NOT (Send(what,@GetItemUseType) & ITEM_USE_HAND)
+         OR FindListElem(Send(poOwner,@GetPlayerUsing),self) = 0
+      {
+         return TRUE;
+      }
+
+      iHandUsed = 0;
+      for i in Send(poOwner,@GetPlayerUsing)
+      {
+         if Send(i,@GetItemUseType) & ITEM_USE_HAND
+         {
+            iHandUsed = iHandUsed + Send(i,@GetItemUseAmount);
+         }
+      }
+
+      if iHandUsed + Send(what,@GetItemUseAmount) > 2
       {
          return Send(poOwner,@TryUnuseItem,#what=self);
       }
--- a/kod/object/item.kod
+++ b/kod/object/item.kod
@@ -518,7 +518,19 @@ messages:
    }
 
    ReqUseSomething(what = $)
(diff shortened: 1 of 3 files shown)


neiltaco:
The equipment switching definitely feels smooth when I tried it out so I think it's a great change!  I realize this also means that players cannot use lutes to safely pick up cursed weapons, so they may be a bit more careful around unidentified weapons if they don't have curse removal.

kod/object/item.kod
```diff
@@ -518,7 +518,34 @@ messages:
...
+      {
+         if Send(i,@GetItemUseType) & ITEM_USE_HAND
+         {
+            iHandUsed = iHandUsed + Send(i,@GetItemUseAmount);
+         }
+      }
+
+      if iHandUsed + Send(what,@GetItemUseAmount) > 2
+      {
+         return Send(poOwner,@TryUnuseItem,#what=self);
```
akirmse: This looks a little dangerous.  Presumably ReqUseSomething is being called inside a loop over plUsing, which is then calling TryUnuseItem, which then modifies plUsing, which is the very list that's being iterated over.  You have to be very careful when modifying lists that are being iterated over.  Let me know if the danger isn't obvious or if you can't find a way around it.

PR #1424, opened 2026-05-02
fix: use regular helm instead of MSH for cavern shrine puzzle

adrienlaws:
## What
- small change to swap `&Helm` for `&SimpleHelm` in the KA0 "A dark, humid cavern" shrine puzzle item-vent list

## Why
- the KA0 shrine puzzle asks players to drop a randomly chosen item from a list of items
- `&Helm` is the "magic spirit helmet"
  - it is extremely rare and expensive
- the rest of the list contains common items (mace, rose, mushroom, jewel, etc.)
  - `&Helm` is out of place and was almost certainly meant to be `&SimpleHelm` (the ordinary "helm")
- this issue was mentioned by players on the official Discord server

### DIFF
--- a/kod/object/active/holder/room/monsroom/kcforest/ka0.kod
+++ b/kod/object/active/holder/room/monsroom/kcforest/ka0.kod
@@ -89,7 +89,7 @@ messages:
                        &ForgetPotion,
                        &JewelofFroz,
                        &Gauntlet,
-                       &Helm,
+                       &SimpleHelm,
                        &IvyCirclet,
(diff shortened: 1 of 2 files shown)


akirmse:
Since this is set up in Constructed, the room needs to be deleted and rebuilt before this will take effect.  Any hand-created items in the room will be destroyed.  If that's OK, add a call to the appropriate function (RecreateOneRoom in System?  Something like that) in the post-update function near the bottom of System that we'll call after the next update.

PR #1425, opened 2026-05-02
feature: add client theme infrastructure with starter dark theme (theming part 1)

adrienlaws:
## What

- Adds theme switching capability to the client
- Ships a starter dark theme alongside the existing default theme
- Adds a `Theme` setting to `meridian.ini` and a preferences dropdown to pick one
- Adds `docs/themes.md` describing the architecture
- Related to https://github.com/Meridian59/Meridian59/pull/1397
  - Splitting this enormous closed PR into smaller ones

## Why

- Sets up the framework so future visual work can target a named theme instead of editing the one global palette
- Visuals get iterated on one surface at a time in follow-up PRs that can be reviewed and reverted independently
- The starter dark theme is intentionally narrow.  It recolors the main background and the chat edit box and leaves most surfaces using the default look so this PR stays small

## How

- Config: `theme` field on the `Config` struct typed as the scoped enum `Theme` (`Theme::Default`, `Theme::Dark`).  `INITheme` key in the `[Interface]` section.  `LoadSettings` calls `ConfigLoad` before `ColorsCreate` so the theme is known when colors load
- Color tables: `colorinfo_default[]` and `colorinfo_dark[]` are parallel tables of `MAXCOLORS` entries indexed by the `COLOR_*` enum.  `ColorSectionForTheme` and `ColorDefaultsForTheme` pick the right pair at runtime based on `config.theme`
- Color version: `THEME_COLOR_VERSION` is stamped into each theme's INI section by `ColorsSave`.  On load, if the saved version doesn't match the compiled version, saved colors are discarded and compiled defaults are used.  This keeps stale per-user tweaks from masking later palette improvements
- Palette snap: `GetColor`/`SetColor` apply `MAKEPALETTERGB`/`GetNearestPaletteColor` only for `Theme::Default`.  Other themes round-trip raw 24-bit RGB so dark colors are not collapsed against the indexed game palette
- Dialog scoping: `DialogCtlColor` returns stock white/black brushes for any non-default theme, so dialog controls never inherit theme colors.  Only the main window and its chat edit box pick up theme colors via `MainCtlColor`
- Server message colors: `srvrstr.c` has a parallel `code_table_dark[]` for the `^r ^g ^b ...` color codes, picked at runtime by `ColorCodeTableForTheme`
- `ThemeApply`: tears down and recreates the color table, refires `EVENT_COLORCHANGED`, and invalidates the main window so a theme swap takes effect without a restart
- Settings dialog: `IDC_THEME` combo box added to all three localized `IDD_SETTINGS` dialogs (English, German, Portuguese).  `IDS_THEME_DEFAULT` and `IDS_THEME_DARK` strings are localized

> [!NOTE]
>  I can remove the dropdown theme selector if it is preferred that the dark theme is finalized with all updates before being used.  It does work, albeit without the complete dark theme implementation.

## Diagram

```mermaid
flowchart TD
    INI[meridian.ini<br/>Theme=N]:::file --> CL[ConfigLoad]:::fn
    CL --> CFG[config.theme]:::data
    UI[Settings IDC_THEME]:::ui --> TA[ThemeApply]:::fn
    TA --> CC

    CFG --> CC[ColorsCreate]:::fn
    SEC[Colors / ColorsDark<br/>INI sections]:::file --> CC
    CC --> COLORS[colors array]:::data

    COLORS --> GC[GetColor / GetBrush]:::fn
    GC --> MCC[MainCtlColor]:::fn
    GC --> DCC[DialogCtlColor]:::fn
    GC --> SRV[DisplayMessage<br/>code table]:::fn
    TA --> REP[Repaint main window]:::fn

    classDef file fill:#1f3a5f,color:#fff
    classDef fn fill:#2f9e44,color:#fff
    classDef data fill:#5c5470,color:#fff
    classDef ui fill:#7048e8,color:#fff
```

## Scope

Example follow-up PRs this enables:

- Replacement bitmaps that don't read well on a dark background
- DWM dark title bar via `DwmSetWindowAttribute`
- Larger individually-colored health/mana/vigor stat bars, recolored for dark theme
- Replacement health/mana/vigor icons that read on a dark background
- A button-shading helper so existing buttons can adopt theme colors

None of those are in this PR.  The dark theme today will look mostly like the default theme with a black main background and a dark chat edit window, but it be a starting point to iterate on.

## Example
meridian_voTO8wXe2h

### DIFF
--- a/clientd3d/color.c
+++ b/clientd3d/color.c
@@ -28,9 +28,10 @@ static COLORREF CustColors[16];  /* User's custom colors for Choose Color dialog
 // Theme color tables.  See docs/themes.md for details.
 // Each theme has a parallel table of MAXCOLORS "R,G,B" strings indexed
 // by the COLOR_* enum from color.h.
+static constexpr size_t COLOR_STR_LEN = 15;
 
 // Default theme.
-static char colorinfo_default[][15] = {
+static char colorinfo_default[][COLOR_STR_LEN] = {
 	{ "0,0,0"},         /* COLOR_BGD */
 	{ "255,255,255"},   /* COLOR_FGD */
 	{ "255,255,255"},   /* COLOR_LISTSELBGD */
@@ -67,7 +68,7 @@ static char colorinfo_default[][15] = {
 };
 
 // Dark theme.
-static char colorinfo_dark[][15] = {
+static char colorinfo_dark[][COLOR_STR_LEN] = {
 	{ "0,0,0"},         /* COLOR_BGD */
 	{ "212,212,212"},   /* COLOR_FGD */
 	{ "255,255,255"},   /* COLOR_LISTSELBGD - matches default for dialog list highlights */
@@ -108,13 +109,13 @@ static char INIColorVersion[]       = "ColorVersion";
 static const int THEME_COLOR_VERSION = 5;
 
 // Returns the INI section name for the active theme.
-static const char *ColorSectionForTheme(Theme theme)
+static char *ColorSectionForTheme(Theme theme)
 {
 	return (theme == Theme::Default) ? color_section_default : color_section_dark;
 }
 
 // Returns the default color table for the active theme.
-static char (*ColorDefaultsForTheme(Theme theme))[15]
+static char (*ColorDefaultsForTheme(Theme theme))[COLOR_STR_LEN]
 {
 	return (theme == Theme::Default) ? colorinfo_default : colorinfo_dark;
 }
@@ -149,7 +150,7 @@ void ColorsCreate(bool use_defaults)
 	bool success;
 
 	const char *section = ColorSectionForTheme(config.theme);
-	char (*defaults)[15] = ColorDefaultsForTheme(config.theme);
+	char (*defaults)[COLOR_STR_LEN] = ColorDefaultsForTheme(config.theme);
 
 	// Discard saved colors when the version doesn't match.
 	if (!use_defaults)
@@ -271,7 +272,7 @@ void ColorsSave(void)
 {
 	int i;
 	char str[100], name[10];
-	const char *section = ColorSectionForTheme(config.theme);
+	char *section = ColorSectionForTheme(config.theme);
 
 	for (i=0; i < MAXCOLORS; i++)
 	{
@@ -283,8 +284,7 @@ void ColorsSave(void)
 	}
 
 	// Stamp the version so future loads can detect stale saved colors.
-	WritePrivateProfileString(section, INIColorVersion,
-		std::to_string(THEME_COLOR_VERSION).c_str(), ini_file);
+	WriteConfigInt(section, INIColorVersion, THEME_COLOR_VERSION, ini_file);
 }
 /************************************************************************/
 void ColorsRestoreDefaults(void)
@@ -438,9 +438,9 @@ HBRUSH MainCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type)
 
 	case CTLCOLOR_SCROLLBAR:
 		// Default paints scrollbars black; other themes use the system look.
-		if (config.theme != Theme::Default)
-			return (HBRUSH) FALSE;
-		return (HBRUSH) GetStockObject( BLACK_BRUSH );
+		if (config.theme == Theme::Default)
+			return (HBRUSH) GetStockObject( BLACK_BRUSH );
+		return (HBRUSH) FALSE;
 
 	default:
 		SelectPalette(hdc, hPal, FALSE);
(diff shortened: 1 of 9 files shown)


clientd3d/color.c
```diff
@@ -199,15 +271,20 @@ void ColorsSave(void)
...
 		snprintf(str, sizeof(str), "%d,%d,%d",
 			GetRValue(colors[i]), GetGValue(colors[i]), GetBValue(colors[i]));
 		snprintf(name, sizeof(name), "Color%d", i);
 
-		WritePrivateProfileString(color_section, name, str, ini_file);
+		WritePrivateProfileString(section, name, str, ini_file);
 	}
+
+	// Stamp the version so future loads can detect stale saved colors.
+	snprintf(str, sizeof(str), "%d", THEME_COLOR_VERSION);
```
akirmse: Use std::to_string() instead of this.

clientd3d/config.h
```diff
@@ -24,6 +24,12 @@
 
 static const int CONFIG_MAX_TEXT_ZOOM_LIMIT = 100;  // Max value of map text zoom limit
 
+// UI themes
+enum {
```
akirmse: For new enums I think we should make them strongly typed.  Read about c++ "scoped enums".

```
enum class Theme : int {
...
```

This prevents accidentally using them as ints where you don't mean to, and it's more self-documenting.

clientd3d/color.c
adrienlaws: Fixed - changed as requested.

clientd3d/config.h
adrienlaws: Fixed - converted to scoped enums. Very interesting.

clientd3d/color.c
```diff
@@ -25,9 +25,12 @@ static HBRUSH hDefaultBrush;
...
 
-/* Default color information structures */
-// default colors changed for grey background in new client - mistery
-static char colorinfo[][15] = {
+// Theme color tables.  See docs/themes.md for details.
+// Each theme has a parallel table of MAXCOLORS "R,G,B" strings indexed
+// by the COLOR_* enum from color.h.
+
+// Default theme.
+static char colorinfo_default[][15] = {
```
akirmse: I see this number 15 at least four times.. seems like there should be a constant for it.

clientd3d/color.c
```diff
@@ -199,15 +271,20 @@ void ColorsSave(void)
...
 		snprintf(str, sizeof(str), "%d,%d,%d",
 			GetRValue(colors[i]), GetGValue(colors[i]), GetBValue(colors[i]));
 		snprintf(name, sizeof(name), "Color%d", i);
 
-		WritePrivateProfileString(color_section, name, str, ini_file);
+		WritePrivateProfileString(section, name, str, ini_file);
 	}
+
+	// Stamp the version so future loads can detect stale saved colors.
+	WritePrivateProfileString(section, INIColorVersion,
```
akirmse: Whoops, I should have noticed this before.  We have a function WriteConfigInt that you should use instead.

clientd3d/color.c
```diff
@@ -359,9 +436,11 @@ HBRUSH MainCtlColor(HWND hwnd, HDC hdc, HWND hwndChild, int type)
 		SetBkColor(hdc, GetColor(COLOR_MAINEDITBGD));
 		return GetBrush(COLOR_MAINEDITBGD);
 
-	case CTLCOLOR_SCROLLBAR:  /* Don't color scrollbars */
-		//return (HBRUSH) FALSE;
-		return (HBRUSH) GetStockObject( BLACK_BRUSH );		//	xxx
+	case CTLCOLOR_SCROLLBAR:
+		// Default paints scrollbars black; other themes use the system look.
+		if (config.theme != Theme::Default)
```
akirmse: Change != to = and reverse the clauses

docs/themes.md
```diff
@@ -0,0 +1,131 @@
...
+## Status
+
+This is the initial theming infrastructure.  Two themes ship today:
+
+- `Theme::Default` (0): the original Meridian 59 palette.  Unchanged
+  visual behavior.
+- `Theme::Dark` (1): a starter dark theme that recolors only the main
+  background and the chat edit box.  Dialog list backgrounds, stat
+  bars, bitmaps, and most other UI surfaces still use the default
+  look.  Those land in follow-up work, one surface at a time.
```
akirmse: FYI I think you may regret this level of detail in the documentation.  It means you're going to be constantly having to keep it up to date with every change.

adrienlaws: Simplified the documentation as suggested. I kept my own detailed notes outside of the codebase for my own reference.

clientd3d/color.c
adrienlaws: Fixed

clientd3d/color.c
adrienlaws: Fixed - created `COLOR_STR_LEN` instead of using the number directly.

clientd3d/color.c
adrienlaws: Fixed - changed to use `WriteConfigInt` as suggested.

PR #1361, opened 2026-01-21
feat: allow dropping multiple inventory items in one network message

adrienlaws:
## What

- Changes `BP_REQ_DROP` from accepting a single object to accepting an object list, allowing players to drop multiple inventory items in one network message.
- Also fixes the issue where players could not drop more than 5 items at a time.

## Why

- Previously, dropping multiple items sent one `BP_REQ_DROP` message per item.
- The server's spam throttle (5 messages/second) would reject messages beyond this limit.
- Items silently failed to drop when selecting many items at once.

## How

- Modified the existing `BP_REQ_DROP` protocol command to accept `PARAM_OBJECT_LIST` instead of `PARAM_OBJECT`.
- Added `UserDropItems()` as a thin iteration wrapper that calls the existing `UserDrop()` for each item.

### Two Functions Work Together to Drop User Items

```
UserDropItems()  -->  Iterates the list, extracts amounts for NumberItems
       |
       v
   UserDrop()    -->  Contains all drop logic
```

### Flow Diagram

```mermaid
flowchart LR
    subgraph BEFORE["Before (per-item messages)"]
        direction TB
        C1[Client selects 10 items] --> M1[BP_REQ_DROP item1]
        C1 --> M2[BP_REQ_DROP item2]
        C1 --> M3[BP_REQ_DROP item3]
        C1 --> M4[...]
        C1 --> M10[BP_REQ_DROP item10]
        M1 --> S1[Server processes]
        M2 --> S1
        M3 --> S1
        M4 --> S1
        M5[items 6-10] --> THROTTLE[THROTTLED - 5/sec limit]
        THROTTLE --> LOST[Items not dropped]
    end
```

```mermaid
flowchart LR
    subgraph AFTER["After (single list message)"]
        direction TB
        C2[Client selects 10 items] --> M11["BP_REQ_DROP [item1..item10]"]
        M11 --> UDI[UserDropItems]
        UDI --> UD1[UserDrop item1]
        UDI --> UD2[UserDrop item2]
        UDI --> UD3[...]
        UDI --> UD10[UserDrop item10]
        UD1 --> OK1[Dropped]
        UD2 --> OK2[Dropped]
        UD3 --> OK3[...]
        UD10 --> OK10[Dropped]
    end
```

### Data Flow for NumberItems

NumberItems (stackable items like shillings, mushrooms) include amounts.  The wire format only sends amounts for NumberItems:

```mermaid
flowchart TB
    subgraph CLIENT["Client Message"]
        direction LR
[...]

PR #1427, opened 2026-05-07
feature: add inventory hotkey groups

adrienlaws:
## What
- adds inventory hotkey groups for quick item toggles
  - `Control + number` to assign a control group 0-9
  - `Control + x` to clear all control groups
  - Hit `number` when 3D window is selected to activate / deactivate selected item groups
- adds one batched client packet for group toggle so one key press sends one request (like #1361 did)
- keeps group bindings stable through inventory redraws in the same client session
- blocks wand-type items from group toggle
- consumes single-use items in inventory display order (top-left first)
> [!Note]
> Does retain groups through system saves.  Does not if the connection is reset (restarting the client, for example).

## Why
- swapping gear one item at a time is slow during normal play
- sending one packet for a group avoids per-item packet drops from the spam throttle
- wand usage from grouped toggles creates balance risk and is blocked in this change

## How
- client hotkeys
  - `InventoryHotkeyAssign` stores group bindings on inventory items
  - `InventoryHotkeyTrigger` builds one object list and sends `BP_REQ_TOGGLE_GROUP`
  - number-object ids are normalized before send so ammo and other number items resolve correctly on server
  - the client builds the wire list in reverse inventory order so that after the server's parse-time list reversal, top-left items end up at the head of the Blakod list and are processed first
    - this means for example if you have scrolls in a control group and you toggle the group, the first scroll in the group going from top left will be consumed first
- protocol and routing
  - added `BP_REQ_TOGGLE_GROUP` in shared protocol headers
  - added client parameter table entry in `protocol.c`
  - added server parse table entry in `sprocket.c`
- server behavior
  - `UserToggleItemGroup` in `user.kod` validates list input and returns early on empty
  - `UserToggleItemGroup` ignores duplicate entries and wand-type items
  - `UserToggleItemGroup` chooses one target state for the group and applies it item by item

### Client and Server Flow
```mermaid
flowchart TD
    A[Hotkey Press in Client] --> B[InventoryHotkeyTrigger]
    B --> C[BP_REQ_TOGGLE_GROUP]
    C --> D[Server ParseClientSendBlakod]
    D --> E[user.kod UserToggleItemGroup]
    E --> F{items in use}
    F -->|Yes| G[Unuse eligible items]
    F -->|No| H[Use eligible items]

    classDef step fill:#1f6feb,color:#fff
    classDef decision fill:#b08900,color:#fff
    classDef safe fill:#2f9e44,color:#fff

    class A,B,C,D,E step
    class F decision
    class G,H safe
```

### Gameplay Example
- open inventory, hover over an item, press `Ctrl+1` to assign it to group 1
- assign up to 5 items to the same group number (groups are capped at 5 items)
- press `Ctrl+X` in the inventory window to clear all group assignments
- press `1` from the 3D view or inventory to toggle the whole group at once
- press `1` again to toggle the group back to the previous state

### Recorded Example
meridian_lo6Ne4rZP4

- An example showing how a user can create and switch groups for melee , ranged, and caster items

### DIFF
--- a/blakserv/sprocket.c
+++ b/blakserv/sprocket.c
@@ -36,6 +36,7 @@ client_def_table_type client_def_table[] =
 	{ BP_REQ_GET,              { {4, TAG_OBJECT}, {0, DONE_PARM} } },
 	{ BP_REQ_GET_FROM_CONTAINER,	{ {4, TAG_OBJECT}, {0, DONE_PARM} } },
 	{ BP_REQ_DROP,             { {2, LIST_OBJ_PARM}, {0, DONE_PARM} } },
+	{ BP_REQ_TOGGLE_GROUP,     { {2, LIST_OBJ_PARM}, {0, DONE_PARM} } },
 	{ BP_REQ_INVENTORY_MOVE,   { {2, TAG_INT}, {2, TAG_INT}, {0, DONE_PARM} } },
 	{ BP_REQ_PUT,              { {4, TAG_OBJECT}, {4, TAG_OBJECT}, {0, DONE_PARM} } },
 	{ BP_REQ_LOOK,             { {4, TAG_OBJECT}, {0, DONE_PARM} } },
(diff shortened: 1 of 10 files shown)


akirmse:
I don't think this approach works.  Object IDs are not stable; they can change at each garbage collection.

There's also a question whether macro capabilities like this are healthy for the game... arguments on both sides.

adrienlaws:
Thanks for the feedback.  I tried to do the mapping by client side inventory slot but this could still break and annoy players.

In terms of gameplay I guess it could be more controversial than I first thought.

I'll save the branch in case it could be referenced for something else later.  Closing.

PR #1430, opened 2026-05-10
feature: add dark theme resources and theme helper functions (theming part 2)

adrienlaws:
## What
- Follow-on to #1425.  Four changes:
  - Adds dark theme BMPs (main background, inventory texture, sidebar frame pieces)
  - Adds per-module bitmap resolvers and capability flags so call sites stay theme-blind
  - Smooths mid-session theme switching
  - Documents the architecture in `docs/themes.md`

## Why
- The dark theme infrastructure was implemented in #1425 but no dark resources existed yet, so the main window background, inventory texture, and frame ornaments still used the default-theme tiles
- Without helpers, every bitmap-loading site would need its own theme branch and every capability check would hard-code a theme name
- A mid-session theme switch only repainted the palette, not the cached bitmaps, so the user had to log out and back in to see theme changes

## How
- Resources: new `_DARK` BMPs loaded through per-module resolvers (`MainThemeResourceId` in `clientd3d/color.c` for client-owned IDs, `InterfaceThemeResourceId` in `module/merintr/theme.c` for merintr-owned IDs).  Each switches over `ThemeCurrent()`, a new accessor that returns the active `Theme` value and replaces `ThemeIsDark()`
- Capability flags: `ThemeSidebarUsesInventoryFill` and `ThemeSkipStatsAreaFrame` are switches over `ThemeCurrent()`. 
- Sidebar fill: portrait, enchantment buttons, and stat icons paint on the inventory texture in dark theme so they match the surrounding sidebar.  `OffscreenSidebarBackground` in `drawint.c` consolidates the per-element fill that was duplicated in `userarea.c`, `statmain.c`, and `enchant.c`.  `InterfaceDrawSidebarBackground` paints the whole sidebar in one pass
- Color-change sentinel: named `COLOR_ID_ALL` in `color.h`.  `InventoryChangeColor(WORD color_id)` reloads themed bitmaps when it sees the sentinel
- Mid-session switch: `ThemeApply` reloads the main background and forces a full repaint with `RedrawWindow(RDW_ALLCHILDREN)`; `EVENT_COLORCHANGED` with `COLOR_ID_ALL` propagates to merintr so the inventory texture and frame elements refresh
- `docs/themes.md` covers color theming, bitmap surfaces, sidebar fill, per-module resolvers, and a list of the main requirements for adding a new theme

## Example

meridian_BMydZtuskH
Before / Default theme on left, Dark theme on right


adrienlaws:
If this PR is accepted, I will submit a fast-follow PR for dark mode resources for the 9 client buttons
- help/get/drop/rest/mail + inventory/spells/skills/stats buttons

They're still silver/grey from the default theme.

clientd3d/client.c
```diff
@@ -275,18 +275,30 @@ void ClearMessageQueue(void)
 }
 /************************************************************************/
 /*
- * ThemeApply:  Reload the color palette for the active theme and force
- *   a repaint of the main window.  Called when the user changes themes
- *   in the Preferences dialog.
+ * ThemeApply:  Reload the color palette and main window background
+ *   bitmap for the active theme, fire EVENT_COLORCHANGED so modules
+ *   refresh their themed resources, and force a repaint of the main
+ *   window.  Called when the user changes themes in the Preferences
```
akirmse: So, this only works if called from the Preferences dialog?  It can never be called from anywhere else?

clientd3d/color.c
```diff
@@ -287,6 +287,33 @@ void ColorsSave(void)
...
+ *   Single source of truth for theme-aware code paths in the client and
+ *   in modules that link against the client's exports.
+ */
+bool ThemeIsDark(void)
+{
+	return config.theme == Theme::Dark;
+}
+/************************************************************************/
+/*
+ * ThemeResourceId:  Returns the variant of a bitmap resource ID for the
```
akirmse: This function seems pretty unnecessary.  Is the idea that you're going to add more entries here?  If so, add a TODO so that readers aren't confused.

clientd3d/color.c
```diff
@@ -287,6 +287,33 @@ void ColorsSave(void)
 	WriteConfigInt(section, INIColorVersion, THEME_COLOR_VERSION, ini_file);
 }
 /************************************************************************/
+/*
+ * ThemeIsDark:  Returns true if the active theme is the dark theme.
+ *   Single source of truth for theme-aware code paths in the client and
```
akirmse: The second sentence of this (AI-generated?) comment is pointless.  Why is it telling us that modules link against the client's exports in a Theme-related function?

module/merintr/drawint.c
```diff
@@ -782,6 +804,15 @@ void InterfaceDrawElements(HDC hdc)
 	  if( i >= ELEMENT_IULTOP && i <= ELEMENT_ILRRIGHT )
 		  continue;
 
+	  // Skip silver stats frame corners for themes that opt out via
```
akirmse: "silver"?  Where is this nomenclature coming from?  Does the code only work if the bitmaps are silver?

module/merintr/inventry.c
```diff
@@ -191,10 +191,10 @@ void InventoryBoxCreate(HWND hParent)
       selftrgt_bits = NULL;
    else selftrgt_bits = ((BYTE *) ptr) + sizeof(BITMAPINFOHEADER) + NUM_COLORS * sizeof(RGBQUAD);
 
-   if (!GetBitmapResourceInfo(hInst, IDB_INVBKGND, &inventory_bkgnd))
+   if (!GetBitmapResourceInfo(hInst, ThemeResourceId(IDB_INVBKGND), &inventory_bkgnd))
      debug(("InventoryBoxCreate couldn't load inventory background bitmap\n"));
 
-	if( !( ptr = GetBitmapResource( hInst, IDB_INVBKGND ) ) )
+	if( !( ptr = GetBitmapResource( hInst, ThemeResourceId(IDB_INVBKGND) ) ) )
```
akirmse: Use a local variable rather than calling this same function on consecutive lines.

module/merintr/mermain.c
```diff
@@ -106,6 +106,15 @@ void InterfaceFontChanged(WORD font_id, LOGFONT *font)
 /****************************************************************************/
 void InterfaceColorChanged(WORD color_id, COLORREF color)
 {
+   // For a global theme change (color_id == -1), reload theme-dependent
+   // bitmaps so the inventory texture and frame elements pick up the new
+   // theme without needing to relog.
+   if (color_id == (WORD)-1)
```
akirmse: color_id == -1 means a theme change?  Is that documented somewhere?

Would it be better to have InventoryChangeColor do this instead, so you don't need to special-case theme changes?

module/merintr/theme.c
```diff
@@ -0,0 +1,80 @@
...
+   return ThemeIsDark();
+}
+/****************************************************************************/
+/*
+ * ThemeSkipSilverFrame:  Returns true if the silver stats frame elements
+ *   (corners and edge repeaters) should be skipped during interface
+ *   element drawing.  Used by themes whose sidebar fill or color palette
+ *   makes the silver artwork visually wrong.
+ */
+bool ThemeSkipSilverFrame(void)
```
akirmse: This function and its comment seem pretty poorly thought out.

module/merintr/userarea.c
```diff
@@ -60,8 +60,9 @@ void UserAreaRedraw(void)
       debug(("DrawUserBitmap got NULL player object or area\n"));
    else
    {
-      // Set up window background under user bitmap
-      OffscreenWindowBackground(NULL, user_area.x, user_area.y, user_area.cx, user_area.cy);
+      // Paint the portrait background to match the surrounding sidebar fill.
+      RawBitmap *bg = ThemeSidebarUsesInventoryFill() ? pinventory_bkgnd() : NULL;
+      OffscreenWindowBackground(bg, user_area.x, user_area.y, user_area.cx, user_area.cy);
```
akirmse: This pattern is repeated three times in this PR.  Make a function for it?

module/merintr/theme.c
```diff
@@ -0,0 +1,80 @@
...
+#include "merintr.h"
+
+/****************************************************************************/
+/*
+ * ThemeResourceId:  Returns the variant of a merintr bitmap resource ID
+ *   that matches the active theme.  When no override exists for the
+ *   given ID (or the default theme is active), returns the input ID
+ *   unchanged.  Callers stay theme-blind.
+ */
+int ThemeResourceId(int id)
```
akirmse: There are two functions called ThemeResourceId?  Uhhh....

akirmse (commented):
Please review function comments before sending PRs, even if they're AI-generated.  I'm spending a lot of time just trying to understand comments.

clientd3d/client.c
adrienlaws: Fixed - removed mention of how it's called.

clientd3d/color.c
adrienlaws: Fixed - added TODO.

clientd3d/color.c
adrienlaws: Fixed - sentence removed. ThemeIsDark has also been removed and replaced by ThemeCurrent.

module/merintr/drawint.c
adrienlaws: Fixed - renamed ThemeSkipSilverFrame to ThemeSkipStatsAreaFrame. This was sloppy of me.  I cleaned up references to "silver" to be more specific.

module/merintr/inventry.c
adrienlaws: Fixed - changed to use local variable.

module/merintr/mermain.c
adrienlaws: Refactored as recommended.  Named the sentinel COLOR_ID_ALL in color.h with a comment. Also moved the inventory bitmap reload into InventoryChangeColor.  It now reloads themed resources when it sees the sentinel.

module/merintr/theme.c
adrienlaws: Refactored all this.  Renamed to ThemeSkipStatsAreaFrame, adjusted the comment, and now the body switches over ThemeCurrent instead of returning ThemeIsDark.

module/merintr/userarea.c
adrienlaws: Fixed - created OffscreenSidebarBackground in drawint.c. Now calling it from 3 areas.

module/merintr/theme.c
adrienlaws: Fixed - this was terrible, sorry.  Made specifically named functions for both modules instead of using the same function name. Now using MainThemeResourceId for clientd3d and MerintrThemeResourceId for merintr.

adrienlaws:
This PR is ready for another review, when possible. I will review the comments better next time before submitting a PR. Sorry about that.  I think it's in a much cleaner state now.

module/merintr/drawint.c
```diff
@@ -782,6 +816,13 @@ void InterfaceDrawElements(HDC hdc)
 	  if( i >= ELEMENT_IULTOP && i <= ELEMENT_ILRRIGHT )
 		  continue;
 
+	  // Skip stats_area frame corners if the active theme opts out.
+	  if (ThemeSkipStatsAreaFrame())
```
akirmse: Seems like these two if checks can be combined.

module/merintr/drawint.c
```diff
@@ -801,6 +842,13 @@ void InterfaceDrawElements(HDC hdc)
 	if( i == ELEMENT_BBOTTOM )
 		continue;
 
+	// Skip stats_area edge repeaters if the active theme opts out.
+	if (ThemeSkipStatsAreaFrame())
```
akirmse: These can be combined.

module/merintr/inventry.c
```diff
@@ -1434,3 +1439,28 @@ RawBitmap* pinventory_bkgnd()
 {
 	return &inventory_bkgnd;
 }
+/************************************************************************/
+/*
+ * InventoryReloadBackground:  Re-fetch the inventory background bitmap
```
akirmse: fetch -> load, maybe?  Fetch sounds like you're getting it from the network.

module/merintr/theme.c
```diff
@@ -0,0 +1,90 @@
...
+/****************************************************************************/
+/*
+ * MerintrThemeResourceId:  Returns the variant of a merintr bitmap
+ *   resource ID that matches the active theme.  Returns the input ID
+ *   unchanged when no variant exists for the active theme.  Callers
+ *   stay theme-blind.
+ *
+ *   See docs/themes.md.
+ */
+int MerintrThemeResourceId(int id)
```
akirmse: "Merintr" is a horrible name that we came up with because it had to fit into DOS's 8-character filename limit.  Maybe "Interface"?

adrienlaws: Fixed - renamed MerintrThemeResourceId to InterfaceThemeResourceId

module/merintr/inventry.c
adrienlaws: Fixed - changed as requested

module/merintr/drawint.c
adrienlaws: Combined.

module/merintr/drawint.c
adrienlaws: Combined.

clientd3d/client.h
```diff
@@ -143,6 +143,7 @@ M59EXPORT void _cdecl dprintf(const char *fmt,...);
 #include "font.h"
 #include "editbox.h"
 #include "maindlg.h"
+#include "config.h"
```
akirmse: Is this change necessary?  Revert this file if not.

docs/themes.md
```diff
@@ -1,11 +1,118 @@
...
+## Sidebar fill
+
+The right sidebar (enchantments, portrait, stat bars) sits inside the main window.  Two fill strategies exist for the sidebar area:
+
+- Show the main window background through any gaps between drawn elements.
+- Paint the sidebar with the inventory texture so it has its own fill.
+
+Each theme picks one.  The choice depends on whether the main window background reads well behind the portrait and stat bars.  For example, the dark theme paints the inventory texture across the sidebar because the dark main window background lacks contrast against the portrait; the default theme leaves the main window background visible.
+
+The choice lives in `ThemeSidebarUsesInventoryFill` in `module/merintr/theme.c`.  It is a switch over `ThemeCurrent()` that returns true for themes that paint with the inventory texture, false for the rest.  To activate the inventory fill for a new theme, add a `case Theme::<NewName>: return true;` to that switch.
```
akirmse: If you want to maintain this level of per-line documentation here, then you need to promise to monitor all future changes to the codebase, and update the documentation if anything changes in the code that affects it.  There's no way anyone else will know that this needs updating.

module/merintr/inventry.c
```diff
@@ -1434,3 +1439,28 @@ RawBitmap* pinventory_bkgnd()
 {
 	return &inventory_bkgnd;
 }
+/************************************************************************/
+/*
+ * InventoryReloadBackground:  Reload the inventory background bitmap
+ *   for the active theme and rebuild the scrollbar pattern brush.
+ */
+static void InventoryReloadBackground(void)
```
akirmse, line 1434: It looks like this function essentially duplicates the code at line 195.  Can you remove the duplication (e.g. have one call the other, or similar)?

clientd3d/client.h
adrienlaws: I had to do this to build without errors.

`color.h` now declares `Theme ThemeCurrent`, and `Theme` is defined as `enum class Theme : int` in `config.h`. 

In the previous ordering with `color.h` before `config.h`, the parser hit the `Theme` return type before the type was known and the build failed with:
`C4430: missing type specifier - int assumed`
`C2146: missing ';' before identifier 'ThemeCurrent'.`

The easiest fix was to move `config.h` up higher in `client.h` but I could also do a forward declaration of `Theme` in `color.h` if that is better.

akirmse:
Yes, please do a forward declaration and revert client.h.

clientd3d/client.h
adrienlaws: Fixed - did forward declaration instead and reverted `client.h`

docs/themes.md
adrienlaws: Fixed ( I think ) - removed a lot of super specific details and saved them in my personal notes for reference.  Tried to make it more robust as "general" documentation.

module/merintr/inventry.c
adrienlaws: Fixed - `InventoryBoxCreate` now calls `InventoryReloadBackground` instead of duplicating the load and brush-creation code.

clientd3d/color.h
```diff
@@ -12,6 +12,10 @@
 #ifndef _COLOR_H
 #define _COLOR_H
 
+// Forward declaration of Theme required for ThemeCurrent() to
```
akirmse: This comment doesn't add anything; all it can do is get out of date if something changes elsewhere in the codebase.  People understand the purpose of forward declarations.

Is the ": int" necessary in a forward declaration?  I honestly don't know.

adrienlaws: Fixed - remove the comment.  I had searched the codebase for similar situations and found a comment that I reused here from https://github.com/Meridian59/Meridian59/blob/13d1b5ea4586d1fa798417da5dfaf08c618dad2c/clientd3d/draw3d.h#L19  Just sharing in case we would want to remove that.

The ": int" was also not necessary.  I removed it.  I guess int type is implicit.

PR #1431, opened 2026-05-10
Refactor particle system

neiltaco:
This PR refactors the particle system to improve readability and performance.  Its randomness is also updated to provide better distribution and also offer more customization on how to vary a particle's position, rotation, and velocity.  These changes only affect the sandstorm effect for now, and helps set up the code to include the future weather particle setup.

Refactoring
- Overhauled randomness to use mt19937 for better distributions.  Added randomizing particle position/rotation/velocity.
- Particle creation now uses circular buffering instead of a for-loop to optimize fetching of new particles.
- Refactored sandstorm emitter setup to use a for-loop to set up emitters, making it easier to read and adjust.
- Extracted parts of `D3DParticleSystemUpdate()` and made them into their own functions.
- Updated C-style code to use C++ equivalents.
- Fixed possible leak from emitters that were never freed upon game shutdown.

Cleanup
- Reorganized `d3dparticle.c` and `d3drender_fx.c` 
- Removed unused variables and redundant if-statements.

### DIFF
--- a/clientd3d/d3dparticle.c
+++ b/clientd3d/d3dparticle.c
@@ -7,8 +7,9 @@
 // Meridian is a registered trademark.
 #include "client.h"
 
-void D3DParticleDestroy(particle *pParticle);
-
+//////////////////////
+// Public Functions //
+//////////////////////
 void D3DParticleSystemReset(particle_system *pParticleSystem)
 {
 	list_destroy(pParticleSystem->emitterList);
@@ -92,7 +93,7 @@ void D3DParticleSystemUpdate(particle_system *pParticleSystem, d3d_render_pool_n
 
 			if (--pParticle->energy <= 0)
 			{
-				D3DParticleDestroy(pParticle);
+				pParticle->energy = 0;
 				pEmitter->numParticles--;
 			}
 			else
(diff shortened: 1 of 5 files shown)


clientd3d/d3drender_fx.c
```diff
@@ -7,244 +7,98 @@
 // Meridian is a registered trademark.
 #include "client.h"
 
-// Variables
+///////////////
+// Variables //
+///////////////
+static std::vector<particle_system*> particleSystemList;
```
akirmse: Why is there a list when there's only ever one particle system?

Also, by making this a list of pointers, a reader would expect that particle systems are heap-allocated.  But anything put on this list is leaked, because there's no cleanup.  So its expected usage is confusing.

clientd3d/d3dparticle.c
```diff
@@ -6,254 +6,175 @@
...
+// Constants //
+///////////////
+static constexpr uint32_t PARTICLE_INDICES = 2;
+static constexpr uint32_t PARTICLE_VERTICES = 2;
+static constexpr uint32_t PARTICLE_PRIMITIVES = 1;
+
+///////////////
+// Variables //
+///////////////
+static std::mt19937 gen(static_cast<uint32_t>(std::chrono::high_resolution_clock::now().time_since_epoch().count()));
```
akirmse: Maybe a better name than "gen"?

neiltaco: I think 'rng' works.

clientd3d/d3drender_fx.c
neiltaco: Ah!  Originally I brought in that list from the weather PR to get the particle system ready for future particle systems.  Now that you mention it, the list seems redundant until there's something else besides the sandstorm effect.  I can remove that list for now.

PR #1435, opened 2026-05-16
QOL: Reduce shroud concentration time

pauly-barns:
**What**
This change updates the concentration time to seven seconds, down from thirty.  

**Why**
A recent uptick in use of Shatter and Swap has caused a good deal of conversation about the painful nature of Shrouding items; the counterplay.

**Rationale**
I settled on seven seconds in terms of relative 'battle preparation' tasks.  Shrouding your shillings, weapon, armor and most expensive reagent now takes the same amount of time as enchanting your weapon.

### DIFF
--- a/kod/object/passive/spell/shroud.kod
+++ b/kod/object/passive/spell/shroud.kod
@@ -40,7 +40,7 @@ classvars:
    viSpell_level = 3
    viMana = 10
    viSpellExertion = 10
-   viCast_time = 30000
+   viCast_time = 7000
 
    viChance_To_Increase = 25
 


akirmse:
We're going to leave this at its current value.

pauly-barns:
> We're going to leave this at its current value.

Understood.  Why?  Feedback is very helpful for contribution guidelines / guard rails.

akirmse:
I didn't hear a whole lot of justification other than just making the game easier, which is generally what happens over time when players suggest ideas.  Maybe have a discussion on Discord first next time to gather different points of view; maybe there are better ideas that are more balanced.

pauly-barns:
Got it.  Appreciate the insight!

PR #1436, opened 2026-05-17
Add centralized delta time and modernize particle timers.

neiltaco:
This PR adds a separate new source file called `clock.c` that is dedicated to calculating delta time.  It also updates the particle system to use seconds instead of frames, which will help prepare it in the future for doing physics calculations.

Several source files were calculating their own delta time using `timeGetTime()` independently, which risked having desynced timings.  `move.c` in particular depended on the delta time calculated by `animate.c`.  `timeGetTime()` is also known to be less precise and has some performance overhead.  By having a centralized delta time, source files can simply retrieve the exact same delta time already calculated from `clock.c`.

This PR updates only the particle FX.  Certain source file such as `move.c` had unintended side effects when using the new clock and were left out of this PR for now.

New
- Created `clock.c` to centralize tracking delta time.
- Added `GetDeltaTime()` to calculate in seconds.

Refactoring
- Refactored particle system to use seconds instead of frames for particle and emitter timers.
  - Adjusted sandstorm effect based on changes to timing.  Also tightened its radius and particle lifetime, and added a slight curvature to particle flight.

### DIFF
--- a/clientd3d/clock.c
+++ b/clientd3d/clock.c
@@ -8,12 +8,15 @@
 #include "client.h"
 #include <chrono>
 
-using steady_clock_time_point = std::chrono::time_point<std::chrono::steady_clock>;
+//////////////////////
+// Local Namespaces //
+//////////////////////
+using namespace std::chrono;
 
 ///////////////
 // Variables //
 ///////////////
-static steady_clock_time_point lastEndFrame;
+static time_point<steady_clock> lastEndFrame;
 
 // Elapsed time between frames (in seconds).
 static float deltaTime_s;
@@ -45,33 +48,32 @@ void InitializeTime()
 	deltaTime_ms = 0;
 	msAccumulator = 0.0f;
 
-	lastEndFrame = std::chrono::steady_clock::now();
+	lastEndFrame = steady_clock::now();
 }
 
 void UpdateTime()
 {
 	// Get current time first.
-	auto currentTime = std::chrono::steady_clock::now();
+	auto currentTime = steady_clock::now();
 
 	// Calculate delta time (in seconds) since last frame.
 	auto frameTime = currentTime - lastEndFrame;
-	deltaTime_s = std::chrono::duration<float>(frameTime).count();
+	deltaTime_s = duration<float>(frameTime).count();
 
 	// Cap the max allowed delta time to roughly an equivalent to 10 FPS. This prevents things
 	// like game logic or animations from "spiking" if the window halts (e.g. during a resize).
-	if (deltaTime_s > 0.1f)
-	{
-		deltaTime_s = 0.1f;
+	static constexpr duration<float> DELTA_TIME_CAP = duration<float>(0.1f);
 
-		// Sync frameTime duration back down to match the safety cap.
-		frameTime = std::chrono::duration_cast<std::chrono::steady_clock::duration>(std::chrono::duration<float>(0.1f));
-	}
+	// Sync frameTime duration back down to match the safety cap.
+	frameTime = std::min(frameTime, duration_cast<steady_clock::duration>(DELTA_TIME_CAP));
+	deltaTime_s = duration<float>(frameTime).count();
+
 	// Always slide the time window completely forward to prevent catch-up loops.
 	lastEndFrame = currentTime;
 
 	// Also calculate delta time in milliseconds for systems that still use it.
 	// msAccumulator keeps track of fractional time so it isn't lost between frames.
-	float totalMs = (std::chrono::duration<float, std::milli>(frameTime).count()) + msAccumulator;
+	float totalMs = (duration<float, std::milli>(frameTime).count()) + msAccumulator;
 	deltaTime_ms = static_cast<int>(totalMs);
 	msAccumulator = totalMs - static_cast<float>(deltaTime_ms);
 }
--- a/clientd3d/clock.c
+++ b/clientd3d/clock.c
@@ -17,6 +17,7 @@ using namespace std::chrono;
 // Variables //
 ///////////////
 static time_point<steady_clock> lastEndFrame;
+static time_point<steady_clock> lastCappedEndFrame;
 
 // Elapsed time between frames (in seconds).
 static float deltaTime_s;
@@ -42,7 +43,6 @@ float GetDeltaTime()
 }
 
 // Returns delta time in milliseconds.
-// Used in legacy systems that still use milliseconds instead of seconds.
 int GetDeltaTime_Ms()
 {
 	return deltaTime_ms;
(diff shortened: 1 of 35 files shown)


neiltaco:
I believe this PR should be ready for review.  I tested the systems that were updated to use the delta time in `clock.c` and addressed the following bugs resulting from the changes:
- Now that `graphics.c` uses a centralized uncapped delta time instead of the one calculated from rendering a room, it's possible for `msSleep` to be negative, so it has a new safety check to ensure it has positive value.  It is offset by 1 because `if (msSleep > 0)` meant that '1' can be passed into `Sleep()`.  It seemed to result in framerate drops in GPU Efficiency Mode.
- A new constant `MAX_STEP_THRESHOLD` was added to clamp both capped and uncapped delta time to a max of 250ms.
  - This fixes a newly introduced sliding movement bug for moving objects such as enemies.  They would use capped delta time in milliseconds that would max out at 100ms, which didn't appear to be enough to trigger the coordinate snap with `MoveSingle()` within `moveobj.c`.  This resulted in objects rapidly sliding into position if the window resize was held for several seconds then released.
  - For uncapped delta time, this prevents a speed burst bug.  In `graphics.c`, it was possible for a long enough delay to have `Sleep()` skipped while `msSleep` is negative.  The player could move incredibly fast while there was a 'time debt'.

akirmse:
I think I want @sicoderealm to take a look at this too (he's out for a week), as he did most of the recent FPS changes.

I don't understand the "capped time" concept; it seems to come out of nowhere.

neiltaco:
Originally, capped delta time was meant to solve weird bugs like enemies sliding fast into position if delta time was abnormally large.  Eventually `MAX_STEP_THRESHOLD` was added to help trigger the position snap that is built into `MoveSingle()`, and also prevent the speed exploit.

Now that you mention it, the 'capped delta time' name might be confusing since there's also another cap for the large 250ms delay.  I also tested out both uncapped and capped delta time by adding a temporary artificial 150ms stutter every five seconds, and tested with GPU Efficiency both on and off.  There seems to be little to no difference.

I think 'capped' delta time should be safe to remove.  I'll also be happy to hear Si's input as well if anything seems off in this PR.

Meridian59:
I think I was hoping to see a simpler change that just replaced all the
individual clocks with a single, consistent one.  With all these extra
changes in here, it's really hard to know what all the side effects are.

neiltaco:
I think I see what you mean.  I thought it would be straightforward to have the separate delta times switched to a centralized one.  Although the newly introduced speed exploit and the "sliding" bug should have been big red flags for me that there are major unintended side effects.

I can dial back this PR to just add `clock.c`, and just calculate raw delta time (both sec and ms).  I think the updates to `d3drender_fx.c` and `d3dparticle.c` should be safe to include since they are purely aesthetic.  I think everything else such as `move.c` can be tested in PRs that are separate from this one.

clientd3d/d3drender_fx.c
```diff
@@ -23,26 +23,31 @@ static particle_system sandParticleSystem;
...
 	static constexpr int SAND_EMITTER_COUNT = 8;
-	static constexpr int SAND_LIFETIME = 40;
-	static constexpr float SAND_EMITTER_RADIUS = 10000.0f;
-	static constexpr float SAND_Z_VARIANCE = 1000.0f;
-	static constexpr float SAND_VELOCITY = 400.0f;
-	static constexpr float SAND_TIMER = 1;
-	static constexpr float SAND_RAND_ROT = PI / 1000.0f;
+	static constexpr float SAND_EMITTER_TIMER_S = 0.015f;
+
+	// Sand particle fly really fast, so their duration are short to use up less resources.
```
akirmse: particle -> particles, duration -> durations

Not sure I understand this comment.  Why does flying fast use more resources?

clientd3d/clock.c
```diff
@@ -0,0 +1,68 @@
...
+//////////////////////
+
+// Returns delta time in seconds.
+float GetDeltaTime()
+{
+	return deltaTime_s;
+}
+
+// Returns delta time in milliseconds.
+int GetDeltaTime_Ms()
```
akirmse: This simplification of the PR is good because it makes clearer a fundamental problem.  At super high frame rates, the diff between frames could be 0 ms, which will cause all sorts of problems.  So I think you should remove this function and instead fix anyplace that is using integer milliseconds, and change it to use floating point seconds.  It looks like there aren't many.

clientd3d/d3drender_fx.c
neiltaco: I updated the comment so it's more about the visuals, and that fast-moving particles just need a short duration since they are only seen briefly on-screen.

clientd3d/d3drender_fx.c
```diff
@@ -117,7 +117,7 @@ void D3DPostOverlayEffects(const FxRenderSystemStructure& fxrss)
 	{
 		custom_bgra	bgra;
 
-		effects.duration -= GetDeltaTime_Ms();
+		effects.duration -= static_cast<int>(GetDeltaTime() * 1000.0f);
```
akirmse: This just moves the 0ms problem here instead.  What would be involved in making effects.duration floating point seconds?

neiltaco: Oh!  I think I was originally concerned with how much would need to change if the packets needed to change to allow floats.  Now that I think about it, the packet doesn't need to change, but the int milliseconds passed in can just be converted to float seconds in PerformEffect().

clientd3d/clock.h
```diff
@@ -0,0 +1,18 @@
...
+//
+// Meridian is a registered trademark.
+
+#ifndef _CLOCK_H
+#define _CLOCK_H
+
+// Returns delta time in seconds.
+float GetDeltaTime();
+
+void InitializeTime();
```
akirmse, line 16: You could use some comments here explaining how to use these functions.  i.e. call one at startup, and one on each frame (beginning of frame?  end?)

clientd3d/effect.h
```diff
@@ -37,10 +38,10 @@ typedef struct {
    ID   wadingsound;  // ID of wav resource for wading in current room
    ID   ambientsound; // ID of wav resource played constantly in current room
 
-   int  flashxlat;
-   int	duration;
-   int  xlatOverride;
-} Effects;
+   int  	flashxlat;
+   float	flashxlatDuration;
```
akirmse: yuck, this is the only one in seconds---needs a comment.

So, the real change would be to make all of these into floating point seconds, then change everything in animate.c to also use floating point seconds.  And come to think of it, the animation timer is kind of weird, since it's asynchronous with frames actually being drawn.  Might need @sicoderealm  to weigh in on this.  But the time intervals in Animate should definitely be floating point seconds in an ideal world.

sicoderealm:
Agree with the direction. The end state should be float seconds throughout animate.c, and the timer being async with frames is the right design - but that's exactly why dt should come from the animation timer's own tick, not the renderer's; otherwise dialogs freeze the dt. A few things I'd want before this lands:

1. Tighten the scope. Land clock.{c,h} plus the conversion of one subsystem (particles is the natural pick as already float-native). Convert the remaining subsystems in follow-up PRs so each one is reviewable on its own. @akirmse's "hard to know the side effects" worry is valid; some of the integer-modulo accumulators (e.g. animate.c:228–231 bounceTime/TIME_FLASH) are deliberately integer-cyclic and will drift under naive float conversion.
 
2. UpdateTime()-via-GameIdle() leaves animation with a stale dt during dialogs. AnimationTimerProc runs from two places: GameIdle() during normal play, and GameTimer(TIMER_ANIMATE) when a menu/dialog blocks the message pump (statgame.c:367, 389, 399–400). If UpdateTime() is only called from GameIdle(), the SetTimer path reads a frozen GetDeltaTime() for the entire time the dialog is up - animations either stall or fast-forward depending on what the last idle frame measured. Simplest fix: keep AnimationTimerProc's pre-PR local last_animate_time and compute its own dt. clock.c is the right home for the renderer's dt (particles, fx) - animation is a separate timeline and shouldn't share it.

 3. Cap dt at something like 0.1 s before exposing it: alt-tab / breakpoint pauses currently let dt grow unbounded, which will warp particles and projectiles on resume.
 
4. ModuleEvent(EVENT_ANIMATE, …) payload changes type - worth confirming with @akirmse whether the module interface is considered stable, since existing modules compiled against the old int ms signature will misread the new double seconds.

Separately: the frame-pacing in graphics.c:346–360 is still Sleep(msSleep) which IMO is the bigger source of FPS instability than the dt scattering. We have been discussing and looking at this for some time. The standard fix is a hybrid sleep-then-spin against an intended target time, not last-wakeup time, worth a follow up PR and I'm happy to write that one up if useful.

neiltaco:
Thank you both for the feedback.  I admit I interpreted the 'change to seconds' response as a green light to start converting the rest of animate.c to only affect animations, though that quickly grew in scope and it still ended up affecting movement code.

I'll keep this PR simple once more: adding the infrastructure for clock.c and also having the particle system switch from frames to seconds.  The delta time cap was also readded to keep it from growing to absurd durations.

neiltaco:
This PR is ready for review, and I'm happy to address anything else that is needed.

I tested the delta time by casting sandstorm and the movement appears working as intended.  I also tested the delta time cap by temporarily slowing down the sandstorm particles to a velocity of 10.0f, and then holding on the game window for several seconds before releasing it.  The particles don't "leap" forward and continue roughly from where they were last positioned before the pause.

Of course, this capped delta time will cause issues with Effects, where it can actually extend how long something like the blur effect will last.  I plan to test and address that in a separate PR.

clientd3d/clock.h
```diff
@@ -0,0 +1,23 @@
...
+// Meridian is a registered trademark.
+
+#ifndef _CLOCK_H
+#define _CLOCK_H
+
+// Returns delta time in seconds.
+float GetDeltaTime();
+
+// Sets up baseline clock snapshot.
+// Called within WinMain() in client.c right before entering primary game update loop.
```
akirmse: You should not add comments saying where this function must be called from.  Do you mean that the game will instantly crash if it's not called "right before entering primary game update loop"?

clientd3d/clock.h
```diff
@@ -0,0 +1,23 @@
...
+void InitializeTime();
+
+// Refreshes clock snapshot for the current frame.
+// Called every game loop at the very start of GameIdle() in statgame.c.
```
akirmse: Do not do this.  Instead say that it should be called at the start of each frame (if that's what you mean).

clientd3d/clock.h
neiltaco: Whoops, I forgot to update the PR here after your last feedback asking not to state exactly where functions are called.

I adjusted the comments for both functions to instead state the order they should be called in.

PR #1437, opened 2026-05-18
Add admin setting to toggle flickering light effects

sicoderealm:
Adds a global `pbFlickeringLightsEnabled` admin setting that controls whether flickering light objects (torches, braziers, candles, lamps, campfires, etc.) advertise their flicker effect to clients.

- When enabled (default), flickering objects report `FLICKERING_YES` as before.
- When disabled, they report as steady lights so no flicker/flash is drawn - useful for players sensitive to flashing light.

Takes effect per player the next time they load a room.


akirmse:
I am confused why this would be a user setting.

kod/util/settings.kod
```diff
@@ -151,6 +151,14 @@ properties:
    % If disabled, only the player opening the door hears the sound.
    pbDoorSoundShared = TRUE
 
+   % If enabled, flickering light objects (torches, braziers, candles, lamps,
+   % campfires, etc.) tell clients to render their light as flickering/flashing.
+   % If disabled, those objects report themselves as steady lights instead, so
+   % no flicker effect is drawn.  Admins can flip this to globally turn the
+   % flickering light effect on or off (e.g. for players sensitive to flashing
+   % light).  Takes effect for each player as they next load a room.
```
akirmse: There's a lot of "what" here, but no "why".  Why would we want to disable all flickering lights?  I mean, we don't have a setting to disable all monsters, which seems just about as sensible.

sicoderealm: comment updated with a little more context, ty

kod/object/passive/flikerer.kod
```diff
@@ -97,8 +97,12 @@ messages:
       local iFlags;
 
       iFlags = viObject_flags;
-      
+
+      % Only advertise the flickering light effect to clients when this object
+      % is set to flicker AND the global admin setting allows it.  When the
+      % setting is off the object reports as a steady light instead.
       if pbIsFlickering
+         AND Send(Send(SYS,@GetSettings),@FlickeringLightsEnabled)
```
akirmse, line 105: Did you try this, and it works?  I don't see a function called "FlickeringLightsEnabled".

sicoderealm: not sure how that got lost, reinstated - testing and working as expected

PR #1438, opened 2026-05-20
Refactor: Centralize game unit to radian rendering constants for hardware renderer.

neiltaco:
This PR defines the magic numbers used for angle calculations and conversions in the D3D source files and centralizes them in `d3drender.h`.  It also moves the degree/radian conversion functions into `drawdefs.h`, and updates the macros used for angle calculations to be static constexpr instead.

- Added new constants used for angle calculations in D3D.
- Move degree/radian conversion functions from `d3drender.c` into `drawdefs.h`.
- Refactored part of `drawdefs.h`:
  - Converted specific macros in `drawdefs.h` used for angle calculation into static constexpr.
  - Refactored existing `DegToRad` and `RadToDeg` macros to be helper functions.
  - Renamed the new helper functions to clarify that they convert game units to radians (and vice versa).

### DIFF
--- a/clientd3d/drawdefs.h
+++ b/clientd3d/drawdefs.h
@@ -1,4 +1,4 @@
-// Meridian 59, Copyright 1994-2012 Andrew Kirmse and Chris Kirmse.
+// Meridian 59, Copyright 1994-2026 Andrew Kirmse and Chris Kirmse.
 // All rights reserved.
 //
 // This software is distributed under a license that is described in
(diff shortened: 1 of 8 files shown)


clientd3d/drawdefs.h
```diff
@@ -63,9 +63,28 @@
 #define VIEWER_DISTANCE  (FINENESS >> 1)         /* Distance from viewer to screen in pixels */
 #define LOG_VIEWER_DISTANCE  (LOG_FINENESS - 1)
 
-/* Convert from pseudo degrees to radians */
-#define DegToRad(x) ((float) (x) * PITWICE / NUMDEGREES)
-#define RadToDeg(x) ((long)  ((x) * NUMDEGREES / PITWICE))
+constexpr float deg_to_rad(float degrees)
+{
+	constexpr float DEG_TO_RAD_FACTOR = PITWICE / 360.0f;
```
akirmse: These extra expressions inside these 4 functions are kind of overkill.. just do the calculation in-line.

clientd3d/drawdefs.h
```diff
@@ -13,12 +13,12 @@
...
 
-#define NUMDEGREES 4096
+constexpr int NUMDEGREES = 4096;
 #define LOG_NUMDEGREES 12
 #define NUMDEGREES_MASK 0xfff
 
-#define PI      3.1415926f
-#define PITWICE 6.2831853f
+constexpr float PI = 3.1415926f;
+constexpr float PITWICE = 6.2831853f;
```
akirmse: PI * 2 ?

neiltaco: I was going to do the same to PIHALF, though I found out it currently isn't used at all so I removed it to clear up the header.

PR #1440, opened 2026-05-22
feature: add client theme buttons helpers and dark theme button resources (theming part 3)

adrienlaws:
## What

- Follow-on to #1430.  
- Two related changes:
  - 1/ Adds the capability for each toolbar button to declare its own per-theme bitmap lookup.
  - 2/ Adds dark theme art: 11 new dark BMP files for the 9 main-window buttons (5 top: help, drop, get, rest, mailbox; 4 tab: inventory, spells, skills, stats).

When the dark theme is active, the dark icons render automatically.  Switching themes mid-session swaps the icons without restarting the client.

## Why

- Allows for extensibility.  Putting the lookup on each button keeps the toolbar code theme-blind and lets future themes  register their own button variants without touching the toolbar code.
- PR #1430 darkened the sidebar and frame, but the buttons themselves still rendered with the original light icons against the dark sidebar.  Adding dark variants finishes the look.

## How

- 11 new dark BMP files.  The 4 tab buttons each render as 3 pieces (left icon, middle, right edge), and the middle and right pieces are shared across all 4 tabs, so 4 tabs use 6 BMPs (4 icons + 1 shared middle + 1 shared right).  The 5 top buttons use 1 BMP each.  Total: 11 BMPs for 9 visible buttons.
- Each toolbar button has an optional function pointer that returns the bitmap ID for the active theme.  When set, the toolbar code calls it before loading the bitmap; when unset, the toolbar uses the original ID.  This keeps the toolbar code from needing to know which themes exist.
- The merintr module sets this pointer to the existing `InterfaceThemeResourceId` for its toolbar buttons.  The mailnews module uses its own small function for the mailbox icon.
- `InterfaceThemeResourceId` is extended to map the 10 merintr-owned button bitmap IDs to their dark counterparts (4 top buttons + 6 tab pieces).  The mailbox is handled inside the mailnews module.
- When the user switches themes mid-session, two new functions reload the icons: `ToolbarReloadBitmaps` for the top toolbar and `StatButtonsReload` for the tab buttons.

### Default theme toolbar buttons
image

### Dark theme toolbar buttons
image

### Default theme stat buttons
image

### Dark theme stat buttons
image


akirmse:
I feel like this is probably not worth the extra complexity.  It's a bunch of code and assets for a tiny visual difference; the existing buttons seem like they would work just fine in either mode.

Also, the callback function on each toolbar button is huge overkill IMO.

adrienlaws:
Fair enough, closing this.  I'll save the branch in case we ever want to revisit it in some form.

PR #1441, opened 2026-05-25
feature: add per-theme title bar and menu bar customization capability (theming part 4)

adrienlaws:
## What

- The client title bar style can now be set per theme (light or dark text)
- The client menu bar style and color can also be set per theme (any RGB)
- Dark theme now uses both (light text on dark-colored title and menu bars)
- The client menu bar is now owner drawn

## Why

- The title bar and menu bar previously were system grey/white
  - This caused glare / mismatch with darker themes and prevented customization
- Module menus (Spells, Actions) used to render in the system style after the player logged in
  - Modules add their menus after the main client has already themed the menu bar, so they need to re-trigger the themed rendering
- Themes had no clean way to set the title bar style or menu bar color
  - This change allows for both color and text customizations of the title and menu bar beyond dark theme
    - Example: implementing a gold/brown theme at a later date
- Changing the client to use an owner drawn menu bar allows for live-swapping of the menu bar color (themed to unthemed)
  - The menu bar is owner-drawn in all themes, not just themed ones
  - Default-theme menu bars still match the system look after switching to being owner drawn

### New functions

- `color.c` (where the theme decides to use title or menu bar configurations)
    - `ThemeMenuBarColor` — per-theme menu bar background color, or `CLR_INVALID` for system default
    - `ThemeUsesDarkTitleBar` — per-theme title bar style (true = light text on dark, false = dark text on light)
    - These follow the same shape as the existing per-theme helpers
        - A theme can pick either, both, or neither

- `client.c`
    - `ThemeApplyTitleBar` — applies the title bar style for the active theme

- `winmenu.c`
    - `MenuBarApply` — builds the brushes for the current theme and sets the menu's background.  Marks any top-level items not yet owner-drawn, allocating a text buffer per item that `MenuBarShutdown` frees at exit.  Separators are skipped
    - `MenuBarMeasureItem` — Windows callback reporting the size of each item
    - `MenuBarDrawItem` — Windows callback painting each item
    - `MenuBarShutdown` — frees the cached brushes, font, and per-item text buffers at client shutdown.  Called from `MainQuit`
    - `GetMenuFont` (static helper) — returns a cached HFONT for the system menu font.  Used by MeasureItem and DrawItem

### Title bar

`ThemeApplyTitleBar` in `clientd3d/client.c` reads `ThemeUsesDarkTitleBar` and calls `DwmSetWindowAttribute` with `DWMWA_USE_IMMERSIVE_DARK_MODE`.  Runs from `ThemeApply` on a theme switch and from `WinMain` on first paint.

Note: Title bar background color is not controlled — only the binary dark/light style.  An exact background color would need `DWMWA_CAPTION_COLOR` (Win11+), which is not in this change.

### Menu bar

`MenuBarApply`, `MenuBarMeasureItem`, and `MenuBarDrawItem` in `clientd3d/winmenu.c` make top-level menu bar items owner-drawn and paint them.  Items stay owner-drawn for their entire lifetime.

- For themes with a menu bar color: hover background uses `COLOR_EDITBGD`, text uses `COLOR_FGD`, both read from the theme's color table
- For themes without a menu bar color: fall back to `GetSysColor(COLOR_WINDOW)`, `COLOR_HIGHLIGHT`, `COLOR_MENUTEXT`, `COLOR_HIGHLIGHTTEXT` so the menu bar matches the system default look
- `MenuBarApply` is exported via `client.def`.  `SpellsInit` and `ActionsInit` call it after inserting their own top-level menus
- `ThemeApply` calls `MenuBarApply` again on a theme switch to rebuild the brushes with the new theme's colors
- The menu font, menu bar brush, and selected-item brush are cached.  The brushes are rebuilt each `MenuBarApply` call.  The font is built once

Owner-drawing top-level items is the standard Win32 approach.  Win32 has no API to set menu bar colors directly.

- Since the client now users owner drawn menu bars, we had to use COLOR_WINDOW for the default theme menu bar background so it remained the same light grey as it was before this change.  Without that, the menu bar and the title bar would become one single color which was overbearing.
- Menu bar item height comes from `GetSystemMetrics(SM_CYMENU)`.  DPI-aware on modern Windows and avoids a hard-coded pixel value.  Item width is the text width as measured.  Trade-off: the gap between top-level items is slightly wider than the system-drawn menu bar.

### OS requirement

- `DWMWA_USE_IMMERSIVE_DARK_MODE` only works on Windows 10 1809 or newer.
- On older Windows versions the call returns an error and the system default title bar shows.  Unlikely to affect many players.
- The name is not about "dark mode" — it controls whether the title bar uses light text on dark or dark text on light.  The name comes from the Win32 API.

### Accelerator underlines

Windows hides the menu accelerator underlines (the underlined letter in "&Game") until the user presses Alt, following the system "underline keyboard shortcuts" setting.  Owner-drawn items have to handle this themselves: `MenuBarDrawItem` checks the `ODS_NOACCEL` flag on the draw item state and passes `DT_HIDEPREFIX` to `DrawText` when set.  Without this, items would always show underlines.

### Why this approach

Win32 has no API to set menu bar colors.  Taking over the rendering means owner-drawing each top-level item, handling `WM_MEASUREITEM` and `WM_DRAWITEM`, and managing GDI handles.  Other Win32 apps with themed menu bars use the same pattern.  Examples include Notepad++ and 7-Zip.

## Example

image
Default theme (left) vs Dark theme (right)


clientd3d/client.c
```diff
@@ -196,6 +201,64 @@ LRESULT CALLBACK WndProc (HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 	case BK_MODULEUNLOAD:
 		ModuleUnloadById(lParam);
 		break;
+
+	case WM_NCPAINT:
+	{
+		// Let Windows paint the non-client area first.
+		LRESULT res = DefWindowProc(hwnd, message, wParam, lParam);
+
+		// Windows draws bright 1px separator lines above and below the
```
akirmse: This seems super, super fragile.  You're depending on the exact pixel positioning of the undocumented Windows menu bar drawing routines?  Is this guaranteed to be the same in all versions of Windows, past and future?

I remember custom-drawing a Windows menu in a game many, many years ago, but I don't recall needing to know the default behavior of Windows to this level of detail.  Why is this necessary?

clientd3d/client.c
```diff
@@ -274,6 +337,19 @@ void ClearMessageQueue(void)
 	}
 }
 /************************************************************************/
+/*
+ * ThemeApplyTitleAndMenu:  Apply the active theme's title bar style
+ *   and menu bar.
+ */
+static void ThemeApplyTitleAndMenu(void)
+{
+	BOOL dwmDark = ThemeUsesDarkTitleBar();
```
akirmse: bool not BOOL (notice that auto solves this problem for you)

clientd3d/winmenu.c
```diff
@@ -236,3 +244,284 @@ void MenuCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
...
+   MENUINFO mi;
+   memset(&mi, 0, sizeof(mi));
+   mi.cbSize = sizeof(mi);
+   mi.fMask = MIM_BACKGROUND;
+   mi.hbrBack = NULL;
+   SetMenuInfo(hMenu, &mi);
+}
+/************************************************************************/
+/*
+ * RebuildMenuItems:  Reads each top-level item out of hMenu, removes
```
akirmse: This is pretty gross.  Is it only needed to apply a theme change without restarting the client?  If so, is it worth it?

clientd3d/winmenu.c
```diff
@@ -236,3 +244,284 @@ void MenuCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
...
+
+   struct SavedMenuItem
+   {
+      char text[256];
+      UINT id;
+      HMENU subMenu;
+      UINT state;
+   };
+
+   SavedMenuItem *items = (SavedMenuItem *)SafeMalloc(count * sizeof(SavedMenuItem));
```
akirmse: Using new[] / delete[] is more standard for this kind of thing.

clientd3d/client.c
akirmse: This should be a static_cast, since ThemeUsesDarkTitleBar returns a bool.

adrienlaws: I had to keep this as BOOL because it needs a 4 byte argument and bool is 1 byte.  `auto` did do the conversion but then since it was 1 byte it broke the call.

I added a comment to explain why we need BOOL.

clientd3d/winmenu.c
adrienlaws: Refactored - this was a mess because I was trying to switch between owner and system drawn menu bars. I tried a different approach by making the menu bar always owner-drawn. 

`RebuildMenuItems`, `ThemedMenuBarRemove`, and `ThemedMenuBarDestroy` are removed.  Live switching still works.

clientd3d/winmenu.c
adrienlaws: Removed one and Fixed one - the mentioned allocation was inside `RebuildMenuItems` and is gone as part of refactor. The other allocation in `winmenu.c` (per-item text buffer) is now new[] / delete[]. The delete[] happens in a new function `MenuBarShutdown `in `winmenu.c` during shutdown.

clientd3d/client.c
adrienlaws: Fixed - I removed the `WM_NCPAINT` and `WM_NCACTIVATE` handlers entirely.  There is one visible line along the bottom edge but it doesn't look bad (IMO).  Screenshot attached.
image

adrienlaws:
This PR is ready for another review, when possible. Thank you.

clientd3d/client.c
akirmse: This should say 

`BOOL dwmDark = static_cast<BOOL>(ThemeUsesDarkTitleBar());`

to indicate that you're intentionally converting types, rather than by accident.

adrienlaws: Fixed - added the static_cast to make the conversion explicit.  Thank you.

PR #1442, opened 2026-05-31
feature: add configurable theme borders for chat and stats areas (theming part 5)

adrienlaws:
## What

- Adds additional theming controls for the chat box and stats area
    - Adds `ThemeSkipChatBoxFrame()` to complement the existing `ThemeSkipStatsAreaFrame` (added in #1430)
    - Adds a customizable border color, `ThemeBorderColor`, that can be used to draw a border around the chat box and/or the stats area in place of the skipped ornament
- Dark theme now uses the above to skip the chat box and stats area ornamental frames and to draw a border specific to the dark theme
- Default theme is unchanged
- related to #1425 

## Why

- #1430 removed the stats area ornamental frame in dark theme, but that left the stats area floating without a border
- The chat box had a border, but it used ornamental frame resources that did not look good in dark theme
- The replacement border should be configurable so future themes can set their own color

### New functions

- `module/merintr/theme.c`
    - `ThemeSkipChatBoxFrame` - returns true when the active theme skips the `B*` chat box ornament.  Same shape as the existing `ThemeSkipStatsAreaFrame`
- `clientd3d/color.c`
    - `ThemeBorderColor` - per-theme border color for the stats area and chat box, or `CLR_INVALID` for the default border.  Exported via `client.def` so both the client and the interface module read it.  One shared color so the two borders match

### Borders

Both borders use the existing `DrawBorder`, converting the color to the nearest palette index with `GetClosestPaletteIndex`.

- Chat box: drawn by `TextInputDrawBorder` in `clientd3d/textin.c`.  No new drawing code.  The chat box already drew its own border, it just uses `ThemeBorderColor` now and keeps its focus highlight (highlight color while focused, theme color while not)
- Stats area: drawn by `InterfaceDrawElements` in `module/merintr/drawint.c`.  New code, since the stats area had no border before

## Notes

- The border color is not a `COLOR_*` table entry.  The table is for user-customizable colors.  This border is a fixed per-theme decision.
- The frame and the border are independent.  Setting a border color without skipping the frame draws both, which may not look good.  Whether to combine them is the theme author's choice.
- The chat box border sits a couple of pixels outside the text area, leaving a thin gap that shows the window background.  This is the existing border geometry, not new to this change.  In the default theme the ornament filled that region, so the gap was hidden.  With the ornament skipped it is faintly visible.  It is minor so I left it out of scope for this change.

#### Chat Box (before)
image

- default border clashes with dark theme and is not configurable per-theme

#### Chat Box (after)
image

#### Stat Area (before)
image

- floats without a border (see lower left)

#### Stat Area (after)
image

### DIFF
--- a/module/merintr/theme.c
+++ b/module/merintr/theme.c
@@ -76,11 +76,23 @@ bool ThemeSidebarUsesInventoryFill(void)
 }
 /****************************************************************************/
 /*
- * ThemeSkipStatsAreaFrame:  Returns true when the active theme does
- *   not draw the ornamental frame (corners and edge repeaters) around
- *   stats_area (the right-side tabbed panel).
+ * ThemeSkipStatsAreaFrame:  Returns true when the active theme does not want
+ *   an ornamental frame around the stats area.
  */
 bool ThemeSkipStatsAreaFrame(void)
+{
+   switch (ThemeCurrent())
+   {
+   case Theme::Dark: return true;
+   default:          return false;
+   }
+}
+/****************************************************************************/
+/*
+ * ThemeSkipChatBoxFrame:  Returns true when the active theme does not want
+ *   an ornamental frame around the chat box.
+ */
+bool ThemeSkipChatBoxFrame(void)
 {
    switch (ThemeCurrent())
    {
--- a/module/merintr/theme.c
+++ b/module/merintr/theme.c
@@ -76,27 +76,27 @@ bool ThemeSidebarUsesInventoryFill(void)
 }
 /****************************************************************************/
 /*
- * ThemeSkipStatsAreaFrame:  Returns true when the active theme skips the
- *   ornamental frame around the stats area.
+ * ThemeDrawsStatsAreaOrnamentalFrame:  Returns true when the active theme draws
+ *   its ornamental frame around the stats area.
  */
-bool ThemeSkipStatsAreaFrame(void)
+bool ThemeDrawsStatsAreaOrnamentalFrame(void)
 {
    switch (ThemeCurrent())
    {
-   case Theme::Dark: return true;
-   default:          return false;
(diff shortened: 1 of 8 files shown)


module/merintr/theme.c
```diff
@@ -76,9 +76,8 @@ bool ThemeSidebarUsesInventoryFill(void)
 }
 /****************************************************************************/
 /*
- * ThemeSkipStatsAreaFrame:  Returns true when the active theme does
- *   not draw the ornamental frame (corners and edge repeaters) around
- *   stats_area (the right-side tabbed panel).
+ * ThemeSkipStatsAreaFrame:  Returns true when the active theme skips the
```
akirmse: Rather than true meaning "don't draw", it would be more intuitive if true meant "draw" and you flipped the values.  One less negative to keep track of in your head.

module/merintr/theme.c
```diff
@@ -88,3 +87,16 @@ bool ThemeSkipStatsAreaFrame(void)
    default:          return false;
    }
 }
+/****************************************************************************/
+/*
+ * ThemeSkipChatBoxFrame:  Returns true when the active theme skips the
```
akirmse: here too

module/merintr/theme.c
adrienlaws: Fixed - flipped both to positive and adjusted the values. I also added Ornamental to the names since that is the codebase term for the bitmap frame. We still draw a frame when the value is false (a plain border in place of the bitmap ornament) so I figured the distinction might be helpful.

I also updated the PR description to match this change.  Seems to build and work ok.

module/merintr/theme.c
adrienlaws: Fixed - same as other comment.

PR #1443, opened 2026-05-31
Fix password reset for emails containing underscore

sicoderealm:
- Use | as the email/account separator in lookup_username values. The previous _ separator was ambiguous because _ is a legal character in email local-parts; | is illegal in emails, so the split is unambiguous. Updates both the encode site (m59utils.py) and the decode site (reset_password/app.py).
 - Fix missing success page on password reset submit. do_success() was redirecting back to /password-reset (the form page itself), so users got no feedback after submitting. Now redirects to /password-reset-success, matching the existing failure-page pattern and the sibling verify_reset Lambda.

### DIFF
--- a/aws/src/m59utils/python/lib/python3.12/site-packages/m59utils.py
+++ b/aws/src/m59utils/python/lib/python3.12/site-packages/m59utils.py
@@ -44,7 +44,7 @@ def util_load_account_db(lookup_email, lookup_username, db_check, db_check_lates
                 try:
 
                     username_key = f"{parts[0].lower()}_{parts[2]}"
-                    lookup_username[username_key] = f"{parts[1]}_{parts[3]}"
+                    lookup_username[username_key] = f"{parts[1]}|{parts[3]}"
 
                     email_key = f"{parts[1].lower()}_{parts[2]}"
                     lookup_email[email_key] = 1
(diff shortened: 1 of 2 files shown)


aws/src/reset_password/app.py
akirmse, line 100: replace underscore here?

aws/src/m59utils/python/lib/python3.12/site-packages/m59utils.py
```diff
@@ -44,7 +44,7 @@ def util_load_account_db(lookup_email, lookup_username, db_check, db_check_lates
                 try:
 
                     username_key = f"{parts[0].lower()}_{parts[2]}"
-                    lookup_username[username_key] = f"{parts[1]}_{parts[3]}"
+                    lookup_username[username_key] = f"{parts[1]}|{parts[3]}"
 
                     email_key = f"{parts[1].lower()}_{parts[2]}"
```
akirmse, line 49: is this underscore still OK?

aws/src/reset_password/app.py
sicoderealm: usernames are not allowed to have underscores in our validation so this one is fine

aws/src/m59utils/python/lib/python3.12/site-packages/m59utils.py
sicoderealm: yes this is for looking up the email associated with an account username and server number - so this one is also fine

PR #1445, opened 2026-06-02
feature: add custom stat bars and enable for dark theme (theming part 6)

adrienlaws:
## What

- Adds the capability to draw custom stat bars in addition to the default stat bars
    - Examples: health, mana, vigor, character stats (might, stamina, etc)
  - Adds this capability to the dark theme
- Fixes a rare bug where a low-vigor bar could lose its red warning color after a theme switch
- The default theme and default stat bars are unchanged in appearance
- Net effects:
    - Under the dark theme the stat bars now draw as a flat rounded bar with a gradient fill and a clean outline instead of the original rectangular bar inside an ornamental gold frame
    - The health, mana, and vigor bars above the map draw in per-stat colors: red for health, blue for mana, gold for vigor
    - The character stat bars in the stats tab (might, intellect, stamina, and so on) draw in green
    - The headroom between the value and the maximum draws in the stat's own color, darkened, so the value edge stays easy to find
    - The bar background past the maximum is a very dark grey, and the gold ornament frame around each bar is skipped
    - The filled portion has a rounded right end rather than a hard vertical cut
    - The value sits at the right edge of the bar instead of following the filled portion
    - The custom stat bars are taller to make up for the lack of the default gold ornamental border
    - Gradient effect added to custom stat bars

## Why

- Gives themes a new capability
- The default stat bar drawing clashed with dark theme and would likely clash with future themes' color schemes
- Readability
    - The M59 player population is getting older so more visibility can be helpful to an aging cohort
    - Having independent colors for health, mana, and vigor makes it easier to quickly discern a particular bar
    - The custom bars drop the ornamental frame, so a small height increase keeps them as visually substantial as the framed default bars

## How

- There are 3 elements to this change:
    - 1/ The infrastructure to allow a theme to enable custom stat bars
    - 2/ Changes to stat bar drawing logic to allow decisions on which type of bars to draw (classic or custom)
    - 3/ The implementation of dark theme using custom stat bars

#### Files touched

- `clientd3d/graphctl.c`: shared render flow with custom rounded gradient bars, plus the `GraphCtlFillSlice` and `GraphCtlDrawNumber` helpers.  The custom render rounds the right end of each slice and reserves space at the right of the bar for the value
- `clientd3d/color.c`: `ThemeUsesCustomStatBars` capability switch; dark `COLOR_BAR1`/`COLOR_BAR2`/`COLOR_BAR3`; seven entries in both color tables (six per-stat colors plus a stat-bar frame color); `THEME_COLOR_VERSION` bump
- `clientd3d/color.h`: seven `COLOR_*` enum entries (per-stat plus `COLOR_STATBARFRAME`); export `ThemeUsesCustomStatBars`
- `clientd3d/graphctl.h`: the `GCS_CUSTOM` style flag
- `clientd3d/client.def`: export `ThemeUsesCustomStatBars`
- `docs/themes.md`: a stat bars section covering the capability and the value placement
- `clientd3d/makefile`: link `msimg32.lib` for `GradientFill`
- `module/merintr/statmain.c`: per-stat fill and limit color helpers, frame color, taller bar, state-based vigor emergency color
- `module/merintr/stats.c`: `StatsBarHeight`, frame color for the character stat bars
- `module/merintr/stats.h`: `STATS_BAR_HEIGHT_CUSTOM`, `StatsBarHeight` declaration
- `module/merintr/statnum.c`: taller character stat bars
- `module/merintr/drawint.c`: skip the ornament bar frame for themes using custom stat bars
- `clientd3d/preferences.c`: relayout the stat bars on a theme change so the custom height takes effect

### The graph control render, before and after

Before this change, `GraphCtlPaint` drew every bar one way: the background, the value and limit slices through `FillRect`, a rectangular frame, the number, and the slider handle for slider graphs.  There was no custom branch.

```mermaid
flowchart TD
    P[GraphCtlPaint] --> G[Compute geometry]
    G --> T[FillRect background]
    T --> V[FillRect value slice]
    V --> L[FillRect limit slice]
    L --> F[Rectangle frame]
    F --> N[Plain number]
    N --> S[Slider handle for slider graphs]
    S --> Z[End]
```

After this change, `GraphCtlPaint` computes one `custom` flag, true only when the control sets `GCS_CUSTOM` and the theme draws custom stat bars, then walks the same spine.  The geometry is computed once.  Four steps fork on `custom`: the background, the limit and value slices (through `GraphCtlFillSlice`), the frame, and the number (through `GraphCtlDrawNumber`).  The slider path is untouched.

A custom bar paints its limit slice first, running from the left, then paints the value over it.  That way both slices can have a rounded right end instead of the value being cut square where the limit begins.

```mermaid
flowchart TD
    P[GraphCtlPaint] --> G[Compute geometry]
    G --> C{custom?}
    C -->|yes| T1[RoundRect background]
    C -->|no| T2[Inset FillRect background]
    T1 --> S1[Gradient limit then value<br/>rounded right end]
    T2 --> S2[Flat value then limit]
    S1 --> F1[RoundRect frame]
    S2 --> F2[Rectangle frame]
    F1 --> N1[Number at the right edge]
    F2 --> N2[Number at the value]
    N1 --> J[Slider handle for slider graphs]
    N2 --> J
    J --> Z[End]
```

#### StatsBarSetColors

Both the main stat bars (`statmain.c`) and the character stat bars (`stats.c`) now apply their colors through one helper, `StatsBarSetColors`, which sends the four graph-control colors.  The caller passes the fill and limit, and the helper supplies the shared background (`COLOR_BAR3`) and outline.  This keeps the background color and the outline-reset logic in one place instead of duplicated across the two files.

##### Example of Drawing the Health Bar
For a main bar the path is:

```text
StatsMainSetColor(s)
   -> StatsBarSetColors( s->hControl,
                         StatBarFillColor(s),    // per-stat fill  (e.g. health = red)
                         StatBarLimitColor(s) )  // per-stat limit (e.g. health = dim red)
         -> sets BAR, LIMITBAR, BKGND (COLOR_BAR3), FRAME (custom outline or -1)
```

The character stat bars take the same path from `StatsChangeColor`, passing `COLOR_BAR1` / `COLOR_BAR2` instead of per-stat colors.

## Notes
- The dark theme's bar colors are taken from [Bootstrap](https://github.com/twbs/bootstrap), which is the most widely used open source UI project of its kind and is MIT licensed. 
- The value sits at the right edge of the bar rather than following the filled portion.  A number that moves with the fill has to stay readable against both the filled and empty parts of the bar, which no single text color does well
- Vigor bar coloring when low
    - When vigor drops below ten, the filled value turns red in either theme
    - The headroom past the value keeps its normal color: dim gold in dark theme, dark red in default theme

## Example (Custom vs Default Stat Bars)
| Dark theme w/ custom stat bars | Default theme |
| :---: | :---: |
| Dark theme custom stat bars | Default theme stat bars |


akirmse:
Why do we need two different versions of stat bars?  Should we just always use whichever one is better (or improve the existing one)?

adrienlaws:
Fair question regarding why we would have two different versions of the stat bars.

Throughout the theme system implementation I hoped to preserve the original (now called `default theme`) client.

The thought was that since the game has been around so long some players might balk at a redesign of the interface.  Also some people feeling nostalgic might want to see what the game looked like "before" or "originally."  I realize there were a lot of interfaces that existed before what we're calling `default theme` now, but the intent was to preserve some of the history.

Being able to swap between themes live provides a really neat contrast (IMO) and lets a player choose how they want to experience the game.

Monkey Island, Halo, and Age of Empires allow you to do similar switches.

I realize it is an antipattern to create multiple "things" when one will suffice and it does create additional complexity and maintenance burden.  I figured in this case it might be forgivable to attract both old and new players.

Anyway, that was my thinking, but if it doesn't make sense I can instead modify this PR to use the modern stat bars across all themes, with different colors between themes.  Perhaps in that case, using gold trimmed stat bars with green coloring for the `default theme.`

akirmse:
These new stat bars appear just different to me, not "dark".  The gray bars have less contrast and thus are harder to see.  So I see it less as theming, and more a wholesale replacement of the stat bars.  Maybe ask around if people want this change?

adrienlaws:
I made some tweaks based on the feedback here and elsewhere.  This PR description and the screenshots have been updated.  

In regards to feedback the contrast one made a lot of sense so I switched dark theme stat bars from dark grey to dark green. I also tried to find an existing UI/UX framework to reference since I am not an expert in the area.

[Bootstrap](https://github.com/twbs/bootstrap) with 175k+ GitHub stars seemed like a good reference candidate to try and incorporate best practices.

I also started a [Feedback Request: Stat bar dark theme proposal Post](https://discord.com/channels/650824489885630475/1525592964938731671) in the official Meridian 59 Discord.  Generally people seemed interested.

I realize that there will be a lot of opinions on what is the "best" way.  For example, one player has already posted that maybe they could keep the grey stat bars (which don't have high contrast), likely because they match a grey theme.

My current hope / strategy is to:
- 1/ add capability for theming
- 2/ demonstrate "what's possible" in this starter Dark theme
- 3/ eventually empower the community to create their own themes as many will probably be better at it than me and they might make some neat ones

PR #1449, opened 2026-06-10
Harden table API against NULL resource table to fix post-download crash

sicoderealm:
After an update download, some clients crashed at `table.c:91` via `DownloadFiles → FreeResources → table_destroy`.

PR #1314 added a `FreeResources()` call to the download path, but there source table `t` is NULL until `LoadResources()` runs at game start, and the `table_*` functions dereferenced `t->size` with no NULL check.

**Fix:** NULL-guard `table_destroy` and `table_delete`. Each returns its existing empty/not-found result, so callers are unaffected on a valid table and a NULL table becomes a safe no-op instead of a crash.

### DIFF
--- a/clientd3d/table.c
+++ b/clientd3d/table.c
@@ -88,6 +88,9 @@ void table_destroy(Table *t, TableDestroyProc destructor)
    DWORD i;
    list_type l;
 
+   if (t == NULL)
+      return;
+
    for (i=0; i < t->size; i++)
    {
       for (l = t->entries[i]; l != NULL; l = l->next)
--- a/clientd3d/table.c
+++ b/clientd3d/table.c
@@ -49,8 +49,6 @@ Table *table_create(DWORD size)
  */
 int table_insert(Table *t, void *data, HashProc hasher, CompareProc compare)
 {
-   if (t == NULL)
-      return 1;
 
    int hashval = (*hasher)(data, t->size);
    Entry p = t->entries[hashval];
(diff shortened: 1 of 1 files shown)


clientd3d/table.c
```diff
@@ -49,6 +49,9 @@ Table *table_create(DWORD size)
  */
 int table_insert(Table *t, void *data, HashProc hasher, CompareProc compare)
 {
+   if (t == NULL)
```
akirmse: Revert this--giving false information is far worse than crashing.

clientd3d/table.c
```diff
@@ -71,6 +74,9 @@ int table_insert(Table *t, void *data, HashProc hasher, CompareProc compare)
  */
 void *table_lookup(Table *t, void *data, HashProc hasher, CompareProc compare)
 {
+   if (t == NULL)
```
akirmse: Revert

clientd3d/table.c
```diff
@@ -118,8 +130,10 @@ void table_delete(Table *t)
  */
 void table_delete_item(Table *t, void *data, HashProc hasher, CompareProc compare)
 {
-   DWORD hashval = (*hasher)(data, t->size);
+   if (t == NULL)
```
akirmse: I'd say revert this one too--the caller shouldn't be doing this on a null table.

clientd3d/table.c
```diff
@@ -88,6 +94,9 @@ void table_destroy(Table *t, TableDestroyProc destructor)
    DWORD i;
    list_type l;
 
+   if (t == NULL)
```
akirmse: nullptr

clientd3d/table.c
```diff
@@ -108,6 +117,9 @@ void table_delete(Table *t)
 {
    DWORD i;
 
+   if (t == NULL)
```
akirmse: nullptr

PR #1455, opened 2026-06-16
Fix avatar layering in D3D renderer

sicoderealm:
Co-located avatars no longer interleave. Objects sharing a tile each get a per-frame depthBin that shifts their whole z-bias band (depthBin * DEPTH_BIN_STRIDE), applied in both body and overlay passes. Overlay-carrying objects also draw at base depth to avoid z-fighting their own underlays.

Before
overlays-npc
After
overlays-npc-fixed

### DIFF
--- a/clientd3d/d3drender_objects.c
+++ b/clientd3d/d3drender_objects.c
@@ -114,6 +114,37 @@ long D3DRenderObjects(
 	const auto room = objectsRenderParams.room;
 	const auto params = objectsRenderParams.params;
 
+	// Assign every visible object a per-tile "stack bin". Objects that share a tile (same motion x/y)
+	// would otherwise be drawn with identical z-bias bands, causing their bodies and overlays to
+	// interleave (one avatar's head appearing over another's body). The body and overlay passes each
+	// run as separate loops in different orders, so we compute the bins once here and stash the result
+	// on the node (depthBin) for both passes to read. We sort by object id first so the front/back
+	// assignment is stable frame-to-frame and doesn't flicker.
+	{
+		const auto* localPlayer = GetPlayerInfo();
+		std::vector<room_contents_node*> stackNodes;
+		std::unordered_set<int> seenIds;
+		for (long i = 0; i < gameObjectDataParams.numItems; i++)
+		{
+			if (gameObjectDataParams.drawData[i].type != DrawObjectType)
+				continue;
+			room_contents_node* node = gameObjectDataParams.drawData[i].u.object.object->draw.obj;
+			if (node == NULL || node->obj.id == localPlayer->id)
+				continue;
+			node->depthBin = 0;
+			if (seenIds.insert(node->obj.id).second)
+				stackNodes.push_back(node);
+		}
+		std::sort(stackNodes.begin(), stackNodes.end(),
+			[](const room_contents_node* a, const room_contents_node* b) { return a->obj.id < b->obj.id; });
+		std::unordered_map<int64, int> tileCount;
+		for (room_contents_node* node : stackNodes)
+		{
+			int64 key = ((int64)node->motion.x << 32) | (int)(node->motion.y & 0xFFFFFFFF);
+			node->depthBin = tileCount[key]++;
+		}
+	}
+
 	if (config.draw_names)
 	{
 		IDirect3DDevice9_SetVertexShader(gpD3DDevice, NULL);
@@ -967,7 +998,9 @@ void D3DRenderOverlaysDraw(
 					pChunk->numPrimitives = pChunk->numVertices - 2;
 					pChunk->xLat0 = xLat0;
 					pChunk->xLat1 = xLat1;
-					pChunk->zBias = zBias;
+					// Apply the same per-object band shift used for the body (D3DRenderObjectsDraw) so this
+					// overlay stays grouped with its own avatar and doesn't interleave with a co-located one.
+					pChunk->zBias = zBias + (BYTE)(pRNode->depthBin * DEPTH_BIN_STRIDE);
 
 					zBias++;
 
@@ -1367,12 +1400,6 @@ void D3DRenderObjectsDraw(
 
 	auto drawdata = gameObjectDataParams.drawData;
 
-	// We track all objects that are in similar positions in the 3D world.
-	// This is to mitigate z-fighting by incrementing z-depths for each unique object in the same position.
-	// The key is composed of the x and y coordinates of the object and the value is the current
-	// count of objects found at that location.
-	std::unordered_map<int64, int> depth_adjustment_map;
-
 	// As we receive objects in different orders from the BSP walk this can cause inconsistent z-depth ordering.
 	// We now attempt to maintain a consistent view by sorting draw data by their ids.
 	// For translucent objects, the caller has already sorted by distance (back-to-front)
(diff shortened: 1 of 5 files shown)


clientd3d/object.h
```diff
@@ -106,6 +106,9 @@ typedef struct {
    int         distance;          /* Distance from player to object; may not be valid */
    bool        visible;           // True when object is visible in current frame
    int			boundingHeightAdjust;	// adjustment in height from overlays
+   int         depthBin;          // Per-frame stack index among objects sharing this object's tile.
+                                   // Used to give co-located avatars non-overlapping z-bias bands so
+                                   // their bodies/overlays don't interleave. Computed in D3DRenderObjects.
```
akirmse: Do not document the name of the file in which this is computed; it's not relevant and will become wrong over time.

But if this is computed in D3DRenderObjects, i.e. the hardware renderer and not software, what is it doing in the definition of every object?

clientd3d/d3drender_objects.c
```diff
@@ -1564,16 +1591,20 @@ void D3DRenderObjectsDraw(
...
-		pChunk->zBias = (pRNode->obj.flags & OF_PLAYER) || (pRNode->boundingHeightAdjust != 0) ? ZBIAS_BASE : ZBIAS_DEFAULT;
-
-
-		// All objects are drawn at the default depth offset by the number of items already drawn at this location.
-		// Combine objects x and y position into a single int64 for the map key.
-		int64 key = ((int64)pRNode->motion.x << 32) | (int)(pRNode->motion.y & 0xFFFFFFFF);
-
-		// Increment the counter at the appropriate bin and assign the appropriate zBias.
-		pChunk->zBias += (BYTE)depth_adjustment_map[key]++;
+		// Any object that carries overlays (e.g. an NPC soldier with armor, shield and weapon) is the same kind of
```
akirmse: Do not reference particular Blakod class names in the renderer.

clientd3d/d3drender_objects.c
```diff
@@ -967,7 +998,9 @@ void D3DRenderOverlaysDraw(
 					pChunk->numPrimitives = pChunk->numVertices - 2;
 					pChunk->xLat0 = xLat0;
 					pChunk->xLat1 = xLat1;
-					pChunk->zBias = zBias;
+					// Apply the same per-object band shift used for the body (D3DRenderObjectsDraw) so this
```
akirmse: Body?  Avatar?  What are those?

clientd3d/d3drender_objects.c
```diff
@@ -114,6 +114,37 @@ long D3DRenderObjects(
...
+			node->depthBin = 0;
+			if (seenIds.insert(node->obj.id).second)
+				stackNodes.push_back(node);
+		}
+		std::sort(stackNodes.begin(), stackNodes.end(),
+			[](const room_contents_node* a, const room_contents_node* b) { return a->obj.id < b->obj.id; });
+		std::unordered_map<int64, int> tileCount;
+		for (room_contents_node* node : stackNodes)
+		{
+			int64 key = ((int64)node->motion.x << 32) | (int)(node->motion.y & 0xFFFFFFFF);
```
akirmse: This is rather mysterious and deserves a comment.

clientd3d/object.h
sicoderealm: Removed the "Computed in D3DRenderObjects" reference. Comment now says it's hardware-renderer scratch, recomputed each frame, ignored by the software renderer - no file/function name.

clientd3d/d3drender_objects.c
sicoderealm: Dropped "NPC soldier with armor, shield and weapon" (and the "Tos Fountain" example). Now reads generically: "Any object that carries overlays."

clientd3d/d3drender_objects.c
sicoderealm: Standardized the jargon everywhere: body → "main sprite", avatar → "object". Applied to object.h, d3drender.h (the layer-band comment), and all three comment blocks in d3drender_objects.c.

clientd3d/d3drender_objects.c
sicoderealm: Added above tileCount: each object's bin = count of objects already at its location (so co-located objects get 0,1,2…), and the x/y position is packed into a single int64 to serve as the map key.

clientd3d/object.h
```diff
@@ -106,6 +106,10 @@ typedef struct {
    int         distance;          /* Distance from player to object; may not be valid */
    bool        visible;           // True when object is visible in current frame
    int			boundingHeightAdjust;	// adjustment in height from overlays
+   int         depthBin;          // Hardware-renderer scratch: per-frame stack index among objects
```
akirmse: The main object structure doesn't seem like a great place for per-frame, hardware-renderer-specific scratch memory.  See what I'm saying?

sicoderealm: Good call - moved it out. The depth bins now live in a renderer-local map in d3drender_objects.c, keyed by object id, so room_contents_node is back to being free of per-frame renderer scratch. Thanks!

clientd3d/d3drender_objects.c
```diff
@@ -114,6 +127,40 @@ long D3DRenderObjects(
 	const auto room = objectsRenderParams.room;
 	const auto params = objectsRenderParams.params;
 
+	// Assign every visible object a per-tile "stack bin". Objects that share a tile (same motion x/y)
+	// would otherwise be drawn with identical z-bias bands, causing their main sprites and overlays to
+	// interleave (one object's overlay appearing over a different object's sprite). The main-sprite and
+	// overlay passes each run as separate loops in different orders, so we compute the bins once here into
+	// objectDepthBins for both passes to read. We sort by object id first so the front/back assignment is
+	// stable frame-to-frame and doesn't flicker.
+	{
```
akirmse: Ah, the long-awaited return of the needless brace.

Can you review your PRs for this before sending them to me?

sicoderealm: Fair hit. The brace wasn't pointless - it was fencing off four locals so they didn't leak into a rest of the function - but that's exactly the tell that it wanted to be a function. I've pulled it into D3DRenderBuildObjectDepthBins(...) - the brace is gone and D3DRenderObjects reads as a sequence of steps now. Thanks for the nudge.

clientd3d/d3drender_objects.c
```diff
@@ -94,6 +94,52 @@ static void updateRenderChunkAnimationIntensity(d3d_render_chunk_new* pChunk)
...
+	std::sort(stackNodes.begin(), stackNodes.end(),
+		[](const room_contents_node* a, const room_contents_node* b) { return a->obj.id < b->obj.id; });
+	// Each object's bin is the number of objects already placed at its location, so co-located
+	// objects get successive bins (0, 1, 2, ...). Combine the object's x and y position into a
+	// single int64 to use as the per-location map key.
+	std::unordered_map<int64, int> tileCount;
+	for (room_contents_node* node : stackNodes)
+	{
+		int64 key = ((int64)node->motion.x << 32) | (int)(node->motion.y & 0xFFFFFFFF);
+		objectDepthBins[node->obj.id] = tileCount[key]++;
```
akirmse: Is postincrement correct here?  The first time through, you want to put 0 in objectDepthBins?  (This is why relying on ++ side effects is generally frowned upon).

clientd3d/d3drender_objects.c
```diff
@@ -94,6 +94,52 @@ static void updateRenderChunkAnimationIntensity(d3d_render_chunk_new* pChunk)
 
 // Implementations
 
+// Per-frame stack index for each object, keyed by object id. Objects sharing a tile get successive
+// bins so their z-bias bands don't overlap (see DEPTH_BIN_STRIDE). Rebuilt every frame by
+// D3DRenderObjects before the draw passes read it; lives here rather than on room_contents_node so the
+// shared object struct stays free of hardware-renderer scratch.
```
akirmse: "stays free of hardware-renderer scratch" does not make any sense to me.  Why are you documenting what this data structure is NOT?

akirmse (commented):
I think the intention here is good but the change introduces a lot of confusion and is very hard to read.  The names are poor and the AI-generated comments are large yet unhelpful.

clientd3d/d3drender_objects.c
sicoderealm: Fair. Fixed - it no longer relies on ++ side effects. Now it reads the current count (0 on first visit), stores that as the bin, then advances the counter

sicoderealm:
> I think the intention here is good but the change introduces a lot of confusion and is very hard to read. The names are poor and the AI-generated comments are large yet unhelpful.

I've reworked this entire PR - hopefully looking better, thanks for the feedback

PR #1456, opened 2026-06-16
fix: keep shield equipped when swapping weapons

sicoderealm:
## Summary
Swapping one weapon for another while a shield was equipped could unequip the shield instead of the old weapon.

With both hand slots full (shield + one-handed weapon), equipping a new weapon calls `FreeHandSlotsFor` to free a single slot. It walked the held hand items in arbitrary order and yielded the first one willing to come off - which could be the shield rather than the weapon being replaced.

## Fix
`FreeHandSlotsFor` now buckets held hand items by kind (weapon vs. non-weapon) and yields the *same kind* as the item being equipped first, only giving up the other kind if still short on space. Swapping weapons drops the old weapon and keeps the shield; swapping shields drops the old shield; two-handed equips still clear both hands as needed.

## Repro (fixed)
Equip a small round shield + hammer, then switch to an axe - the axe equips and the shield stays on.

Regression from ecf49560 (#1423), which introduced the auto-unequip behavior.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -3438,20 +3438,31 @@ messages:
 
    FreeHandSlotsFor(what = $)
    "Unequips currently-held hand items until there is enough free hand space "
-   "to equip `what`, or until no more held hand items are willing to yield."
+   "to equip `what`, or until no more held hand items are willing to yield. "
+   "Prefers to yield items of the same kind (weapon vs. non-weapon) as `what`, "
+   "so e.g. swapping one weapon for another does not drop an equipped shield."
    {
-      local i, iHandUsed, iNeeded, lHandItems;
+      local i, iHandUsed, iNeeded, lSameKind, lOtherKind, bWantWeapon;
 
       iNeeded = Send(what,@GetItemUseAmount);
+      bWantWeapon = IsClass(what,&Weapon);
 
       iHandUsed = 0;
-      lHandItems = $;
+      lSameKind = $;
+      lOtherKind = $;
       for i in plUsing
       {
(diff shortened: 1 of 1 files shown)


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -3460,7 +3471,21 @@ messages:
          return;
       }
 
-      for i in lHandItems
+      % Yield same-kind items first (e.g. swap weapon for weapon), only
+      % giving up the other kind (e.g. a shield) if still short on space.
+      for i in lSameKind
```
akirmse, line 3476: Isn't a guard against $ needed here?  Do you get server errors when not holding anything?

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -3460,7 +3471,21 @@ messages:
...
+         {
+            iHandUsed = iHandUsed - Send(i,@GetItemUseAmount);
+            if iHandUsed + iNeeded <= viHand_space
+            {
+               return;
+            }
+         }
+      }
+
+      for i in lOtherKind
```
akirmse, line 3488: Here too?

kod/object/active/holder/nomoveon/battler/player.kod
sicoderealm: I believe no guard is needed as the compiler emits if (temp = $) goto end at the top of every for in loop (codegen.c), so iterating a $ list is a no-op, not an error. That's why there are no server errors. The pre-existing code already iterated for i in lHandItems when it was $, so the behavior is unchanged.

kod/object/active/holder/nomoveon/battler/player.kod
sicoderealm: also as above ty

PR #1458, opened 2026-06-18
Better feedback when a kill is too weak to advance

sicoderealm:
Helps players know when a creature is no longer worth fighting:

  - New occasional message on the killing blow when your weapon's proficiency can no longer improve from a creature this weak (fired once at the kill, like the spit-on-corpse message - not every swing).
  - Spit-on-corpse message now fires when the monster is at or below your HP (the point where it can no longer toughen you), instead of the old 5-HP buffer. The small near-level +1 gain bonus is unchanged.
  - Minor comment typo fix in stroke.kod.

  ▎ You feel you've learned all you can about hammer wielding from fighting foes this weak.  You should seek tougher opponents.
  
image

Inspired by a suggestion made by Gar at https://discord.com/channels/650824489885630475/1515569949522006106

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -144,6 +144,12 @@ resources:
    player_improve_maxhealth = "~I~BYou suddenly feel a little tougher."
    player_improve_health_invigorate = "~I~BYou are invigorated by your success."
    player_spits = "You spit on the corpse of your unworthy foe."
+   % Shown after killing a creature whose level is too low to train the
+   % wielded weapon's proficiency any further.  %s is the proficiency name
+   % (e.g. "fencing").
+   player_weapon_prof_capped = \
+      "~IYou feel you've learned all you can about %s from fighting foes this "
+      "weak.  You should seek tougher opponents."
    player_regain_angel = \
       "Due to your weakness, your protective guardian angel returns."
    player_no_angel = "You suddenly feel more vulnerable."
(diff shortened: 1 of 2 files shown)


kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -7709,19 +7690,72 @@ messages:
...
-         
+
          poKill_target=$;
          Send(self,@ResetGainFlags);
       }
       
       propagate;
    }
 
+   NotifyWeaponProfCapped(victim=$)
```
akirmse: A weapon professor?  How do I become one of those?

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -7709,19 +7690,72 @@ messages:
...
+   "Called when this player lands the killing blow.  If the wielded weapon's"
+   "proficiency can no longer improve from a creature this weak, tell the"
```
akirmse: multi-line comments are appended to each other, so there are some missing spaces at the end of start of lines here.

kod/object/active/holder/nomoveon/battler/player.kod
```diff
@@ -7709,19 +7690,72 @@ messages:
...
+
          poKill_target=$;
          Send(self,@ResetGainFlags);
       }
       
       propagate;
    }
 
+   NotifyWeaponProfCapped(victim=$)
+   "Called when this player lands the killing blow.  If the wielded weapon's"
```
akirmse: How do you know when this function is called?  Does it not work if it's called from elsewhere?

akirmse (commented):
I don't have a good sense of whether this is good or bad gameplay-wise.  Does @davidklaw or @Aesica want to weigh in?

davidklaw:
> I don't have a good sense of whether this is good or bad gameplay-wise. Does @davidklaw or @Aesica want to weigh in?

I do like what this offers for players who are unfamiliar with this mechanic. However, I do have one suggestion: reference skills generally rather than by name. My thinking is that that's more in line with the original design sensibilities while still giving players enough to go on. Let the players put two and two together.

**Original:**
> You feel you've learned all you can about hammer wielding from fighting foes this weak. You should seek tougher opponents.

**Suggestion:**
> You feel you've learned all you can from foes this weak. Seek worthier opponents.

**Update:**
After speaking with @sicoderealm, the specificity works for me as the message context is similar to that of "You have improved", where there's precedence for including the skill.

kod/object/active/holder/nomoveon/battler/player.kod
sicoderealm: Good call, updated the comment. ty

PR #1459, opened 2026-06-20
Fix scale for attached POV overlays in hardware renderer.

neiltaco:
This PR fixes the scale of attached POV overlays in hardware rendering.  This currently only affects the spellcasting glow that attaches to the player's hand.

In the current game, the casting glow in hardware rendering mode is much smaller than the one displayed in the original software renderer.  In `d3drender_objects.c`, hotspot-attached POV overlays aren't use the same scaling calculation as the one used for most POV overlays.

To fix this inconsistency, the scaling calculation was moved into a helper function so it can be used by both `D3DComputePlayerOverlayArea()` and `D3DRenderPlayerOverlayOverlaysDraw()`.  The following GIF shows the comparison between the original glow, the current glow, and the updated glow:

hand_glow_comparison

### DIFF
--- a/clientd3d/d3drender_objects.c
+++ b/clientd3d/d3drender_objects.c
@@ -75,10 +75,9 @@ static bool IsInvisibleEffect(int flags) {
 	return (flags & (OF_INVISIBLE | OF_DITHERINVIS)) == OF_INVISIBLE;
 }
 
-// Calculate scaling factors for POV overlays based on current resolution.
-// Shared with D3DComputePlayerOverlayArea() and D3DRenderPlayerOverlayOverlaysDraw()
-// so a base overlay and any overlay attached to its hotspot are aligned.
-static void D3DGetPlayerOverlayScreenScale(const PlayerViewParams& playerViewParams, float& screenW, float& screenH)
+// Computes scaling factors for POV overlays based on current resolution.
+// Then writes the results to *screenW and *screenH.
+static void D3DComputePlayerOverlayScreenScale(const PlayerViewParams& playerViewParams, float* screenW, float* screenH)
 {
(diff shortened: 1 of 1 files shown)


clientd3d/d3drender_objects.c
```diff
@@ -75,6 +75,31 @@ static bool IsInvisibleEffect(int flags) {
 	return (flags & (OF_INVISIBLE | OF_DITHERINVIS)) == OF_INVISIBLE;
 }
 
+// Calculate scaling factors for POV overlays based on current resolution.
+// Shared with D3DComputePlayerOverlayArea() and D3DRenderPlayerOverlayOverlaysDraw()
```
akirmse: Do not document who calls a function!!!!!!!  It is not part of the function's contract and can change anytime.

I have seen this so, so many times in recent PRs.  I think AI loves to add this, but then it is your job to think about what makes sense.

clientd3d/d3drender_objects.c
```diff
@@ -75,6 +75,31 @@ static bool IsInvisibleEffect(int flags) {
...
+// so a base overlay and any overlay attached to its hotspot are aligned.
+static void D3DGetPlayerOverlayScreenScale(const PlayerViewParams& playerViewParams, float& screenW, float& screenH)
```
akirmse: It's more common to pass inout parameters like screenW and screenH as pointers, otherwise at the call site it's not obvious that these are getting filled in.  Also, the comment should document that these are filled in (the function is called "Get" but doesn't return anything -- confusing!)

neiltaco: Done!  I think it's also worth renaming the helper function  to make it extra clear it isn't a getter function.

PR #1465, opened 2026-06-25
Fix slow, distance-dependent projectile speed at close/normal range

sicoderealm:
#1356 changed the projectile speed divisor from /1000 to /10000 to "match MoveObject2", but projectiles are meant to fly ~10× faster than walking objects. This made every projectile 10× slower in the normal (uncapped) case, while the new max-travel-time cap only masked it at long range. The result was crawling arrows up close and an obvious speed-vs-distance regime switch ("goofy" flight, very noticeable next to a monster).

Restores the projectile-specific /1000 divisor so close/normal shots fly at their original speed. The `MAX_PROJECTILE_TRAVEL_TIME` long-range cap is preserved, so far shots still arrive within 2 s - and the cap now only kicks in at genuine long range (distance > 2*speed) instead of speed/5.

Client-only change (ProjectileAdd); no server/KOD rebuild needed.

### DIFF
--- a/clientd3d/moveobj.h
+++ b/clientd3d/moveobj.h
@@ -14,6 +14,13 @@
 
 #define GRAVITY_ACCELERATION (- 5 * FINENESS)   // In FINENESS units / second / second
 
+// Smooth-motion increment is the fraction of the journey covered per millisecond:
+// (speed / period) / distance.  A "speed" value is grid squares per time window.
+// Walking objects measure speed over a 10-second window; projectiles over 1 second,
+// so the same speed value moves a projectile ten times as far per millisecond.
+static const float PROJECTILE_SPEED_PERIOD_MSEC = 1000.0f;
+static const float OBJECT_SPEED_PERIOD_MSEC     = 10.0f * PROJECTILE_SPEED_PERIOD_MSEC;
+
 void MoveObject2(ID object_id, int x, int y, BYTE speed, BOOL turnToFace);
 bool ObjectsMove(int dt);
 bool MoveSingle(Motion *m, int dt);
--- a/clientd3d/moveobj.h
+++ b/clientd3d/moveobj.h
@@ -18,8 +18,8 @@
 // (speed / period) / distance.  Object motion is given in # of grid squares per 10
 // seconds; projectiles are given per 1 second, so the same speed value moves a
 // projectile ten times as far per millisecond.
-static const float PROJECTILE_SPEED_PERIOD_MSEC = 1000.0f;
-static const float OBJECT_SPEED_PERIOD_MSEC     = 10.0f * PROJECTILE_SPEED_PERIOD_MSEC;
+static const float PROJECTILE_INTERVAL_MS = 1000.0f;
+static const float OBJECT_INTERVAL_MS     = 10.0f * PROJECTILE_INTERVAL_MS;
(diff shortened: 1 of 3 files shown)


clientd3d/projectile.c
```diff
@@ -90,11 +90,10 @@ void ProjectileAdd(Projectile *p, ID source_obj, ID dest_obj, BYTE speed, WORD f
...
       distance = sqrtf(dx_f * dx_f + dy_f * dy_f + dz_f * dz_f) / FINENESS;
 
-      // Calculate normal increment based on speed
-      // The server sends speed as "grid squares per 10 seconds", and increment is
-      // "progress per millisecond", so we divide by 10000 (10 seconds = 10000 milliseconds).
-      // This matches the calculation in MoveObject2() in moveobj.c.
-      float normal_increment = ((float) speed) / 10000.0f / distance;
+      // Calculate normal increment based on speed, where increment is "progress per
+      // millisecond". Projectiles fly faster than walking objects, so they use a /1000
+      // divisor rather than the /10000 MoveObject2() uses for object motion.
```
akirmse: If this is supposed to be 10x faster than MoveObject2 (whatever that is), then perhaps the constants should be linked with a *10 multiplier, rather than hard-coded?

sicoderealm: Good call - linked them so the relationship is explicit. PROJECTILE_SPEED_PERIOD_MSEC = 1000 and OBJECT_SPEED_PERIOD_MSEC = 10 * PROJECTILE_SPEED_PERIOD_MSEC now live together in moveobj.h, and both projectile.c and MoveObject2 reference them instead of the raw 1000/10000. Change one, both stay in sync.

clientd3d/moveobj.h
```diff
@@ -14,6 +14,13 @@
 
 #define GRAVITY_ACCELERATION (- 5 * FINENESS)   // In FINENESS units / second / second
 
+// Smooth-motion increment is the fraction of the journey covered per millisecond:
+// (speed / period) / distance.  Object motion is given in # of grid squares per 10
+// seconds; projectiles are given per 1 second, so the same speed value moves a
+// projectile ten times as far per millisecond.
+static const float PROJECTILE_SPEED_PERIOD_MSEC = 1000.0f;
```
akirmse: The naming could use some work here.  You have three different units in a single name: speed, period, and msec.  Can you come up with a name that describes what this means more closely?

Also, I'm a little concerned that such a big problem slipped through your testing when you made the previous speed adjustments.  Was the testing sufficient this time?

sicoderealm: I went with `PROJECTILE_INTERVAL_MS` / `OBJECT_INTERVAL_MS`. Naming this is genuinely tough: the value is just the time window a speed is quoted over - 1 second for projectiles, 10 for objects - so every accurate word for it (period, interval, window) is a near-synonym. If you've got a name you'd prefer, I'm happy to use it.

On the testing - fair question! Worth noting this wasn't a crash: monsters still took damage and arrows were still visible in flight, so nothing was actually broken. The bug only showed up at extremely close range, so it slipped past normal-combat testing where everything looked fine. This time I've put it through its paces from a variety of ranges, point-blank through to long-distance, to make sure it holds up across the board.

PR #1468, opened 2026-07-09
justicar: Add adjustable playtime and ballot cost gates

davidklaw:
The intent of this PR is to make the Justicar election harder to take advantage of.

Players have pointed out that the Justicar election is too easy to cheat, specifically by players with an army of alt accounts. There were a number of ideas shared in several threads in Discord, but I opted for a few simple changes to the voting itself to get things moving in a positive direction. If deployed and no adjustments are made, the only difference will be the new playtime requirement. From there, an administrator can adjust the ballot cost, HP limit for candidates, for voters, and adjust the amount of playtime required to vote and run.

### What changed

* You now need real time played on the character to vote or run. This is perhaps the biggest change and one I think could even be looked at differently, like recent time played. The required amount is adjustable and defaulted to 160 hours (4 x 40hr "work weeks").
* Ballot cost is defaulted to 1,000, but should be adjusted higher if this change ships (e.g., 10,000). The thought is that anyone with an army of voters should feel pain in their pocketbook. This cost is adjustable by an admin. The current cost of a ballot is 16 shillings.
* Voters and candidates have independent, adjustable HP requirements. Both of these thresholds are set to 100 to match the current state.
* Caramo's announcements and the room signage have been updated to speak to these changes.

### Post Patch Notes
* All existing ballots will be deleted.

### DIFF
--- a/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqclerk.kod
+++ b/kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqclerk.kod
@@ -68,10 +68,12 @@ resources:
       "--Caramo"
    barloque_clerk_election_open_subject = "The Election Campaign Hath Begun"
    barloque_clerk_election_open = \
-      "The election for Justicar hath begun.  Thou mayest now buy a ballot from me. "
-      "Write the name of the candidate of thy preference and return it for official"
-      " tally.  Do remember, under current Meridian law, neither new citizens nor "
-      "outlaws can vote or hold office."
+      "The election for Justicar hath begun.  Thou mayest now purchase a ballot from "
+      "me.  Write the name of the candidate of thy preference and return it for "
+      "official tally.  Do remember, under current Meridian law, a citizen must be "
+      "hardy enough and have spent enough time in the lands to vote or hold office, "
+      "by the measure the Crown doth set; neither new citizens nor outlaws can do "
+      "either."
       "\n\n"
       "--Caramo"
    barloque_clerk_tally_day = \
@@ -169,7 +171,9 @@ resources:
 
    barloqueclerk_hands_back = "~bCaramo hands back the ballot."
    barloqueclerk_bad_candidate = "%q is ineligible for the office of Justicar."
-   barloqueclerk_no_newbie_voter = "You are not yet eligible to vote."
+   barloqueclerk_no_newbie_voter = \
+      "You are not yet eligible to vote.  By the Crown's present measure, thou must "
+      "be hardier and have spent more time in the lands ere thou may cast a ballot."
    barloqueclerk_no_outlaw_voter = "Criminals have no voting rights.  Begone."
 
    barloque_clerk_notify_ad = \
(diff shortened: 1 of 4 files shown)


pauly-barns:
From one players perspective; this is a perfect mix of prior suggestions and dynamism.  Excellent!

kod/object/active/holder/nomoveon/battler/monster/towns/barlqtwn/bqclerk.kod
```diff
@@ -1376,12 +1395,28 @@ messages:
...
    {
-      return Send(who, @GetBaseMaxHealth) >= 
-             Send(Send(SYS, @GetSettings), @GetMinHPForJusticar);
+      if Send(who,@GetBaseMaxHealth) < Send(Send(SYS,@GetSettings),@GetMinHPForJusticarVoter)
+      {
+         return FALSE;
+      }
+
+      return Send(who,@GetTimeLoggedIn) >=
+             Send(Send(SYS,@GetSettings),@GetMinHoursOnlineForJusticar) * 3600;
```
sicoderealm, line 1407: Nice changes - curious why you went for time online over HP? They are related but time online is easy to obtain while HP requires active engagement. I’d go for adjustable HP threshold

davidklaw: HP is still a factor, but it's a separate check above this one. I should combine with this logic to make that more obvious.

Same below for candidates, who have their own HP threshold.

sicoderealm: Oh nice all good then - adjustable HP threshold is enough imo with ballet increase but either way all good :) thanks

pauly-barns: > Oh nice all good then - adjustable HP threshold is enough imo with ballet increase but either way all good :) thanks

The most common scenario is players with access to dozens of “built” accounts handed down from former players who have quit over years past logging them all in, casting a vote, and logging them back off.

Since these will already be past any reasonable HP requirement, the time played and ballot cost are the most critical pieces here.

sicoderealm: Understood - so the time played check here for 140 hours is for the total time or recent time played? It would be a little mean to ban people from voting who haven’t played too much recently

davidklaw: The time requirement is **total time logged in** across all sessions.

As such, it won't catch your example @pauly-barns. Built accounts will pass both the HP _and_ time requirements, at which point ballot cost becomes the only deterrent.

pauly-barns: Ah, understood.  Even a base of something relatively low like your default of 1000 that can be adjusted should still move the needle in the right direction then.

PR #1484, opened 2026-08-14
Add weapon cast animation

neiltaco:
This PR adds spellcasting animations while wielding weapons.  It uses the first three bitmap groups of each melee weapon's POV overlay to show the casting swing effect.  This PR also plans to add hotspots to those three frames so that the casting glow can be attached to them.

This PR also allows the casting hand to be shown along with the bow.  This is done by changing the bow's hotspot definition to PWO_LEFT_HAND, and the only other item that uses that hotspot definition is a shield, which can't be used along with bows anyways.

I asked Claude on how to address a crash that occurs in hardware rendering if at least one bitmap group is missing a hotspot while the rest of the groups in the animation have hotspots.  It suggests that in the definition for D3DRenderPlayerOverlayOverlaysDraw(), I can replace assert(0) with continue, which appears to skip the setup for rendering overlays that attach to other overlays.  Software rendering meanwhile appears to play the animation fine without crashing.

The .bbg files were updated to include the new hotspots near the weapon hilt areas for the casting glow, and the new .bgf files are ready to be sent in a .zip file.  Currently there is a minor graphical bug that occurs once when a player casts their first spell in the session using a melee weapon.  I consider it a minor bug since subsequent casts doesn't show it for the rest of the game session.

weaponcasttest1

### DIFF
--- a/clientd3d/d3drender_objects.c
+++ b/clientd3d/d3drender_objects.c
@@ -2292,9 +2292,9 @@ void D3DRenderPlayerOverlayOverlaysDraw(
 				}
 				else
 				{
-					assert(0);
-					// the hotspot wasn't in the base object pdib
-					// must be an overlay on an overlay, so find base overlay
+					// The hotspot wasn't found on this group so it is skipped instead.
+					// Used for animations where not every frame has a hotspot defined.
+					continue;
 				}
 			}
 
--- a/kod/object/active/holder/nomoveon/battler/player.kod
+++ b/kod/object/active/holder/nomoveon/battler/player.kod
@@ -8559,30 +8559,33 @@ messages:
          AddPacket(1,HS_SE);
          AddPacket(4,PWO_RIGHT_HAND);
 
-		 oWeapon = Send(self,@LookupPlayerWeapon);	 
-		 if oWeapon <> $
-		 {
+		 oWeapon = Send(self,@LookupPlayerWeapon);
+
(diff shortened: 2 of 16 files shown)


neiltaco:
I tested out the updated .bbg files by using makebgf.exe to make the new .bgf files.  The magic glow attaches properly to the new hotspots in-game, and I've posted a video of all the new melee casting animations in the discord feedback forum.

However, this PR shows a minor graphical glitch that occurs when the player casts their first spell in the current game session using a melee weapon.  There's about a frame or two where the magic glow properly follows the hotspot but doesn't swap to the next bitmap group in time.  This also cause the casting hand to also show this glitch at first.  On subsequent casts, the glow then displays normally for both the hand and the melee weapon.


https://github.com/user-attachments/assets/9c109ac5-b557-4202-8c4c-7bc67b79d402



If this is still acceptable, I have the .bgf files in a zip file ready to be send on Discord.  For now I plan to address this bug either in this PR or in a separate PR.

neiltaco:
After some testing, I believe this PR is ready for review.  The desync glitch mentioned earlier doesn't seem to be caused by this PR and is actually be a rare pre-existing bug.  I believe that bug may need to be addressed in a separate PR to keep the scope of this PR short.

I'm also not fully sure if I should include the new .bgf files in this PR, but they can be added in if needed.

PR #1487, opened 2026-08-17
Fix No Look Through walls in the hardware renderer

sicoderealm:
"No Look Through" on a transparent wall means: show the background through the holes, but not the room on the other side. The hardware renderer does not do this, so the room beyond shows through.

The old mask code built geometry every frame and discarded it undrawn, so none of it had run for years, and it was aimed at painting the masked area with a solid texture. That cannot blend a translucent wall against the background, and it needs the mask drawn topmost, which fights the depth buffer.

D3DRenderNoLookThroughMask is a simpler replacement: one depth-only pass between the sky and the world geometry, colour writes off. Sky already in the frame buffer survives, anything further away fails the depth test, and translucent walls blend against the background for free. 

Tested across a range of rooms locally and on live.

### DIFF
--- a/clientd3d/d3drender.c
+++ b/clientd3d/d3drender.c
@@ -492,6 +492,16 @@ HRESULT D3DRenderInit(HWND hWnd)
 	gpD3DDevice->CreateTexture(gFullTextureSize, gFullTextureSize, 1, D3DUSAGE_RENDERTARGET, D3DFMT_A8R8G8B8,
 									D3DPOOL_DEFAULT, &gpBackBufferTexFull, nullptr);
 
+	// A wall texture set not to tile vertically leaves a gap above it, which the no look
+	// through pass fills with a quad of its own. That quad has no texture to be drawn with,
+	// so it borrows this one; the pass writes depth only, so this texel is never seen.
+	gpD3DDevice->CreateTexture(1, 1, 1, 0, D3DFMT_A8R8G8B8, D3DPOOL_MANAGED, &gpNoLookThrough, nullptr);
+
+	D3DLOCKED_RECT lockedRect;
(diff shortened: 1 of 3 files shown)


clientd3d/d3drender.c
```diff
@@ -719,17 +723,16 @@ void D3DRenderBegin(room_type *room, Draw3DParams *params)
 
 	// Prepare our rendering parameters
 	WorldCacheSystemParams worldCacheSystemParams(&gWorldCacheSystem, &gWorldCacheSystemStatic,
-		&gLMapCacheSystem, &gLMapCacheSystemMoving, &gLMapCacheSystemStatic, &gWallMaskCacheSystem);
+		&gLMapCacheSystem, &gLMapCacheSystemStatic, &gWallMaskCacheSystem);
 
 	WorldPoolParams worldPoolParams(&gWorldPool, &gWorldPoolStatic, &gLMapPool, &gLMapPoolStatic, &gWallMaskPool);
```
akirmse, line 725: Looks like maybe D3DRenderLightsGetOrange is now no longer used?

akirmse (commented):
This seems to change a lot of unrelated lighting code---just double checking that you didn't accidentally include some changes not related to no-lookthrough here.

clientd3d/d3drender.c
sicoderealm: I need to follow up to completely remove `D3DRenderLightsGetOrange` - I didn't want to pollute this PR any more

sicoderealm:
> This seems to change a lot of unrelated lighting code---just double checking that you didn't accidentally include some changes not related to no-lookthrough here.

Thank you for checking, we are good. I will send a follow up to clean up that orange light stuff. Good to merge as is

Issue #222, opened 2013-08-13
Safety On&Off


Pkers:
Hi!

I really think it would facilitate if safety on and off was individualized on one single character.
Now if I for example turns safety off on a character, the other one on the same account will aswell, automatecly have it's safety off too. Which means I can accidently hit other players and get orange, which are not appreciated and often lead to suicide.

//Martin

davidklaw:
If this is a common issue but the implementation of character specific settings is a big investment, perhaps updating the following onboarding messages is an option:

> Your safety is now ON: You can no longer strike innocents.
>Your safety is now **OFF**: Be careful, you are now able to hurt those around you.

To something like:

> Your safety is now **ON**: You can no longer strike innocents. (type "safetyoff" to disable)
>Your safety is now **OFF**: Be careful, you are now able to hurt those around you. (type "safetyon" to enable)

jrotunda:
Suggest to close this.

This is missing some perspective on other player types - as a former player who enjoyed the management of each character, and doesn't like "automatic", this would get weird.

This is more for a player scared of slipping snd turning orange - which is its own meta to meridian survival.

Also if you did turn orange - so what? This may be a necessary first introduction to the games social politics.

Dying needs to be experienced as well - and overcame or you're going to have a very stressful time enjoying the game. A player can always quietly kill themselves to go white too. There are options and they add something intangible but necessary.

Every character should have their own specific safety setting, and settings managed, this is how players now expect separate various identities and character types and play styles. 

Also the Safety On:Off text command is broken - it's probably where the passion for this Issue came from, since it doesn't save. It currently does nothing. @adrienlaws

PR #1500, opened 2026-09-09
fix: safety not turning on after switching characters

adrienlaws:
## What

- Record the safety value in `pinfo.aggressive` every time the client sends one to the server, so the two stay synchronized 

## Why

- Multiple reports of players switching characters under the impression that their safety is set one way, when it is actually the reverse, leading them to become orange (and sometimes red) by accident.
- The client keeps two copies of the safety, and they do not live the same length of time
  - `config.aggressive` is the saved preference, and it survives a quit to the character select screen
  - `pinfo.aggressive` is the client's note of what it last told the server, and `ModulesClose` wipes it on logoff
  - so the second character starts with the two already disagreeing
- Three of the places that send the safety never wrote down what they sent
  - `InterfaceInit` and both mismatch branches of `InterfaceUserChanged` called `SendSafety` and left the note alone
  - `InterfaceConfigChanged` sends only when the setting differs from that note, so a stale note makes a real change look like no change
- You hear nothing when it happens
  - the server plays the sound when it receives a safety change, so silence means the client sent nothing

## How

- Added `SendConfiguredSafety`, which sends the setting and writes it down in one place
- Replaced the three sends that did not
  - the two that already did, in `InterfaceConfigChanged` and the `safetyon` and `safetyoff` commands, are untouched
- No server change

## The values involved

Only one of these is the actual safety.  Two of the others are named against it.

| Value | Where | What it holds | Set means |
|---|---|---|---|
| `PFLAG_SAFETY` | Server, per character | The real safety, and the only value that decides whether an attack lands | This character cannot attack innocents |
| `config.aggressive` | Client, saved to the INI | The Preferences box, shared by every character on the installation | Safety off.  The box reads "Can attack innocent players", so it is inverted against the game term |
| `pinfo.aggressive` | Client, memory only | The last value the client sent.  Its only reader is the check for whether a Preferences edit is a real change | The server was last told safety off |
| `safety_flipped` | Client, memory only | Not a safety state.  A note that a correction has already been sent, so the client does not resend in a loop | Do not send another correction |

Only `UC_SAFETY` crosses the wire.  Everything else is the client talking to itself.

## Example

Both characters on the account start with the safety on.  The client stays open throughout, and step 2 is a quit to character select rather than a log off.

1. Log on to the first character and turn the safety off
2. Quit to the character select screen
3. Log on to the second character
4. Turn the safety on, which means clearing the "Can attack innocent players" box

Only `UC_SAFETY` crosses the wire.  Everything else is the client talking to itself.

### Before

Step 4 does nothing.  No message, no sound, and the server still has the safety off.

```mermaid
sequenceDiagram
    autonumber
    actor P as Player
    participant C as Client
    participant S as Server

    Note over C: config = safety on<br/>note = safety on
    P->>C: log on to character A
    P->>C: tick the box, safety off
    C->>S: UC_SAFETY, safety off
    S->>P: "your safety is now off" and the sound
    Note over C: config = safety off<br/>note = safety off

    P->>C: quit to the character select screen
    Note over C: ModulesClose wipes the note back to safety on<br/>config is saved elsewhere and survives

    P->>C: log on to character B
    Note over C: config = safety off<br/>note = safety on, they now disagree
    C->>S: UC_SAFETY, safety off
    S->>P: "your safety is now off" and the sound
    C->>S: UC_SAFETY, safety off, the client correcting the mismatch it sees
    S->>P: "your safety is now off" and the sound
    Note over C: neither send updated the note

    P->>C: untick the box, safety on
    Note over C: config = safety on<br/>note = safety on<br/>they match, so the client sends nothing
    Note over S: safety is still off
    S->>P: a reflection attacks an innocent, you turn murderer
```

### After

Each send updates the note, so step 4 finds a real change and sends it.

```mermaid
sequenceDiagram
    autonumber
    actor P as Player
    participant C as Client
    participant S as Server

    Note over C: config = safety on<br/>note = safety on
    P->>C: log on to character A
    P->>C: tick the box, safety off
    C->>S: UC_SAFETY, safety off
    Note over C: the send updates the note to safety off
    S->>P: "your safety is now off" and the sound

    P->>C: quit to the character select screen
    Note over C: ModulesClose wipes the note back to safety on<br/>config is saved elsewhere and survives

    P->>C: log on to character B
    Note over C: config = safety off<br/>note = safety on, they now disagree
    C->>S: UC_SAFETY, safety off
    S->>P: "your safety is now off" and the sound
    C->>S: UC_SAFETY, safety off, the client correcting the mismatch it sees
    S->>P: "your safety is now off" and the sound
    Note over C: both sends updated the note to safety off

    P->>C: untick the box, safety on
    Note over C: config = safety on<br/>note = safety off<br/>they differ, so send
    C->>S: UC_SAFETY, safety on
    S->>P: "your safety is now on" and the sound
    Note over S: safety is on, the reflection cannot attack
```

### DIFF
--- a/module/merintr/mermain.c
+++ b/module/merintr/mermain.c
@@ -32,6 +32,19 @@ static bool safety_flipped;
 
 /****************************************************************************/
 /*
+ * SendConfiguredSafety:  Send the configured safety state to the server and
+ *   record it in pinfo.aggressive.
+ */
+static void SendConfiguredSafety(void)
+{
+   pinfo.aggressive = cinfo->config->aggressive;
+   if (pinfo.aggressive)
+      SendSafety(0);
+   else
+      SendSafety(1);
+}
+/****************************************************************************/
+/*
  * InterfaceInit:  Called on startup.  Initialize interface.
  */
 void InterfaceInit(void)
@@ -71,10 +84,7 @@ void InterfaceInit(void)
       RequestRest();
    else
       RequestStand();
-   if (cinfo->config->aggressive)
-      SendSafety(0);
-   else
-      SendSafety(1);
+   SendConfiguredSafety();
 }
 /****************************************************************************/
 /*
@@ -175,10 +185,7 @@ void InterfaceResetData(void)
       RequestRest();
    else
       RequestStand();
-   if (cinfo->config->aggressive)
-      SendSafety(0);
-   else
-      SendSafety(1);
+   SendConfiguredSafety();
 
    RequestSpells();
    RequestSkills();
@@ -416,7 +423,7 @@ void InterfaceUserChanged(void)
       if (cinfo->config->aggressive)
       {
          if (!safety_flipped)
-            SendSafety(0);
+            SendConfiguredSafety();
          safety_flipped = true;
       }
       else 
@@ -430,7 +437,7 @@ void InterfaceUserChanged(void)
       if (!cinfo->config->aggressive)
       {
          if (!safety_flipped)
-            SendSafety(1);
+            SendConfiguredSafety();
          safety_flipped = true;
       }
       else 


davidklaw:
I tested and verified from a functionality standpoint.

A clarifying note: the reason involves multiple reports of players switching accounts under the impression that their safety is set one way, when it is actually the reverse, leading them to become orange (and perhaps red) by accident.

Know anything about the safety message appearing twice? Whether that's an intentional notice we added in years past or a bug for another PR? I see no reason for it if the safety toggle is working as intended.

adrienlaws:
Thanks for testing and the clarification. I updated the PR description with your reasoning under the "Why" section.

In regards to the double messages, it is the client sending the safety twice on login.  It sends the safety value from its saved setting and then again to correct a disagreement between server and client.  When the server responds it "announces" each one with the text string and the sound.

I think it'd be worth fixing but probably not in this PR.  If we did implement a fix like the #222 Issue suggests, that is per-character, this issue would go away.

PR #1501, opened 2026-09-11
Add bitmap preloading

neiltaco:
This PR fixes a bug that occurs with multiple bitmaps animations that aren't cached when meant to play together.  It appears the game loads in bitmaps on demand, which works as intended but it appears to cause this rare graphical bug.

In the base game, it is possible to trigger this bug by logging in first, punching something, then casting a spell.  povglow.bgf will attempt to load in while the cached povhand.bgf is already playing its animation, which show a desync in animation by a few frames.  Afterwards, the bug doesn't show anymore now that both bitmaps are cached.  While this specific bug is rarely triggered in the base game, it is made more apparent in #1484.

This PR attempts to address the bug by creating a new server message that tells the client to preload specific bitmaps and cache them upon login.  Ideally, preloading .bgf files should be kept to a minimum, so only povhand and povglow are preloaded.

This short video shows the difference between the current game and the new fix:

https://github.com/user-attachments/assets/2cb25c9a-cf12-45a6-b69b-a216f3e2f06a


akirmse:
I haven't looked into this problem in detail yet but I'm very doubtful that
the right answer to a client-only animation bug is sending an additional ad
hoc message from the server.

neiltaco:
That is fair!  I was using GetObjectBitmap() to load in .bgf files, but I mistakenly thought I had to use resource IDs passed in by the server.  I can close this PR.

